PR #376, opened 2021-04-20
test(sdk): add getAccessToken for SDK integration test

wenytang-ms:


wenytang-ms:
Take time off tomorrow morning to do examination, will WFH in the afternoon to save commute time.

PR #800, opened 2021-05-10
feat(extension): show welcome page at first activation

tecton:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/9888174

### DIFF
--- a/packages/vscode-extension/src/controls/openWelcomePage.ts
+++ b/packages/vscode-extension/src/controls/openWelcomePage.ts
@@ -1,0 +1,21 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import * as vscode from 'vscode';
+import { ext } from "../extensionVariables";
+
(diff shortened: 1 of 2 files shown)


packages/vscode-extension/src/controls/openWelcomePage.ts
```diff
@@ -0,0 +1,21 @@
...
+// Licensed under the MIT license.
+
+import * as vscode from 'vscode';
+import { ext } from "../extensionVariables";
+
+const welcomePageKey = 'teamsfx-extension.welcomePage.shown';
+
+export async function openWelcomePageAfterExtensionInstallation(): Promise<void> {
+    if (!/^en(-us)?$/i.test(vscode.env.language)) {
+        // Don't show: this page is English only
```
HuihuiWu-Microsoft, line 11: Why not show for English only?

tecton: Fixed, thanks!

PR #805, opened 2021-05-10
fix(function): update for template repo move

henzhang-ms:


packages/fx-core/src/common/templates.ts
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+import * as semver from "semver";
+
+export const templatesVersion = "0.1.*";
```
hund030, line 5: All templates share same versioin?

hund030: Yes, we have offline decided it.

packages/fx-core/src/common/templates.ts
```diff
@@ -0,0 +1,20 @@
...
+    "https://github.com/henzhang-ms/action-study/releases/download/template-tag-list/template-tags.txt";
+
+export function selectTag(tags: string[]): string | undefined {
+    const versionPattern = usePrerelease ? `${templatesVersion}-*` : templatesVersion;
+    const versionList = tags.map((tag: string) => tag.replace(tagPrefix, ""));
+    const selectedVersion = semver.maxSatisfying(versionList, versionPattern);
+    return selectedVersion ? (tagPrefix + selectedVersion) : undefined;
+}
+
+export const templateURL = (tag: string, templateName: string) : string =>
```
hund030, line 19: NIT: getTemplateURL would be better for me.

hund030 (commented):
LGTM, pending approve for the tagListURL.

hund030 (approved):

packages/fx-core/src/common/templates.ts
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+import * as semver from "semver";
+
+export const templatesVersion = "0.1.*";
+export const tagPrefix = "templates@";
+export const preRelease = process.env.TEAMSFX_TEMPLATE_PRERELEASE || "";
```
eriolchan, line 7: how about nightly build? I remember the conclusion is that the env variable is a flag about branch (main for nightly build, pre-release for pre-release and release with version for release)?

henzhang-ms: I think, the conclusion is that for nightly, pre-release, we always have one release for each of them. So my decision is that we have 3 release called (0.0.0-nightly, 0.0.0-pre-release, 0.0.0-test). And we use env var TEAMSFX_TEMPLATE_PRERELEASE to specify which release we are using. I only create 0.0.0-test now, which is a test env for dev. (Notice, prerelase is the name of the 'xxx' of '0.0.0-xxx'. In semver, it is the term of the version part).

PR #807, opened 2021-05-10
feat(extension): make quick start and sample page responsive

tecton:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/9888175

### DIFF
--- a/packages/vscode-extension/src/controls/QuickStart.scss
+++ b/packages/vscode-extension/src/controls/QuickStart.scss
@@ -1,5 +1,5 @@
 .quick-start-page{
-    margin: 100px 100px 100px 100px;
+    margin: 100px 0;
 
     a{
         text-decoration: none;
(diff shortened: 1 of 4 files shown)


packages/vscode-extension/src/controls/QuickStart.scss
```diff
@@ -8,6 +8,10 @@
     .section{
         display: flex;
         margin-bottom: 36px;
+
+        .title {
+            min-width: 300px;
```
HuihuiWu-Microsoft, line 13: As you stretch the page, will title div goes wider than contents?

HuihuiWu-Microsoft (approved):

packages/vscode-extension/src/controls/QuickStart.scss
tecton: No, it won't.

PR #824, opened 2021-05-10
feat(local-debug): detect if some ports are already in use

kuojianlu:
detect if some ports are already in use, and if so, stop debugging.

Preview:
image

image

image


packages/vscode-extension/src/debug/commonUtils.ts
```diff
@@ -143,4 +145,49 @@ async function getLocalDebugConfig(key: string): Promise<string | undefined> {
 
 export async function getSkipNgrokConfig(): Promise<string | undefined> {
   return getLocalDebugConfig(constants.skipNgrokConfigKey);
-}
\ No newline at end of file
+}
+
+async function getPortListeningPidWindows(host: string, port: number): Promise<string | undefined> {
+  try {
+    let command = `(Get-NetTCPConnection -LocalPort ${port} -State Listen).OwningProcess`;
```
swatDong, line 152: Try to use existing npm package first, e.g., https://www.npmjs.com/package/tcp-port-used, https://www.npmjs.com/package/portfinder, etc...

swatDong (changes requested):
Leverage npm package to work for multi-os.

packages/vscode-extension/src/debug/commonUtils.ts
kuojianlu: Great solution. I suppose it will be more stable than shell solution. I investigated several npm package, and decide to use `net.createServer` and `net.listen` directly. The reason is that these package cannot handle one thing like, node inspector listens on `127.0.0.1`, it is OK if there is already a process listens on `0.0.0.0`, `[::]` or `[::1]` on the same port.

packages/vscode-extension/src/debug/constants.ts
```diff
@@ -35,4 +35,16 @@ export enum ProgrammingLanguage {
...
 }
 
-export const skipNgrokConfigKey = "fx-resource-local-debug.skipNgrok";
\ No newline at end of file
+export const skipNgrokConfigKey = "fx-resource-local-debug.skipNgrok";
+
+const allAddressIPv4 = "0.0.0.0";
+const allAddressIPv6 = "::";
+const loopbackAddressIPv4 = "127.0.0.1";
+const loopbackAddressIPv6 = "::1";
```
swatDong, line 43: What if the machine does not have ipv6?

packages/vscode-extension/src/debug/commonUtils.ts
```diff
@@ -143,4 +144,34 @@ async function getLocalDebugConfig(key: string): Promise<string | undefined> {
...
+  return new Promise<boolean>((resolve, reject) => {
+    try {
+      const server = net.createServer();
+      server.once("error", (err) => {
+              if (err.message.includes("EADDRINUSE")) {
+                resolve(true);
+              }
+            })
+            .once("listening", () => {
+              server.close();
```
swatDong, line 159: Does `server.close()` release the port immediately? Or, need some timeout?

packages/vscode-extension/src/debug/constants.ts
kuojianlu: Good question. Did not find any document on this. Tried disabling IPv6 on my Windows machine, the result is that `net.listen` can still listens on "::". I am not sure whether disabling IPv6 takes effect.
However, if `net.listen` throws an error, only `EADDRINUSE` will be considered as port in use. For any other errors, simply ignore.

packages/vscode-extension/src/debug/commonUtils.ts
kuojianlu: Practically, `server.close()` will release the port immediately since there is no connection established on this port.

kuojianlu: Added `once("close")` to solve this potential problem:
```
            .once("listening", () => {
              server.close();
            })
            .once("close", () => {
              resolve(false);
            })
```

swatDong (approved):

packages/vscode-extension/src/debug/constants.ts
kuojianlu: Added `resolve(false)` for any other errors:
```
      server.once("error", (err) => {
              if (err.message.includes("EADDRINUSE")) {
                resolve(true);
              } else {
                resolve(false);
              }
            })
```

jayzhang (approved):

PR #857, opened 2021-05-11
fix: cannot read selected subscription correctly

HuihuiWu-Microsoft:
Read config when open so sidebar subscription item can load correctly

ADO: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/9888522

### DIFF
--- a/packages/fx-core/src/core/index.ts
+++ b/packages/fx-core/src/core/index.ts
@@ -408,6 +408,8 @@ class CoreImpl implements Core {
             supported = await this.isSupported();
             if (!supported) {
                 this.ctx.logProvider?.warning(`non Teams project:${workspace}`);
+            } else{
+                await this.readConfigs();
             }
         }
         const t2 = new Date().getTime();


packages/fx-core/src/core/index.ts
```diff
@@ -408,6 +408,8 @@ class CoreImpl implements Core {
             supported = await this.isSupported();
             if (!supported) {
                 this.ctx.logProvider?.warning(`non Teams project:${workspace}`);
+            } else{
+                await this.readConfigs();
```
HuihuiWu-Microsoft, line 412: I tried to read config by setting true here this.runWithErrorHandling<null>("open", **true**, ok(null), () => this.coreImpl.open(workspace)), but seems the root is not assigned before open() called. So I manually read config here.

1openwindow (approved):

LongOddCode (approved):

PR #861, opened 2021-05-11
feat: update webview resources and css styles

HuihuiWu-Microsoft:
1. Update In Meeting sample app screenshot
2. Adjust sample card css style
3. Update Teams command image(https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/9901343)
image


packages/vscode-extension/src/controls/SampleGallery.tsx
```diff
@@ -3,6 +3,7 @@ import { Icon, Stack, Image, PrimaryButton } from '@fluentui/react'
 import './SampleGallery.scss'
 import { Commands } from './Commands'
 import HelloWorld from '../../media/helloworld.gif'
```
1openwindow, line 5: should we need to remove "HelloWorld"

1openwindow (approved):

packages/vscode-extension/src/controls/SampleGallery.tsx
HuihuiWu-Microsoft: Yes, after all resources done. But currently only in meeting app is ready. Other apps still need a placeholder

jayzhang (approved):

PR #868, opened 2021-05-11
fix(tab): disable eslint: exhaustive deps 

hund030:
Either include dependencies or remove the dependency array cause it not work, so disable the rule to fix the warning.
```
src\components\sample\lib\useData.ts
  Line 47:6:  React Hook useEffect has missing dependencies: 'auto' and 'reload'. Either include them or 
remove the dependency array  react-hooks/exhaustive-deps
```


zhenyasav (approved):
LGTM

PR #876, opened 2021-05-11
feat(frontend): refine sending telemetry

hund030:
1. start event doesn't need `success` property
2. error event need contains `error-code`, `error-message` and `error-type`
3. Plugin must send telemetry for api: [pre,post] [scaffold, provision, deploy], get question and local debug don't need to send telemetry. We can offline discuss if our plugin need more telemetry for our own business.

### DIFF
--- a/packages/fx-core/src/plugins/resource/frontend/constants.ts
+++ b/packages/fx-core/src/plugins/resource/frontend/constants.ts
@@ -28,7 +28,7 @@ export class Constants {
 }
 
 export class FrontendPluginInfo {
-    static PluginName = "Tab Frontend plugin";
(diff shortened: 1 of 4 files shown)


packages/fx-core/src/plugins/resource/frontend/constants.ts
```diff
@@ -28,7 +28,7 @@ export class Constants {
 }
 
 export class FrontendPluginInfo {
-    static PluginName = "Tab Frontend plugin";
+    static PluginName = "frontend";
```
hund030, line 31: The plugin name is only for telemetry, pick an easier name.

PR #878, opened 2021-05-11
feat:  fx-core webpack

jayzhang:
1. templates folder access: using getTemplatesFolder() instead of relative path
2. strings.json access: using getStrings() instead of import directly


packages/fx-core/package.json
```diff
@@ -105,6 +106,8 @@
     "@typescript-eslint/parser": "^4.19.0",
     "chai": "^4.2.0",
     "chai-as-promised": "^7.1.1",
+    "clean-webpack-plugin": "^4.0.0-alpha.0",
```
henzhang-ms, line 109: can we use a stable version?

jayzhang: not used, you can ignore it

henzhang-ms (approved):
approve for function modification

xzf0587 (approved):

packages/fx-core/src/index.ts
```diff
@@ -0,0 +1,19 @@
...
+export * from "./core";
+
+
+export function getTemplatesFolder(){
+    return path.join(__dirname, "../templates");
+}
+
+export function getResourceFolder(){
+    return path.join(__dirname, "../resource");
+}
```
YitongFeng-git, line 19: Why put these two functions in index.ts?

jiayi-ruan (approved):
approve for spfx change

KennethBWSong (approved):
approve for simpleauth

packages/fx-core/src/index.ts
jayzhang: to make sure the relative path of resource and templates are the same in development mode and the webpack mode

LongOddCode (approved):

IvanJobs (approved):

nliu-ms (approved):

XiaofuHuang (approved):

LongOddCode (approved):

PR #880, opened 2021-05-11
fix(function): align telemetry

henzhang-ms:


packages/fx-core/src/plugins/resource/function/index.ts
```diff
@@ -101,30 +102,35 @@ export class FunctionPlugin implements Plugin {
...
+            sendTelemetry && telemetryHelper.sendStartEvent(ctx, name);
             const res: FxResult = await fn();
-            telemetryHelper.sendResultEvent(ctx, name, res);
+            sendTelemetry && telemetryHelper.sendResultEvent(ctx, name, res);
             return res;
         } catch (e) {
             if (e instanceof UserError || e instanceof SystemError) {
                 const res = err(e);
-                telemetryHelper.sendResultEvent(ctx, name, res);
+                sendTelemetry && telemetryHelper.sendResultEvent(ctx, name, res);
```
hund030, line 119: I think this kind of errors are from other components, do we send telemetry for them?

henzhang-ms: It is not sent for reporting error. It is sent for that the life cycle exits because of this error. we send it for our life cycle. And we can't assume that library always maintain telemetry.

hund030: Make sense.

hund030 (approved):

PR #881, opened 2021-05-11
fix: simple auth cannot run with .net 5 only

blackchoey:
Set rollForward to `Major` which roll forward to lowest higher major version, and lowest minor version, if requested major version is missing.


adashen (approved):

qinezh:
How about updating the csproj to support multiple target frameworks, which is more like a common solution:

```
<TargetFrameworks>netcoreapp3.1;net5.0</TargetFrameworks>
```

https://github.com/OfficeDev/TeamsFx/blob/main/packages/simpleauth/src/TeamsFxSimpleAuth/Microsoft.TeamsFx.SimpleAuth.csproj#L4

PR #890, opened 2021-05-11
feat(vsc): add tree view telemetery in vsc

1openwindow:
work item: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/9888182

### DIFF
--- a/packages/fx-core/src/core/index.ts
+++ b/packages/fx-core/src/core/index.ts
@@ -444,7 +444,10 @@ class CoreImpl implements Core {
             };
 
             const selectSubscriptionCallback = async (args?: any[]): Promise<Result<null, FxError>> => {
-                this.ctx?.telemetryReporter?.sendTelemetryEvent(TelemetryEvent.SelectSubscription, getTriggerFromProperty(args))
+                this.ctx?.telemetryReporter?.sendTelemetryEvent(TelemetryEvent.SelectSubscription, {
+                    [TelemetryProperty.TriggerFrom]: args && args.toString() === "TreeView" ? TelemetryTiggerFrom.TreeView : TelemetryTiggerFrom.CommandPalette
+                });
+
                 const azureToken = await this.ctx.azureAccountProvider?.getAccountCredentialAsync();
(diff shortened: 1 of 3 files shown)


HuihuiWu-Microsoft (approved):

packages/vscode-extension/src/handlers.ts
```diff
@@ -759,11 +693,11 @@ export async function cmdHdlLoadTreeView(context: ExtensionContext) {
     try {
       switch (node.contextValue) {
         case "signedinM365": {
-          signOutM365();
+          signOutM365(true);
```
chagong, line 696: we do sign in but the function name here is "signOutXXX()", do we do it on purpose?

packages/vscode-extension/src/telemetry/extTelemetryEvents.ts
```diff
@@ -1,9 +1,23 @@
...
+  QuickStart = "quick-start",
+
+  Samples = "samples",
+
+  Documentation = "documentation",
+
   LoginStart = "login-start",
   Login = "login",
 
+  LoginAzure = "login-azure",
```
chagong, line 13: Let's use "login-start" and "login" for event name, and add "account-type" property to indicate account type.

packages/fx-core/src/core/index.ts
```diff
@@ -486,7 +487,8 @@ class CoreImpl implements Core {
                 return ok(null);
             };
 
-            const signinM365Callback = async (): Promise<Result<null, FxError>> => {
+            const signinM365Callback = async (args?: any[]): Promise<Result<null, FxError>> => {
+                this.ctx?.telemetryReporter?.sendTelemetryEvent(TelemetryEvent.LoginM365, getTriggerFromProperty(args));
```
chagong, line 491: And we should have the following properties according to PM's spec: 
"component", "hashed-userid", "internal": "internal-alias"

1openwindow: @chagong can you share the spec? I try to understand what these properties are.

jayzhang (approved):

PR #898, opened 2021-05-12
chore(bot): update teamsfx package for bot and me templates

YitongFeng-git:


templates/bot-msgext/js/default/package.json
```diff
@@ -13,7 +13,7 @@
     "watch": "nodemon ./index.js"
   },
   "dependencies": {
-    "teamsdev-client": "^1.0.0",
+    "@microsoft/teamsfx": "0.1.0-alpha.3",
```
IvanJobs, line 16: Is it possible to use something like ^0.1.0? or why choosing a fixed version?

IvanJobs (approved):

templates/bot-msgext/js/default/package.json
YitongFeng-git: Offline discussed

PR #901, opened 2021-05-12
fix(env-checker): fix ENOENT error when run dotnet-install script with execFile

a1exwang:
- When using relative PATH in execFile, it will throw ENOENT on macOS 11.2.3. Changed to absolute path to workaround this issue
## Tests:
Run local debug without .NET SDK

#### Output Channel
image
#### Final result
image

#### Output Channel
image
#### Final result
image

### Linux
Linux is not affected because we won't install .NET SDK on Linux

### DIFF
--- a/packages/fx-core/src/plugins/resource/function/utils/depsChecker/dotnetChecker.ts
+++ b/packages/fx-core/src/plugins/resource/function/utils/depsChecker/dotnetChecker.ts
@@ -394,7 +394,7 @@ export class DotnetChecker implements IDepsChecker {
   ): Promise<string[]> {
     const command = [
       // path.join does not prepend '.'
-      [".", this.getDotnetInstallScriptName()].join(path.sep),
+      this.getDotnetInstallScriptPath(),
       "-InstallDir",
       isWindows() ? DotnetChecker.escapeFilePath(dotnetInstallDir) : dotnetInstallDir,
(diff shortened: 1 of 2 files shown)


kimizhu:
@a1exwang can you test on Windows/Mac/Linux before checking the code? You can add the test result here to let us know.

packages/fx-core/src/plugins/resource/function/utils/depsChecker/dotnetChecker.ts
```diff
@@ -394,7 +394,7 @@ export class DotnetChecker implements IDepsChecker {
   ): Promise<string[]> {
     const command = [
       // path.join does not prepend '.'
```
qinezh, line 396: this comment can be removed.

qinezh (approved):

a1exwang:
> @a1exwang can you test on Windows/Mac/Linux before checking the code? You can add the test result here to let us know.

Tested on Windows and macOS. Linux is not affected because we won't install .NET SDK on Linux

packages/fx-core/src/plugins/resource/function/utils/depsChecker/dotnetChecker.ts
a1exwang: Removed

PR #903, opened 2021-05-12
fix(function): use public sdk and func ext in templates

henzhang-ms:


templates/function-base/js/default/package.json
```diff
@@ -5,7 +5,7 @@
     "start": "npx func start"
   },
   "dependencies": {
-    "teamsdev-client": "^1.1.0",
+    "@microsoft/teamsfx": "^0.1.0-alpha.3",
```
YitongFeng-git, line 8: Currently only 0.1.0-alpha.3 works, it's better to remove ^

dilin-MS (approved):

PR #911, opened 2021-05-12
feat: add show message interactive operation for CLI

jiayi-ruan:
image

### DIFF
--- a/packages/cli/src/userInterface.ts
+++ b/packages/cli/src/userInterface.ts
@@ -24,7 +24,7 @@ import {
 import inquirer from "inquirer";
 import CLILogProvider from "./commonlib/log";
 import { ProgressHandler } from "./progressHandler";
-import { NotSupportedQuestionType } from "./error";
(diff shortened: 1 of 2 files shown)


packages/cli/src/userInterface.ts
```diff
@@ -151,19 +148,35 @@ export class DialogManager implements Dialog {
...
+        name: DialogType.Show,
+        type: "list",
+        message: msg.description,
+        choices: msg.items
+      }]);
+      if (DialogType.Show in answers) {
+        return answers[DialogType.Show];
+      }
+      else {
+        throw InquirerAnswerNotFound(msg);
```
chagong, line 163: Where do we capture this error?

jiayi-ruan: I think cli upper level(probably the runcommand level) will catch that?
and also I am a bit unsure about whether we need to handle the error, since it seems impossible for the list question returns an answer not in the list.

chagong (approved):

PR #913, opened 2021-05-12
feat(local-debug): enable detecting ports in use

kuojianlu:
enable detecting ports in use introduced by #824 .

### DIFF
--- a/packages/fx-core/src/plugins/resource/localdebug/tasks.ts
+++ b/packages/fx-core/src/plugins/resource/localdebug/tasks.ts
@@ -10,13 +10,19 @@ export function generateTasks(includeFrontend: boolean, includeBackend: boolean,
      * Referenced by launch.json
      *   - Pre Debug Check
      *   - Start Frontend
+     *   - Start Backend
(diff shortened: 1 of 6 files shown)


packages/fx-core/src/plugins/resource/localdebug/tasks.ts
```diff
@@ -27,6 +33,7 @@ export function generateTasks(includeFrontend: boolean, includeBackend: boolean,
                 label: "Pre Debug Check",
                 dependsOn: [
                     "dependency check",
+                    "detect ports in use",
```
swatDong, line 36: Move into predebugcheck, not add new task...

swatDong (changes requested):

packages/fx-core/src/plugins/resource/localdebug/tasks.ts
kuojianlu: Updated. Adding new task makes it more complex...

swatDong (approved):

jayzhang (approved):

PR #923, opened 2021-05-12
fix(apim): fix apim plugin display bug

XiaofuHuang:
1. Change retry command error message
2. Fix the default value of empty api prefix and version


packages/fx-core/src/plugins/resource/apim/constants.ts
```diff
@@ -189,50 +188,50 @@ export enum TeamsToolkitComponent {
...
 
-export enum RetryCommands {
-    Create = "start a project",
-    Update = "add the resource",
-    Provision = "provision resource",
-    Deploy = "deploy package",
-    Login = "login and choose a subscription",
+export enum RetryOperation {
+    Create = "create a new project",
+    Update = "add resources",
```
dooriya, line 193: Add APIM resource?

dooriya (approved):

PR #939, opened 2021-05-13
feat(tab): update docs links & use public sdk

hund030:


templates/tab/js/default/package.json
```diff
@@ -13,7 +13,7 @@
     "react-dom": "^16.14.0",
     "react-router-dom": "^5.1.2",
     "react-scripts": "^4.0.3",
-    "teamsdev-client": "^1.1.0"
```
henzhang-ms, line 16: I think we should sort the dependency.

henzhang-ms (approved):

PR #945, opened 2021-05-13
feat: fix cli login mfa

xiaolang124:
1. Fix cli login azure mfa.


xiaolang124:
image

packages/cli/src/commonlib/azureLogin.ts
```diff
@@ -115,44 +116,51 @@ export class AzureAccountManager extends login implements AzureAccountProvider {
...
-            (tokenJson as any).upn ?? (tokenJson as any).unique_name,
-            undefined,
-            env,
-            memoryDictionary[AzureAccountManager.tenantId]
-          );
-          resolve(credential);
-        }
-      });
+  async getAccountCredentialAsync(showDialog = true, tenantId = ""): Promise<TokenCredentialsBase | undefined> {
+    if (tenantId.length==0) {
```
swatDong, line 120: what if tenantId is undefined?

xiaolang124: It will encounter ts compile error.

packages/cli/src/commonlib/codeFlowLogin.ts
```diff
@@ -59,7 +61,11 @@ export class CodeFlowLogin {
     }
   }
 
-  async login(): Promise<string> {
+  async login(tenantId?: string): Promise<string> {
+    if (tenantId) {
+      this.config!.auth.authority = env.activeDirectoryEndpointUrl + tenantId;
+      this.pca = new PublicClientApplication(this.config!);
```
swatDong, line 67: Change `this.pca` outside of constructor is not a good idea.
How about adding a new class `codeFlowTenantLogin` to handle tenant case only?
Since this flow will be shared by M365 and azure, better to not change the current class.

swatDong:
@timngmsft - please help review the wording of the MFA warning message.
We are using "**The following tenants require Multi-Factor Authentication (MFA). Use 'teamsfx account login azure --tenant TENANT_ID' to explicitly login to a tenant.**" which is copied from the Azure CLI one (https://github.com/Azure/azure-cli/blob/dev/src/azure-cli-core/azure/cli/core/_profile.py#L1019), but not sure if it's enough for teamsfx scenario.

packages/cli/src/commonlib/codeFlowLogin.ts
xiaolang124: Seperate tenant login from codeFlow.

packages/cli/src/commonlib/azureLogin.ts
```diff
@@ -164,20 +190,35 @@ export class AzureAccountManager extends login implements AzureAccountProvider {
...
+      AzureAccountManager.codeFlowInstance !== undefined &&
       AzureAccountManager.codeFlowInstance.account &&
       AzureAccountManager.statusChange !== undefined
     ) {
       const credential = await this.doGetAccountCredentialAsync();
       const accessToken = await credential?.getToken();
       const accountJson = await this.getJsonObject();
       await AzureAccountManager.statusChange("SignedIn", accessToken?.accessToken, accountJson);
+    } else if (
+      AzureAccountManager.codeFlowTenantInstance !== undefined &&
```
swatDong, line 202: It seems `if` and `else if` have the same code block here

packages/cli/src/commonlib/azureLogin.ts
```diff
@@ -235,8 +288,13 @@ export class AzureAccountManager extends login implements AzureAccountProvider {
     return Promise.resolve(undefined);
   }
 
-  async getJsonObject(showDialog = true): Promise<Record<string, unknown> | undefined> {
-    const token = await AzureAccountManager.codeFlowInstance.getToken();
+  async getJsonObject(showDialog = true, tenantId = ""): Promise<Record<string, unknown> | undefined> {
```
swatDong, line 291: Is this a public function? Do not change contract of other public function(s).

packages/cli/src/commonlib/codeFlowLogin.ts
```diff
@@ -218,15 +220,23 @@ export class CodeFlowLogin {
           }
         })
         .catch(async (error) => {
-          CliCodeLogInstance.error("[Login] getTenantToken acquireTokenSilent : " + error.message);
-          const accountList = await this.msalTokenCache?.getAllAccounts();
-          for (let i=0;i<accountList!.length;++i) {
-            this.msalTokenCache?.removeAccount(accountList![i]);
+          if (error.message.indexOf(MFACode) >= 0) {
+            if (this.showMFA) {
```
swatDong, line 224: To show warning only once? What's the side effect if show multiple times?

packages/cli/src/commonlib/codeFlowLogin.ts
```diff
@@ -218,15 +220,23 @@ export class CodeFlowLogin {
...
+              this.showMFA = false;
+            }
+            return undefined;
+          } else {
+            CliCodeLogInstance.error("[Login] getTenantToken acquireTokenSilent : " + error.message);
+            const accountList = await this.msalTokenCache?.getAllAccounts();
+            for (let i=0;i<accountList!.length;++i) {
+              this.msalTokenCache?.removeAccount(accountList![i]);
+            }
+            this.config!.auth.authority = env.activeDirectoryEndpointUrl + tenantId;
```
swatDong, line 235: Why setting tenantId in this common login workflow?

packages/cli/src/commonlib/codeFlowTenantLogin.ts
```diff
@@ -0,0 +1,329 @@
...
+    if (dataCache) {
+      this.msalTokenCache?.removeAccount(dataCache);
+    }
+    if (fs.existsSync(accountPath + this.accountName)) {
+      fs.writeFileSync(accountPath + this.accountName, "", UTF8);
+    }
+    return true;
+  }
+
+  async getToken(tenantId?: string): Promise<string | undefined> {
```
swatDong, line 170: Duplicated `getToken` and `getTenantToken`. This work flow should only work for tenanted login.

packages/cli/src/commonlib/codeFlowLogin.ts
xiaolang124: Will look like this.
```
The following tenants require Multi-Factor Authentication (MFA). Use 'teamsfx account login azure --tenant TENANT_ID' to explicitly login to a tenant.
1111111111111
The following tenants require Multi-Factor Authentication (MFA). Use 'teamsfx account login azure --tenant TENANT_ID' to explicitly login to a tenant.
2222222222222
```

packages/cli/src/commonlib/codeFlowLogin.ts
xiaolang124: When login refresh token expires, try to use user used tenant id to login.

packages/cli/src/commonlib/codeFlowLogin.ts
swatDong: I see...
Make change to existing work flow as less as possible, e.g., catch and show error in `azureLogin`

xiaolang124: Moved to azureLogin.

packages/cli/src/commonlib/codeFlowTenantLogin.ts
xiaolang124: ```getTenantToken``` has been removed.

Alive-Fish:
push a commit to enable `teamsfx account login azure --tenant xxx`.

jayzhang (approved):

packages/cli/src/cmds/account.ts
```diff
@@ -138,14 +144,16 @@ class AccountLogin extends YargsCommand {
     });
     switch (args.service) {
       case "azure": {
-        const result = await outputAzureInfo("login");
+        await AzureTokenProvider.signout();
```
chagong, line 147: Is this to make sure every time login command will open login page?

packages/cli/src/commonlib/azureLogin.ts
xiaolang124: Fixed.

packages/cli/src/commonlib/azureLogin.ts
xiaolang124: Fixed.

packages/cli/src/cmds/account.ts
xiaolang124: Yes, this is for Azure login, should we change M365 login scenario to this?
@chagong @swatDong @Alive-Fish

chagong: I think so. And the change is already there.

xiaolang124: Very nice.

xiaolang124:
I test with my account, and it works fine.
Provision:
image

Deploy:
image

chagong (approved):
Looks good to me, please make sure you have tested it.

swatDong (approved):
The login logic looks good to me and glad to see the happy path works.

One more question maybe I missed - where to validate the input tenant id format? @xiaolang124 @Alive-Fish

xiaolang124:
> The login logic looks good to me and glad to see the happy path works.
> 
> One more question maybe I missed - where to validate the input tenant id format? @xiaolang124 @Alive-Fish

I will talk with zhiyu and add validation later.

PR #949, opened 2021-05-13
fix(apim): fix add resource options for APIM

dooriya:
The option item for APIM in Add Resource command is not correct:

- Current value: Register APIs in Azure API Management
- Updated value: Azure API Management

image

image


jayzhang (approved):

chagong:
Hi @dooriya, You tried it with CLI, right?

XiaofuHuang (approved):

PR #950, opened 2021-05-13
fix(apim): fix apim e2e test and change cli description

XiaofuHuang:
fix apim e2e test and change cli description


Alive-Fish (approved):
Look good to me.

PR #959, opened 2021-05-13
fix: only prompt user to select active subscription when user manual …

HuihuiWu-Microsoft:
…sign in

1> Prompt user to select active sub only when user sign in from 
  a. sidebar
  b. command palette
  c. webview
2> When lazy sign in, provision logic will handle the select sub logic, so don't popup
3> Select sub by default if user only has one sub under current account


packages/fx-core/src/core/index.ts
```diff
@@ -446,6 +446,10 @@ class CoreImpl implements Core {
                     if(subscriptions.length === 0){
                         contextValue = "emptySubscription";
                     }
+
+                    if(subscriptions.length ===1){
```
HuihuiWu-Microsoft, line 450: Select by default if user has only one sub

packages/fx-core/src/core/index.ts
```diff
@@ -531,11 +521,12 @@ class CoreImpl implements Core {
 
             const signinAzureCallback = async (validFxProject: boolean, args?: any[]): Promise<Result<null, FxError>> => {
                 this.ctx?.telemetryReporter?.sendTelemetryEvent(TelemetryEvent.LoginStart, {
-                    [TelemetryProperty.TriggerFrom]: args && args.toString() === "TreeView" ? TelemetryTiggerFrom.TreeView : TelemetryTiggerFrom.CommandPalette,
+                    [TelemetryProperty.TriggerFrom]: args && args[0].toString() === "TreeView" ? TelemetryTiggerFrom.TreeView : TelemetryTiggerFrom.CommandPalette,
                     [TelemetryProperty.AccountType]: AccountType.Azure
                 });
 
-                const token = await this.ctx.azureAccountProvider?.getAccountCredentialAsync(true);
+                const showDialog = args && args[1] !== undefined? args[1]: true;
```
HuihuiWu-Microsoft, line 528: Show dialog when triggered from sidebar and cmp, not webview

packages/fx-core/src/core/index.ts
```diff
@@ -549,6 +540,10 @@ class CoreImpl implements Core {
 
                     const subItem = await getSelectSubItem!(token, validFxProject);
                     this.ctx.treeProvider?.add([subItem[0]]);
+
+                    if (validFxProject && !subItem[1]) {
```
HuihuiWu-Microsoft, line 544: Popup select sub if it's valid teamsfx project and user hasn't select one(don't popup if only one sub)

packages/vscode-extension/src/commandsTreeViewProvider.ts
```diff
@@ -373,7 +373,7 @@ export class TreeViewCommand extends vscode.TreeItem {
       this.command = {
         title: label,
         command: commandId,
-        arguments: [CommandsTreeViewProvider.TreeViewFlag]
+        arguments: [[CommandsTreeViewProvider.TreeViewFlag]]
```
HuihuiWu-Microsoft, line 376: I need second parameter to mark if showDialog when sign in

jayzhang (approved):

packages/fx-core/src/core/index.ts
```diff
@@ -543,11 +530,12 @@ class CoreImpl implements Core {
 
             const signinAzureCallback = async (validFxProject: boolean, args?: any[]): Promise<Result<null, FxError>> => {
                 this.ctx?.telemetryReporter?.sendTelemetryEvent(TelemetryEvent.LoginStart, {
-                    [TelemetryProperty.TriggerFrom]: args && args.toString() === "TreeView" ? TelemetryTiggerFrom.TreeView : TelemetryTiggerFrom.CommandPalette,
+                    [TelemetryProperty.TriggerFrom]: args && args[0].toString() === "TreeView" ? TelemetryTiggerFrom.TreeView : TelemetryTiggerFrom.CommandPalette,
```
chagong, line 533: why this change?

packages/vscode-extension/src/commandsTreeViewProvider.ts
chagong: I see

packages/fx-core/src/core/index.ts
HuihuiWu-Microsoft: Because I need second parameter to indicate if need show dialog when sign in

PR #1007, opened 2021-05-14
chore(solution): format core

1yefuwang1:
format core


LongOddCode (approved):

jayzhang (approved):

IvanJobs (approved):

packages/fx-core/src/plugins/resource/localdebug/certificate.ts
```diff
@@ -14,284 +22,307 @@ import * as ps from "./util/process";
 const continueText: string = "Continue";
 const learnMoreText: string = "Learn More";
 const learnMoreUrl: string = "https://aka.ms/teamsfx-ca-certificate";
-const confirmMessage: string = "To debug applications in Teams, your localhost server must be on HTTPS.\
+const confirmMessage: string =
+  'To debug applications in Teams, your localhost server must be on HTTPS.\
  For Teams to trust the self-signed SSL certificate used by the toolkit, a CA certificate must be added to your certificate store.\
- You may skip this step, but you'll have to manually trust the secure connection in a new browser window when debugging your apps in Teams.\
- For more information \"https://aka.ms/teamsfx-ca-certificate\". You may be asked for your account credentials when installing the certificate.";
+ You may skip this step, but you\'ll have to manually trust the secure connection in a new browser window when debugging your apps in Teams.\
```
swatDong, line 28: Why this "\" added?

swatDong: And I remember we should use `"` for string, not `'`

swatDong (commented):
One comment for local debug part.
And... do merge and resolve conflict carefully...

XiaofuHuang (approved):

YitongFeng-git (approved):

hund030 (approved):

packages/fx-core/src/plugins/resource/localdebug/certificate.ts
1yefuwang1: It's auto generated by prettier...

swatDong: So what's the expected result, `"` or `'`?

a1exwang (approved):

packages/fx-core/src/plugins/resource/localdebug/certificate.ts
1yefuwang1: I think we are using ".

zhenyasav:
Perhaps it might be easier to just run the command on `main` and push straight to main without a PR? this way we don't have to worry about conflicts?

packages/fx-core/src/plugins/resource/localdebug/certificate.ts
swatDong: Agree... then how to make sure next time it won't be auto generated to `'`...

1yefuwang1:
> Perhaps it might be easier to just run the command on `main` and push straight to main without a PR? this way we don't have to worry about conflicts? Theoretically that would reduce this PR down to any "real" changes we had in this (and other) PRs ?

Agreed. But I think only admins can do that. Currently managers and @zhenyasav are the admins.

PR #1037, opened 2021-05-14
fix: unify user cancel handling

jayzhang:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/9935752
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/9941232


packages/fx-core/src/common/tools.ts
```diff
@@ -248,3 +248,12 @@ export function getStrings(): any {
...
 }
+
+export function isUserCancelError(error: Error): boolean {
+  const errorName = "name" in error ? (error as any)["name"] : "";
+  return (
+    errorName === "User Cancel" ||
+    errorName === getStrings().solution.CancelProvision ||
+    errorName === "UserCancel"
+  );
+}
```
jayzhang, line 259: This is the common function to judge whether an error is user cancel event

packages/fx-core/src/plugins/solution/fx-solution/solution.ts
```diff
@@ -865,23 +865,29 @@ export class TeamsAppSolution implements Solution {
...
-        const error = provisionResult.error;
-        error.message +=
-          " " +
-          util.format(getStrings().solution.ProvisionFailGuide, subscriptionId, resourceGroupName);
-        if (error instanceof UserError) {
-          const ue = error as UserError;
-          if (!ue.helpLink) {
-            ue.helpLink = "https://aka.ms/teamsfx-solution-help";
-            (ue.source = "Solution"), (ue.name = "ProvisionFailure");
+        if (!isUserCancelError(provisionResult.error)) {
```
jayzhang, line 868: ignore user cancel case for provision error handling

1openwindow (approved):

nliu-ms (approved):

LongOddCode (approved):

PR #1043, opened 2021-05-15
Refactor lerna setup

zhenyasav:
Upon looking around the setup scripts I thought some of these changes might improve things, but given how complex it is, I'm not sure what got better and what got worse.

Changes I made:
- [x] convert `npm setup` to a `postinstall` script, removing the need for any extra steps after `npm install` at the root
- [x] drop extra `npx lerna run -- npm install --only=dev` work during bootstrap, convert to recommended `--hoist` pattern instead https://github.com/lerna/lerna/issues/1679
- [x] update all .yml and .md files to reflect simplified boostrapping process
- [x] add `.md` to `.prettierignore`
- [x] drop spurious `npm link` commands, unit:tests seem to pass, but idk what i might have broken 

I had a few questions:
- why are there .prettierignore and .prettierrc.js files in every package folder? why not a single policy at the root of the monorepo?
- why are the some packages missing `.prettierignore` and `.prettierrc.js` if we decided to put them in every package?
- why are there `npm link` commands in CI yml definitions? `npm link` is only functional when paired with a matching `npm link <package>` statement in a target folder.

> Also, I have to say - when I recommended we all use the prettier extension, I did not mean to suggest we should go add more precommit hooks and automate prettification of the entire repo on every commit. These additions have really slowed down the dev loop because I now have to wait a long time during every commit for all the stuff we added. It would have been sufficient to just have developers manually prettify whatever files they want while they edit the code and over time we would have reached a reasonable state. 
> I am equally not a fan of the commit message format strictness, I don't understand what value it adds, I end up just prefixing every commit message with `fix:` - clearly that's not valuable to the team, but I value my own developer velocity above commit rules and recommend the team does the same.


LongOddCode:
1. in fact, we only format staged files on every commit by git precommit hook. 
2. I think each subrepo could work independently. Tha's why we add prettier & eslint in each subrepo.

qinezh:
> why are there .prettierignore and .prettierrc.js files in every package folder? why not a single policy at the root of the monorepo?

Actually, the policy defined for eslint and prettier config is already shared across the monorepo: [eslint-config](https://github.com/OfficeDev/TeamsFx/blob/main/packages/eslint-config/index.js), [prettier-config](https://github.com/OfficeDev/TeamsFx/blob/main/packages/prettier-config/index.js), and every packages (only Typescript project) follows this policy.

qinezh:
> I am equally not a fan of the commit message format strictness, I don't understand what value it adds, I end up just prefixing every commit message with fix: - clearly that's not valuable to the team, but I value my own developer velocity above commit rules and recommend the team does the same.

The reason why we prefer the unified commit message style:
* it's easy for teams member to know the purpose of the PR from its title:
  * is it a bug fix or a feature implementation?
  * which component is the PR related to? (especially we have quit a few components now.)
* it's easy for the tool to generate changelog for each release automatically, here's an example from the github release:
  image

PR #1051, opened 2021-05-17
Fix: make login telemetry accurate for extension and CLI

chagong:
fix: fix login telemetry

https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/9951570

### DIFF
--- a/packages/cli/src/commonlib/appStudioLogin.ts
+++ b/packages/cli/src/commonlib/appStudioLogin.ts
@@ -10,6 +10,8 @@ import CLILogProvider from "./log";
 import { getBeforeCacheAccess, getAfterCacheAccess } from "./cacheAccess";
 import { signedIn, signedOut } from "./common/constant";
 import { login, LoginStatus } from "./common/login";
+import CliTelemetry from "./../telemetry/cliTelemetry";
+import { TelemetryAccountType, TelemetryEvent, TelemetryProperty, TelemetrySuccess } from "../telemetry/cliTelemetryEvents";
 
 const accountName = "appStudio";
 const scopes = ["https://dev.teams.microsoft.com/AppDefinitions.ReadWrite"];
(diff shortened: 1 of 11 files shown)


packages/fx-core/src/core/index.ts
```diff
@@ -588,6 +588,22 @@ class CoreImpl implements Core {
               icon: "M365",
             },
           ]);
+
+          this.ctx?.telemetryReporter?.sendTelemetryEvent(TelemetryEvent.Login, {
```
HuihuiWu-Microsoft, line 592: Why are you sending the telemetry in commonlib for cli but in core for extension?

HuihuiWu-Microsoft: Also, the callback here will only be triggered when user click sign in from sidebar, how about other entries(webview, command palette and lazy sign in)?

jayzhang (approved):

packages/cli/src/commonlib/appStudioLogin.ts
```diff
@@ -73,17 +75,51 @@ export class AppStudioLogin extends login implements AppStudioTokenProvider {
...
+
+      if (loginToken) {
+        const array = loginToken.split(".");
+        const buff = Buffer.from(array[1], "base64");
+        const res = JSON.parse(buff.toString("utf-8"));
+        CliTelemetry.sendTelemetryEvent(TelemetryEvent.AccountLogin, {
+          [TelemetryProperty.AccountType]: TelemetryAccountType.M365,
+          [TelemetryProperty.Success]: TelemetrySuccess.Yes,
+          [TelemetryProperty.UserId]: res.oid,
+          [TelemetryProperty.Internal]: res.upn.endsWith("@microsoft.com") ? "true" : "false",
```
swatDong, line 102: Is it possible to get machine domain or machine account via node? That may be more accurate to determine an internal user.

swatDong (approved):
The logic looks good to me. One comment related to the "Internal" property.

packages/fx-core/src/core/index.ts
chagong: OK, change to another place.

packages/cli/src/commonlib/appStudioLogin.ts
chagong: This logic  is confirmed with PM. For now it's enough for our requirement.

HuihuiWu-Microsoft (approved):

packages/vscode-extension/src/commonlib/appStudioLogin.ts
```diff
@@ -17,7 +17,7 @@ import { login, LoginStatus } from "./common/login";
 import * as StringResources from "../resources/Strings.json";
 import * as util from "util";
 
-const accountName = "appStudio";
+const accountName = "m365";
```
xiaolang124, line 20: This will affect cache file name.
For now, there are two login types for M365: appStudio, graph. How about graph login telemetry?

chagong: Got it. Then I actually need 3 values for account type.

swatDong: Good catch! @chagong - I suggest to not change any used variable for now. How about using new variable for telemetry only and adding `TODO` so we can address more post public preview?

Alive-Fish (approved):

swatDong (approved):
LGTM

xiaolang124 (approved):

MSFT-yiz (approved):

PR #1052, opened 2021-05-17
fix(function): avoid to refer templates in ut

henzhang-ms:


packages/fx-core/tests/plugins/resource/function/unit/scaffold.test.ts
```diff
@@ -1,133 +1,148 @@
...
-
-            // Act
-            await plugin.preScaffold(context);
-            const ret: FxResult = await plugin.scaffold(context);
-
-            // Assert
-            chai.assert.isTrue(ret.isOk());
-        });
-
-        it("Test scaffold with fallback in TS", async () => {
```
henzhang-ms, line 116: Remove the ts version test because the templates file is mock, so language does not matter o n this test any more.

packages/fx-core/tests/plugins/resource/function/unit/scaffold.test.ts
```diff
@@ -115,22 +144,5 @@ describe(FunctionPluginInfo.pluginName, () => {
       // Assert
       chai.assert.isTrue(ret.isOk());
     });
-
-    it("Test scaffold with fallback in TS", async () => {
```
henzhang-ms: Remove the ts version test because the templates file is mock, so language does not matter o n this test any more.

hund030 (approved):

eriolchan (approved):

PR #1073, opened 2021-05-18
fix: throw error when download samples to the folder twice

Alive-Fish:
related bug: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/9942418/


Alive-Fish:
When users want to download a same sample app twice to a same folder, it will throw an error like this (follow [this PR's experience](https://github.com/OfficeDev/TeamsFx/pull/1060)): 
image

chagong (approved):

MSFT-yiz (approved):

PR #1088, opened 2021-05-18
fix: update the order of samples in CLI

Alive-Fish:
related PR: [fix: remove faq plus sample app in extension sample gallery page](https://github.com/OfficeDev/TeamsFx/pull/1081) adjust the order of sample apps in CLI.
image

### DIFF
--- a/packages/cli/src/utils.ts
+++ b/packages/cli/src/utils.ts
@@ -44,7 +44,7 @@ export function getParamJson(jsonFilePath: string): { [_: string]: Options } {
       const data = node.data as Question;
       if (isAutoSkipSelect(data)) {
         // set the only option to default value so yargs will auto fill it.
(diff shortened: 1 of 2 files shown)


packages/cli/src/utils.ts
```diff
@@ -53,48 +53,50 @@ export function getParamJson(jsonFilePath: string): { [_: string]: Options } {
...
     return singleOption;
   }
 }
 
 export function toYargsOptions(data: Question): Options {
   const choices = getChoicesFromQTNodeQuestion(data);
-  if (choices && choices.length > 0 && data.default === undefined) {
-    data.default = choices[0];
-  }
+  // if (choices && choices.length > 0 && data.default === undefined) {
```
chagong, line 93: why comment out this?

Alive-Fish: because there the default value will be changed to lower case and it will cause an error.

PR #1091, opened 2021-05-18
fix: add app id to local debug telemetry

swatDong:
Add teams app id as `debug-appid` to local debug related telemetries:
- `debug-precheck` event records local teams app id if exist.
- `debug-start` event records local teams app id for local debug, and remote teams app id for remote launch.


swatDong:
Close this to create another instead of the forked repo.

PR #1092, opened 2021-05-18
fix: add app id to local debug telemetry

swatDong:
Add teams app id as `debug-appid` to local debug related telemetries:
- `debug-precheck` event records local teams app id if exist.
- `debug-start` event records local teams app id for local debug, and remote teams app id for remote launch.

### DIFF
--- a/packages/vscode-extension/src/debug/commonUtils.ts
+++ b/packages/vscode-extension/src/debug/commonUtils.ts
@@ -226,3 +226,17 @@ export function getTeamsAppTenantId(): string | undefined {
 
   return undefined;
 }
+
+export function getLocalTeamsAppId(): string | undefined {
+  if (ext.workspaceUri) {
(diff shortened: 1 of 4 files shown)


kuojianlu (commented):
Maybe [this line](https://github.com/OfficeDev/TeamsFx/blob/qidon/localdebugapptelemetry/packages/vscode-extension/src/handlers.ts#L647) (abort if port in use) should also be updated.

swatDong:
> Maybe [this line](https://github.com/OfficeDev/TeamsFx/blob/qidon/localdebugapptelemetry/packages/vscode-extension/src/handlers.ts#L647) (abort if port in use) should also be updated.

Good catch! Updated.

kuojianlu (approved):

kimizhu (approved):

LongOddCode (approved):

PR #1098, opened 2021-05-19
fix(bot): bot name conflicts

IvanJobs:
1. add specific error message for bot name conflicts, and, remind users to clear resources in previous azure account.
2. align user error principle: If users can possibly recover by themselves, then it should be a user error.
3. refine wording for some errors.
4. remove CommandExecutionError message from low-level Error object, instead, using the message made by bot plugin, to avoid long low-level Error.message.
5. Log message endpoint to be updated manually.

to fix bug: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/9959194

Explanations for the screenshot below:
Since solution and plugins both need to throw error, to reduce duplicated error. Solution will throw an overall error if help link is not set. Plugins will throw a specific error if help link is present.

image

### DIFF
--- a/packages/fx-core/src/plugins/resource/bot/azureOps.ts
+++ b/packages/fx-core/src/plugins/resource/bot/azureOps.ts
@@ -11,13 +11,12 @@ import {
   MessageEndpointUpdatingError,
   MissingSubscriptionRegistrationError,
   FreeServerFarmsQuotaError,
-  BotNameRegisteredError,
+  InvalidBotDataError,
 } from "./errors";
 import { CommonStrings, ConfigNames } from "./resources/strings";
 import * as utils from "./utils/common";
 import { default as axios } from "axios";
 import { ErrorMessagesForChecking } from "./constants";
-import { Messages } from "./resources/messages";
 
 export class AzureOperations {
   public static async CreateBotChannelRegistration(
@@ -40,11 +39,8 @@ export class AzureOperations {
     } catch (e) {
       if (e.code === "MissingSubscriptionRegistration") {
         throw new MissingSubscriptionRegistrationError();
-      } else if (
-        e.code === "InvalidBotData" &&
-        e.message.includes(Messages.BotNameAlreadyRegistered)
-      ) {
-        throw new BotNameRegisteredError();
+      } else if (e.code === "InvalidBotData") {
+        throw new InvalidBotDataError(e);
       } else {
         throw new ProvisionError(CommonStrings.BOT_CHANNEL_REGISTRATION, e);
       }
--- a/packages/fx-core/src/plugins/resource/bot/azureOps.ts
+++ b/packages/fx-core/src/plugins/resource/bot/azureOps.ts
@@ -11,11 +11,13 @@ import {
   MessageEndpointUpdatingError,
   MissingSubscriptionRegistrationError,
   FreeServerFarmsQuotaError,
+  BotNameRegisteredError,
 } from "./errors";
 import { CommonStrings, ConfigNames } from "./resources/strings";
 import * as utils from "./utils/common";
 import { default as axios } from "axios";
 import { ErrorMessagesForChecking } from "./constants";
+import { Messages } from "./resources/messages";
 
 export class AzureOperations {
   public static async CreateBotChannelRegistration(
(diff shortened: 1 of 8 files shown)


packages/fx-core/src/plugins/resource/bot/plugin.ts
```diff
@@ -401,6 +401,9 @@ export class TeamsBotImpl {
           `Before running this bot, please manually update bot's message endpoint(${this.config.provision.siteEndpoint}${CommonStrings.MESSAGE_ENDPOINT_SUFFIX}). Click 'Get Help' button for more details.`,
           Links.UPDATE_MESSAGE_ENDPOINT
         );
+        Logger.info(
+          `Please manually update bot's message endpoint(${this.config.provision.siteEndpoint}${CommonStrings.MESSAGE_ENDPOINT_SUFFIX}).`
```
eriolchan, line 405: Since this sentence is the same as above, I'd like to have a variable to for it to remove duplicate code.

packages/fx-core/src/plugins/resource/bot/resources/messages.ts
```diff
@@ -32,12 +32,12 @@ export class Messages {
     "Please click 'Get Help' button for more details.";
   public static readonly ClickIssueButtonToReportIssue =
     "Please click 'Report Issue' button to report the issue.";
-  public static readonly CommandFailWithMessage = (command: string, message: string): string =>
-    `Run '${command}' failed with message: ${message}`;
+  public static readonly CommandExecutionFailed = (command: string): string =>
+    `Run '${command}' failed.`;
```
eriolchan, line 36: What about "Failed to run ${}" ?

packages/fx-core/src/plugins/resource/bot/resources/messages.ts
```diff
@@ -32,12 +32,12 @@ export class Messages {
...
   public static readonly DoSthBeforeSth = (sth: string, beforeSth: string): string =>
     `Perform command '${sth}' before '${beforeSth}'.`;
-  public static readonly FailToCallAppStudio = (apiName: string): string =>
-    `Failed to execute '${apiName}'.`;
+  public static readonly FailToCallAppStudioForCheckingAADApp =
+    "Failed to call app studio's api to check aad app's existence.";
```
eriolchan, line 40: do we need to use "App Studio" which is aligned with official name? Same for "AAD application existence"

packages/fx-core/src/plugins/resource/bot/azureOps.ts
```diff
@@ -38,6 +40,11 @@ export class AzureOperations {
     } catch (e) {
       if (e.code === "MissingSubscriptionRegistration") {
         throw new MissingSubscriptionRegistrationError();
+      } else if (
+        e.code === "InvalidBotData" &&
+        e.message.includes(Messages.BotNameAlreadyRegistered)
```
eriolchan, line 45: What if the error message is changed from SDK?

packages/fx-core/src/plugins/resource/bot/plugin.ts
IvanJobs: sure.

packages/fx-core/src/plugins/resource/bot/resources/messages.ts
IvanJobs: Ok for me.

packages/fx-core/src/plugins/resource/bot/resources/messages.ts
IvanJobs: Okay for me.

packages/fx-core/src/plugins/resource/bot/plugin.ts
IvanJobs: updated.

eriolchan (approved):

packages/fx-core/src/plugins/resource/bot/errors.ts
```diff
@@ -237,3 +237,15 @@ export class FreeServerFarmsQuotaError extends PluginError {
...
 }
+
+export class InvalidBotDataError extends PluginError {
+  constructor(innerError: Error) {
+    super(
+      ErrorType.User,
+      ErrorNames.INVALID_BOT_DATA_ERROR,
+      innerError.message,
+      [Messages.DeleteExistingBotChannelRegistration, Messages.DeleteBotAfterAzureAccountSwitching],
+      innerError
```
YitongFeng-git, line 248: Where to set its code = 'InvalidBotData'

packages/fx-core/src/plugins/resource/bot/resources/messages.ts
```diff
@@ -107,4 +109,10 @@ export class Messages {
   public static readonly RecreateTheProject = "Please recreate the project.";
   public static readonly CheckCommandOutputAndTryToFixIt =
     "Please check the command output and try to fix it.";
+  public static readonly DeleteExistingBotChannelRegistration =
+    "Please delete existing azure bot channel registrations.";
```
YitongFeng-git, line 113: It is proper none, better to use "Azure Bot Channels Registration", you can confirm with PM.
image

PR #1103, opened 2021-05-19
fix(core): do not readConfigs in progress of provision

jayzhang:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/9948007

### DIFF
--- a/packages/fx-core/src/core/index.ts
+++ b/packages/fx-core/src/core/index.ts
@@ -798,7 +798,7 @@ class CoreImpl implements Core {
 
   private async setSubscription(subscription: SubscriptionInfo | undefined) {
     if (subscription) {
-      await this.readConfigs();
+      // await this.readConfigs();
       this.configs
         .get(this.env!)!
         .get("solution")!
@@ -840,9 +840,10 @@ class CoreImpl implements Core {
   }
 
   public async readConfigs(): Promise<Result<null, FxError>> {
-    if (!fs.existsSync(`${this.ctx.root}/.${ConfigFolderName}`)) {
+    const confFolderPath = path.resolve(this.ctx.root, `.${ConfigFolderName}`);
+    if (!fs.existsSync(confFolderPath)) {
       this.ctx.logProvider?.warning(
-        `[Core] readConfigs() - folder does not exist: ${this.ctx.root}/.${ConfigFolderName}`
+        `[Core] readConfigs() - folder does not exist: ${confFolderPath}`
       );
       return ok(null);
     }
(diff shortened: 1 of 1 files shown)


packages/fx-core/src/core/index.ts
```diff
@@ -798,7 +798,7 @@ class CoreImpl implements Core {
 
   private async setSubscription(subscription: SubscriptionInfo | undefined) {
     if (subscription) {
-      await this.readConfigs();
+      // await this.readConfigs();
```
jayzhang, line 801: do not readConfigs() in the progress of provision

LongOddCode (approved):

packages/fx-core/src/core/index.ts
```diff
@@ -885,30 +893,44 @@ class CoreImpl implements Core {
...
-        const solutionConfig = entry[1];
-        const configJson = mapToJson(solutionConfig);
-        const filePath = `${this.ctx.root}/.${ConfigFolderName}/env.${envName}.json`;
-        const localDataPath = `${this.ctx.root}/.${ConfigFolderName}/${envName}.userdata`;
-        const localData = sperateSecretData(configJson);
-        const content = JSON.stringify(configJson, null, 4);
-        await fs.writeFile(filePath, content);
-        await fs.writeFile(localDataPath, serializeDict(localData));
+    let res: Result<null, FxError> = ok(null);
+    for (let i = 0; i < 5; ++i) {
```
1yefuwang1, line 904: Why are we doing this loop 5 times.

packages/fx-core/src/core/index.ts
```diff
@@ -885,30 +893,44 @@ class CoreImpl implements Core {
...
+          this.ctx.logProvider?.debug(
+            `[Core] writeConfigs()#########file:${jsonFilePath}, content:${content}`
+          );
+        }
+        //write settings
+        await this.writeSettings(this.ctx.root, this.ctx.projectSettings);
+        break;
+      } catch (e) {
+        res = err(error.WriteFileError(e));
+        sleep(10);
```
1yefuwang1, line 930: Why do we need to sleep here.

packages/fx-core/src/core/index.ts
jayzhang: It's guarantee to make sure config is written successfully

packages/fx-core/src/core/index.ts
jayzhang: To avoid concurrent write.

1yefuwang1: Besides，I think await is needed here

1yefuwang1 (approved):

MSFT-yiz (approved):

PR #1105, opened 2021-05-19
fix: toast message if user doesn't provision for spfx project before publish

nliu-ms:
image


jayzhang (approved):

MSFT-yiz (approved):

MSFT-yiz:
@timngmsft FYI. error message update.

PR #1107, opened 2021-05-19
fix(solution): return error when publish or deploy not-provisioned project

jayzhang:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/9968980


Alive-Fish:
Should it check in dev?

PR #1121, opened 2021-05-20
fix: login refresh token expire

xiaolang124:
1. Fix login expire bug
2. Change getStatus silent get token behavior


packages/vscode-extension/src/commonlib/codeFlowLogin.ts
```diff
@@ -235,8 +235,12 @@ export class CodeFlowLogin {
                   error.message
                 )
             );
-            const accessToken = await this.login();
-            return accessToken;
+            await this.logout();
```
swatDong, line 238: Will this trigger `statusChanged`? Or, how to ensure the expiration triggers correct status changed callback.

xiaolang124: No, codeFlow workFlow does not know about notify logic. Notify logic is handled by loginProvider. So now statusChange is handled by code after ```codeFlowInstance.getToken()```.

swatDong: I see, make sense.

swatDong (approved):

PR #1142, opened 2021-05-21
fix(apim): install swagger-jsdoc in version 6

XiaofuHuang:
The swagger-jsdoc pre-release version 7.0.0-rc.6 does not work. Change the document and recommend user to install version 6.

https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/9986060


kimizhu:
I don't think it is a blocking issue even it is for document change only, so let's delay the change towards dev branch instead.

PR #1146, opened 2021-05-22
hotfix: skip join view when local debug

swatDong:
Add `&webjoin=true` to launch url to **skip** the selector page:
image

### DIFF
--- a/packages/fx-core/src/plugins/resource/localdebug/launch.ts
+++ b/packages/fx-core/src/plugins/resource/localdebug/launch.ts
@@ -10,46 +10,6 @@ export function generateConfigurations(
   includeBackend: boolean,
   includeBot: boolean
 ): Record<string, unknown>[] {
-  /* No attach until CLI ready
-    let launchConfigurations: object[] = [
-        {
-            name: "Attach to Frontend (Chrome)",
-            type: LaunchBrowser.chrome,
-            request: "launch",
-            url: "https://teams.microsoft.com/_#/l/app/${input:teamsAppId}?installAppPackage=true",
-            presentation: {
-                group: "partial",
-                order: 2
-            }
-        },
-        {
-            name: "Attach to Frontend (Edge)",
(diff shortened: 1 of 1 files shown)


qinezh (approved):

kimizhu (approved):

zhenjiao-ms (commented):
There are a few other places that have this URL. Could you please check if we needn't change them. They are:
https://github.com/OfficeDev/TeamsFx/blob/fca8bd1751d7c51e5652b432b0ac3bfc90f9bd46/packages/fx-core/src/plugins/resource/localdebug/launch.ts#L19 and L29
https://github.com/OfficeDev/TeamsFx/blob/fca8bd1751d7c51e5652b432b0ac3bfc90f9bd46/packages/vscode-extension/src/debug/teamsfxTaskProvider.ts#L242

zhenjiao-ms:
Another comment: if we are going to have 2.0.1, the version number in https://github.com/OfficeDev/TeamsFx/blob/release/20210514/packages/vscode-extension/CHANGELOG.md needs to be updated.

swatDong:
> There are a few other places that have this URL. Could you please check if we needn't change them. They are:
> https://github.com/OfficeDev/TeamsFx/blob/fca8bd1751d7c51e5652b432b0ac3bfc90f9bd46/packages/fx-core/src/plugins/resource/localdebug/launch.ts#L19
> 
> and L29
> https://github.com/OfficeDev/TeamsFx/blob/fca8bd1751d7c51e5652b432b0ac3bfc90f9bd46/packages/vscode-extension/src/debug/teamsfxTaskProvider.ts#L242

The first one is code comment. I've removed that to not confuse others.
The second one is for codespace only, no need to update.


BTW, version number needs to be updated in both changelog and package.json (or anywhere else?). Better to update by who do the release.

zhenjiao-ms (approved):

PR #1147, opened 2021-05-22
chore: update version in changelog for hotfix

swatDong:
Update from `2.0.0` to `2.0.1`

### DIFF
--- a/packages/vscode-extension/CHANGELOG.md
+++ b/packages/vscode-extension/CHANGELOG.md
@@ -1,11 +1,9 @@
 # Changelog
 
 ## 2.0.1 - May 25 2021
-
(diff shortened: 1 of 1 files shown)


zhenjiao-ms (commented):
@swatDong I guess it is because you are using windows and Zhidi are using Mac, the line breaks are different. Is it okay?

swatDong:
> @swatDong I guess it is because you are using windows and Zhidi are using Mac, the line breaks are different. Is it okay?

Probably due to OS diff. I've changed to only version number diff.

zhenjiao-ms (approved):

kimizhu (approved):

PR #1154, opened 2021-05-24
feat: add action to publish vsix file to vscode marketplace

LongOddCode:


.github/workflows/vscode-marketplace.yml
```diff
@@ -0,0 +1,37 @@
...
+          timeout_minutes: 10
+          max_attempts: 10
+          retry_on: error
+          shell: bash
+          command: |
+            wget ${{ github.event.inputs.URL }} 
+
+      - name: release to VSCode marketplace
+        run: |
+          vsce publish --pat ${{ secrets.VSCE_PAT }} --packagePath *.vsix
```
qinezh, line 37: If an issue occurs during the step **Download VISX files**, it will retry to download it again. Then there will be two downloaded vsix packages:
1. ms-teams-vscode-extension-2.0.1.vsix
2. ms-teams-vscode-extension-2.0.1.vsix.1

`*.vsix` in the **release to VSCode marketplace** step will point to the first package, however, the right package should be the second one (downloaded by the second round) since the first package is broken.

Suggestions:
1. We need to reset the status (clean up the broken downloaded files) before retrying it again.
2. It would be better to use the built-in retry parameter of `wget` to take care of this case.

qinezh (approved):

packages/sdk/src/packageMetadata.ts
```diff
@@ -2,4 +2,4 @@
 // Licensed under the MIT license.
 
 export const name = "@microsoft/teamsfx";
-export const version = "0.1.2";
+export const version = "0.1.3";
```
blackchoey, line 5: in the future, we also need to auto update the version here and run test case again when lerna bumps the version automatically.

blackchoey (approved):

jayzhang (approved):

MSFT-yiz (approved):

packages/sdk/src/packageMetadata.ts
zhenjiao-ms, line 5: What is the relationship between this version and the package publish?

.github/workflows/vscode-marketplace.yml
```diff
@@ -0,0 +1,31 @@
...
+        run: |
+          npm install vsce -g
+
+      - name: Download VISX fils
+        run: |
+          wget --tries=5 --waitretry=5 ${{ github.event.inputs.URL }} 
+
+      - name: release to VSCode marketplace
+        run: |
+          vsce publish --pat ${{ secrets.VSCE_PAT }} --packagePath *.vsix
```
zhenjiao-ms, line 31: Have we tried this pipeline?

packages/sdk/src/packageMetadata.ts
tecton: Hi @zhenjiao-ms, this version should be aligned with SDK version in `packages/sdk/package.json`. So this change is to prevent unit test failure in CI because SDK version has already been updated by `lerna publish`. We still have difficulties reading version directly from package.json and manually maintain the consistency in PR. We'll automate this step soon.

zhenjiao-ms (approved):

PR #1164, opened 2021-05-24
fix(apim): add retry to graph api get operation

XiaofuHuang:
Add graph api get operation retry to e2e test case.


Alive-Fish (approved):

packages/fx-core/src/plugins/resource/apim/services/apimService.ts
```diff
@@ -459,7 +459,7 @@ export class ApimService {
         operation.displayName,
         resourceType.displayName
       );
-      this.logger?.info(LogMessages.operationFailed(operation, resourceType, resourceId));
+      this.logger?.warning(LogMessages.operationFailed(operation, resourceType, resourceId));
```
dooriya, line 462: >logger?.warning [](http://example.com/codeflow?start=11&length=15)

Should it be `logger?.error` since this is an operation error?

packages/cli/tests/commonlib/apimValidator.ts
```diff
@@ -362,3 +399,22 @@ class Config {
...
+  let result: T = await fn(executionIndex);
+  while (executionIndex <= 20 && condition(result)) {
+    await delay(executionIndex * 1000);
+    result = await fn(executionIndex);
+    ++executionIndex;
+  }
+  return result;
+}
+
+function delay(ms: number): Promise<void> {
```
dooriya, line 417: >ms: number [](http://example.com/codeflow?start=15&length=10)

Return `return Promise.resolve();` if `ms < 0`

packages/cli/tests/commonlib/apimValidator.ts
```diff
@@ -362,3 +399,22 @@ class Config {
     return this.config[this.apimPlugin]["apiDocumentPath"];
   }
 }
+
+async function retry<T>(
```
dooriya, line 403: >async function retry<T> [](http://example.com/codeflow?start=0&length=23)

As we discussed before, we need to consilidate the `retry` API for all code components in the future.

packages/cli/tests/commonlib/apimValidator.ts
```diff
@@ -362,3 +399,22 @@ class Config {
...
+  fn: (retries: number) => Promise<T>,
+  condition: (result: T) => boolean
+): Promise<T> {
+  let executionIndex = 1;
+  let result: T = await fn(executionIndex);
+  while (executionIndex <= 20 && condition(result)) {
```
dooriya, line 409: >executionIndex <= 20 [](http://example.com/codeflow?start=9&length=20)

Magic number in the retry numbers.

packages/fx-core/src/plugins/resource/apim/services/apimService.ts
XiaofuHuang: This log is inside the retry handler. If the operation failed after retry, another error log will be printed.

packages/cli/tests/commonlib/apimValidator.ts
XiaofuHuang: Moved it to default value.

packages/cli/tests/commonlib/apimValidator.ts
XiaofuHuang: Yes, we need to put the retry handler into a common place for all plugins to use.
The retry handler used in the test is different from the retry handler used in the plugin. 
One is to retry to get the data until it is refreshed or timeout. The other is to retry the operation if the error is thrown.

packages/fx-core/src/plugins/resource/apim/services/apimService.ts
dooriya: I see. Thanks Xiaofu for the explaination.

XiaofuHuang: Got it, change the log level back to warning

dooriya (approved):

PR #1166, opened 2021-05-24
fix(bot): invalidBot by npm install

IvanJobs:
to fix bug: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/9978406

### DIFF
--- a/packages/fx-core/src/plugins/resource/bot/languageStrategy.ts
+++ b/packages/fx-core/src/plugins/resource/bot/languageStrategy.ts
@@ -101,7 +101,7 @@ export class LanguageStrategy {
     if (programmingLanguage === ProgrammingLanguage.JavaScript) {
       try {
         // fail to npm install @microsoft/teamsfx on azure web app, so pack it locally.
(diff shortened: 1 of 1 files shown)


packages/fx-core/src/plugins/resource/bot/languageStrategy.ts
```diff
@@ -101,7 +101,7 @@ export class LanguageStrategy {
     if (programmingLanguage === ProgrammingLanguage.JavaScript) {
       try {
         // fail to npm install @microsoft/teamsfx on azure web app, so pack it locally.
```
hund030, line 103: Also remove the comment?

IvanJobs: Thanks for the comment, and I believe some comments should be added here since bot used remote build mode for deploy, and original meaning of this comment doesn't change.  I prefer to keep this.

YitongFeng-git (approved):

PR #1169, opened 2021-05-24
chore(sdk): fix bug in pipeline

tecton:


LongOddCode (approved):

.github/workflows/sdk-api-doc.yml
```diff
@@ -50,7 +50,7 @@ jobs:
           # See the `pathspec`-documentation for git
           # - https://git-scm.com/docs/git-add#Documentation/git-add.txt-ltpathspecgt82308203
           # - https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-aiddefpathspecapathspec
-          file_pattern: docs/sdk/*.md,packages/sdk/src
+          file_pattern: docs/sdk/*.md packages/sdk/src/packageMetadata.ts
```
blackchoey, line 53: why we need to commit packageMetadata.ts?

tecton: The packageMetadata.ts will be updated to the same version in package.json during `npm run setup`. This is to automate the version synchronization between them and prevent unit test break in CI.

blackchoey (approved):

PR #1170, opened 2021-05-24
chore: add ut for app studio plugin

nliu-ms:


packages/fx-core/src/plugins/resource/appstudio/appStudio.ts
```diff
@@ -52,39 +52,6 @@ export namespace AppStudioClient {
     }
   }
 
-  /**
```
nliu-ms, line 55: Remove this method, currently we're using the implementation in solution/appstudio.

jayzhang (approved):

PR #1172, opened 2021-05-24
test(env-checker): use vscode-test to run env-checker test cases

a1exwang:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/9769204/

Some bugs are not reproducible with standalone mocha tests. Use vscode-test to run the env-checker test cases in vscode context. This is more realistic to the actual user environment than launching the tests with mocha.

- add test script and test runner script for vscode-test
- add an entry in the extension's `launch.json` to support debugging the test cases
- modify GitHub action config to support vscode-test
- add macOS 11 as an test environment in schedule test

** tradeoff: using vscode-test will slightly decrease the performance. (e.g. Windows test 4min -> 5min)

## Test

- Schedule test passed: https://github.com/OfficeDev/TeamsFx/actions/runs/873636003
- Revert the commit that fixes the "ENOENT" bug and manually run the test in GitHub action, the original bug can be reproduced now 
https://github.com/OfficeDev/TeamsFx/runs/2661893898?check_suite_focus=true#step:11:463

### DIFF
--- a/.github/workflows/env-checker-ci-pr.yml
+++ b/.github/workflows/env-checker-ci-pr.yml
@@ -35,7 +35,7 @@ jobs:
       ## Details about this issue: https://github.com/npm/cli/wiki/%22cb()-never-called%3F--I'm-having-the-same-problem!%22
       ### tl;dr: This error is not one thing, but a category of errors. It means "something broke and we didn't have a way to catch it". We will always need a lot of detail to reproduce an error like this, or we cannot ever fix it. Every instance is unique, and your cb() never called is nothing like any other.
       matrix:
-        os: [windows-latest, macos-latest, ubuntu-latest]
+        os: [windows-latest, macos-11, ubuntu-latest]
         node-version: [14]
         func-version: [none]
         dotnet-version: [none]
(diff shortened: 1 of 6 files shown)


.github/workflows/env-checker-ci-pr.yml
```diff
@@ -49,12 +49,12 @@ jobs:
           fetch-depth: 0
 
       - name: Uninstalling .NET on Windows
-        if: ${{ matrix.os == 'windows-latest' }}
+        if: ${{ runner.os == 'Windows' }}
```
meifans, line 52: I wonder why you don't use latest version after all the vscode test use latest version.

1openwindow (approved):

.github/workflows/env-checker-ci-pr.yml
a1exwang: These are version check statements. The original test state use "windows-latest" which depends on the version "latest". Now we have both macos-11 and macos-latest, use "runner.os == 'macOS'" can match both of them.

meifans (approved):

qinezh (approved):

PR #1175, opened 2021-05-25
refactor(frontend): make frontend plugin stateless

hund030:
1. Make frontend plugin stateless.
2. Remove useless codes and comments.
3. Sync config to context, make it easier to add new persistent configs.

### DIFF
--- a/packages/fx-core/src/plugins/resource/frontend/configs.ts
+++ b/packages/fx-core/src/plugins/resource/frontend/configs.ts
@@ -12,9 +12,7 @@ import {
 import { Utils } from "./utils";
 
 export class FrontendConfig {
-  appName: string;
   subscriptionId: string;
-  resourceNameSuffix: string;
   resourceGroupName: string;
   location: string;
   storageName: string;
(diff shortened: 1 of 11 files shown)


packages/fx-core/src/plugins/resource/frontend/configs.ts
```diff
@@ -79,21 +74,38 @@ export class FrontendConfig {
...
       location,
-      resourceNameSuffix,
       storageName,
       credentials
     );
   }
 
+  public syncToPluginContext(ctx: PluginContext): void {
+    Object.entries(this)
+      .filter((kv) => FrontendConfig.persistentConfigList.find((x: string) => x === kv[0]))
```
henzhang-ms, line 87: Nit: could be `FrontendConfig.persistentConfigList.find(kv[0])`

packages/fx-core/src/plugins/resource/frontend/configs.ts
```diff
@@ -79,21 +74,38 @@ export class FrontendConfig {
...
+      .forEach((kv) => {
+        if (kv[1]) {
+          FrontendConfig.setConfigIfNotExists(ctx, kv[0], kv[1]);
+        }
+      });
+  }
+
+  private static persistentConfigList = Object.entries(FrontendConfigInfo).map((kv) => kv[1]);
```
henzhang-ms, line 95: Nit: could be `Object.values(FrontendConfigInfo)`.

henzhang-ms (approved):

packages/fx-core/src/plugins/resource/frontend/configs.ts
hund030: Updated.

packages/fx-core/src/plugins/resource/frontend/configs.ts
hund030: seems `find` only consume a callback, use includes instead.

PR #1183, opened 2021-05-25
fix(extension): UI CSS bugs

tecton:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/9935756/


packages/vscode-extension/src/controls/QuickStart.scss
```diff
@@ -40,8 +40,13 @@
     margin-left: 12px;
 
     h2 {
+      font-family: 'SF Pro Text';
```
HuihuiWu-Microsoft, line 43: I don't think this is available on Windows. VSC has its own default settings for font family we should adopt that

HuihuiWu-Microsoft:
From the bug, the button text should be vertically centered. maybe we should add some bottom padding for that?

tecton:
> From the bug, the button text should be vertically centered. maybe we should add some bottom padding for that?

Which one are you mentioning? If it's the "next" button, it's vertically centered by setting the same height.
image

HuihuiWu-Microsoft (approved):

PR #1186, opened 2021-05-25
chore: fix e2e tests yml

Alive-Fish:


.github/workflows/e2e-test.yml
```diff
@@ -47,7 +47,14 @@ jobs:
       - name: Setup project
         run: |
           npm run setup
-          npm link
+
+      - name: Package CLI and install
+        run: |
+          npm run package
+          npm pack
+          mv *.tgz teamsfx-cli.tgz
```
LongOddCode, line 55: will npm install -g *.tgz work?

LongOddCode (approved):

.github/workflows/e2e-test.yml
Alive-Fish: Yes

PR #1189, opened 2021-05-26
feat(local-debug): switch to active npm install task automatically

qinezh:
Currently there might be multiple npm-install tasks running in parallel. This PR is to switch the terminal to show other npm-install tasks when the current npm-install task is completed.

Please note, the terminal will be switched even the current npm-install task is failed (completed without exit code 0). This would be addressed in another separated PR (still under investigation).

### DIFF
--- a/packages/vscode-extension/src/debug/teamsfxTaskHandler.ts
+++ b/packages/vscode-extension/src/debug/teamsfxTaskHandler.ts
@@ -21,6 +21,15 @@ const allRunningTeamsfxTasks: Map<IRunningTeamsfxTask, number> = new Map<
   number
 >();
 const allRunningDebugSessions: Set<string> = new Set<string>();
+const activeNpmInstallTasks = new Set<string>();
+
+function isNpmInstallTask(task: vscode.Task): boolean {
+  if (task) {
+    return task.name.trim().toLocaleLowerCase().endsWith("npm install");
+  }
+
+  return false;
+}
 
 function isTeamsfxTask(task: vscode.Task): boolean {
   // teamsfx: xxx start / xxx watch
@@ -56,6 +65,8 @@ function onDidStartTaskProcessHandler(event: vscode.TaskProcessStartEvent): void
         { source: task.source, name: task.name, scope: task.scope },
         event.processId
       );
+    } else if (isNpmInstallTask(task)) {
+      activeNpmInstallTasks.add(task.name);
     }
   }
 }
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/debug/teamsfxTaskHandler.ts
```diff
@@ -64,6 +75,20 @@ function onDidEndTaskProcessHandler(event: vscode.TaskProcessEndEvent): void {
...
   if (task.scope !== undefined && isTeamsfxTask(task)) {
     allRunningTeamsfxTasks.delete({ source: task.source, name: task.name, scope: task.scope });
+  } else if (isNpmInstallTask(task)) {
+    activeNpmInstallTasks.delete(task.name);
+    // when the task in the active terminal is ended.
+    // TODO: only when the task is ended successfully.
+    if (vscode.window.activeTerminal?.name === task.name) {
+      for (const hiddenTaskName of activeNpmInstallTasks) {
+        // show the first hiden terminal.
+        const hiddenTerminal = vscode.window.terminals.find(t => t.name === hiddenTaskName);
```
swatDong, line 85: will `vscode.window.terminals` be undefined? or at least it's an empty array?

packages/vscode-extension/src/debug/teamsfxTaskHandler.ts
```diff
@@ -64,6 +75,20 @@ function onDidEndTaskProcessHandler(event: vscode.TaskProcessEndEvent): void {
   const task = event.execution.task;
   if (task.scope !== undefined && isTeamsfxTask(task)) {
     allRunningTeamsfxTasks.delete({ source: task.source, name: task.name, scope: task.scope });
+  } else if (isNpmInstallTask(task)) {
+    activeNpmInstallTasks.delete(task.name);
+    // when the task in the active terminal is ended.
+    // TODO: only when the task is ended successfully.
+    if (vscode.window.activeTerminal?.name === task.name) {
+      for (const hiddenTaskName of activeNpmInstallTasks) {
```
swatDong, line 83: If a hidden task is not found in ternimals, should it be removed from active task list?

packages/vscode-extension/src/debug/teamsfxTaskHandler.ts
qinezh: It will be an empty array if there's no opened terminals.

packages/vscode-extension/src/debug/teamsfxTaskHandler.ts
qinezh: it can be safely kept in the active task list as it's a real hidden npm-install task, and it won't break the logic here.

swatDong: But it may impact the next run. The active task list is never cleaned until user closes vscode...

swatDong (approved):

packages/vscode-extension/src/debug/teamsfxTaskHandler.ts
qinezh: Offline discussed that the active task list might be impact only when the task.json is modified by user manually.

a1exwang (approved):

kuojianlu (approved):

PR #1190, opened 2021-05-26
feat(frontend): add teams app id field for telemetry

hund030:


chagong (approved):

henzhang-ms (approved):

packages/fx-core/src/plugins/resource/frontend/utils/telemetry-helper.ts
```diff
@@ -12,6 +18,11 @@ export class telemetryHelper {
     measurements: { [key: string]: number } = {}
   ): void {
     properties[TelemetryKey.Component] = FrontendPluginInfo.PluginName;
+    properties[TelemetryKey.AppId] =
```
eriolchan, line 21: Since there are duplicated code, is it possible to extract common logic as a method of 
```
static fillBasicTelemetry(properties, ctx) {
  properties[TelemetryKey.Component] = FrontendPluginInfo.PluginName;
  properties[TelemetryKey.AppId] =  (ctx.configOfOtherPlugins.get(DependentPluginInfo.SolutionPluginName)        ?.get(DependentPluginInfo.RemoteTeamsAppId) as string) ?? "";
}
```

PR #1196, opened 2021-05-26
fix(tab): fix issues for tab template

hund030:
1. The environment description should either be removed or be dynamic based on the actual environment (Azure or local). Remove here for there is already a message indicates the environment at the top of welcome page.
image

2. Remove the broken scripts. The cli command can only be run at the root folder, instead of ./tabs/ folder.

### DIFF
--- a/templates/tab/js/default/public/index.html
+++ b/templates/tab/js/default/public/index.html
@@ -1,7 +1,7 @@
 <html lang="en">
   <head>
     <meta charset="utf-8" />
-    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <link rel="icon" href="%PUBLIC_URL%/favicon.ico" />
     <link
       rel="stylesheet"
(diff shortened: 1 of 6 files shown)


templates/tab/js/default/public/index.html
```diff
@@ -1,7 +1,7 @@
 <html lang="en">
   <head>
     <meta charset="utf-8" />
-    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
```
zhenyasav, line 4: aHA! not exactly. but good catch!  https://stackoverflow.com/questions/19506028/do-you-need-to-close-meta-and-link-tags-in-html

zhenyasav (approved):
LGTM, one nitpick fyi :)

templates/tab/js/default/public/index.html
hund030: Thanks for comment! As it dose no harm and for consistency, let's remain the slash.

PR #1211, opened 2021-05-26
fix: filter login log

xiaolang124:
1. Only output login error log


packages/vscode-extension/src/commonlib/appStudioLogin.ts
```diff
@@ -38,7 +38,9 @@ const config = {
     loggerOptions: {
       // @ts-ignore
       loggerCallback(loglevel, message, containsPii) {
-        VsCodeLogInstance.info(message);
+        if (loglevel<=LogLevel.Error) {
+          VsCodeLogInstance.info(message);
```
swatDong, line 42: always call `info` ?

xiaolang124: Good point, changed to error.

swatDong (approved):

PR #1220, opened 2021-05-27
chore(bot): add checkpoint between app studio call and azure call

IvanJobs:
to reduce risks:
1. remove the save config logic after exceptions.
2. add checkpoint between app studio call and azure call

### DIFF
--- a/packages/fx-core/src/plugins/resource/bot/plugin.ts
+++ b/packages/fx-core/src/plugins/resource/bot/plugin.ts
@@ -701,6 +701,17 @@ export class TeamsBotImpl {
         appStudioToken!,
         aadDisplayName
       );
+
+      if (!this.config.scaffold.botId) {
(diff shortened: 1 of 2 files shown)


packages/fx-core/src/plugins/resource/bot/plugin.ts
```diff
@@ -745,18 +756,6 @@ export class TeamsBotImpl {
     );
     Logger.info(Messages.SuccessfullyProvisionedMsTeamsChannel);
 
-    if (!this.config.scaffold.botId) {
-      this.config.scaffold.botId = botAuthCreds.clientId;
```
YitongFeng-git, line 749: why remove these？

IvanJobs: actually, them were moved up to where these configs generated.
The final goal is to save the new configs into conext.

eriolchan (approved):

PR #1224, opened 2021-05-27
test(simpleauth): wait for url to jump to authentication url with code

wenytang-ms:


blackchoey (approved):

packages/simpleauth/src/TeamsFxSimpleAuth.Tests/Helpers/Utilities.cs
```diff
@@ -101,18 +102,25 @@ private static void ClickElementWithRetry(IWebDriver driver, By elementLocator,
...
+            do
             {
-                throw new Exception($"Failed to get authorization code from redirect url: {redirectedUrl.AbsoluteUri}");
-            }
+                if(!string.IsNullOrEmpty(authorizationCode))
+                {
+                    driver.Close();
+                    return authorizationCode;
+                }
+                Thread.Sleep(200 * ++retryTime);
```
SLdragon, line 117: How about using await Task.delay(200) instead of Thread.Sleep? Since Thread.Sleep is blocking

wenytang-ms: this is a static method, not a task, so I think thread.sleep is better.

PR #1227, opened 2021-05-27
fix(aad): print detailed message in telemetry

KennethBWSong:
feat:
1. print detailed error in telemetry:
image

### DIFF
--- a/packages/fx-core/src/plugins/resource/aad/errors.ts
+++ b/packages/fx-core/src/plugins/resource/aad/errors.ts
@@ -15,8 +15,8 @@ export interface AadError {
 
 export const GetAppError: AadError = {
   name: "AadGetAppError",
-  message: (objectId: string) =>
-    `Failed to get app info with Object ID: ${objectId}. ` +
+  message: () =>
+    `Failed to get app info with current Object Id in env.default.json. ` +
     "Please make sure object id is valid, " +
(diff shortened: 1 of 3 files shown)


xzf0587 (approved):

packages/fx-core/src/plugins/resource/aad/errors.ts
```diff
@@ -15,8 +15,8 @@ export interface AadError {
 
 export const GetAppError: AadError = {
   name: "AadGetAppError",
-  message: (objectId: string) =>
-    `Failed to get app info with Object ID: ${objectId}. ` +
+  message: () =>
```
wenytang-ms, line 18: message has to be a function?

SLdragon (approved):

packages/fx-core/src/plugins/resource/aad/errors.ts
dilin-MS: If message is a function, please follow function naming convention such as `getMessage()`.
Or, make it a property accessors:
```
export 
const GetAppError = {
  name: "AadGetAppError",
  message: ()=> {
      return "Failed to get app info with current Object Id in env.default.json. Please make sure object id is valid, " + `or delete 'objectId' under ${Plugins.pluginNameComplex} in env.default.json and try again.`;
  }
}
```

adashen (approved):

PR #1241, opened 2021-05-31
fix: stdout/stderr TTY

Alive-Fish:
related bugs:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/9893250
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/9893368

description:
1. use `logProvider` instead of `console` in CLI.
2. check whether the `stdout` and `stderr` are `TTY`, if yes, add colors to the logs, otherwise, not add colors.

image

### DIFF
--- a/packages/cli/src/cmds/resource.ts
+++ b/packages/cli/src/cmds/resource.ts
@@ -286,7 +286,7 @@ export class ResourceShowFunction extends YargsCommand {
     const pluginName = "fx-resource-function";
     if (result.isOk()) {
       if (pluginName in result.value) {
-        await CLILogProvider.necessaryLog(LogLevel.Info, result.value[pluginName], true);
+        CLILogProvider.necessaryLog(LogLevel.Info, result.value[pluginName], true);
       }
       return ok(null);
     } else {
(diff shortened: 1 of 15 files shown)


xiaolang124 (approved):

packages/cli/src/cmds/resource.ts
```diff
@@ -339,11 +340,11 @@ export class ResourceList extends YargsCommand {
     pluginNameMap.set("fx-resource-function", "azure-function");
 
     if (result.isOk()) {
-      pluginNameMap.forEach((pluginAlias: string, pluginName: string) => {
+      for (const [pluginAlias, pluginName] of pluginNameMap) {
         if (pluginName in result.value) {
-          console.log(pluginAlias);
+          await CLILogProvider.necessaryLog(LogLevel.Info, pluginAlias, true);
```
a1exwang, line 345: It seems the behavior is changed here. I don't know whether this is what you expect.

Originally this is `console.log()`, but the new one is `console.log(s.white())`.
When the user's default color is not white, these strings will be white.

Alive-Fish: Yeah, the colors of logs will be defined more clearly after. Now the white logs should be the main color theme in the CLI.

chagong (approved):

a1exwang (approved):

PR #1242, opened 2021-05-31
feat(local-debug): show message once react starts before browser started

kuojianlu:
show information message once react starts before browser started, telling user to wait for the browser to start.

https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/9712275

Preview:
image

### DIFF
--- a/packages/vscode-extension/src/resources/Strings.json
+++ b/packages/vscode-extension/src/resources/Strings.json
@@ -107,7 +107,8 @@
     },
     "localDebug": {
       "portAlreadyInUse": "Port: %s is already in use. Close this port and try again.",
-      "portsAlreadyInUse": "Ports: %s are already in use. Close these ports and try again."
+      "portsAlreadyInUse": "Ports: %s are already in use. Close these ports and try again.",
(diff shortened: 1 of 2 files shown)


packages/vscode-extension/src/resources/Strings.json
```diff
@@ -107,7 +107,8 @@
     },
     "localDebug": {
       "portAlreadyInUse": "Port: %s is already in use. Close this port and try again.",
-      "portsAlreadyInUse": "Ports: %s are already in use. Close these ports and try again."
+      "portsAlreadyInUse": "Ports: %s are already in use. Close these ports and try again.",
+      "waitForBrowserStarted": "Please wait while the toolkit is starting your browser..."
```
dilin-MS, line 111: Why the verb `start` is past tense? What about `waitForBrowserToStart`?

kuojianlu: Nice catch, updated.

kuojianlu:
Close this PR since the solution is not that good: the information message will be folded into message center after about 20 seconds; it is also a little confusing.

PR #1243, opened 2021-05-31
fix: vscodeignore

jayzhang:


packages/vscode-extension/.vscodeignore
```diff
@@ -3,10 +3,6 @@
 src/**
 .gitignore
 .yarnrc
-**/tsconfig.json
-**/.eslintrc.json
-**/*.map
```
1openwindow, line 8: what is this change for?

1openwindow: how about exclude ignore template file.

1openwindow: how about just ignore template folder?

1openwindow (approved):

PR #1244, opened 2021-05-31
fix: update telemetry for app studio plugin

nliu-ms:
- update telemetry for app studio plugin

### DIFF
--- a/packages/fx-core/src/plugins/resource/appstudio/utils/telemetry.ts
+++ b/packages/fx-core/src/plugins/resource/appstudio/utils/telemetry.ts
@@ -3,6 +3,7 @@
 
 import { PluginContext, SystemError, UserError } from "@microsoft/teamsfx-api";
 import { Constants } from "./../constants";
+import { GLOBAL_CONFIG, REMOTE_TEAMS_APP_ID } from "../../../solution/fx-solution/constants";
 
 export enum TelemetryPropertyKey {
   component = "component",
@@ -13,6 +14,7 @@ export enum TelemetryPropertyKey {
   updateExistingApp = "update",
   success = "success",
   appId = "appid",
+  buildOnly = "build",
 }
 
 enum TelemetryPropertyValue {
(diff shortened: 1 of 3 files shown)


packages/fx-core/src/plugins/resource/appstudio/utils/telemetry.ts
```diff
@@ -10,11 +10,14 @@ export enum TelemetryPropertyKey {
   errorCode = "error-code",
   errorMessage = "error-message",
   validationResult = "validation-result",
+  success = "success",
 }
 
 enum TelemetryPropertyValue {
   UserError = "user",
   SystemError = "system",
+  success = "yes",
```
dilin-MS, line 19: Why the value is not `success` and `fail`?

nliu-ms: It's aligned with other plugin telemetries

chagong: @dilin-MS The property name is "success", it's like asking if the operation is succeeded, so the answer should be "yes" or "no".

packages/fx-core/src/plugins/resource/appstudio/utils/telemetry.ts
```diff
@@ -39,6 +44,9 @@ export class TelemetryUtils {
       properties = {};
     }
     properties[TelemetryPropertyKey.component] = Constants.PLUGIN_NAME;
+    if (this.ctx.app.id) {
```
chagong, line 47: ctx.app.id is not "remote teams app id", it the id in manifest file, and its value could be a placeholder like "{appid}". we opened a bug for that before. So please make sure you get appid from env config file.

chagong (approved):

jayzhang (approved):

packages/fx-core/src/plugins/resource/appstudio/utils/telemetry.ts
dilin-MS: What about `isSuccessful` as property name and `true` / `false` as possible value so as to increase code readability?

PR #1261, opened 2021-06-01
chore(template): prevent too many sdk logs by setting Error log level

tecton:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/9912111

### DIFF
--- a/packages/sdk/src/util/logger.ts
+++ b/packages/sdk/src/util/logger.ts
@@ -80,7 +80,7 @@ export function getLogLevel(): LogLevel {
 }
 
 class InternalLogger {
-  public level: LogLevel = LogLevel.Error;
+  public level: LogLevel = LogLevel.Info;
 
   public customLogger: Logger | undefined;
   public customLogFunction: LogFunction | undefined;
--- a/packages/sdk/src/util/logger.ts
+++ b/packages/sdk/src/util/logger.ts
@@ -80,7 +80,7 @@ export function getLogLevel(): LogLevel {
 }
 
 class InternalLogger {
-  public level: LogLevel = LogLevel.Info;
+  public level: LogLevel = LogLevel.Error;
 
   public customLogger: Logger | undefined;
   public customLogFunction: LogFunction | undefined;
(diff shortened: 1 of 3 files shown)


templates/tab/js/default/src/components/sample/lib/useTeamsFx.js
```diff
@@ -14,6 +14,7 @@ export function useTeamsFx() {
   const [result] = useTeams({});
   const { error, loading } = useData(async () => {
     if (!initialized) {
+      setLogLevel(LogLevel.Error);
```
hund030, line 17: What's the default log level? Would it be better if we set default log level to error and explicitly set log level to debug or info for development environment? So that the client code could be: 
```
if (DEBUG_MODE) {
    setLogLevel(LogLevel.Debug);
}
```
Which makes more sense for me.

tecton: Thanks, updated.

hund030 (approved):

packages/sdk/src/util/logger.ts
```diff
@@ -80,7 +80,7 @@ export function getLogLevel(): LogLevel {
 }
 
 class InternalLogger {
-  public level: LogLevel = LogLevel.Info;
+  public level: LogLevel = LogLevel.Error;
```
blackchoey, line 83: This is a breaking change to developers. For the bug, seems we don't need to change default log level.
Can we check the default log level of other SDK (like MSAL) and then have some discussion around it before change it?

tecton: Sure I agree. A related bug is https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/9921340

hund030: Agree. We may need more discussion and investigation before making changes.

hund030 (approved):
LGFM

SLdragon (approved):

blackchoey (approved):

wenytang-ms (approved):

PR #1266, opened 2021-06-02
Zhaofeng/fix telemetry

xzf0587:
1. add detail error message and appid property in telemetry.
2. add skip-adding-user property in sql post-provision telemetry.
3. update some systemError to userError.

### DIFF
--- a/packages/fx-core/src/plugins/resource/identity/index.ts
+++ b/packages/fx-core/src/plugins/resource/identity/index.ts
@@ -122,7 +122,7 @@ export class IdentityPlugin implements Plugin {
       ctx.logProvider?.error(
         ErrorMessage.IdentityProvisionError.message(this.config.identity) + `:${_error.message}`
       );
-      const error = ResultFactory.SystemError(
+      const error = ResultFactory.UserError(
         ErrorMessage.IdentityProvisionError.name,
         ErrorMessage.IdentityProvisionError.message(this.config.identity),
         _error
--- a/packages/fx-core/src/plugins/resource/identity/index.ts
+++ b/packages/fx-core/src/plugins/resource/identity/index.ts
@@ -63,7 +63,11 @@ export class IdentityPlugin implements Plugin {
     } catch (error) {
       const errorCode = error.source + "." + error.name;
       const errorType = error instanceof SystemError ? Telemetry.systemError : Telemetry.userError;
-      TelemetryUtils.sendErrorEvent(Telemetry.stage.provision, errorCode, errorType, error.message);
+      let errorMessage = error.message;
+      if (error.innerError) {
+        errorMessage += ` Detailed error: ${error.innerError.message}.`;
+      }
+      TelemetryUtils.sendErrorEvent(Telemetry.stage.provision, errorCode, errorType, errorMessage);
       return err(error);
     }
 
--- a/packages/fx-core/src/plugins/resource/identity/utils/telemetryUtil.ts
+++ b/packages/fx-core/src/plugins/resource/identity/utils/telemetryUtil.ts
@@ -1,5 +1,5 @@
 import { PluginContext } from "@microsoft/teamsfx-api";
-import { Telemetry } from "../constants";
+import { Constants, Telemetry } from "../constants";
 
 export class TelemetryUtils {
   static ctx: PluginContext;
@@ -11,8 +11,8 @@ export class TelemetryUtils {
   public static sendEvent(
     eventName: string,
     success?: boolean,
-    properties?: { [key: string]: string },
-    measurements?: { [key: string]: number }
+    properties?: { [key: string]: string; },
+    measurements?: { [key: string]: number; }
   ) {
     if (!properties) {
       properties = {};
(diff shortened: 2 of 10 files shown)


packages/fx-core/src/plugins/resource/identity/utils/telemetryUtil.ts
```diff
@@ -42,14 +39,17 @@ export class TelemetryUtils {
...
     TelemetryUtils.ctx.telemetryReporter?.sendTelemetryErrorEvent(
       eventName,
       properties,
       measurements
     );
   }
+
+  private static addProperties(properties: { [key: string]: string; }) {
+    properties[Telemetry.properties.component] = Telemetry.componentName;
+    const appId = this.ctx.configOfOtherPlugins.get(Constants.solution)?.get(Constants.remoteTeamsAppId) as string;
```
KennethBWSong, line 52: If remote teams app id is not available, here will be "undefined as string", which I remember will throw error?

packages/fx-core/src/plugins/resource/sql/utils/telemetryUtils.ts
```diff
@@ -42,14 +39,17 @@ export class TelemetryUtils {
...
-    }
+    this.addProperties(properties);
     TelemetryUtils.ctx.telemetryReporter?.sendTelemetryErrorEvent(
       eventName,
       properties,
       measurements
     );
   }
+
+  private static addProperties(properties: { [key: string]: string; }) {
```
SLdragon, line 50: Seems there are duplicate code between sql telemetryUtils and identity telemetryUtils, is this by design?

xzf0587: Yes, currently we don't share code across plugins;

packages/fx-core/src/plugins/resource/identity/utils/telemetryUtil.ts
xzf0587: I debugged it locally, and it will not trigger an error.

packages/fx-core/src/plugins/resource/identity/index.ts
```diff
@@ -63,7 +63,11 @@ export class IdentityPlugin implements Plugin {
     } catch (error) {
       const errorCode = error.source + "." + error.name;
       const errorType = error instanceof SystemError ? Telemetry.systemError : Telemetry.userError;
-      TelemetryUtils.sendErrorEvent(Telemetry.stage.provision, errorCode, errorType, error.message);
+      let errorMessage = error.message;
+      if (error.innerError) {
```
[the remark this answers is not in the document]
xzf0587: The duplicate codes are in sql and identity plugin separately. Currently they may not be refactored.

KennethBWSong (approved):

wenytang-ms (approved):

packages/fx-core/src/plugins/resource/identity/utils/telemetryUtil.ts
adashen: node env should throw error.....

packages/fx-core/src/plugins/resource/sql/plugin.ts
```diff
@@ -289,7 +292,10 @@ export class SqlPluginImpl {
 
     await managementClient.deleteLocalFirewallRule();
 
-    TelemetryUtils.sendEvent(Telemetry.stage.postProvision, true);
+    TelemetryUtils.sendEvent(Telemetry.stage.postProvision,
+      true,
+      { [Telemetry.properties.skipAddingUser]: this.config.skipAddingUser ? Telemetry.resultYes : Telemetry.resultNo }
```
blackchoey, line 297: Maybe create new constants for the skipAddingUser property? Though the string we're going to set is same, but `Telemetry.resultYes` represents different meaning.

xzf0587: I thought yes and no are the general keywords for telemetry to search and filter. Maybe I can change the name for the constants to make the meaning clear.

packages/fx-core/src/plugins/resource/identity/utils/telemetryUtil.ts
xzf0587: The value type of config map is like string | number | undefined, I tested it and it was safe. But I didn't find docs to support this casting. I will fix it.

packages/fx-core/src/plugins/resource/identity/utils/telemetryUtil.ts
```diff
@@ -32,24 +29,31 @@ export class TelemetryUtils {
...
+    if (appId) {
```
adashen, line 53: Do we need the following lines here? in 52 it already as string.... So we will provide empty string instead of undefined?

KennethBWSong: We found that it will not raise error here when undefined as string, but we can not find any official doc on it. So we decide to add logic to know whether the value is undefined and assign empty string when undefined, which will have the same result in query.

adashen: If that is the case line 52 don't need as string right?

KennethBWSong: Yeah, from my side I think we don't need as string in line 52

xzf0587: It is fixed, thanks.

adashen (approved):

PR #1270, opened 2021-06-02
fix(aad, simpleauth): add app id in telemetry

KennethBWSong:
fix: add app id in telemetry:
image

### DIFF
--- a/packages/fx-core/src/plugins/resource/aad/utils/telemetry.ts
+++ b/packages/fx-core/src/plugins/resource/aad/utils/telemetry.ts
@@ -59,6 +59,7 @@ export class TelemetryUtils {
 
   private static addAppIdInProperty(properties:{ [key: string]: string }, ctx: PluginContext): void {
     const appId = ctx.configOfOtherPlugins.get(Plugins.solution)?.get(ConfigKeysOfOtherPlugin.remoteTeamsAppId);
+    
     if (appId) {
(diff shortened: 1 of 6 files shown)


xzf0587 (approved):

formulahendry (approved):

packages/fx-core/src/plugins/resource/aad/utils/telemetry.ts
```diff
@@ -54,4 +56,9 @@ export class TelemetryUtils {
       "error-message": errorMessage,
     };
   };
+
+  private static addAppIdInProperty(properties:{ [key: string]: string }, ctx: PluginContext): void {
+    const appId = ctx.configOfOtherPlugins.get(Plugins.solution)?.get(ConfigKeysOfOtherPlugin.remoteTeamsAppId);
+    properties[Telemetry.appId] = appId as string;
```
adashen, line 62: undefined as string may throw exception?

KennethBWSong: Here it will not throw exception in my test. And also confirmed with zhaofeng in his pr.

adashen: Better no to do like this. It is not best practice I think.

adashen (approved):

PR #1284, opened 2021-06-07
refactor(function): normalize suffix when generate service name

henzhang-ms:
1. Normalize suffix when generating resource name. it is not a bug in current stage because suffix and resource name are generated in same lifecycle, user has no chance to inject. In case it becomes a bug someday, let's refine it.
2. Unify the max length of function app name and storage name because there is no benefit to separate them. Unification makes it simpler.

### DIFF
--- a/packages/fx-core/src/plugins/resource/function/constants.ts
+++ b/packages/fx-core/src/plugins/resource/function/constants.ts
@@ -165,8 +165,8 @@ export class DefaultProvisionConfigs {
 }
 
 export class AzureInfo {
-  public static readonly storageAccountNameLenMax: number = 24;
-  public static readonly functionAppNameLenMax: number = 32;
+  public static readonly resourceNameLenMax: number = 24;
+  public static readonly suffixLenMax: number = 12;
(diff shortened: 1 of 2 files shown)


hund030 (approved):

packages/fx-core/src/plugins/resource/function/constants.ts
```diff
@@ -165,8 +165,8 @@ export class DefaultProvisionConfigs {
 }
 
 export class AzureInfo {
-  public static readonly storageAccountNameLenMax: number = 24;
-  public static readonly functionAppNameLenMax: number = 32;
+  public static readonly resourceNameLenMax: number = 24;
+  public static readonly suffixLenMax: number = 12;
```
hund030, line 169: Why 12?

henzhang-ms: no reason, just pick a number that is half of resourceNameLenMax. I think the value only has two requirements:
1. no more than half of the full length, because suffix string is not human-readable.
2. no less than our current suffix length (6).

PR #1296, opened 2021-06-08
fix: cannot provision with the wrong existing sub && validate preset answers

Alive-Fish:
Also returns APIM undefined preset answers.


packages/cli/src/activate.ts
```diff
@@ -32,7 +32,9 @@ export default async function activate(rootPath?: string): Promise<Result<Core,
   if (rootPath) {
     const subscription = await getSubscriptionIdFromEnvFile(rootPath);
     if (subscription) {
-      await AzureAccountManager.setSubscription(subscription);
+      try {
+        await AzureAccountManager.setSubscription(subscription);
```
Alive-Fish, line 36: This means if it throws an error when setting sub, CLI will do nothing and core will ask sub again.

chagong (approved):

PR #1300, opened 2021-06-08
feat: Use core getQuestions to get CLI help param

jiayi-ruan:
1. Initialize core and get questions for help param of all the stage at the begining.
2. Switch all the cmd(except init) builder to use params from core.
3. Reorganize all the runCommand() to be consistent.
4. Fix updatePresetAnswers(): use questions/params to set the answer instead of the yargs args.

Plugin Change:
1. Add APIM getQuestions() for CLI_HELP platform @XiaofuHuang 
2. Add SQL getQuestions() for CLI_HELP platform @xzf0587 
3. Change publish getQuestions().(remove build-or-publish for CLI platform) @nliu-ms 

Core Change:
1. Add subscription question node in provision stage. @jayzhang  @LongOddCode  @1yefuwang1

### DIFF
--- a/packages/cli/src/helpParamGenerator.ts
+++ b/packages/cli/src/helpParamGenerator.ts
@@ -1,0 +1,188 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+"use strict";
+
+import { Result, FxError, err, ok, Inputs, Tools, Stage, Platform, QTreeNode, Question, isAutoSkipSelect, SingleSelectQuestion, MultiSelectQuestion, OptionItem } from "@microsoft/teamsfx-api";
+
+import { FxCore } from "@microsoft/teamsfx-core";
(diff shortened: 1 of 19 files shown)


packages/cli/src/helpParamGenerator.ts
```diff
@@ -0,0 +1,188 @@
...
+  type: "folder",
+  name: "folder",
+  title: "Select root folder of the project",
+  default: "./",
+});
+
+export class HelpParamGenerator {
+  private static core: FxCore | undefined;
+  private static questionsMap: Map<string, QTreeNode> = new Map<string, QTreeNode>();
+  public static async activate(): Promise<Result<FxCore, FxError>> {
```
chagong, line 29: Can it be changed to sync function?

jiayi-ruan: seems it can. thx for the reminder, will commit that in the merged branch.

PR #1315, opened 2021-06-09
feat(solution): telemetry for capability and resource

1yefuwang1:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_sprints/taskboard/Teams%20Extensibility%20E2E%20Team/Microsoft%20Teams%20Extensibility/CY21-6.1?workitem=10087728


packages/fx-core/src/plugins/solution/fx-solution/constants.ts
```diff
@@ -295,3 +295,24 @@ export const TEAMS_APP_SHORT_NAME_MAX_LENGTH = 30;
...
+  AddResource = "add-resource",
+
+  AddCapabilityStart = "add-capability-start",
+  AddCapability = "add-capability",
+}
+
+export enum SolutionTelemetryProperty {
+  Component = "component",
+  Resources = "resources",
+  ResourcesToAdd = "resources-to-add",
```
chagong, line 313: feel no need for "ResourceToAdd" and "CapabilitiesToAdd", just use "Resources" and "Capabilities". People should know the meaning.

packages/fx-core/src/plugins/solution/fx-solution/solution.ts
```diff
@@ -376,6 +383,11 @@ export class TeamsAppSolution implements Solution {
         JSON.stringify(manifest, null, 4)
       );
       await fs.writeJSON(`${ctx.root}/permissions.json`, DEFAULT_PERMISSION_REQUEST, { spaces: 4 });
+      ctx.telemetryReporter?.sendTelemetryEvent(SolutionTelemetryEvent.Create, {
```
chagong, line 386: You have to send telemetry before every return. And if return result is ok, add "success": "yes" to property. If return result is err, add "success": "no", and "error-type", "error-code" and "error-message" properties.

packages/fx-core/src/plugins/solution/fx-solution/solution.ts
```diff
@@ -2262,9 +2279,18 @@ export class TeamsAppSolution implements Solution {
               notifications.join(",")), 
           false);
     }
+
+    ctx.telemetryReporter?.sendTelemetryEvent(SolutionTelemetryEvent.AddResource, {
```
chagong, line 2283: Same as above.

packages/fx-core/src/plugins/solution/fx-solution/solution.ts
```diff
@@ -2262,9 +2279,18 @@ export class TeamsAppSolution implements Solution {
...
+      [SolutionTelemetryProperty.Component]: SolutionTelemetryComponentName,
+      [SolutionTelemetryProperty.ResourcesToAdd]: addResourcesAnswer.join(";"),
+    });
     return ok(Void);
   }
+
   async executeAddCapability(func: Func, ctx: SolutionContext): Promise<Result<any, FxError>> {
+    ctx.telemetryReporter?.sendTelemetryEvent(SolutionTelemetryEvent.AddCapabilityStart, {
```
chagong, line 2291: Same as above.

packages/fx-core/src/plugins/solution/fx-solution/constants.ts
1yefuwang1: "Resources" alone could be ambiguous, because it could mean "resources that have already been added" and "resources to add"

chagong: that's why you need "success": "yes/no". I believe for "yes" means resources been added. 'no' means to be added.

chagong (approved):

jayzhang (approved):

PR #1324, opened 2021-06-09
fix(sdk): update trim-newlines to 3.0.1

tecton:


packages/sdk/src/packageMetadata.ts
```diff
@@ -2,4 +2,4 @@
 // Licensed under the MIT license.
 
 export const name = "@microsoft/teamsfx";
-export const version = "0.1.3";
+export const version = "0.2.0-alpha.0";
```
blackchoey, line 5: not blocking this PR: can we book sometime next sprint to improve the dev experience around this file?

blackchoey (approved):

packages/sdk/src/packageMetadata.ts
tecton: Yes, it should also be automatically updated in a daily CI job.

PR #1339, opened 2021-06-10
feat: support ansi color for logprovider info

nliu-ms:
1) Support color for LogProvider.info, others like error/warning will have default color.
2) Replace colors with chalk package.

E.g. CLI:
image

VSCode output channel (default color):
image

### DIFF
--- a/packages/cli/src/commonlib/log.ts
+++ b/packages/cli/src/commonlib/log.ts
@@ -3,13 +3,12 @@
 
 "use strict";
 
-import colors from "colors";
-
-import { LogLevel, LogProvider } from "@microsoft/teamsfx-api";
+import { LogLevel, LogProvider, Colors } from "@microsoft/teamsfx-api";
 
 import { CLILogLevel } from "../constants";
+import { getColorizedString } from "./../utils";
+import chalk from "chalk";
 
-colors.white.green.yellow.red;
 
 export class CLILogProvider implements LogProvider {
   private static instance: CLILogProvider;
@@ -44,7 +43,16 @@ export class CLILogProvider implements LogProvider {
     return this.log(LogLevel.Debug, message);
   }
 
-  info(message: string): Promise<boolean> {
+  info(message: Array<{content: string, color: Colors}>): Promise<boolean>;
+  
+  info(message: string): Promise<boolean>;
+
+  info(message: string | Array<{content: string, color: Colors}>): Promise<boolean> {
+    if (message instanceof Array) {
+      message = getColorizedString(message);
+    } else {
+      message = this.white(message);
+    }
     return this.log(LogLevel.Info, message);
   }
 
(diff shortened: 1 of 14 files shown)


chagong (approved):

packages/cli/src/utils.ts
```diff
@@ -225,3 +227,40 @@ export function getTeamsAppId(rootfolder: string | undefined): any {
 
   return undefined;
 }
+
+export function getColorizedString(message: Array<{content: string, color: Colors}>): string {
+  // Only TTY support ANSI color
+  if (process.stdout.isTTY) {
+    let colorizedMessage = "";
+    (message as Array<{content: string, color: Colors}>).map(function(item) {
```
a1exwang, line 235: 1. The type conversion is unnecessary since `message` is already `Array<{content: string, color: Colors}>`.
2. It seems what you want to achieve is to map the array and then join the string, so why not use `message.map(...).join()` to make this more readable just like what you did in the `else` branch? The current code is [an anti-pattern on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map#when_not_to_use_map)

packages/cli/src/commonlib/log.ts
```diff
@@ -62,35 +70,35 @@ export class CLILogProvider implements LogProvider {
 
   white(msg: string): string {
     if (process.stdout.isTTY) {
```
a1exwang, line 72: It seems `chalk` library has [auto-detection of ANSI escape sequence support](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map#when_not_to_use_map), so this check is redundant now.

a1exwang (approved):
ANSI color part looks good to me.

packages/cli/src/utils.ts
```diff
@@ -225,3 +227,28 @@ export function getTeamsAppId(rootfolder: string | undefined): any {
...
+      case Colors.BRIGHT_GREEN:
+        return chalk.greenBright(item.content);
+      case Colors.BRIGHT_RED:
+        return chalk.redBright(item.content);
+      case Colors.BRIGHT_YELLOW:
+        return chalk.yellowBright(item.content);
+      case Colors.BRIGHT_CYAN:
+        return chalk.cyanBright.underline(item.content);
+      default:
+        return ""
```
a1exwang, line 250: return `item.content` for unsupported colors?

jayzhang (approved):

KennethBWSong (approved):
approve for changes in aad plugin

xiaolang124 (approved):

swatDong (approved):

hund030 (approved):

IvanJobs (approved):

PR #1355, opened 2021-06-15
feat: add color for showMessage

nliu-ms:
image

### DIFF
--- a/packages/api/src/qm/ui.ts
+++ b/packages/api/src/qm/ui.ts
@@ -97,6 +97,10 @@ export interface UserInteraction {
     modal: boolean,
     ...items: string[]
   ): Promise<Result<string|undefined,FxError>>;
+  /**
+   * 
+   * @param message Only works for CLI info level, color for other log levels and VSCode window will be ignored
+   */
   showMessage(
     level: "info" | "warn" | "error",
     message: Array<{content: string, color: Colors}>,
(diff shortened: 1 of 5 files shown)


packages/vscode-extension/src/qm/vsc_ui.ts
```diff
@@ -493,27 +494,44 @@ export class VsCodeUI implements UserInteraction {
...
+  ): Promise<Result<string|undefined,FxError>>;
+
+  public async showMessage(
+    level: "info" | "warn" | "error",
+    message: Array<{content: string, color: Colors}>,
+    modal: boolean,
+    ...items: string[]
+  ): Promise<Result<string|undefined,FxError>>;
+
+  async showMessage(
```
HuihuiWu-Microsoft, line 511: Do we still need the above two functions?

chagong (approved):

packages/api/src/qm/ui.ts
```diff
@@ -96,6 +97,16 @@ export interface UserInteraction {
     modal: boolean,
     ...items: string[]
   ): Promise<Result<string|undefined,FxError>>;
+  /**
+   * 
+   * @param message Only INFO level supports message color, color for other log levels will be ignored
```
HuihuiWu-Microsoft, line 102: Color only works for CLI?

jayzhang (approved):

packages/vscode-extension/src/qm/vsc_ui.ts
nliu-ms: This is the way to overload function, as chagong suggested.
image

packages/api/src/qm/ui.ts
nliu-ms: comments updated.

HuihuiWu-Microsoft (approved):

PR #1360, opened 2021-06-16
fix: update componentId when download todo-spfx sample

nliu-ms:
When users in the same tenant download todo-list-SPFx sample, the same componentId will trigger some bug.


packages/cli/src/cmds/new.ts
```diff
@@ -234,6 +237,21 @@ class NewTemplete extends YargsCommand {
         })
     );
   }
+
+  private async downloadSampleHook(sampleId: string, sampleAppPath: string) {
+    // A temporary solution to avoid duplicate componentId
+    if (sampleId === "todo-list-SPFx") {
+      const originalId = "c314487b-f51c-474d-823e-a2c3ec82b1ff";
+      const componentId = uuid.v4();
+      const files = glob.sync(`${sampleAppPath}/**/*.json`, { nodir: true, dot: true });
```
HuihuiWu-Microsoft, line 246: I think better to use await + async

1openwindow (approved):

packages/cli/src/cmds/new.ts
```diff
@@ -234,6 +237,21 @@ class NewTemplete extends YargsCommand {
...
+      files.forEach(async function (file) {
```
Alive-Fish, line 247: This may cause an error when use foreach and async together.

jayzhang (approved):

Alive-Fish (approved):

PR #1365, opened 2021-06-16
feat(function): add telemetry for fallback

henzhang-ms:
add telemetry for fallback, we can use event `scaffold-fallback` to track it.

### DIFF
--- a/packages/fx-core/src/plugins/resource/function/utils/telemetry-helper.ts
+++ b/packages/fx-core/src/plugins/resource/function/utils/telemetry-helper.ts
@@ -7,6 +7,7 @@ import { CommonConstants, DependentPluginInfo, FunctionPluginInfo } from "../con
 import { FxResult } from "../result";
 import { FunctionEvent, TelemetryKey, TelemetryValue } from "../enums";
 import { DepsCheckerEvent } from "./depsChecker/common";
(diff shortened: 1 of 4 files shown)


hund030 (approved):

packages/fx-core/src/plugins/resource/function/utils/telemetry-helper.ts
```diff
@@ -75,4 +76,16 @@ export class TelemetryHelper {
...
+  static sendFallbackEvent(
+    e: FunctionPluginError = new UnknownFallbackError(),
+    properties: { [key: string]: string } = {},
+    measurements: { [key: string]: number } = {}
+  ) {
+    this.fillCommonProperty(properties);
+    properties[TelemetryKey.ErrorMessage] = e.message;
+    properties[TelemetryKey.ErrorCode] = e.code;
+
+    this.ctx?.telemetryReporter?.sendTelemetryEvent(FunctionEvent.scaffoldFallback, properties, measurements);
```
hund030, line 89: NIT: The event name doesn't match the method name, you send a specific `scaffoldFallback` event in a general `sendFallbackEvent` method.

henzhang-ms: fixed

PR #1371, opened 2021-06-16
refactor(function): rename telemetryHelper and make ctx static

henzhang-ms:
1. rename `telemetryHelper` to `TelemetryHelper`
2. make ctx static to avoid broadcast


a1exwang (approved):
The environment checker part looks good to me.

hund030 (approved):

PR #1376, opened 2021-06-18
feat: integrate tas to support a/b test

nliu-ms:


packages/vscode-extension/src/exp/index.ts
```diff
@@ -0,0 +1,43 @@
...
+    public setSharedProperty(name: string, value: string): void {
+        this.sharedProperties[name] = value;
+    }
+
+    public postEvent(eventName: string, props: Map<string, string>): void {
+        let properties = { ...this.sharedProperties };
+        props.forEach((value, key) => {
+            properties[key] = value;
+        });
+        ExtTelemetry.sendTelemetryEvent(eventName, properties);
```
chagong, line 25: does this mean Exp has a series of its own defined telemetries?

nliu-ms: Yes, here is an example:
image

chagong (approved):

packages/vscode-extension/src/exp/index.ts
```diff
@@ -0,0 +1,43 @@
...
+    }
+}
+
+let expService: IExperimentationService;
+
+export function getExpService(): IExperimentationService {
+    return expService
```
HuihuiWu-Microsoft, line 32: Semicolon

packages/vscode-extension/src/exp/index.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    constructor() {
+        this.sharedProperties = {};
+    }
+
+    public setSharedProperty(name: string, value: string): void {
+        this.sharedProperties[name] = value;
+    }
+
+    public postEvent(eventName: string, props: Map<string, string>): void {
```
HuihuiWu-Microsoft, line 20: Will Exp set shared properties and postEvent automatically?

packages/vscode-extension/src/exp/index.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export function getExpService(): IExperimentationService {
+    return expService
+}
+
+export async function initialize(context: vscode.ExtensionContext) {
+    expService = await getExperimentationServiceAsync(
+        pkg.name,
+        pkg.version,
+        TargetPopulation.Public,
```
HuihuiWu-Microsoft, line 39: Does this mean public users will be in the test?

packages/vscode-extension/src/exp/index.ts
nliu-ms: Yes, these two functions will be called automatically, as this telemetry object is passed to ExperimentationService.

packages/vscode-extension/src/exp/index.ts
nliu-ms: I think our target is testing changes for public population.

HuihuiWu-Microsoft (approved):

HuihuiWu-Microsoft (approved):

packages/vscode-extension/src/exp/index.ts
```diff
@@ -0,0 +1,44 @@
...
+            pkg.version,
+            TargetPopulation.Public,
+            new ExperimentationTelemetry(),
+            context.globalState
+        );
+    }
+
+    class ExperimentationTelemetry implements IExperimentationTelemetry {
+        public setSharedProperty(name: string, value: string): void {
+            sharedProperties[name] = value;
```
daytonellwanger, line 33: These shared properties need to be attached to all telemetry events your extension sends - it's how we can tie an event to a specific experiment group and create a scorecard. It looks like they're just being added to `sharedProperties` and never used?

daytonellwanger: Oh I see - they're being attached to all events sent via `postEvent`. They need to be attached to those events *and* all other events your extension sends.

nliu-ms: Got it, thanks!

packages/vscode-extension/src/exp/index.ts
```diff
@@ -0,0 +1,39 @@
...
+        public setSharedProperty(name: string, value: string): void {
+            ExtTelemetry.addSharedProperty(name, value);
+        }
+    
+        public postEvent(eventName: string, props: Map<string, string>): void {
+            let properties: {[key: string]: string} = {};
+            props.forEach((value, key) => {
+                properties[key] = value;
+            });
+            ExtTelemetry.sendTelemetryEvent(eventName, properties);
```
HuihuiWu-Microsoft, line 36: One question, the props here are not the shared properties that you attached  to vscode telemetry reporter?

swatDong (approved):

packages/vscode-extension/src/exp/index.ts
nliu-ms: here is the example of the props, other shared properties will be added by the reporter.
image

HuihuiWu-Microsoft (approved):

packages/vscode-extension/src/exp/index.ts
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import * as vscode from "vscode";
+import { IExperimentationService, IExperimentationTelemetry, TargetPopulation, getExperimentationServiceAsync } from "vscode-tas-client";
+import * as pkg from "./../../package.json";
+import { ExtTelemetry } from "./../telemetry/extTelemetry"
+
+export namespace exp {
```
1openwindow, line 9: what is this new folder for? what is it stand for?

PR #1382, opened 2021-06-21
feat(frontend): send scaffold fallback event

hund030:
Align to function plugin's changes.

### DIFF
--- a/packages/fx-core/src/plugins/resource/frontend/ops/scaffold.ts
+++ b/packages/fx-core/src/plugins/resource/frontend/ops/scaffold.ts
@@ -9,6 +9,7 @@ import fs from "fs-extra";
 import {
   FetchTemplateManifestError,
   FetchTemplatePackageError,
+  FrontendPluginError,
   InvalidTemplateManifestError,
   runWithErrorCatchAndThrow,
 } from "../resources/errors";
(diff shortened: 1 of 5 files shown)


henzhang-ms (approved):

packages/fx-core/src/plugins/resource/frontend/ops/scaffold.ts
```diff
@@ -94,7 +96,15 @@ export class FrontendScaffold {
       );
       return await FrontendScaffold.fetchZipFromUrl(templateUrl);
     } catch (e) {
+      Logger.debug(e.toString());
```
eriolchan, line 99: Is this logging only for debug and we need to remove it from PROD?

packages/fx-core/src/plugins/resource/frontend/ops/scaffold.ts
```diff
@@ -94,7 +96,15 @@ export class FrontendScaffold {
...
       Logger.warning(Messages.FailedFetchTemplate());
+
+      if (e instanceof FrontendPluginError) {
```
eriolchan, line 102: I think Heng will have some change for this part, and you can follow his change

packages/fx-core/src/plugins/resource/frontend/ops/scaffold.ts
hund030: As we synced offline, remain this.

PR #1384, opened 2021-06-21
feat(cli): add task wrapper for preview command

kuojianlu:
1. spawn shell command
2. no timeout support
3. no kill support

### DIFF
--- a/packages/cli/src/cmds/preview/task.ts
+++ b/packages/cli/src/cmds/preview/task.ts
@@ -1,4 +1,4 @@
-import { ChildProcess, spawn, SpawnOptions } from "child_process";
+import { spawn, SpawnOptions } from "child_process";
 
 interface TaskOptions {
     cwd?: string;
--- a/packages/cli/src/cmds/preview/task.ts
+++ b/packages/cli/src/cmds/preview/task.ts
@@ -1,10 +1,5 @@
 import { ChildProcess, spawn, SpawnOptions } from "child_process";
 
-function delay(ms: number) {
-    // tslint:disable-next-line no-string-based-set-timeout
-    return new Promise((resolve) => setTimeout(resolve, ms));
-}
-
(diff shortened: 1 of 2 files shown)


packages/cli/src/cmds/preview/task.ts
```diff
@@ -0,0 +1,87 @@
...
+            // TODO: log
+            this.stderr += data;
+        });
+    }
+
+    /**
+     * wait for the task to end
+     */
+    public async wait(): Promise<TaskResult> {
+        while (true) {
```
swatDong, line 46: Instead of `while(true)...delay`, can we leverage existing `exit` or `close` event? E.g.,

``` node
const task = spawn(...);
return new Promise<number>((resolve, reject) => {
    task.on('exit', code => resolve(code));
});
```

packages/cli/src/cmds/preview/task.ts
```diff
@@ -0,0 +1,87 @@
...
+            }
+            await delay(1000);
+        }
+    }
+
+    /**
+     * wait until stdout of the task matches the pattern or the task ends
+     */
+    public async waitFor(pattern: RegExp): Promise<TaskResult> {
+        while (true) {
```
swatDong, line 64: The same as above comment, leverage existing event and `resolve` instead of `while(true)...wait...`

packages/cli/src/cmds/preview/task.ts
kuojianlu: Great idea! Moved `spawn` to `wait` and updated the implementation.

packages/cli/src/cmds/preview/task.ts
kuojianlu: Updated.

swatDong (approved):

Alive-Fish (approved):
look good to me

PR #1385, opened 2021-06-21
fix: revert the unexpected merged code that get initial Azure account…

HuihuiWu-Microsoft:
… status


packages/vscode-extension/src/accountTree.ts
```diff
@@ -203,16 +203,6 @@ export async function registerAccountTreeHandler(): Promise<Result<Void, FxError
     return ok(null);
   };
 
-  let azureAccountLabel = "Sign in to Azure";
```
HuihuiWu-Microsoft, line 206: No need to get initial status since the setStatusMap will auto fire

1openwindow (approved):

PR #1386, opened 2021-06-21
refactor(function): explicit the error on sending scaffold telemetry

henzhang-ms:


hund030 (approved):

packages/fx-core/src/plugins/resource/function/resources/errors.ts
```diff
@@ -237,7 +237,7 @@ export class UploadZipError extends FunctionPluginError {
 
 export class UnknownFallbackError extends FunctionPluginError {
   constructor() {
-    super(ErrorType.User, "UnknownFallbackError", `Trigger fallback caused by unknown reason.`, []);
+    super(ErrorType.System, "UnknownFallbackError", `Trigger fallback caused by unknown reason.`, []);
```
hund030, line 240: No need to use ` for error message.

henzhang-ms: udpated

PR #1387, opened 2021-06-21
refactor: cleanup account tree telemetry codes

HuihuiWu-Microsoft:


packages/vscode-extension/src/accountTree.ts
```diff
@@ -2,33 +2,16 @@
...
-
-enum TelemetryEvent {
-  SelectSubscription = "select-subscription",
-}
-
-export enum AccountType {
-  M365 = "m365",
-  Azure = "azure",
-}
+import {
```
HuihuiWu-Microsoft, line 10: Use definitions in extension to avoid duplication since it's moved to extension now

1openwindow (approved):

PR #1388, opened 2021-06-21
fix: azure initialize status

xiaolang124:
1. Fix Azure notify status when first open Vs Code and not signed in.  (This is because Azure Account will be ready after toolkit extension starts.)

Bug: When open new VsCode and do not sign in Azure after treeView move from core to extension.
https://dev.azure.com/msazure/Microsoft%20Teams%20Extensibility/_workitems/edit/10192921/

### DIFF
--- a/packages/vscode-extension/src/commonlib/azureLogin.ts
+++ b/packages/vscode-extension/src/commonlib/azureLogin.ts
@@ -343,10 +343,10 @@ export class AzureAccountManager extends login implements AzureAccountProvider {
     AzureAccountManager.currentStatus = azureAccount.status;
     azureAccount.onStatusChanged(async (event) => {
       if(AzureAccountManager.currentStatus === "Initializing") {
-        AzureAccountManager.currentStatus = azureAccount.status;
+        AzureAccountManager.currentStatus = event;
         return;
       }
-      AzureAccountManager.currentStatus = azureAccount.status;
+      AzureAccountManager.currentStatus = event;
       if (event === loggedOut) {
         if (AzureAccountManager.statusChange !== undefined) {
           await AzureAccountManager.statusChange(signedOut, undefined, undefined);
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/commonlib/azureLogin.ts
```diff
@@ -339,6 +339,8 @@ export class AzureAccountManager extends login implements AzureAccountProvider {
   async addStatusChangeEvent() {
     const azureAccount: AzureAccount =
       vscode.extensions.getExtension<AzureAccount>("ms-vscode.azure-account")!.exports;
+    // wait for Azure Account extension initialize
+    await azureAccount.waitForSubscriptions();
```
swatDong, line 343: This may increase our toolkit activation time... If this is the only place to fix, try a safer way:

``` typescript
private string currentStatus;
...
async addStatusChangeEvent() {
    ...
    this.currentStatus = azureAccount.status;
    azureAccount.onStatusChanged(async (event) => {
        if (event === loggedOut && this.currentStatus !== initializing) {
            this.currentStatus = event;
            ...
        }
    });
}
```

xiaolang124: Good idea, changed to this behavior.

packages/vscode-extension/src/commonlib/azureLogin.ts
```diff
@@ -339,7 +340,13 @@ export class AzureAccountManager extends login implements AzureAccountProvider {
   async addStatusChangeEvent() {
     const azureAccount: AzureAccount =
       vscode.extensions.getExtension<AzureAccount>("ms-vscode.azure-account")!.exports;
+    AzureAccountManager.currentStatus = azureAccount.status;
     azureAccount.onStatusChanged(async (event) => {
+      if(AzureAccountManager.currentStatus === "Initializing") {
+        AzureAccountManager.currentStatus = azureAccount.status;
```
swatDong, line 346: AzureAccountManager.currentStatus = **event**

The same for line 349

xiaolang124: Fixed.

swatDong (approved):

PR #1394, opened 2021-06-22
refactor(sdk): remove packageMetadata.ts

tecton:
Package version is controlled by Lerna and it's not synchronized with packageMetadata.ts file. The version in log output is not correct. This PR removes the version in log output to fix the problem.


packages/sdk/src/util/logger.ts
```diff
@@ -116,7 +114,7 @@ class InternalLogger {
       return;
     }
     const timestamp = new Date().toUTCString();
-    const logHeader = `[${timestamp}] : ${name}@${version} : ${LogLevel[logLevel]} - `;
+    const logHeader = `[${timestamp}] : @microsoft/teamsfx : ${LogLevel[logLevel]} - `;
```
tecton, line 117: Use hard coded string here because package name is unlikely to change.

blackchoey (approved):
Not sure whether the `fix` conventional commit type is good for this PR. Maybe `refactor` is a better one?

wenytang-ms (approved):

PR #1397, opened 2021-06-22
feat(frontend): Add Azure error code to error message and send for telemetry

hund030:
There are many CreateStorageAccountError in telemetry but we don't know why. Add Azure error code to error message and the error message will be added to telemetry, which may help us know what's happening.

### DIFF
--- a/packages/fx-core/src/plugins/resource/frontend/resources/errors.ts
+++ b/packages/fx-core/src/plugins/resource/frontend/resources/errors.ts
@@ -146,11 +146,13 @@ export class StorageAccountAlreadyTakenError extends FrontendPluginError {
 }
 
 export class CreateStorageAccountError extends FrontendPluginError {
-  constructor(innerErrorCode: string) {
+  constructor(innerErrorCode?: string) {
     super(
       ErrorType.User,
       "CreateStorageAccountError",
-      `Failed to create Azure Storage Account, Azure error code: ${innerErrorCode}.`,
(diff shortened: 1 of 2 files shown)


henzhang-ms (approved):

packages/fx-core/src/plugins/resource/frontend/resources/errors.ts
```diff
@@ -146,10 +146,13 @@ export class StorageAccountAlreadyTakenError extends FrontendPluginError {
...
 export class CreateStorageAccountError extends FrontendPluginError {
-  constructor() {
-    super(ErrorType.User, "CreateStorageAccountError", "Failed to create Azure Storage Account.", [
-      tips.checkLog,
-    ]);
+  constructor(innerErrorCode: string) {
+    super(
+      ErrorType.User,
+      "CreateStorageAccountError",
+      `Failed to create Azure Storage Account, Azure error code: ${innerErrorCode}.`,
```
eriolchan, line 153: Is there any update from Chaoyi on the best practice to mask sensitive data from telemetry? If that's ready, maybe it would be helpful to output the inner exception into the telemetry.

hund030: This issue is assigned to Changyong now. He is requesting vscode-extension-telemetry to filter out user info in error stack. Seems it still takes some time. Let's put the error code to see if it helps first.

PR #1399, opened 2021-06-22
feat(cli): init preview command

kuojianlu:
image

### DIFF
--- a/packages/cli/src/cmds/preview/commonUtils.ts
+++ b/packages/cli/src/cmds/preview/commonUtils.ts
@@ -1,0 +1,17 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+"use strict";
+
+import * as path from "path";
+import * as fs from "fs-extra";
+import { ConfigFolderName } from "@microsoft/teamsfx-api";
+
+import * as constants from "./constants";
+
+export async function getActiveResourcePlugins(workspaceFolder: string): Promise<string[]> {
+    const settingsPath = path.join(workspaceFolder, `.${ConfigFolderName}`, constants.settingsFileName);
+    await fs.writeFile("temp.txt", settingsPath);
+    const settings = await fs.readJson(settingsPath);
+    return settings.solutionSettings.activeResourcePlugins;
+}
--- a/packages/cli/src/cmds/preview/commonUtils.ts
+++ b/packages/cli/src/cmds/preview/commonUtils.ts
@@ -11,7 +11,6 @@ import * as constants from "./constants";
 
 export async function getActiveResourcePlugins(workspaceFolder: string): Promise<string[]> {
     const settingsPath = path.join(workspaceFolder, `.${ConfigFolderName}`, constants.settingsFileName);
-    await fs.writeFile("temp.txt", settingsPath);
     const settings = await fs.readJson(settingsPath);
     return settings.solutionSettings.activeResourcePlugins;
 }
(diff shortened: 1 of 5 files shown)


packages/cli/src/cmds/preview/commonUtils.ts
```diff
@@ -0,0 +1,17 @@
...
+
+import * as path from "path";
+import * as fs from "fs-extra";
+import { ConfigFolderName } from "@microsoft/teamsfx-api";
+
+import * as constants from "./constants";
+
+export async function getActiveResourcePlugins(workspaceFolder: string): Promise<string[]> {
+    const settingsPath = path.join(workspaceFolder, `.${ConfigFolderName}`, constants.settingsFileName);
+    await fs.writeFile("temp.txt", settingsPath);
```
swatDong, line 14: Redundant code?

packages/cli/src/cmds/preview/preview.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    public async runCommand(args: { [argName: string]: boolean | string | string[] | undefined }): Promise<Result<null, FxError>> {
+        const workspaceFolder = path.resolve(args.folder as string);
+        if (!utils.isWorkspaceSupported(workspaceFolder)) {
+            return err(WorkspaceNotSupported(workspaceFolder));
+        }
+
+        CliTelemetry.setReporter(CliTelemetry.getReporter().withRootFolder(workspaceFolder));
+
+        if (args.local || (!args.local && !args.remote)) {
```
swatDong, line 50: what if both `local` and `remote` are set to true? better to handle such case as user error

packages/cli/src/cmds/preview/commonUtils.ts
kuojianlu: Forgot to remove this line for test. Good catch. Removed.

packages/cli/src/cmds/preview/preview.ts
kuojianlu: Current behavior regards it as `local`. I think command like `teamsfx preview --type local` is better than `teamsfx preview --local`. However, since we have made the decision, let me just handle the exclusion case.

kuojianlu: Added. Also updated the option description.

swatDong (approved):
The code logic looks good to me.
One more concern is - any way to disable this command for this release? Since the functionality is not added...

packages/cli/src/cmds/preview/preview.ts
```diff
@@ -0,0 +1,92 @@
...
+            description: "Select root folder of the project",
+            string: true,
+            default: "./",
+        });
+
+        return yargs.version(false);
+    }
+
+    public async runCommand(args: { [argName: string]: boolean | string | string[] | undefined }): Promise<Result<null, FxError>> {
+        if (args.local && args.remote) {
```
chagong, line 43: Should you send out telemetries for preview command?

kuojianlu: Nice reminder, thanks. All telemetries will be added after the implementation in one single PR. Will let you know then.

kuojianlu: I loop you in this PR mainly to review the preview command (see the picture above).

packages/cli/src/cmds/preview/constants.ts
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+"use strict";
+
+export const settingsFileName = "settings.json";
+
+export const FrontendHostingPluginName = "fx-resource-frontend-hosting";
```
chagong, line 8: Can we get these constants from core?

kuojianlu: Nice idea. But neither api nor core provides the method to get these information.

chagong (approved):

PR #1401, opened 2021-06-22
refactor(bot): merge bot-msgext template code automatically

YitongFeng-git:
### Description
Use bash + Mustache to generate Bot/ME/Bot-ME code from Mustache template

### Test
https://github.com/YitongFeng-git/TeamsFx/releases/tag/templates%406.30.2-yife

image


.github/scripts/merge.py
```diff
@@ -0,0 +1,62 @@
...
+        bot_fh = open(bot_file)
+        bot_content = bot_fh.readlines()
+        me_fh = open(me_file)
+        me_content = me_fh.readlines()
+
+        # merge pkg reference
+        for line in me_content:
+            if not line.startswith(("import", "const")):
+                break
+            if "botbuilder" not in line and line not in bot_content:
```
eriolchan, line 30: How do we handle this line in ME file? skip or merge the content of "TeamsActivityHandler, CardFactory" to Bot file?
```
const { TeamsActivityHandler, CardFactory } = require("botbuilder");
```

.github/scripts/template-zip-autogen.sh
```diff
@@ -1,6 +1,8 @@
 #!/usr/bin/env bash
 set -x
 
+python ./.github/scripts/merge.py
```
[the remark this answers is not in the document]
YitongFeng-git: Tested, it works

.github/scripts/merge.py
YitongFeng-git: Yes Bot's pkg include ME's. Use Bot's statement is ok

.github/scripts/template-zip-autogen.sh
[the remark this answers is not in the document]
henzhang-ms: better to use the path independent on cwd
https://stackoverflow.com/questions/242538/unix-shell-script-find-out-which-directory-the-script-file-resides

.github/scripts/merge.py
```diff
@@ -0,0 +1,88 @@
...
+def merge_pkg_reference(bot_lines, me_lines, merged_lines):
+    for line in me_lines:
+        if not line.startswith(("import", "const")):
+            break
+        if "botbuilder" not in line and line not in bot_lines:
+            merged_lines.insert(0, line)
+
+
+# merge functions in TeamsActivityHandler class
+def merge_functions_in_class(bot_lines, me_lines, merged_lines):
```
henzhang-ms, line 24: NIT: the solution looks like an ad-hoc string analyze, it highly depends on some implicit code constraints. In fact, my prefer solution is using a template engine (like jinja since you use python, or mustache which we are using now) to generate code based on parameters. So user who know the template engine can easily review your templates.

YitongFeng-git: Good suggestion, I will consider this solution

.github/scripts/template-zip-autogen.sh
```diff
@@ -33,6 +40,24 @@ for LANGUAGE in ${LANGUAGE_LIST[@]}; do
             exit -1
         fi
 
+        # Copy bot code to msgext-bot, except readme and images
+        rsync -avz --recursive --exclude "*.md" --exclude "*images/*" ./templates/bot/ ./templates/bot-msgext/
+        
+        # Generate code from Mustache templates
+        if [ ${SCOPE} == "bot" ]; then
```
henzhang-ms, line 47: NIT: put it outside of this loop, create a new loop for it.

```
BOT_FEATURES=(bot me)
for lan ...
  for feat ...
     
```

YitongFeng-git: ok

YitongFeng-git: I think put in the loop is ok, and can avoid extra for loop definition.

henzhang-ms (approved):
script part looks good to me

eriolchan (approved):

.github/scripts/template-zip-autogen.sh
```diff
@@ -12,6 +14,9 @@ TEMPLATE_LIST=(
     bot-msgext.default
 )
 
+# Copy bot code to msgext-bot, except readme and images
+rsync -az --recursive --exclude "*.md" --exclude "*images/*" ./templates/bot/ ./templates/bot-msgext/
```
IvanJobs, line 18: I may need more context about this.  
1. mustache templates stay in bot-msgext folder.
2. copy bot-msgext to bot folder.

how about the msgext folder?

Nit: In addition, what's the reason behind the exclude logic?

YitongFeng-git: discussed offline

PR #1409, opened 2021-06-23
fix(vsc): update notice file

1openwindow:


HuihuiWu-Microsoft:
Do we need to update the file each time 3rd party dependencies changed? If so, is that better if we integration in CI?

1openwindow:
> Do we need to update the file each time 3rd party dependencies changed? If so, is that better if we integration in CI?

This notice file is generated from Azure Devops pipeline. Didn't find a way to integration this so far.

HuihuiWu-Microsoft (approved):

PR #1411, opened 2021-06-23
fix(aad): add retry telemetry

KennethBWSong:
Will send telemetry every time in retry:
new key: method-name, retry-times
image

After discuss with Wei, will only send telemetry when successfully called rest api to see usually how many times will it take.


packages/fx-core/src/plugins/resource/aad/aadAppClient.ts
```diff
@@ -218,6 +229,7 @@ export class AadAppClient {
         if (retries === 0) {
           throw error;
         } else {
+          TelemetryUtils.sendEvent(stage, {[Telemetry.methodName]: fn.toString(), [Telemetry.retryTimes]: (10-retries).toString()});
```
adashen, line 232: not use magic number. we should have a param for retrynumber

packages/fx-core/src/plugins/resource/aad/aadAppClient.ts
```diff
@@ -218,6 +229,7 @@ export class AadAppClient {
         if (retries === 0) {
           throw error;
```
formulahendry, line 230: When `throw error`, need also add telemetry?

packages/fx-core/src/plugins/resource/aad/aadAppClient.ts
formulahendry: When the request succeed without retry, the retryTimes should be `0`.
So `10-retries` should be `10-retries-1`?

packages/fx-core/src/plugins/resource/aad/aadAppClient.ts
KennethBWSong: Here will cause fail, and telemetry will be sent in index.ts

packages/fx-core/src/plugins/resource/aad/aadAppClient.ts
KennethBWSong: If succeed, seems will not go into catch.

packages/fx-core/src/plugins/resource/aad/aadAppClient.ts
formulahendry: The telemetry sent in index.ts would not have info of retry times. I think we'd better to add the info as well? Otherwise, I may miss the telemetry of retry times is 10.

formulahendry: Oh, just see that we will send telemetry every time it is retrying. So we may not need to do that.

packages/fx-core/src/plugins/resource/aad/aadAppClient.ts
```diff
@@ -218,6 +229,7 @@ export class AadAppClient {
...
         } else {
+          TelemetryUtils.sendEvent(stage, {[Telemetry.methodName]: fn.toString(), [Telemetry.retryTimes]: (10-retries).toString()});
           await delay(5000);
```
formulahendry, line 233: `we will send telemetry every time it is retrying`. Just want to confirm whether the behavior is expected?
If we retry 7 times, then we will have 7 telemetry event.

blackchoey: Maybe we can only send 1 telemetry?

KennethBWSong: Actually adding this telemetry is for get how many times some user will wait. In some cases user will just close the session and the error telemetry will not be sent.

xzf0587: Just a reminder. If we send multiple events, maybe we also should consider how to utilize it and show it in Power BI.

packages/fx-core/src/plugins/resource/aad/plugin.ts
```diff
@@ -86,23 +86,23 @@ export class AadAppForTeamsImpl {
...
     if (!config.password) {
       await DialogUtils.progress?.next(ProgressDetail.CreateAadAppSecret);
-      await AadAppClient.createAadAppSecret(config);
+      await AadAppClient.createAadAppSecret(ctx, isLocalDebug ? Messages.EndLocalDebug.telemetry : Messages.EndProvision.telemetry, config);
       ctx.logProvider?.info(Messages.getLog(Messages.CreateAadAppPasswordSuccess));
     }
 
     await DialogUtils.progress?.next(ProgressDetail.UpdatePermission);
-    await AadAppClient.updateAadAppPermission(config.objectId as string, permissions);
+    await AadAppClient.updateAadAppPermission(ctx, isLocalDebug ? Messages.EndLocalDebug.telemetry : Messages.EndProvision.telemetry, config.objectId as string, permissions);
```
SLdragon, line 105: The same expression are in different place, can we use a variable for this?
```
isLocalDebug ? Messages.EndLocalDebug.telemetry : Messages.EndProvision.telemetry
```

KennethBWSong: Fixed. Thank you for pointing that out.

xzf0587 (approved):

packages/fx-core/src/plugins/resource/aad/aadAppClient.ts
blackchoey: Since we sent multiple error telemetries for one operation, the error rate will be **higher** than actual value (not sure whether we can workaround this issue via Kusto query). Can we confirm whether it's fine?
Another option is send a new event for the errors and exclude this new event in our report to avoid possible confusion on error report..

KennethBWSong: Here I send a telemetry event without success = no, thus I think it seems will not cause higher error rate. To get detailed info, we can do like this: count(retry = 1) = telemetry(retry = 1) - telemetry(retry=2). Where count() stands for the actual number of user retry 1 time, and telemetry() means the number of telemetry with specific retry times.

blackchoey (approved):

adashen (approved):

PR #1423, opened 2021-06-24
fix: regen lock files & install local deps when setup

LongOddCode:


packages/eslint-config/package.json
```diff
@@ -4,6 +4,7 @@
   "description": "Shared ESLint configuration",
   "private": true,
   "devDependencies": {
+    "@shared/prettier-config": "file:../prettier-config",
```
adashen, line 7: why do we need add this ref? seems circulate ref between eslint-confg and prettier-config

adashen (approved):

wenytang-ms (approved):

PR #1424, opened 2021-06-24
fix(local-debug): add error message for npm install tasks telemetry

kuojianlu:
**debug-npm-install-node-version**: "undefined" or string value
**debug-npm-install-npm-version**: "undefined" or string value
**debug-npm-install-error-message**: "undefined" or string value


packages/vscode-extension/src/debug/npmLogHandler.ts
```diff
@@ -0,0 +1,54 @@
...
+import * as fs from "fs-extra";
+import * as path from "path";
+
+function getNpmCachePath(): Promise<string> {
+  const command = spawn("npm config get cache", {
+    shell: true,
+  });
+  return new Promise((resolve, reject) => {
+    command.stdout.on("data", (data) => {
+      resolve(`${data.toString().trim()}`);
```
a1exwang, line 14: stdout.on("data") returns whenever the process write to stdout. This does not mean the process exits. You should append the result returned from stdout.on("data") to a buffer and use the buffer when command.on("exit") or command.on("error"). see [cputils code](https://github.com/OfficeDev/TeamsFx/blob/f841ea22ad6b66c6f8cebbbff8b7d12de6949c32/packages/vscode-extension/src/debug/depsChecker/cpUtils.ts#L64)

packages/vscode-extension/src/debug/npmLogHandler.ts
```diff
@@ -0,0 +1,54 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { spawn } from "child_process";
+import * as fs from "fs-extra";
+import * as path from "path";
+
+function getNpmCachePath(): Promise<string> {
```
swatDong, line 8: Use existing utilities, e.g., *depsChecker/cpUtils*

packages/vscode-extension/src/debug/npmLogHandler.ts
kuojianlu: In normal cases, `npm config get cache` only has one line output. I wonder when will this command have more than one line output.

packages/vscode-extension/src/debug/npmLogHandler.ts
```diff
@@ -0,0 +1,54 @@
...
+    shell: true,
+  });
+  return new Promise((resolve, reject) => {
+    command.stdout.on("data", (data) => {
+      resolve(`${data.toString().trim()}`);
+    })
+  });
+}
+
+async function getLatestNpmLogFile(npmLogPath: string): Promise<string | undefined> {
```
swatDong, line 19: Instead of always getting the latest one, is it possible to get via timestamp?
Is log file named the start timestamp or the end timestamp? Npm install may run several minutes and the latest one may not be the matched one.

packages/vscode-extension/src/debug/npmLogHandler.ts
```diff
@@ -0,0 +1,54 @@
...
+    return undefined;
+  }
+  const latestNpmLogFile = await getLatestNpmLogFile(path.join(npmCachePath, "_logs"));
+  if (latestNpmLogFile === undefined) {
+      return undefined;
+  }
+  const log = (await fs.readFile(latestNpmLogFile)).toString();
+  const cwdPattern = /\d+ verbose cwd (.*)/;
+  const cwdResult = log.match(cwdPattern);
+  // TODO: handle case sensitive path
```
swatDong, line 42: Seems already done. What else to do?

packages/vscode-extension/src/debug/npmLogHandler.ts
```diff
@@ -0,0 +1,54 @@
...
+  if (!cwdResult || cwdResult[1].trim().toLowerCase() !== cwd.toLowerCase()) {
+      return undefined;
+  }
+  const errorPattern = /\d+ error .*/g;
```
swatDong, line 46: Add following info as well:
- `/\d+ verbose node .*/g` : to check the node version
- `/\d+ verbose npm .*/g` : to check the npm version
- `/\d+ verbose exit .*/g` : to double check the exit code

packages/vscode-extension/src/debug/npmLogHandler.ts
kuojianlu: Converting to lower case is for Windows, however, file path on Linux is case-sensitive, so this may cause error in some corner case.

packages/vscode-extension/src/debug/npmLogHandler.ts
a1exwang: This has nothing todo with how many lines it returns.

If the node process run this:
```
process.stdout.write("/home");
process.stdout.write("/alexwang");
process.stdout.write("/.npm");
```

You will get 3 `data` event.

Or more common cases, if node use `process.stdout.write("/home/alexwang/.npm")` to output but the underlying buffer of stdin is full when it writes only part of the string, node or the underlying libc will write the first part of the string to stdout, which will trigger you `data` event.

Anyway, I think we should not depend on the implementation details and you can reuse cpUtils to prevent this.

packages/vscode-extension/src/debug/npmLogHandler.ts
swatDong: To compare two paths, better to use built-in methods of *path*, instead of string ones.
E.g., `path.relative` or `path.parse` or `path.format`, etc.

packages/vscode-extension/src/debug/teamsfxTaskHandler.ts
```diff
@@ -90,18 +92,37 @@ function onDidStartTaskProcessHandler(event: vscode.TaskProcessStartEvent): void
...
   if (task.scope !== undefined && isTeamsfxTask(task)) {
     allRunningTeamsfxTasks.delete({ source: task.source, name: task.name, scope: task.scope });
   } else if (isNpmInstallTask(task)) {
     try {
-      ExtTelemetry.sendTelemetryEvent(TelemetryEvent.DebugNpmInstall, {
-        [TelemetryProperty.DebugNpmInstallName]: task.name,
-        [TelemetryProperty.DebugNpmInstallExitCode]: event.exitCode + "",
-      });
+      let npmInstallErrorMessage: string | undefined;
+      if (event.exitCode) {
```
a1exwang, line 104: maybe `event.exitCode !== 0` to handle cases like `event.exitCode` is undefined?

packages/vscode-extension/src/debug/npmLogHandler.ts
kuojianlu: The error log is generated when install fails, so the duration between when the log was generated and when to get the file is short.
Getting via timestamp is not that easy, since the timestamp is in `2021-06-24T02_55_46_184Z` format. We are not able to get the accurate timestamp when the error log was generated.

packages/vscode-extension/src/debug/teamsfxTaskHandler.ts
kuojianlu: `event.exitCode` being `undefined` means the npm install process is terminated before completion, so there will be no err log.

packages/vscode-extension/src/debug/npmLogHandler.ts
kuojianlu: Make sense.

packages/vscode-extension/src/debug/teamsfxTaskHandler.ts
swatDong: Maybe `if (event.exitCode && event.exitCode !== 0) {` , since we don't care 0 as well

packages/vscode-extension/src/debug/npmLogHandler.ts
swatDong: I see, so the log file name is the end timestamp? That makes sense.

packages/vscode-extension/src/debug/teamsfxTaskHandler.ts
kuojianlu: `if (event.exitCode)` implicitly includes `event.exitCode !== 0`.

packages/vscode-extension/src/debug/npmLogHandler.ts
kuojianlu: Updated to use `cpUtils.executeCommand`.

packages/vscode-extension/src/debug/npmLogHandler.ts
kuojianlu: Good idea. Updated to use `cpUtils.executeCommand`.

packages/vscode-extension/src/debug/npmLogHandler.ts
kuojianlu: Verified that the error log timestamp is when install fails.

packages/vscode-extension/src/debug/teamsfxTaskHandler.ts
```diff
@@ -90,18 +92,37 @@ function onDidStartTaskProcessHandler(event: vscode.TaskProcessStartEvent): void
...
 
   if (task.scope !== undefined && isTeamsfxTask(task)) {
     allRunningTeamsfxTasks.delete({ source: task.source, name: task.name, scope: task.scope });
   } else if (isNpmInstallTask(task)) {
     try {
-      ExtTelemetry.sendTelemetryEvent(TelemetryEvent.DebugNpmInstall, {
-        [TelemetryProperty.DebugNpmInstallName]: task.name,
-        [TelemetryProperty.DebugNpmInstallExitCode]: event.exitCode + "",
-      });
+      let npmInstallErrorMessage: string | undefined;
```
swatDong, line 103: Add try cache to the get-error-message part since we do not want to loss any npm install complete telemetry.

kuojianlu: Added.

packages/vscode-extension/src/debug/npmLogHandler.ts
kuojianlu: Added.

packages/vscode-extension/src/debug/npmLogHandler.ts
kuojianlu: Used `path.relative` to check.

swatDong (approved):

a1exwang (approved):

swatDong (approved):

chagong (approved):

jayzhang (approved):

PR #1426, opened 2021-06-24
chore: revert to stable version

wenytang-ms:


packages/fx-core/package-lock.json
```diff
@@ -1813,7 +1813,7 @@
     },
     "archy": {
       "version": "1.0.0",
-      "resolved": "https://registry.npmjs.org/archy/-/archy-1.0.0.tgz",
+      "resolved": "https://msazure.pkgs.visualstudio.com/9660fff2-2363-48b0-9e15-64df2283e932/_packaging/MODS-NPM-Registry/npm/registry/archy/-/archy-1.0.0.tgz",
```
adashen, line 1816: remove the npmrc file, seems the registry is not correct.

wenytang-ms: fixed.

adashen (approved):

PR #1427, opened 2021-06-24
[WIP] chore(bot): change deploy from remote build to local

YitongFeng-git:
### Note

This solution not working, in investigating.

### DIFF
--- a/packages/fx-core/src/plugins/resource/bot/languageStrategy.ts
+++ b/packages/fx-core/src/plugins/resource/bot/languageStrategy.ts
@@ -67,13 +67,13 @@ export class LanguageStrategy {
     }
 
     appSettings.push({
-      name: "SCM_DO_BUILD_DURING_DEPLOYMENT",
-      value: "true",
+      name: "WEBSITE_NODE_DEFAULT_VERSION",
+      value: "12.13.0",
     });
 
     appSettings.push({
-      name: "WEBSITE_NODE_DEFAULT_VERSION",
-      value: "12.13.0",
+      name: "WEBSITE_RUN_FROM_PACKAGE",
+      value: "1",
     });
 
     appSettings.forEach((p: NameValuePair) => {
@@ -91,7 +91,7 @@ export class LanguageStrategy {
     if (programmingLanguage === ProgrammingLanguage.TypeScript) {
       //Typescript needs tsc build before deploy because of windows app server. other languages don"t need it.
       try {
-        await utils.execute("npm install", packDir);
+        await utils.execute(`${Commands.NPM_INSTALL}`, packDir);
         await utils.execute("npm run build", packDir);
       } catch (e) {
         throw new CommandExecutionError(`${Commands.NPM_INSTALL},${Commands.NPM_BUILD}`, e);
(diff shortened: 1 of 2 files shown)


packages/fx-core/src/plugins/resource/bot/languageStrategy.ts
```diff
@@ -66,11 +66,6 @@ export class LanguageStrategy {
       appSettings = [];
     }
 
-    appSettings.push({
-      name: "SCM_DO_BUILD_DURING_DEPLOYMENT",
-      value: "true",
-    });
-
     appSettings.push({
       name: "WEBSITE_NODE_DEFAULT_VERSION",
```
eriolchan, line 70: What if user local node version is not 12 but 14? Will it lead to some compatible issue from remote to local? In Function plugin, it will detect user local machine node version and create the same node version in Azure. If this node version is not supported in Azure, we will use 12 as default value. Can you help sync with Heng to check if we need to apply the same to Bot plugin?

packages/fx-core/src/plugins/resource/bot/plugin.ts
```diff
@@ -480,9 +480,7 @@ export class TeamsBotImpl {
     );
 
     await handler?.next(ProgressBarConstants.DEPLOY_STEP_ZIP_FOLDER);
-    const zipBuffer = utils.zipAFolder(packDir, DeployConfigs.UN_PACK_DIRS, [
-      `${FolderNames.NODE_MODULES}/${FolderNames.KEYTAR}`,
```
eriolchan, line 484: If some constants such as FolderNames.KEYTAR are no longer used, we can safely remove them.

packages/fx-core/src/plugins/resource/bot/languageStrategy.ts
```diff
@@ -91,7 +86,7 @@ export class LanguageStrategy {
     if (programmingLanguage === ProgrammingLanguage.TypeScript) {
       //Typescript needs tsc build before deploy because of windows app server. other languages don"t need it.
       try {
-        await utils.execute("npm install", packDir);
+        await utils.execute(`${Commands.NPM_INSTALL}`, packDir);
         await utils.execute("npm run build", packDir);
```
eriolchan, line 90: If we make "npm install" string to constant, will we do the same as "npm run build"?

packages/fx-core/src/plugins/resource/bot/languageStrategy.ts
```diff
@@ -101,7 +96,7 @@ export class LanguageStrategy {
     if (programmingLanguage === ProgrammingLanguage.JavaScript) {
       try {
         // fail to npm install @microsoft/teamsfx on azure web app, so pack it locally.
-        await utils.execute("npm install", packDir);
+        await utils.execute(`${Commands.NPM_INSTALL}`, packDir);
```
eriolchan, line 99: Since we put build from remote to local, have we handled the incremental build/deployment case in the current code? e.g. if there is no new file change, there is no need to build the artifact again and we don't need to deploy the same artifact to Azure. This is another performance improvement of resource plugin, you can find more details with Heng and check if it apply to Bot plugin

eriolchan: Beside, since we have changed the build and deployment strategy/solution, it is a good opportunity for us to review the current Bot design on provision/deployment again to check if there is any other place need to be updated along with this design change?

PR #1431, opened 2021-06-24
feat(cli-preview): implement prepare dev env stage

kuojianlu:
Happy path:
image

Error path:
image

### DIFF
--- a/packages/cli/src/cmds/preview/commonUtils.ts
+++ b/packages/cli/src/cmds/preview/commonUtils.ts
@@ -5,12 +5,36 @@
 
 import * as path from "path";
 import * as fs from "fs-extra";
-import { ConfigFolderName } from "@microsoft/teamsfx-api";
+import { ConfigFolderName, FxError, IProgressHandler, LogLevel } from "@microsoft/teamsfx-api";
 
 import * as constants from "./constants";
+import { TaskResult } from "./task";
+import cliLogger from "../../commonlib/log";
+import { TaskFailed } from "./errors";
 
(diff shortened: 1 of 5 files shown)


swatDong:
> Happy path:
> image
> 
> Error path:
> image

kuojianlu:
> > Happy path:
> > image
> > Error path:
> > image
> 
> One question about the progress UX - is it possible to merge all preview steps into one progress? So that user can know how many remaining steps. E.g.,
> 
> ```
> [TeamsfxCLI] Preview [0/10]: Starting preview
> [TeamsfxCLI] Preview [1/10]: NPM install tabs
> [TeamsfxCLI] Preview [2/10]: NPM install api
> [TeamsfxCLI] Preview [3/10]: AAD
> ...
> ```

One question is that some steps run in parallel. For example,  since`NPM install` and `NPM install api` run in parallel, what is the UX of these two steps?
```
[TeamsfxCLI] Preview [0/10]: Starting preview
[TeamsfxCLI] Preview [1/10]: NPM install tabs
[TeamsfxCLI] Preview [2/10]: NPM install api
```
This UX means that `NPM install tabs` is done and `NPM install api` is running.

swatDong:
I see... the same for multiple plugins. That should be a global improvement but not for `preview` only. Ok to follow existing pattern.

swatDong (approved):

kuojianlu:
> I see... the same for multiple plugins. That should be a global improvement but not for `preview` only. Ok to follow existing pattern.

For multiple plugins, each plugin stage has its own title, so it is possible to have a nice UX.

PR #1438, opened 2021-06-25
fix(bot): no detailed error message when deploy failed

YitongFeng-git:
fix bug https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10211407

### DIFF
--- a/packages/fx-core/src/plugins/resource/bot/azureOps.ts
+++ b/packages/fx-core/src/plugins/resource/bot/azureOps.ts
@@ -17,6 +17,7 @@ import { CommonStrings, ConfigNames } from "./resources/strings";
 import * as utils from "./utils/common";
 import { default as axios } from "axios";
 import { ErrorMessagesForChecking } from "./constants";
+import { Logger } from "../../../core";
 
 export class AzureOperations {
   public static async CreateBotChannelRegistration(
@@ -32,7 +33,7 @@ export class AzureOperations {
         location: "global",
         kind: "bot",
         properties: {
-          displayName: displayName?? botChannelRegistrationName,
+          displayName: displayName ?? botChannelRegistrationName,
           endpoint: "",
           msaAppId: msaAppId,
         },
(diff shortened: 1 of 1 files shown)


Alive-Fish (approved):

henzhang-ms (approved):

packages/fx-core/src/plugins/resource/bot/azureOps.ts
```diff
@@ -199,10 +200,12 @@ export class AzureOperations {
     try {
       res = await axios.post(zipDeployEndpoint, zipBuffer, config);
     } catch (e) {
+      Logger.error(e.message);
       throw new ZipDeployError(e);
     }
 
     if (!res || !utils.isHttpCodeOkOrCreated(res.status)) {
+      Logger.error(JSON.stringify(res));
```
henzhang-ms, line 208: Nit: do we need all the res? shall we just log the error code and message?

YitongFeng-git: There is no message in res, only status code.
image

eriolchan: Agree that we don't need to print headers and request which are not very helpful for trouble shooting. I think the error message should be in data field. Can we trigger some failure on this zip deploy step (or just ask bug creator how to repro this) and check what's the value of data when status code is not 200?

In general, status code and data is enough to be print

eriolchan: Agree that we don't need to print headers and request which are not very helpful for trouble shooting. I think the error message should be in data field. Can we trigger some failure on this zip deploy step (or just ask bug creator how to repro this) and check what's the value of data when status code is not 200?

In general, status code and data is enough to be print

packages/fx-core/src/plugins/resource/bot/azureOps.ts
```diff
@@ -205,7 +205,9 @@ export class AzureOperations {
     }
 
     if (!res || !utils.isHttpCodeOkOrCreated(res.status)) {
-      Logger.error(JSON.stringify(res));
+      Logger.error(
+        `Deploy endpoint: ${zipDeployEndpoint}, status code: ${res.status}, response data: ${res.data}`
```
eriolchan, line 209: If res.data is JSON format, can it be print correctly?

eriolchan (approved):

PR #1443, opened 2021-06-28
fix(sql): remove sql skip-adding-user question

xzf0587:
1. remove sql skip-adding-user question.
2. set default skip-adding-user value based on azureAccountProvider?.getIdentityCredentialAsync

### DIFF
--- a/packages/fx-core/src/plugins/resource/sql/plugin.ts
+++ b/packages/fx-core/src/plugins/resource/sql/plugin.ts
@@ -8,6 +8,7 @@ import {
   Stage,
   QTreeNode,
   Platform,
+  Func,
 } from "@microsoft/teamsfx-api";
 import { ManagementClient } from "./managementClient";
 import { ErrorMessage } from "./errors";
@@ -21,6 +22,11 @@ import { Constants, HelpLinks, Telemetry } from "./constants";
 import { Message } from "./utils/message";
 import { TelemetryUtils } from "./utils/telemetryUtils";
 import { adminNameQuestion, adminPasswordQuestion, confirmPasswordQuestion } from "./questions";
+import {
+  sqlConfirmPasswordValidatorGenerator,
+  sqlPasswordValidatorGenerator,
+  sqlUserNameValidator,
+} from "./utils/checkInput";
 
 export class SqlPluginImpl {
   config: SqlConfig = new SqlConfig();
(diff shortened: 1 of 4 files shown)


packages/fx-core/src/plugins/resource/sql/plugin.ts
```diff
@@ -134,14 +98,15 @@ export class SqlPluginImpl {
...
-      const skipAddingUser = ctx.answers![Constants.questionKey.skipAddingUser] as string;
-      if (skipAddingUser) {
-        this.config.skipAddingUser = skipAddingUser === "true" ? true : false;
-        ctx.config.set(Constants.skipAddingUser, this.config.skipAddingUser);
-      }
+    const skipAddingUser = ctx.config.get(Constants.skipAddingUser);
+    if (skipAddingUser === undefined) {
+      this.config.skipAddingUser = (await ctx.azureAccountProvider?.getIdentityCredentialAsync())
+        ? false
+        : true;
```
Alive-Fish, line 105: Is that the opposite?

Alive-Fish:
I think it's OK for the most cases, but some corner cases may have problems, such as provisioning in CLI first and then provision in VSCode extension.

xzf0587 (commented):
> I think it's OK for the most cases, but some corner cases may have problems, such as provisioning in CLI first and then provision in VSCode extension.

If developers haven't modify the skip-adding-user config in env file, leave it as default. The project will work.
Only  specifying the skip-adding-user true in env file will make the cli provision failed.

wenytang-ms:
it's better merge as a hotfix not a feature

adashen:
> > I think it's OK for the most cases, but some corner cases may have problems, such as provisioning in CLI first and then provision in VSCode extension.
> 
> If developers haven't modify the skip-adding-user config in env file, leave it as default. The project will work.
> Only specifying the skip-adding-user false in env file will make the cli provision failed.

Let's add both test cases and do the verification

xzf0587:
> > > I think it's OK for the most cases, but some corner cases may have problems, such as provisioning in CLI first and then provision in VSCode extension.
> > 
> > 
> > If developers haven't modify the skip-adding-user config in env file, leave it as default. The project will work.
> > Only specifying the skip-adding-user false in env file will make the cli provision failed.
> 
> Let's add both test cases and do the verification

Added the uint test for skip-adding-user.
The e2e test does not support test cli scenario which the identity credential is undefined.

adashen (approved):

Alive-Fish (approved):
look good to me

packages/fx-core/src/plugins/resource/sql/plugin.ts
```diff
@@ -134,14 +124,15 @@ export class SqlPluginImpl {
...
-    this.config.skipAddingUser = ctx.config.get(Constants.skipAddingUser) as boolean;
-    if (ctx.answers?.platform === Platform.CLI) {
-      const skipAddingUser = ctx.answers![Constants.questionKey.skipAddingUser] as string;
-      if (skipAddingUser) {
-        this.config.skipAddingUser = skipAddingUser === "true" ? true : false;
-        ctx.config.set(Constants.skipAddingUser, this.config.skipAddingUser);
-      }
+    const skipAddingUser = ctx.config.get(Constants.skipAddingUser);
+    if (skipAddingUser === undefined) {
+      this.config.skipAddingUser = (await ctx.azureAccountProvider?.getIdentityCredentialAsync())
```
SLdragon, line 129: Can this be simplified to :
```
this.config.skipAddingUser = !(await ctx.azureAccountProvider?.getIdentityCredentialAsync())
```

PR #1452, opened 2021-06-29
feat: add 'run and debug' item on sidebar for a/b testing

HuihuiWu-Microsoft:
UX design:
https://www.figma.com/file/qqVegbnWWBsqHhR8Qijs4O/Teams-Toolkit---GA?node-id=1%3A297

### DIFF
--- a/packages/vscode-extension/src/commandsTreeViewProvider.ts
+++ b/packages/vscode-extension/src/commandsTreeViewProvider.ts
@@ -6,6 +6,7 @@ import * as path from "path";
 import { ext } from "./extensionVariables";
 import { TreeItem, TreeCategory, Result, FxError, ok } from "@microsoft/teamsfx-api";
 import * as StringResources from "./resources/Strings.json";
+import { exp } from "./exp/index";
 
 class TreeViewManager {
   private static instance: TreeViewManager;
@@ -22,7 +23,7 @@ class TreeViewManager {
     return TreeViewManager.instance;
   }
 
-  public registerTreeViews() {
+  public async registerTreeViews() {
     const disposables = [];
 
     const getStartTreeViewCommand = [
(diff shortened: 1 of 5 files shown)


packages/vscode-extension/src/commandsTreeViewProvider.ts
```diff
@@ -143,9 +144,24 @@ class TreeViewManager {
         undefined,
         undefined,
         { name: "publish", custom: true }
-      )
+      ),
     ];
 
+    if (
+      true /*await exp.getExpService().getTreatmentVariableAsync("flight", "treatmentVariable", true)*/
```
1openwindow, line 151: always true?

packages/vscode-extension/src/extension.ts
```diff
@@ -36,7 +36,9 @@ export async function activate(context: vscode.ExtensionContext) {
   );
   context.subscriptions.push(createCmd);
 
-  // 1.2 Register the creating command.
+  const debugCmd = vscode.commands.registerCommand("fx-extension.debug", handlers.debugHandler);
```
1openwindow, line 39: why this format is different to others? e.g.
  const updateCmd = vscode.commands.registerCommand(
    "fx-extension.update",
    handlers.addResourceHandler
  );

packages/vscode-extension/src/commandsTreeViewProvider.ts
HuihuiWu-Microsoft: Because the flight name and treatment variable hasn't been confirmed, so will update later

packages/vscode-extension/src/extension.ts
HuihuiWu-Microsoft: Because the debug line is shorter than others. I tested locally to make the total length of updateCmd line the same as debugCmd and it changed format to one line.

nliu-ms (approved):

1openwindow (approved):

PR #1453, opened 2021-06-29
chore: Separate lerna version and publish

wenytang-ms:
…ain branch operation

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -22,7 +22,7 @@ jobs:
       - name: Validate CD branch
         if: ${{ github.event_name == 'workflow_dispatch' && github.ref != 'refs/heads/main' && github.ref != 'refs/heads/dev' }}
         run: |
-          echo It's not allowed to run CD on other branch besides main and dev.
+          echo It's not allowed to run CD on other branch except main and dev.
           exit 1
 
       - name: Validate inputs for main
@@ -40,7 +40,7 @@ jobs:
       - name: Validate schedule
         if: ${{ github.event_name == 'schedule' && github.ref != 'refs/heads/dev' }}
         run: |
-          echo It's not allowed to run schedule release besides dev branch.
+          echo It's not allowed to run schedule release except dev branch.
           exit 1
 
       - name: Checkout branch
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -7,8 +7,8 @@ on:
         description: "preid name(alpha, rc, stable)"
         required: true
         default: "rc"
-      skip-rc-version:
-        description: "skip bump up rc version by lerna conventional commits(yes or no)"
+      skip-version-rc:
+        description: "release RC version just with lerna publish, skip run lerna version(yes or no)"
         required: false
         default: "no"
   schedule:
@@ -22,21 +22,25 @@ jobs:
       - name: Validate CD branch
         if: ${{ github.event_name == 'workflow_dispatch' && github.ref != 'refs/heads/main' && github.ref != 'refs/heads/dev' }}
         run: |
+          echo It's not allowed to run CD on other branch besides main and dev.
           exit 1
 
       - name: Validate inputs for main
         if: ${{ github.event_name == 'workflow_dispatch' && github.ref == 'refs/heads/main' && github.event.inputs.preid != 'rc' && github.event.inputs.preid != 'stable' }}
         run: |
+          echo It's not allowed to release alpha on main branch.
           exit 1
 
       - name: Valiadte inputs for dev
         if: ${{ github.event_name == 'workflow_dispatch' && github.ref == 'refs/heads/dev' && github.event.inputs.preid != 'alpha' }}
         run: |
+          echo It's not allowed to release RC or stable on dev branch.
           exit 1
 
       - name: Validate schedule
         if: ${{ github.event_name == 'schedule' && github.ref != 'refs/heads/dev' }}
         run: |
+          echo It's not allowed to run schedule release besides dev branch.
           exit 1
 
       - name: Checkout branch
@@ -111,7 +115,7 @@ jobs:
           npx lerna publish from-package --dist-tag=alpha --yes --allow-branch dev
 
       - name: version rc npm packages to npmjs.org
-        if: ${{ github.event_name == 'workflow_dispatch' && github.ref == 'refs/heads/main' && github.event.inputs.preid == 'rc' && github.event.inputs.skip-rc-version == 'no'}}
+        if: ${{ github.event_name == 'workflow_dispatch' && github.ref == 'refs/heads/main' && github.event.inputs.preid == 'rc' && github.event.inputs.skip-version-rc == 'no'}}
         run: |
           npx lerna version --no-private --conventional-commits --conventional-prerelease --preid=rc --no-changelog --yes
 
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -3,21 +3,13 @@ name: CD
 on:
   workflow_dispatch:
     inputs:
-      version:
-        description: "which version to bump(major, minor, patch, premajor, preminor, prepatch, prerelease)"
-        required: false
-        default: "prerelease"
       preid:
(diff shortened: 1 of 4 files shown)


.github/workflows/cd.yml
```diff
@@ -112,15 +112,10 @@ jobs:
...
+      - name: publish npm packages to npmjs.org by conventional-commits
+        if: ${{ github.event_name == 'workflow_dispatch' && github.ref == 'refs/heads/main' && github.event.inputs.conventional-commits == 'yes'}}
         run: |
-          npx lerna publish --conventional-prerelease --preid=rc --yes
-
-      - name: release stable npm packages to npmjs.org by conventional-commits
-        if: ${{ github.event_name == 'workflow_dispatch' && github.ref == 'refs/heads/main' && github.event.inputs.conventional-commits == 'yes' && github.event.inputs.preid == 'stable'}}
-        run: |
-          npx lerna publish --conventional-graduate --yes
+          npx lerna publish from-package --yes
```
adashen, line 118: Did we test whether it will release all packages? What is the manual step if we need update the version manually?

wenytang-ms: yes, "from-package" will look inside the latest tag which lerna version put on the branch, and will publish the version tagged on the latest commits.
manually step is:
1. release manager fetch origin main, and run "npm run version-rc"
2. make sure the rc/stable version change, then input "yes" or "y"
3. push to the origin main branch, and trigger the CD pipeline, must choose "conventional-commits" as "yes"

.github/workflows/cd.yml
```diff
@@ -98,30 +100,25 @@ jobs:
           git config --global user.email 'yiz@microsoft.com'
 
       - name: release alpha npm packages to npmjs.org
-        if: ${{ github.event_name == 'schedule' }}
+        if: ${{ github.event_name == 'schedule' || (github.event_name == 'workflow_dispatch' && github.ref == 'refs/heads/dev' && github.event.inputs.preid == 'alpha')}}
         run: |
           npx lerna version --no-private prerelease --preid=alpha.$(git rev-parse --short HEAD) --no-push --allow-branch dev --yes
```
adashen, line 105: add no tag

wenytang-ms: seem with no-git-tag-version will product some uncommited files, then lerna publish will fail

.github/workflows/cd.yml
```diff
@@ -3,21 +3,13 @@ name: CD
...
-      force:
-        description: "force release even if no update(set --force-publish if necessary)"
-        required: false
-        default: ""
-      conventional-commits:
-        description: "bump up release version by lerna conventional commits(yes or no)"
         required: true
+        default: "rc"
+      skip-rc-version:
+        description: "skip bump up rc version by lerna conventional commits(yes or no)"
```
qinezh, line 11: what's the scenario to skip bump rc version in this CD pipeline?

.github/workflows/cd.yml
```diff
@@ -98,30 +105,25 @@ jobs:
...
 
-      - name: release stable npm packages to npmjs.org
-        if: ${{ github.event_name == 'workflow_dispatch' && github.ref == 'refs/heads/main' && !startsWith(github.event.inputs.version, 'pre') && github.event.inputs.conventional-commits == 'no'}}
-        run: |
-          npx lerna publish --no-private --dist-tag=latest ${{ github.event.inputs.version }} --no-changelog --yes ${{ github.event.inputs.force }}
-
-      - name: release rc npm packages to npmjs.org by conventional-commits
-        if: ${{ github.event_name == 'workflow_dispatch' && github.ref == 'refs/heads/main' && github.event.inputs.conventional-commits == 'yes' && github.event.inputs.preid == 'rc'}}
+      - name: publish rc npm packages to npmjs.org
+        if: ${{ github.event_name == 'workflow_dispatch' && github.ref == 'refs/heads/main' && github.event.inputs.preid == 'rc' }}
```
qinezh, line 119: do we need to check `skip-rc-version` here?

package.json
```diff
@@ -5,6 +5,7 @@
         "bootstrap": "lerna bootstrap",
         "lint:staged": "lerna run lint:staged --stream",
         "lint": "lerna run lint --stream",
+        "manual-version": "lerna version --preid=rc --no-push --no-changelog --no-private",
         "version": "lerna bootstrap --ignore-scripts -- --package-lock-only --no-audit && git add packages/*/package-lock.json",
```
qinezh, line 9: how about replacing with the original "version" script to avoid confusion?

.github/workflows/cd.yml
```diff
@@ -27,8 +19,23 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - name: Validate inputs
-        if: ${{ github.event_name == 'workflow_dispatch' && ((github.ref != 'refs/heads/dev' && github.ref != 'refs/heads/main') || (github.ref == 'refs/heads/dev' && (!startsWith(github.event.inputs.version, 'pre') || github.event.inputs.preid != 'alpha')) || (github.ref == 'refs/heads/main' && startsWith(github.event.inputs.version, 'pre') && github.event.inputs.preid != 'rc') || (github.ref == 'refs/heads/dev' && github.event.inputs.conventional-commits == 'yes') || (github.ref == 'refs/heads/main' && github.event.inputs.conventional-commits == 'yes' && !(github.event.inputs.preid == 'rc' || github.event.inputs.preid == 'stable'))) }}
+      - name: Validate CD branch
+        if: ${{ github.event_name == 'workflow_dispatch' && github.ref != 'refs/heads/main' && github.ref != 'refs/heads/dev' }}
+        run: |
+          exit 1
```
qinezh, line 25: would it better to output error message that why the pipeline exits?

.github/workflows/cd.yml
qinezh: offline discussed, may be we could rename it like `enable-auto-versioning`?

wenytang-ms: just rename it as "skip-version-rc" and modify the description
actually, this config work for rc version, so I think it's better add rc on it

.github/workflows/cd.yml
wenytang-ms: skip-rc-verison is optional, not required, so there is no need to check

package.json
wenytang-ms: this 'version' is removed, already sync up wth @LongOddCode

.github/workflows/cd.yml
wenytang-ms: sure!

package.json
```diff
@@ -6,7 +6,6 @@
         "lint:staged": "lerna run lint:staged --stream",
         "lint": "lerna run lint --stream",
         "manual-version": "lerna version --preid=rc --no-push --no-changelog --no-private",
```
qinezh, line 8: would it be better to update to `version:rc` as the script will only version with preid `rc`?

wenytang-ms: sure

qinezh (approved):

LongOddCode (approved):

.github/workflows/cd.yml
```diff
@@ -62,7 +69,7 @@ jobs:
             ./*.zip
 
       - name: release templates' rc version to github
-        if: ${{ github.ref == 'refs/heads/main' && startsWith(github.event.inputs.version, 'pre') }}
+        if: ${{ github.ref == 'refs/heads/main' && github.event.inputs.preid == 'rc' }}
         uses: marvinpinto/action-automatic-releases@latest
```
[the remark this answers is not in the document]
wenytang-ms: sure, I will investigate this later.

.github/workflows/cd.yml
```diff
@@ -62,7 +69,7 @@ jobs:
             ./*.zip
 
       - name: release templates' rc version to github
-        if: ${{ github.ref == 'refs/heads/main' && startsWith(github.event.inputs.version, 'pre') }}
+        if: ${{ github.ref == 'refs/heads/main' && github.event.inputs.preid == 'rc' }}
```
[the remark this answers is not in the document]
wenytang-ms: sure, it's seems will fail, later I will work on this workaround

PR #1455, opened 2021-06-29
feat(cli-preview): support typescript

kuojianlu:
- support preview for typescript
- remove node inspector for function and bot
- add backend extensions install

### DIFF
--- a/packages/cli/src/cmds/preview/preview.ts
+++ b/packages/cli/src/cmds/preview/preview.ts
@@ -155,8 +155,12 @@ export default class Preview extends YargsCommand {
     }
 
     /* === start services === */
+    const programmingLanguage = config?.config
+      ?.get(constants.solutionPluginName)
+      ?.get(constants.programmingLanguageConfigKey) as string;
     result = await this.startServices(
       workspaceFolder,
+      programmingLanguage,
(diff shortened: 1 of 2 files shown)


packages/cli/src/cmds/preview/preview.ts
```diff
@@ -326,6 +349,10 @@ export default class Preview extends YargsCommand {
       core.localDebug(inputs),
       frontendInstallTask?.wait(frontendInstallStartCb, frontendInstallStopCb),
       backendInstallTask?.wait(backendInstallStartCb, backendInstallStopCb),
+      backendExtensionsInstallTask?.wait(
```
swatDong, line 352: Does backend npm install depend on backend extension install?

kuojianlu: No, `backend npm install` does not depend on `backend extension install`.
If we use `npx func extensions install` command to install functions binding, then `backend extension install` will depend on `backend npm install`. However, we use `dotnet build` to install functions binding.

swatDong: Make sense... just curious why in .vscode/tasks.json it has dependency. That could be another improvement.

swatDong (approved):

PR #1463, opened 2021-06-29
fix(cli-preview): merge process.env when starting services

kuojianlu:
merge process.env when starting services, otherwise there will be an error on macos: `env: node: No such file or directory` or `/bin/sh: dotnet: command not found`

### DIFF
--- a/packages/cli/src/cmds/preview/commonUtils.ts
+++ b/packages/cli/src/cmds/preview/commonUtils.ts
@@ -162,16 +162,12 @@ export async function getPortsInUse(
 export function mergeProcessEnv(
   env: { [key: string]: string | undefined } | undefined
 ): { [key: string]: string | undefined } | undefined {
-  const processEnv = process.env;
   if (env === undefined) {
     return process.env;
   }
-  const keys = Object.keys(env);
-  for (const key of Object.keys(processEnv)) {
-    const existing = keys.some((k) => k === key);
-    if (!existing) {
-      env[key] = processEnv[key];
-    }
+  const result = Object.assign({}, process.env);
+  for (const key of Object.keys(env)) {
+    result[key] = env[key];
   }
-  return env;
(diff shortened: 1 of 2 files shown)


packages/cli/src/cmds/preview/commonUtils.ts
```diff
@@ -158,3 +158,20 @@ export async function getPortsInUse(
...
+
+export function mergeProcessEnv(
+  env: { [key: string]: string | undefined } | undefined
+): { [key: string]: string | undefined } | undefined {
+  const processEnv = process.env;
+  if (env === undefined) {
+    return process.env;
+  }
+  const keys = Object.keys(env);
+  for (const key of Object.keys(processEnv)) {
```
swatDong, line 170: How about `Object.assign()`?

``` typescript
const result = Object.assign({}, process.env);
for (const key of Object.keys(env)) {
    result[key] = env[key];
}
```

packages/cli/src/cmds/preview/commonUtils.ts
```diff
@@ -158,3 +158,20 @@ export async function getPortsInUse(
...
+    const existing = keys.some((k) => k === key);
+    if (!existing) {
+      env[key] = processEnv[key];
```
swatDong, line 173: Is it expected to modify input env?

packages/cli/src/cmds/preview/commonUtils.ts
kuojianlu: Nice code. Updated.

packages/cli/src/cmds/preview/commonUtils.ts
kuojianlu: Safe here, but it is not a good practice to modify input argument. Updated.

swatDong (approved):

PR #1466, opened 2021-06-30
chore(frontend&function): send skip deploy telemetry event

hund030:


packages/fx-core/src/plugins/resource/frontend/ops/deploy.ts
```diff
@@ -64,8 +65,9 @@ export class FrontendDeployment {
   }
 
   public static async skipBuild(): Promise<void> {
-    const progressHandler = ProgressHelper.deployProgress;
     Logger.info(Messages.SkipBuild());
+
+    const progressHandler = ProgressHelper.deployProgress;
```
henzhang-ms, line 70: I like this kind of adjustment, it makes the code more compact even it has no function impact and no line reduce.

henzhang-ms (approved):
looks good to me

PR #1469, opened 2021-06-30
fix(cli-preview): terminate background tasks when error occurs

kuojianlu:
use `tree-kill` to terminate background tasks when error occurs.

### DIFF
--- a/packages/cli/src/cmds/preview/preview.ts
+++ b/packages/cli/src/cmds/preview/preview.ts
@@ -73,10 +73,20 @@ export default class Preview extends YargsCommand {
 
     CliTelemetry.setReporter(CliTelemetry.getReporter().withRootFolder(workspaceFolder));
 
-    if (args.local || (!args.local && !args.remote)) {
-      return await this.localPreview(workspaceFolder);
+    try {
+      const previewType = args.local || (!args.local && !args.remote) ? "local" : "remote";
+      const result =
+        previewType === "local"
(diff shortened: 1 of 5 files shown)


packages/cli/src/cmds/preview/preview.ts
```diff
@@ -145,12 +151,14 @@ export default class Preview extends YargsCommand {
...
     );
     if (result.isErr()) {
+      await this.terminateTasks();
       return result;
     }
 
     /* === check ports === */
     const portsInUse = await commonUtils.getPortsInUse(includeFrontend, includeBackend, includeBot);
     if (portsInUse.length > 0) {
+      await this.terminateTasks();
```
swatDong, line 161: Call `terminateTasks` via process.on("exit") instead of everywhere...

kuojianlu: Tried this way. One issue is that the exit callback logic must be synchronous: https://nodejs.org/dist/latest-v14.x/docs/api/process.html#process_event_exit, but the tree-kill operation is synchronous. So this way cannot terminate the processes as expected.

swatDong (approved):

kuojianlu:
Know issue: if killing the preview process directly (not closing terminal or pressing Ctrl+C), the subprocesses will not be terminated.

Alive-Fish (approved):
look good to me

PR #1470, opened 2021-06-30
fix(function): remove notification for skipping deployment

hund030:


packages/fx-core/src/plugins/resource/function/resources/message.ts
```diff
@@ -1,6 +1,11 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
-import { FunctionPluginPathInfo as PathInfo } from "../constants";
+import * as path from "path";
+import { FunctionPluginInfo, FunctionPluginPathInfo as PathInfo } from "../constants";
+
+export function normalizeMessage(message: string): string {
```
henzhang-ms, line 6: We'd better not add customized format in our plugin unless we synced with solution.

packages/fx-core/src/plugins/resource/function/resources/message.ts
```diff
@@ -14,10 +19,11 @@ export class InfoMessages {
...
   public static readonly installFunctionExtensions: string = "Installing Azure Functions binding.";
-  public static readonly noChange: string =
-    "Nothing to deploy; no changes detected since last deployment.";
-  public static readonly skipDeployment: string =
-    "Skip deployment.";
+  public static readonly noChange: string = "No changes detected since last deployment.";
+  public static readonly skipDeployment: string = `Skip deployment for no changes detected. Remove ${path.join(
+    PathInfo.solutionFolderName,
+    PathInfo.funcDeploymentFolderName
+  )} folder to force deployment.`;
```
henzhang-ms, line 26: maybe: `To fully redeploy the function project, please remove xxx and rerun the command`.

packages/fx-core/src/plugins/resource/function/utils/dialog.ts
```diff
@@ -1,17 +1,20 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
-import { DialogMsg, DialogType, MsgLevel, PluginContext } from "@microsoft/teamsfx-api";
+import { PluginContext } from "@microsoft/teamsfx-api";
+import { normalizeMessage } from "../resources/message";
+
+enum MsgLevel {
```
henzhang-ms, line 6: don't we have type from api?

hund030: No. They didn't define such type.

packages/fx-core/src/plugins/resource/function/resources/message.ts
hund030: Agree. Updated.

packages/fx-core/src/plugins/resource/function/resources/message.ts
hund030: Good point. Removed the format for notification messages.

henzhang-ms (approved):

PR #1472, opened 2021-06-30
chore(sdk): add coverage for node unit tests

tecton:
1. Deprecate the wording of "integration test" and use "e2e" because it involves Azure resource.
2. Use nyc to generate coverage report.


packages/sdk/package.json
```diff
@@ -26,15 +26,15 @@
     "integration-test:browser-case": "karma start --single-run --glob=./dist-test/index.integration.browser.js",
     "integration-test:browser": "node test/integration/scripts/DownloadSimpleAuth.js && concurrently --kill-others --success first \"node test/integration/scripts/StartSimpleAuth.js\" \"npm run integration-test:browser-case\"",
     "integration-test:node": "mocha -r test/mocha.env.ts -r ts-node/register \"./test/integration/node/*.spec.ts\" --timeout 100000 --exit --reporter ./tools/mocha-multi-reporter.js",
-    "test:integration": "npm run integration-test:node && npm run integration-test:browser",
+    "test:e2e": "npm run integration-test:node && npm run integration-test:browser",
```
wenytang-ms, line 29: if the test:integration rename to test:2e2, the integraiont-test:browser should be rename to e2e-test:browser too?

tecton: Yes, I'll rename "integration" wording.

packages/sdk/package.json
```diff
@@ -23,20 +23,20 @@
...
     "lint:staged": "lint-staged",
     "lint": "eslint \"**/*.ts\"",
     "prebuild": "npm run clean",
-    "test:browser": "npm run build:test && npm run unit-test:browser && npm run integration-test:browser",
-    "test:node": "npm run build:test && npm run unit-test:node && npm run integration-test:node",
-    "test": "npm run test:unit && npm run test:integration",
+    "test:browser": "npm run build:test && npm run unit-test:browser && npm run e2e-test:browser",
+    "test:node": "npm run build:test && npm run unit-test:node && npm run e2e-test:node",
+    "test": "npm run test:unit && npm run test:e2e",
     "unit-test:browser": "karma start --single-run --glob=./dist-test/index.unit.browser.js",
```
blackchoey, line 36: can we have coverage report for browser libs?

packages/sdk/.nycrc
```diff
@@ -1,19 +0,0 @@
-{
-    "include": [
-      "dist-esm/src/**/*.js"
```
blackchoey, line 3: what config will be used after we remove this file?

packages/sdk/package.json
tecton: I'll investigate how to add coverage for browser tests later.

packages/sdk/.nycrc
tecton: I think it will use the default configuration of nyc which is similar to other packages.

blackchoey: Seems by default nyc only includes files that were required: https://www.npmjs.com/package/nyc#selecting-files-for-coverage.
So if we add a new file but don't add test case for it, the code coverage won't reduce. This might not be the expected result. Not sure whether other default configuration meets our requirement.

tecton: Thank you! I agree that it is not expected, adding @LongOddCode for awareness.

tecton: Added the nycrc file back. @blackchoey

blackchoey (approved):

PR #1473, opened 2021-06-30
fix(aad): change system error to user error

KennethBWSong:
Change system error to user error for some cases.
Add new error type for failing to call app studio api:
image

### DIFF
--- a/packages/fx-core/src/plugins/resource/aad/aadAppClient.ts
+++ b/packages/fx-core/src/plugins/resource/aad/aadAppClient.ts
@@ -12,13 +12,13 @@ import {
   UpdateRedirectUriError,
   GetAppError,
   GetAppConfigError,
+  AppNotExistError,
 } from "./errors";
 import { GraphClient } from "./graph";
 import { IAADPassword } from "./interfaces/IAADApplication";
 import { IAADDefinition, RequiredResourceAccess } from "./interfaces/IAADDefinition";
 import { ResultFactory } from "./results";
 import { ProvisionConfig } from "./utils/configs";
-import { DialogUtils } from "./utils/dialog";
 import { TokenAudience, TokenProvider } from "./utils/tokenProvider";
 
 function delay(ms: number) {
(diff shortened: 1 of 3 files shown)


packages/fx-core/src/plugins/resource/aad/aadAppClient.ts
```diff
@@ -47,7 +47,10 @@ export class AadAppClient {
       config.clientId = provisionAadResponse.appId;
       config.objectId = provisionAadResponse.id;
     } catch (error) {
-      throw ResultFactory.SystemError(CreateAppError.name, CreateAppError.message(), error);
+      if (error?.response?.data?.errorMessage.includes(Constants.appStudioErrorMessage)) {
```
blackchoey, line 50: can we treat 5xx error as system error?

KennethBWSong: Yeah I think we can classify error type by status code in error message. I will update logic here.

blackchoey (approved):

packages/fx-core/src/plugins/resource/aad/aadAppClient.ts
```diff
@@ -47,6 +46,9 @@ export class AadAppClient {
       config.clientId = provisionAadResponse.appId;
       config.objectId = provisionAadResponse.id;
     } catch (error) {
+      if (error?.response?.status >= 400 && error?.response?.status < 500) {
```
xzf0587, line 49: It is better to use constant to replace 400 and 500.

KennethBWSong: Fixed. Thank you for pointing that out.

xzf0587 (approved):

packages/fx-core/src/plugins/resource/aad/aadAppClient.ts
```diff
@@ -123,12 +143,22 @@ export class AadAppClient {
         );
       }
     } catch (error) {
-      throw ResultFactory.UserError(
+      if (
```
adashen, line 146: seems previously it is always user error?

KennethBWSong: Yes, since in my test I only found two errors:
1. sharing the bot, which will cause the same app id uri of two aad apps;
2. storage endpoint can not be set as app id uri;
And now it will classify the errors with status code.

adashen (approved):

PR #1474, opened 2021-06-30
refactor(apim): refactor apim plugin according to core refactor

XiaofuHuang:
Refactor APIM plugin according to the changes of teamsfx-api.
1. Use `input` instead of `answer` in the question model
2. Use `LocalFunc` instead of `callFunc` in the question model
3. Remove the usage of `Stage.update`
4. Add lifecycle `GetQuestionsForUserTask`

### DIFF
--- a/packages/fx-core/src/plugins/resource/apim/index.ts
+++ b/packages/fx-core/src/plugins/resource/apim/index.ts
@@ -143,7 +143,7 @@ async function _getQuestionsForUserTask(
   const apimConfig = new ApimPluginConfig(ctx.config);
   const questionManager = await Factory.buildQuestionManager(ctx, solutionConfig);
   if (func.method === "addResource") {
-    return await questionManager.update(ctx, apimConfig);
+    return await questionManager.addResource(ctx, apimConfig);
   }
   return undefined;
 }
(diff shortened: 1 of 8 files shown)


packages/fx-core/src/plugins/resource/apim/index.ts
```diff
@@ -38,12 +38,17 @@ export class ApimPlugin implements Plugin {
...
   }
-  
+
   public async getQuestionsForUserTask(
     func: Func,
     ctx: PluginContext
   ): Promise<Result<QTreeNode | undefined, FxError>> {
-    return await this.executeWithFxError(PluginLifeCycle.GetQuestions, _getQuestionsForUserTask, ctx, func);
+    return await this.executeWithFxError(
+      PluginLifeCycle.GetQuestions,
```
dooriya, line 47: Is this supposed to be `PluginLifeCycle.GetQuestionsForUserTask`?
If not, seems we can remove `GetQuestionsForUserTask` in the `PluginLifeCycle` definition

XiaofuHuang: Yes, it should be `PluginLifeCycle.GetQuestionsForUserTask`

dooriya (approved):

PR #1482, opened 2021-07-01
chore(sql): remove useless callFunc code

xzf0587:
1. remove useless callFunc code 
2. add unit test.

### DIFF
--- a/packages/fx-core/tests/plugins/resource/sql/helper.ts
+++ b/packages/fx-core/tests/plugins/resource/sql/helper.ts
@@ -28,6 +28,10 @@ export class TestHelper {
           console.debug(message);
           return true;
         },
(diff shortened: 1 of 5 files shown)


packages/fx-core/tests/plugins/resource/sql/helper.ts
```diff
@@ -28,6 +28,10 @@ export class TestHelper {
           console.debug(message);
           return true;
         },
+        async warning(message: string): Promise<boolean> {
```
adashen, line 31: when is this method called seems not in this pr?

xzf0587: it is a mock for logProvider for test.
It is used in ctx.logProvider?.warning()

adashen (approved):

PR #1484, opened 2021-07-01
fix(apim): change resource not found error message

XiaofuHuang:
Change `Failed to find resource` to `Resource does not exist` in the log message.

### DIFF
--- a/packages/fx-core/src/plugins/resource/apim/log.ts
+++ b/packages/fx-core/src/plugins/resource/apim/log.ts
@@ -42,8 +42,8 @@ export class LogMessages {
 
   public static readonly resourceNotFound = (resourceType: IName, resourceId?: string): string =>
     !resourceId
-      ? `[${ProjectConstants.pluginDisplayName}] Failed to find resource ${resourceType.displayName} in Azure.`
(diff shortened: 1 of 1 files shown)


packages/fx-core/src/plugins/resource/apim/log.ts
```diff
@@ -42,8 +42,8 @@ export class LogMessages {
 
   public static readonly resourceNotFound = (resourceType: IName, resourceId?: string): string =>
     !resourceId
-      ? `[${ProjectConstants.pluginDisplayName}] Failed to find resource ${resourceType.displayName} in Azure.`
-      : `[${ProjectConstants.pluginDisplayName}] Failed to find resource ${resourceType.displayName} '${resourceId} in Azure.`;
+      ? `[${ProjectConstants.pluginDisplayName}] Resource ${resourceType.displayName} does not exist in Azure. Try to create a new one.`
+      : `[${ProjectConstants.pluginDisplayName}] Resource ${resourceType.displayName} '${resourceId}' does not exist in Azure. Try to create a new one.`;
```
dooriya, line 46: If this message template is also used for AAD app, then 'in Azure' is not that appropriate here.

dooriya (approved):

PR #1486, opened 2021-07-01
chore(cli-preview): add telemetry for preview command

kuojianlu:
add telemetry for preview command.


packages/cli/src/cmds/preview/cpUtils.ts
```diff
@@ -0,0 +1,149 @@
+/* eslint-disable @typescript-eslint/no-namespace */
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+// NOTE:
+// DO NOT EDIT this file in function plugin.
```
swatDong, line 6: Not function plugin...

kuojianlu: How about just remove these comments?

swatDong: It depends if you need the copy logic for this file as well

kuojianlu: I think not. Just removed them for simplicity.

swatDong (approved):

packages/cli/src/cmds/preview/preview.ts
```diff
@@ -63,27 +70,44 @@ export default class Preview extends YargsCommand {
...
+      cliTelemetry.sendTelemetryEvent(TelemetryEvent.PreviewStart);
+      const error = errors.ExclusiveLocalRemoteOptions();
+      cliTelemetry.sendTelemetryErrorEvent(TelemetryEvent.Preview, error);
+      return err(error);
     }
 
+    const previewType = args.local || (!args.local && !args.remote) ? "local" : "remote";
+    this.telemetryProperties[TelemetryProperty.PreviewType] = previewType;
+
     const workspaceFolder = path.resolve(args.folder as string);
```
chagong, line 82: You need to enable RootFolder for preview command telemetries, I think. This is to make sure we can get some config context in telemetry like "appid". Do something like this: CliTelemetry.withRootFolder(rootFolder).sendTelemetryEvent(TelemetryEvent.ProvisionStart);

kuojianlu: Sure. Updated to use `cliTelemetry.withRootFolder`.

chagong (approved):

PR #1492, opened 2021-07-02
chore(sql): improve sql error message

xzf0587:
1. improve error message for azure api call scenarios.


adashen (commented):
Can we have some screen shot to show the current fixed behavior? What will be shown in the alert box, what will be shown in the output.

xzf0587:
> Can we have some screen shot to show the current fixed behavior? What will be shown in the alert box, what will be shown in the output.

the error message caused by azure api will show as below.
image

adashen (approved):

PR #1493, opened 2021-07-02
fix(bot): do not have detailed error message

YitongFeng-git:
fix bug https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10273591
1. Fix provision/deploy/scaffold operation failed do not have detail error message
2. Fix unhandled error telemetry do not have detail error message

### DIFF
--- a/packages/fx-core/src/plugins/resource/bot/index.ts
+++ b/packages/fx-core/src/plugins/resource/bot/index.ts
@@ -145,6 +145,45 @@ export class TeamsBot implements Plugin {
     );
   }
 
+  private wrapError(
+    e: any,
+    context: PluginContext,
+    sendTelemetry: boolean,
+    name: string
+  ): FxResult {
+    if (e.innerError) {
+      e.message += ` Detailed error: ${e.innerError.message}.`;
+      if (e.innerError.response?.data?.errorMessage) {
+        e.message += ` Reason: ${e.innerError.response?.data?.errorMessage}`;
+      }
+    }
+    Logger.error(e.message);
+    if (e instanceof UserError || e instanceof SystemError) {
+      const res = err(e);
+      sendTelemetry && telemetryHelper.sendResultEvent(context, name, res);
+      return res;
+    }
+
+    if (e instanceof PluginError) {
+      const result =
+        e.errorType === ErrorType.System
+          ? ResultFactory.SystemError(e.name, e.genMessage(), e.innerError)
+          : ResultFactory.UserError(e.name, e.genMessage(), e.showHelpLink, e.innerError);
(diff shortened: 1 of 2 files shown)


packages/fx-core/src/plugins/resource/bot/index.ts
```diff
@@ -158,31 +197,7 @@ export class TeamsBot implements Plugin {
...
-        const UnhandledErrorCode = "UnhandledError";
-        sendTelemetry &&
-          telemetryHelper.sendResultEvent(
-            context,
-            name,
-            ResultFactory.SystemError("Got an unhandled error", UnhandledErrorCode)
-          );
-        return ResultFactory.SystemError(e.name, e.message, e);
-      }
+      return this.wrapError(context, sendTelemetry, name);
```
IvanJobs, line 200: Where is the first parameter e?

YitongFeng-git: resolved

IvanJobs (approved):
lgtm.

packages/fx-core/src/plugins/resource/bot/index.ts
```diff
@@ -145,6 +145,45 @@ export class TeamsBot implements Plugin {
...
+      sendTelemetry && telemetryHelper.sendResultEvent(context, name, result);
+      return result;
+    } else {
+      // Unrecognized Exception.
+      const UnhandledErrorCode = "UnhandledError";
+      sendTelemetry &&
+        telemetryHelper.sendResultEvent(
+          context,
+          name,
+          ResultFactory.SystemError(UnhandledErrorCode, `Got an unhandled error: ${e.message}`)
```
hund030, line 181: It's not recommend to send telemetry with unhandled error messages. We can't ensure the message doesn't contain user information.

packages/fx-core/src/plugins/resource/bot/index.ts
```diff
@@ -145,6 +145,45 @@ export class TeamsBot implements Plugin {
...
+    sendTelemetry: boolean,
+    name: string
+  ): FxResult {
+    if (e.innerError) {
+      e.message += ` Detailed error: ${e.innerError.message}.`;
+      if (e.innerError.response?.data?.errorMessage) {
+        e.message += ` Reason: ${e.innerError.response?.data?.errorMessage}`;
+      }
+    }
+    Logger.error(e.message);
```
hund030, line 160: I think solution will log the e.message since we explicitly return err(e). Please make sure there are no duplicated log. In my opinion, we only need to log the innerError in plugin.

PR #1494, opened 2021-07-02
chore(login): prepare msal-node-extension to encrypt/decrypt token

swatDong:
Per security requirement, local token cache should be protected.
This implementation uses `msal-node-extensions` to create protected local token cache, and this PR is the first one to prepare dependencies and does not break any existing flow.

@xiaolang124 - please help review the new APIs.
@Alive-Fish @chagong - please help review the package.json change in cli
@1openwindow - please help review the package.json change in vscode-extension

### DIFF
--- a/packages/cli/package-lock.json
+++ b/packages/cli/package-lock.json
@@ -259,6 +259,27 @@
         "uuid": "^8.3.0"
       }
     },
+    "@azure/msal-node-extensions": {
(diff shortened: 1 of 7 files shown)


xiaolang124 (approved):

Alive-Fish (approved):
Look good to me.

1openwindow (approved):

PR #1497, opened 2021-07-02
refactor(localdebug): dialog->ut

jayzhang:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10283222

### DIFF
--- a/packages/fx-core/src/plugins/resource/localdebug/certificate.ts
+++ b/packages/fx-core/src/plugins/resource/localdebug/certificate.ts
@@ -5,12 +5,9 @@
 import * as fs from "fs-extra";
 import {
   ConfigFolderName,
-  Dialog,
-  DialogMsg,
-  DialogType,
   LogProvider,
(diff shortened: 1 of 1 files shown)


packages/fx-core/src/plugins/resource/localdebug/certificate.ts
```diff
@@ -298,25 +295,19 @@ export class LocalCertificateManager {
...
-          )
-        ).getAnswer();
+        const res = await this.ui.showMessage(
+          "info",
+          confirmMessage,
+          true,
+          learnMoreText,
+          continueText
+        );
+        userSelected = res.isOk() ? res.value : undefined;
```
swatDong, line 308: Does `undefined` mean cancel/close ?

jayzhang: yes

swatDong (approved):

PR #1499, opened 2021-07-02
fix(cli): make cli case insensitive for param name and option value

jiayi-ruan:
Main fix: Make cli case insensitive for param name and option value
Related task: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/9721031
Notice there are some known issues due to yargs framework limitation:
1. command , subcommand and positional param doesn't support case insensitive. 
2. If user input multi value for a multi-select param and the param name is not lowercase, it will throw error. like:
image

while other cases are supported, like:
image
image

Other fix:
1. Fix bug: when the given option value is not aligned with the dynamic option value, throw error
2. Hide "samples" and "scratch" param in "teamsfx new -h" view because of duplication of the "teamsfx new template -h"

### DIFF
--- a/packages/cli/src/cmds/new.ts
+++ b/packages/cli/src/cmds/new.ts
@@ -11,23 +11,13 @@ import { Argv, Options } from "yargs";
 import * as uuid from "uuid";
 import { glob } from "glob";
 
-import {
-  FxError,
-  err,
-  ok,
-  Result,
-  Question,
-  LogLevel,
-  Stage,
-} from "@microsoft/teamsfx-api";
+import { FxError, err, ok, Result, Question, LogLevel, Stage } from "@microsoft/teamsfx-api";
 
-import activate  from "../activate";
+import activate from "../activate";
 import * as constants from "../constants";
 import { NotFoundInputedFolder, SampleAppDownloadFailed, ProjectFolderExist } from "../error";
 import { YargsCommand } from "../yargsCommand";
-import {
-  getSystemInputs,
-} from "../utils";
+import { getSystemInputs } from "../utils";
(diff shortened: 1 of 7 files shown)


packages/cli/src/cmds/new.ts
```diff
@@ -56,7 +47,7 @@ export default class New extends YargsCommand {
         .options({
           interactive: {
             description: "Select the options interactively",
-            boolean: true,
+            boolean: booleanParamNames.has("interactive"),
             default: true,
```
Alive-Fish, line 51: I think `true` is enough. Because if you let `boolean` be from `booleanParamNames`, you also need to let `default` be as it.

Alive-Fish:
Another thing: the design of `helpParamGenerator.ts` may need to adjust, because it needs to be initialized first but the interfaces allow users to call other functions first.

packages/cli/src/constants.ts
```diff
@@ -107,3 +107,10 @@ export enum CLILogLevel {
 
 export const sqlPasswordQustionName = "sql-password";
 export const sqlPasswordConfirmQuestionName = "sql-confirm-password";
+export const booleanParamNames = new Set<string>([
```
chagong, line 110: If we do this, all newly added boolean options need to update this array, right?

jiayi-ruan: Yes. if not, there will be error if user input capital case parameter name. (but if user input lowercase param name, yargs will help parse it and all will be fine.)

chagong: Can we find another way to judge if an option is a Boolean option? Just don't wanna introduce another hard  code here. Let's discuss it.

PR #1507, opened 2021-07-05
chore(login): revert "prepare msal-node-extension to encrypt/decrypt token (#1494)"

swatDong:
Revert the changes in #1494 
The "@azure/msal-node-extensions" contains native module "dpapi" which may not work in vscode extension. Revert the change and to not break vsix.

### DIFF
--- a/packages/cli/package-lock.json
+++ b/packages/cli/package-lock.json
@@ -259,27 +259,6 @@
         "uuid": "^8.3.0"
       }
     },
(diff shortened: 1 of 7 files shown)


Alive-Fish (approved):
Also, the webpacked package with this package will cause errors.

xiaolang124 (approved):

1openwindow (approved):

PR #1509, opened 2021-07-05
fix: add npm install show error

xiaolang124:
1. When NPM install encounters error, show error notify message and add a report link.

image

image

### DIFF
--- a/packages/vscode-extension/src/debug/teamsfxTaskHandler.ts
+++ b/packages/vscode-extension/src/debug/teamsfxTaskHandler.ts
@@ -121,16 +121,8 @@ async function onDidEndTaskProcessHandler(event: vscode.TaskProcessEndEvent): Pr
             npmInstallLogInfo
           ));
         }
-      } catch (e) {
+      } catch {
         // ignore any error
-        showError(new SystemError(
-          npmInstall,
-          npmInstallErrorMessage,
-          task.name,
-          e.stack,
-          issueLink,
-          e
-        ));
       }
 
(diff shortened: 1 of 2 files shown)


packages/vscode-extension/src/debug/teamsfxTaskHandler.ts
```diff
@@ -110,9 +112,25 @@ async function onDidEndTaskProcessHandler(event: vscode.TaskProcessEndEvent): Pr
...
+            task.name,
+            JSON.stringify(npmInstallLogInfo),
+            issueLink,
+            npmInstallLogInfo
+          ));
         }
-      } catch {
+      } catch (e) {
         // ignore any error
+        showError(new SystemError(
```
swatDong, line 126: This is internal error when getting log file, no need to show the same as above.
Telemetry could be a good improvement, but can be in later PR.

xiaolang124: Already remove inner error show error logic.

swatDong (approved):

kimizhu:
This notification is too simple: Npm install encounters error.
It is better to add more info (at least to tell users the action item is to submit GitHub issue).

kimizhu:
@xiaolang124 here is one example:
https://github.com/OfficeDev/TeamsFx/blob/bf681633893f2e170579d6e524354969310e824e/packages/fx-core/src/plugins/resource/bot/resources/messages.ts#L34

PR #1511, opened 2021-07-05
test: add unit test for extension - part I

HuihuiWu-Microsoft:


packages/vscode-extension/test/index.integration.ts
```diff
@@ -15,7 +15,7 @@ export async function run(): Promise<void> {
     reporterOptions: {
       reporterEnabled: "spec, mocha-junit-reporter",
       mochaJunitReporterReporterOptions: {
-        mochaFile: path.resolve(__dirname, "..", "..", "test-results.nofolder.xml"),
+        mochaFile: path.resolve(__dirname, "..", "..", "test-results.integration.xml"),
```
1openwindow, line 18: do we have integration test now?

packages/vscode-extension/src/controls/webviewPanel.ts
```diff
@@ -69,7 +65,7 @@ export class WebviewPanel {
         // Enable javascript in the webview
         enableScripts: true,
         retainContextWhenHidden: true,
-        localResourceRoots: [vscode.Uri.file(path.join(this.extensionPath, "out"))],
+        localResourceRoots: [vscode.Uri.file(path.join(ext.context.extensionPath, "out"))],
```
1openwindow, line 68: is any side effect on webpack. such as the path is different after webpack?

packages/vscode-extension/test/index.integration.ts
HuihuiWu-Microsoft: Currently there's one. The original structure is <nofolder> = <unit test no folder> + <integration test no folder>, I separated according to unit test and integration test now. Should I clear the integration part in this PR?

1openwindow (approved):

packages/vscode-extension/src/controls/webviewPanel.ts
HuihuiWu-Microsoft: Tested and works well

PR #1514, opened 2021-07-05
fix(localdebug): update run icon display status when the active text editor is changed

XiaofuHuang:
update run icon display status when the active text editor is changed


packages/vscode-extension/src/debug/runIconHandler.ts
```diff
@@ -22,6 +22,7 @@ export async function selectAndDebug(args?: any[]): Promise<Result<null, FxError
 }
 
 export function registerRunIcon(): void {
+  ext.context.subscriptions.push(vscode.window.onDidChangeActiveTextEditor(enableRunIcon));
```
swatDong, line 25: `isValidProject` is a bit heavy (contains I/O operation) so avoid to call it frequently.
How about only validate project when `onDidChangeWorkspaceFolders` and `onDidSaveTextDocument` of any *.fx/* file?

XiaofuHuang: Good idea, All the condition has been changed.

swatDong: `isValidProject` is required for correctly checking the whole project status, but just avoid calling it frequently.
Some options or you can add more:
1. consider `vscode.workspace.createFileSystemWatcher` to catch *.fx* changes, even out of vscode
2. set a timer to call `isValidProject` and save current status, and `onDidXxx` just refer the status flag

XiaofuHuang: According to the [document](https://code.visualstudio.com/api/references/vscode-api), the FileSystemWatcher cannot track the folder change.
> Note that only files within the current workspace folders can be watched. Note that when watching for file changes such as '*/.js', notifications will not be sent when a parent folder is moved or deleted (this is a known limitation of the current implementation and may change in the future).

XiaofuHuang: Add a cache. If the editor changed in 5s, the project status will not be checked.

packages/vscode-extension/src/debug/runIconHandler.ts
```diff
@@ -22,12 +32,35 @@ export async function selectAndDebug(args?: any[]): Promise<Result<null, FxError
 }
 
 export function registerRunIcon(): void {
-  ext.context.subscriptions.push(vscode.workspace.onDidOpenTextDocument(enableRunIcon));
+  ext.context.subscriptions.push(vscode.window.onDidChangeActiveTextEditor(enableRunIcon));
   ext.context.subscriptions.push(vscode.workspace.onDidChangeWorkspaceFolders(enableRunIcon));
   enableRunIcon();
 }
 
+const projectStatusCache = new NodeCache({ stdTTL: 5 });
```
swatDong, line 40: 5 seconds? a bit small..

swatDong (approved):

packages/vscode-extension/src/debug/runIconHandler.ts
```diff
@@ -1,14 +1,25 @@
...
+import * as path from "path";
+import { ExtTelemetry } from "../telemetry/extTelemetry";
+import { TelemetryEvent, TelemetryProperty } from "../telemetry/extTelemetryEvents";
 
-export async function selectAndDebug(args?: any[]): Promise<Result<null, FxError>> {
+export async function selectAndDebug(): Promise<Result<null, FxError>> {
   if (ext.workspaceUri && isValidProject(ext.workspaceUri.fsPath)) {
-    vscode.commands.executeCommand("workbench.view.debug");
-    vscode.commands.executeCommand("workbench.action.debug.selectandstart");
+    await vscode.commands.executeCommand("workbench.view.debug");
```
HuihuiWu-Microsoft, line 21: Do we need to await here, what is the point?

packages/vscode-extension/src/debug/runIconHandler.ts
```diff
@@ -22,12 +33,51 @@ export async function selectAndDebug(args?: any[]): Promise<Result<null, FxError
 }
 
 export function registerRunIcon(): void {
-  ext.context.subscriptions.push(vscode.workspace.onDidOpenTextDocument(enableRunIcon));
-  ext.context.subscriptions.push(vscode.workspace.onDidChangeWorkspaceFolders(enableRunIcon));
-  enableRunIcon();
+  ext.context.subscriptions.push(
```
HuihuiWu-Microsoft, line 36: Is it better to just set the status of whether project is valid teamsfx project and don't care about whether active text editor changed. Do you use onDidChangeActiveTextEditor() here for situations when user checkout code or something else?

XiaofuHuang: Yes, the user may change the project, and the run icon status will be changed.
When the project is open, the project status will be force checked

packages/vscode-extension/src/debug/runIconHandler.ts
XiaofuHuang: Here we execute two commands, if we do not use await, the command `workbench.view.debug` may be execute later than `workbench.action.debug.selectandstart`, which may case the options to be closed.

HuihuiWu-Microsoft (approved):

PR #1516, opened 2021-07-06
chore(sdk): remove default log level

tecton:
1. No default log level and logger.
2. If user set log level, we’ll update the log level and set a default logger that outputs to console.
3. If user set logger, SDK will use that logger. (No output even if user set logger but no log level)
4. If user set logFunction, the callback function will handle each message.


packages/sdk/src/util/logger.ts
```diff
@@ -128,7 +134,7 @@ class InternalLogger {
 export const internalLogger: InternalLogger = new InternalLogger();
 
 /**
- * Set custom logger. Use the output function if it's set. Priority is higher than setLogFunction.
+ * Set custom logger. Use the output function if it's set. Priority is lower than setLogFunction.
```
blackchoey, line 137: Add instructions that people also need to set log level?

tecton: Updated example and README.

packages/sdk/README.md
```diff
@@ -238,37 +238,37 @@ dialogs.add(
...
 Set log level using the snippet below:
 
 ```ts
 // Only need the warning and error messages.
 setLogLevel(LogLevel.Warn);
+// Can also set another logger if you want to redirect to Application Insights in Azure Function
+setLogger(context.log);
 ```
 
+#### Enable log by setting log function
```
blackchoey, line 256: Mention setLogLevel won't take effect (both the log level filter and default logger) when setLogFunction executed?
By the way, I'm curious why the log level filter does not apply to setLogFunction... Maybe it's discussed before but I lost it. Can you share again?

blackchoey:
Also, is there any conclusion on how should sdk bump the minor version in this release?

packages/sdk/README.md
tecton: Oh I misunderstood it, I'll fix this one. Thanks.

wenytang-ms (approved):

wenytang-ms:
it's better to reply the sprint feature email to let release manager know every sprint's feature.
so it's better to bump up patch version as disscussed

blackchoey (approved):
Seems we need more test cases for the changes. Please help make sure related tests are covered manually in this Sprint. And we may add unit tests later.

PR #1519, opened 2021-07-06
feat(function): register required RPs for function app in provision step

hund030:
Have tested below scenario and all get expected result.

<body lang=EN-US style='tab-interval:.5in;word-wrap:break-word'>

  | User can create resource but is not granted to register   resource provider | User have full permission to the subs
-- | -- | --
Required RPs have been registered | Pass | Pass
Required RPs are in Unregistering, Registering or   Unregistered status. | Failed with   RegisterResourceProviderError | Pass

</body>

### DIFF
--- a/packages/fx-core/src/plugins/resource/function/resources/errors.ts
+++ b/packages/fx-core/src/plugins/resource/function/resources/errors.ts
@@ -3,7 +3,7 @@
 import * as path from "path";
 import { ConfigFolderName, SystemError, UserError } from "@microsoft/teamsfx-api";
 
-import { FunctionPluginPathInfo as PathInfo } from "../constants";
+import { AzureInfo, FunctionPluginPathInfo as PathInfo } from "../constants";
 import { Logger } from "../utils/logger";
 
 export enum ErrorType {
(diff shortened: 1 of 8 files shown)


packages/fx-core/src/plugins/resource/function/utils/azure-client.ts
```diff
@@ -70,6 +83,30 @@ export class AzureLib {
     return _t;
   }
 
+  public static async findResourceProvider(
+    client: Providers,
+    namespace: string
+  ): Promise<Provider | undefined> {
+    const provider = await client.get(namespace);
```
henzhang-ms, line 90: Question: is it possible that user has no access to ResourceProvider?

henzhang-ms (approved):

packages/fx-core/src/plugins/resource/function/resources/errors.ts
```diff
@@ -132,6 +135,18 @@ export class UnzipError extends FunctionPluginError {
...
 }
 
+// TODO: help link for the error
+export class RegisterResourceProviderError extends FunctionPluginError {
+  constructor() {
+    super(
+      ErrorType.User,
+      "RegisterResourceProviderError",
+      "Failed to register required resource provider for function app.",
+      [tips.registerRequiredRP, tips.checkLog]
```
henzhang-ms, line 145: Nit: due to the access control, better to mention that user should have enough rights

packages/fx-core/src/plugins/resource/function/utils/azure-client.ts
hund030: Good question. I didn't find any permission named `**/ResourceProvider/read` and provision succeed even I limit my permissions to only ResourceGroup/Read&write, Web/* and Storage/*. I assume user can always get resource provider.

eriolchan (approved):

PR #1521, opened 2021-07-06
chore(sql): add access error help

xzf0587:
1. add sql access error when check database user.
2. improve doc for access error.


xzf0587:
the error message of access server will be changed as below. 
image
the help link will be 
image

adashen (approved):

PR #1534, opened 2021-07-07
feat(collaborate): improve local collaborate experience

KennethBWSong:
1. add key for secret
2. check context and userdata version


swatDong (approved):
Local debug part just format change...

packages/fx-core/src/common/tools.ts
```diff
@@ -90,18 +119,29 @@ export function objectToConfigMap(o?: Json): ConfigMap {
 }
 
 const SecretDataMatchers = [
+  "solution.localDebugTeamsAppId",
+  "solution.teamsAppTenantId",
   "fx-resource-aad-app-for-teams.clientSecret",
   "fx-resource-aad-app-for-teams.local_clientSecret",
+  "fx-resource-aad-app-for-teams.local_clientId",
```
SLdragon, line 126: The naming seems not consistent, some parts of the name using “local_" and the other use "localXXXX"， is there a reason for that?

packages/fx-core/src/common/tools.ts
```diff
@@ -110,23 +150,28 @@ export function sperateSecretData(configJson: Json): Record<string, string> {
     if (!resourceConfig) continue;
     if ("*" !== item) {
       const configValue = resourceConfig[item];
```
SLdragon, line 152: Can we use this to avoid type conversion in if condition?
```
const configValue = resourceConfig[item] as string;
```

packages/fx-core/src/common/tools.ts
```diff
@@ -110,23 +150,28 @@ export function sperateSecretData(configJson: Json): Record<string, string> {
...
-        res[keyName] = configValue;
+        dict[keyName] = configValue;
         resourceConfig[item] = `{{${keyName}}}`;
       }
     } else {
       for (const itemName of Object.keys(resourceConfig)) {
         const configValue = resourceConfig[itemName];
-        if (configValue !== undefined) {
+        if (
+          configValue !== undefined &&
```
SLdragon, line 165: We can add a helper function for this if condition because it appear multiple times

packages/fx-core/src/common/tools.ts
```diff
@@ -162,6 +207,36 @@ export function mergeSerectData(dict: Record<string, string>, configJson: Json):
...
+    const resourceConfig: any = configJson[resourceId];
+    if (resourceConfig) {
+      if ("*" !== item) {
+        const configValue = resourceConfig[item];
+        if (configValue) {
+          delete resourceConfig[item];
+        }
+      } else {
+        for (const itemName of Object.keys(resourceConfig)) {
+          if (resourceConfig[item]) {
```
SLdragon, line 226: Did you mean that:
```
resourceConfig[item] -> resourceConfig[itemName] 
```

packages/fx-core/src/core/middleware/configWriter.ts
```diff
@@ -5,42 +5,50 @@
...
       if (!solutionSettings.activeResourcePlugins) solutionSettings.activeResourcePlugins = [];
       if (!solutionSettings.azureResources) solutionSettings.azureResources = [];
       const envName = solutionContext.projectSettings?.currentEnv;
       const solutionConfig = solutionContext.config;
       const configJson = mapToJson(solutionConfig);
       const envJsonFile = path.resolve(confFolderPath, `env.${envName}.json`);
       const userDataFile = path.resolve(confFolderPath, `${envName}.userdata`);
-      const localData = sperateSecretData(configJson);
+      const localData: Record<string, string> = {};
+      sperateSecretData(localData, configJson);
```
SLdragon, line 51: How about return result instead of using reference to store result.

packages/fx-core/src/common/tools.ts
KennethBWSong: For some reason some plugins did not follow "local_" type. We may fix this when moving to Arm.

packages/fx-core/src/core/middleware/projectUpgrader.ts
```diff
@@ -0,0 +1,131 @@
...
+          "[core]: context version is too low. Will update context and move some config from env to userdata."
+        );
+      }
+
+      // Move info from env.default.json to userdata
+      sperateSecretData(userData, env);
+      await saveContextAndUserData(confFolderPath, envName as string, env, userData);
+    }
+
+    return;
```
SLdragon, line 85: This return is not necessary

PR #1545, opened 2021-07-07
refactor(appstudioplugin): migrate glue codes from Solution to App Studio Plugin

yiqing-zhao:
1. Add error definition in `packages/fx-core/src/plugins/resource/appstudio/errors.ts`
2. Move `IAppDefinition` from Solution to `packages/fx-core/src/plugins/resource/appstudio/index.ts`
3. Migrate `createManifestForRemote()` and `createAndConfigTeamsManifest()` to App Studio Plugin
4. Read and write `REMOTE_TEAMS_APP_ID` to config of app studio plugin context, instead of solutionConfig
5. Slightly modify unit test to suit changes
6. Remove redundant codes and comments


packages/fx-core/src/plugins/resource/appstudio/plugin.ts
```diff
@@ -345,6 +356,260 @@ export class AppStudioPluginImpl {
...
+        ctx.logProvider,
+        ctx.root
+      );
+      if (result.isErr()) {
+        return result.map((_) => appDefinition);
+      }
+
+      ctx.logProvider?.info(`Teams app created ${result.value}`);
+      appDefinition.appId = result.value;
+      ctx.config.set(REMOTE_TEAMS_APP_ID, result.value);
```
nliu-ms, line 465: @1yefuwang1 , will this cause compatibility issue?

This config is moved from solution to plugin, then the configurations in env.default.json file is changed. So the old project may be effected?

1yefuwang1: That's a good point. Maybe we can find a way to sync appstudioPlugin.REMOTE_TEAMS_APP_ID to solution.REMOTE_TEAMS_APP_ID to maintain backward compatibility.

yiqing-zhao: I add REMOTE_TEAMS_APP_ID to solution after the appstudio plugin write to the config.
```ts
        const result = await appStudioPlugin.createAndConfigTeamsManifest(
          pluginCtx,
          maybeSelectedPlugins
        );
        if (result.isErr()) {
          return err(result.error);
        }
        ctx.config.get(GLOBAL_CONFIG)?.set("remoteTeamsAppId", result.value.appId);
```

packages/fx-core/src/plugins/resource/appstudio/plugin.ts
```diff
@@ -55,23 +64,22 @@ import { REMOTE_TEAMS_APP_ID, REMOTE_MANIFEST } from "../../solution/fx-solution
...
-  ): Promise<IAppDefinition | undefined> {
-    return await AppStudioClient.createApp(
-      appDefinition,
-      appStudioToken,
-      logProvider,
-      colorIconContent,
-      outlineIconContent
-    );
-  }
+export enum PluginNames {
```
nliu-ms, line 67: import PluginNames, do not define it again.

yiqing-zhao: Thanks! I have moved PluginNames to /solution/constants.ts, and changed a few files that originally import PluginNames from solution.

1yefuwang1 (approved):

nliu-ms (approved):

PR #1546, opened 2021-07-07
chore: change AIKey in source to testing AIKey and add script for CD

chagong:
chore: change AIKey in source to testing AIKey and add script for CD

https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10308939


wenytang-ms:
could these AIKeys be checkout into code ?

chagong:
> could these AIKeys be checkout into code ?
We could. But for CLI public key it may changed after onboarding to DataX. I just don't wanna create PRs to update those keys.

wenytang-ms:
> > could these AIKeys be checkout into code ?
> > We could. But for CLI public key it may changed after onboarding to DataX. I just don't wanna create PRs to update those keys.

so can we commit the AI keys change into code first, and just restore these codes later ?

wenytang-ms:
https://github.com/marketplace/actions/github-action-for-package-json-update 
this GitHub cation seems can update the AI key, so it's no need to use other script

chagong:
> https://github.com/marketplace/actions/github-action-for-package-json-update
> this GitHub cation seems can update the AI key, so it's no need to use other script

Then let me remove that script.

1openwindow (approved):

Alive-Fish (approved):

PR #1548, opened 2021-07-08
chore(login): persist token cache with encryption

swatDong:
This is an improvement for security requirement - token file should be encrypted. Since Windows Credential Manager has secret size limitation, and `msal-node-extensions` cannot be used in VSCode extension (due to cross-platform issue), this PR provides a solution to use `keytar` to store a secret and use *"aes-256-gcm"* with the secret to encrypt/decrypt token file.

For VSCode extension, VSCode has built-in keytar support.
For CLI, keytar itself has cross-platform support.

Kindly try the attachments in [this work item](https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10262346/) if you are interested in the new login implementation.

@xiaolang124 - please help review the main cache access and encryption part.
@Alive-Fish @chagong - add keytar dependency in CLI
@1openwindow - add keytar dependency in vscode-extension


xiaolang124 (approved):

Alive-Fish (approved):
look good to me

1openwindow (approved):

PR #1549, opened 2021-07-08
chore: add cz cli for conventional-commits

wenytang-ms:
add cz for conventional-commits

### DIFF
--- a/.husky/exec-git-cz.cmd
+++ b/.husky/exec-git-cz.cmd
@@ -1,0 +1,19 @@
+@ECHO off
+SETLOCAL
+
+CALL :find_dp0
+set "_prog=%dp0%..\node_modules\.bin\git-cz.cmd"
+IF NOT EXIST "%_prog%" (
+    ECHO "%_prog% does not exist, please install commitizen"
+    ENDLOCAL
+    EXIT /b 1
+)
+
+call "%_prog%" %*
+
+ENDLOCAL
+EXIT %errorlevel%
+
+:find_dp0
+SET dp0=%~dp0
+EXIT /b
--- a/.husky/prepare-commit-msg
+++ b/.husky/prepare-commit-msg
@@ -1,0 +1,7 @@
+#!/bin/sh
+VAR="$(uname -a)"
+if [[ $VAR == *"Msys"* || $VAR == *"uname"* ]] ; then
+  ./.husky/prepare-commit-msg.cmd
+else
+  exec < /dev/tty && npx --no-install git cz --hook || true
+fi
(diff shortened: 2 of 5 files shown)


wenytang-ms:
this pr enable when developers run git commit -m “……” will trigger an prepare commit msg hook, and cz-cli will give then a conventional-commit rule to select, just like 
image

qinezh:
If I have already used the correct commit message format like `git commit -m "chore: add cz cli for conventional-commits"`, will the hook still be triggered?

.husky/exec-git-cz.cmd
```diff
@@ -0,0 +1,19 @@
+@ECHO off
```
qinezh, line 1: Looks only *.cmd files checked in, will it work on Linux/macOS?

wenytang-ms: default will run by prepare-commit-msg file, in prepare-commit-msg will judge which platform, windows will run cmd, and *nix will run prepare-commit-msg file

wenytang-ms:
> If I have already used the correct commit message format like `git commit -m "chore: add cz cli for conventional-commits"`, will the hook still be triggered?

yes, besides the normal content, cz-cli help to add scope(like "api" or "sdk"), verify import change(to bump up major version), verify issue link.
if we wanna skip this hook with full message, ctr+c will skip this step and go on git commit normally

.husky/exec-git-cz.cmd
qinezh: got it, thanks for the explanation.

qinezh:
> > If I have already used the correct commit message format like `git commit -m "chore: add cz cli for conventional-commits"`, will the hook still be triggered?
> 
> yes, besides the normal content, cz-cli help to add scope(like "api" or "sdk"), verify import change(to bump up major version), verify issue link.
> if we wanna skip this hook with full message, ctr+c will skip this step and go on git commit normally

The dev experience seems to be not so good for me, if I have already used the right commit message format (including the scope). 

As for the issue link, we could add it in the GitHub PR page and it's not the scope of conventional commits from my understanding:
image

wenytang-ms:
> > > If I have already used the correct commit message format like `git commit -m "chore: add cz cli for conventional-commits"`, will the hook still be triggered?
> > 
> > 
> > yes, besides the normal content, cz-cli help to add scope(like "api" or "sdk"), verify import change(to bump up major version), verify issue link.
> > if we wanna skip this hook with full message, ctr+c will skip this step and go on git commit normally
> 
> The dev experience seems to be not so good for me, if I have already used the right commit message format (including the scope).
> 
> As for the issue link, we could add it in the GitHub PR page and it's not in the scope of conventional commits from my understanding:
> image

ok, it seems a personal choice, as a mandatory project tool is not a better choice, so close this pr.
maybe @LongOddCode  can share this as best practice to team

PR #1550, opened 2021-07-08
fix: change graph login to 1st party

xiaolang124:
1. Change graph login to 1st party application.


swatDong:
Any change on UI/solution side? E.g., when signin is triggered via tree view, which one is called, `graphLogin` or `appStudioLogin`?

xiaolang124:
> Any change on UI/solution side? E.g., when signin is triggered via tree view, which one is called, `graphLogin` or `appStudioLogin`?

For this pr, it only fixes APIM graph login problem. Other scenarios, like login graph will have UI update problems. I will do more work later.

swatDong (approved):

PR #1553, opened 2021-07-08
chore: support publish package with ai key update

wenytang-ms:


chagong (approved):

.github/workflows/cd.yml
```diff
@@ -115,15 +115,42 @@ jobs:
         run: |
           npx lerna version --no-private --conventional-commits --conventional-prerelease --preid=rc --no-changelog --yes
 
+      - name: version stable npm packages to npmjs.org
+        if: ${{ github.event_name == 'workflow_dispatch' && github.ref == 'refs/heads/main' && github.event.inputs.preid == 'stable' }}
+        run: |
+          npx lerna version --no-private --conventional-commits --conventional-graduate --no-changelog --yes
+
+      - name: update cli ai key
```
adashen, line 123: the two steps share the same condition cannot be written in one step?

.github/workflows/cd.yml
```diff
@@ -115,15 +115,42 @@ jobs:
...
         run: |
           npx lerna publish from-package --dist-tag=rc --yes
 
-      - name: release stable npm packages to npmjs.org
-        if: ${{ github.event_name == 'workflow_dispatch' && github.ref == 'refs/heads/main' && github.event.inputs.preid == 'stable'}}
+      - name: publish stable npm packages to npmjs.org
+        if: ${{ github.event_name == 'workflow_dispatch' && github.ref == 'refs/heads/main' && github.event.inputs.preid == 'stable' }}
         run: |
-          npx lerna publish --no-private --conventional-commits --conventional-graduate --no-changelog --yes
+          npx lerna publish from-package --yes
```
adashen, line 153: by default the dist tag is latest right?

wenytang-ms: yes

.github/workflows/cd.yml
wenytang-ms: this step use action can only modify one file once. or rewrite this action for two files' field change.

qinezh (approved):

adashen (approved):

PR #1563, opened 2021-07-09
fix(vscode): security alerts of trim-newlines by upgrading to 3.0.1

tecton:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10109465


HuihuiWu-Microsoft (approved):
LGTM

PR #1565, opened 2021-07-09
test: update nyc coverage config

LongOddCode:


packages/sdk/.nycrc
```diff
@@ -1,15 +1,16 @@
...
-    "produce-source-map": true,
-    "instrument": true,
-    "all": true
-}
\ No newline at end of file
+  "extends": "@istanbuljs/nyc-config-typescript",
+  "all": true,
+  "include": [
+    "src/**/*.ts",
+    "src/**/*.js"
```
blackchoey, line 6: can you share what's the purpose of including both ts and js files? Since js is compiled from ts, should we only include one of them?

LongOddCode: Indeed we only develop ts file recently. But some tsconfig.json set "allowJs: true" in our monorepo, which mean we can use javascript file in our typescript project. So I just to be compatible with them.

tecton (approved):

1openwindow (approved):

1yefuwang1 (approved):

Alive-Fish (approved):

PR #1566, opened 2021-07-12
[WIP] feat(vsc, cli, core): Encrypt secrets stored in .userdata file

tecton:
Use symmetric encryption to protect secret data. Since keytar and DPAPI both have some blocking issues, we choose a simplified way that selects project Id + fixed suffix as encryption key.
In VS Code, CodeLens is used to provide a better experience.
In CLI, we can use `teamsfx config get [keyName]` to view and `teamsfx config set [keyName] [keyValue]` command to update.

@jayzhang - added new API in core for crypto
@Alive-Fish @chagong - (new feature) update "config" command in CLI
@1openwindow @HuihuiWu-Microsoft - (new feature) added CodeLens in vscode-extension

VS Code Preview (gif):
vsc-crypto
CLI Preview (gif):
cli-crypto


tecton:
Closing this PR because it's split into 3 PRs:
https://github.com/OfficeDev/TeamsFx/pull/1618
https://github.com/OfficeDev/TeamsFx/pull/1624
https://github.com/OfficeDev/TeamsFx/pull/1630

PR #1575, opened 2021-07-12
fix(cli): set background color for the rest of the line

nliu-ms:
When the output `scrolling the window`, the rest of the line will be filled with color of the first character.

image

Reference: https://stackoverflow.com/a/53741857/16437100

### DIFF
--- a/packages/cli/src/utils.ts
+++ b/packages/cli/src/utils.ts
@@ -298,5 +298,5 @@ export function getColorizedString(message: Array<{ content: string; color: Colo
       }
     })
     .join("");
-  return colorizedMessage + "\u00A0\u001B[K";
+  return colorizedMessage + (process.stdout.isTTY ? "\u00A0\u001B[K" : "");
 }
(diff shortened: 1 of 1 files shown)


packages/cli/src/utils.ts
```diff
@@ -298,5 +298,8 @@ export function getColorizedString(message: Array<{ content: string; color: Colo
       }
     })
     .join("");
-  return colorizedMessage + "\u00A0";
+  // Pad the last line to avoid tailing color
+  const padWidth = (s: string, stream = process.stdout) =>
+    s + " ".repeat(stream.columns - (s.length % stream.columns));
```
Alive-Fish, line 303: If this msg is not beginning at the left of the command line, the padding width will be wrong.

nliu-ms: Do you mean like `[TeamsFx]` at the beggining?

Alive-Fish (approved):

PR #1579, opened 2021-07-12
refactor(cli-preview): refactor task wrapper to support dependency checker

kuojianlu:


a1exwang (approved):

a1exwang (approved):
The task API should be sufficient for env checker

PR #1584, opened 2021-07-13
feat(env-checker): [part-3] add telemetry implementation in CLI env checker

a1exwang:
The telemetry implemenation is similar to the extension's, with just slight differences replacing `ExtTelemetry` with `cliTelemtry`

https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10338893/

### DIFF
--- a/packages/cli/src/cmds/preview/depsChecker/cliTelemetry.ts
+++ b/packages/cli/src/cmds/preview/depsChecker/cliTelemetry.ts
@@ -4,11 +4,13 @@
 import * as os from "os";
 import { performance } from "perf_hooks";
 import { SystemError, UserError } from "@microsoft/teamsfx-api";
+import { TelemetryProperty } from "../../../telemetry/cliTelemetryEvents";
+import cliTelemetryInstance from "../../../telemetry/cliTelemetry";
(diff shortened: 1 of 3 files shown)


kuojianlu (approved):

packages/cli/src/cmds/preview/depsChecker/cliTelemetry.ts
```diff
@@ -32,19 +39,28 @@ export class CLITelemetry implements IDepsTelemetry {
...
 
   public sendSystemErrorEvent(
     eventName: DepsCheckerEvent,
     errorMessage: string,
     errorStack: string
   ): void {
-    // TODO: send system error event
+    const error = new SystemError(
+      eventName,
+      `errorMsg=${errorMessage},errorStack=${errorStack}`,
```
chagong, line 53: why do you want to put callstack into error message?

a1exwang: 1. This code are from the extension. 
2. The reason in the extension is that the error stack is not in the telemetry properties. We put the stack in the error message to work around this issue.

Does CLI telemetry have this issue too? If not, we can remove this.

chagong (approved):

PR #1593, opened 2021-07-13
feat(collaborate): improve local collaborate experience

KennethBWSong:


packages/fx-core/src/common/tools.ts
```diff
@@ -119,13 +119,25 @@ export function objectToConfigMap(o?: Json): ConfigMap {
...
+  "fx-resource-aad-app-for-teams.local_tenantId",
+  "fx-resource-aad-app-for-teams.local_applicationIdUris",
   "fx-resource-simple-auth.filePath",
   "fx-resource-simple-auth.environmentVariableParams",
   "fx-resource-local-debug.*",
   "fx-resource-bot.botPassword",
   "fx-resource-bot.localBotPassword",
+  "fx-resource-bot.localBotId",
+  "fx-resource-bot.localObjectId",
+  "fx-resource-bot.local_redirectUri",
```
blackchoey, line 138: Though related logic should be changed when multi-env support is ready. Is there any way to guarantee resource plugin developer adds local debug related settings here? Or it's just a temporary solution, and we no longer need to consider this when multi-env support is ready?

packages/fx-core/src/core/middleware/projectUpgrader.ts
```diff
@@ -0,0 +1,202 @@
...
+};
+
+export async function upgradeContext(ctx: CoreHookContext): Promise<void> {
+  try {
+    const inputs = ctx.arguments[ctx.arguments.length - 1] as Inputs;
+    if (!inputs.projectPath) {
+      ctx.result = err(NoProjectOpenedError());
+      return;
+    }
+    const projectPathExist = await fs.pathExists(inputs.projectPath);
```
blackchoey, line 50: I'm not sure about the design philosophy about middleware, is these checks (whether project is opened or project path exists) already happened globally? If every middleware needs to check this, there would be many duplicate logic...

packages/fx-core/src/core/middleware/projectUpgrader.ts
```diff
@@ -0,0 +1,202 @@
...
+      // Can quit directly.
+      return;
+    }
+
+    // Some keys updated.
+    // Save the updated context and send log.
+    await saveContext(contextPath, context);
+    const core = ctx.self as FxCore;
+    const logger =
+      core !== undefined && core.tools !== undefined && core.tools.logProvider !== undefined
```
blackchoey, line 78: is `core?.tools?.logProvider` helps?

packages/fx-core/src/common/tools.ts
KennethBWSong: Here is temporary solution to enable collaboration for users in local scenario. The logic will change after multi-env is supported.

packages/fx-core/src/core/middleware/projectUpgrader.ts
KennethBWSong: Here I add the logic since this part may be added to some cmds in cli and I can not make sure whether other middlewares will do the check. Maybe @jayzhang can consider to make the check in a separate middleware?

packages/fx-core/src/core/middleware/projectUpgrader.ts
```diff
@@ -0,0 +1,202 @@
...
+
+function mergeKeysToUserDate(
+  userData: Record<string, string>,
+  updatedKeys: Map<string, any>
+): void {
+  // Move resource context first to userdata
+  let moved = false;
+  for (const item of resourceContext) {
+    // Check whether corresponding secret exists.
+    if (!userData[getUserDataKey(item.plugin, item.secret)]) {
```
blackchoey, line 165: can you share the purpose of checking whether secret exists here?

packages/fx-core/src/core/middleware/projectUpgrader.ts
```diff
@@ -0,0 +1,202 @@
...
+      continue;
+    }
+
+    for (const key of item.relatedKeys) {
+      const userDataKey = getUserDataKey(item.plugin, key);
```
blackchoey, line 170: since the key of `updatedKeys` parameter is already `userDataKey` calculated here, can we just iterate the items in `updatedKeys`?

packages/fx-core/src/core/middleware/projectUpgrader.ts
```diff
@@ -0,0 +1,202 @@
...
+      // Merge will only happen when userData does not contain certain key.
+      // Otherwise, value in userData will be regarded as source of truth.
```
blackchoey, line 172: Though it's very rare case, but this may change local debug behavior. Is PM aware of this?

packages/fx-core/src/core/middleware/projectUpgrader.ts
```diff
@@ -0,0 +1,202 @@
...
+      if (!userData[userDataKey] && updatedKeys.has(userDataKey)) {
+        moved = true;
+        userData[userDataKey] = updatedKeys.get(userDataKey);
+      }
+    }
+  }
+
+  // If any key moved, means at least one secret exists.
```
blackchoey, line 180: Can you share why we only move solution context when at least one secret exists? Sorry that I'm not familiar with the actual local debug scenario.

packages/fx-core/src/core/middleware/projectUpgrader.ts
KennethBWSong: This scenario means that this user has already run local debug with latest extension, and is using a env.default.json that created with old extension (e.g. pulled code from github and the other user is using old extension). We will treat info in userdata as source of truth since this file is generated by this user.

packages/fx-core/src/core/middleware/projectUpgrader.ts
KennethBWSong: Here we want to separate config of aad/bot, thus we still need to add check logic if we iterate the items in updatedKeys. I think the performance should be the same and this way may be easier to understand.

packages/fx-core/src/core/middleware/projectUpgrader.ts
```diff
@@ -0,0 +1,202 @@
...
+  }
+}
+
+async function readUserData(userDataPath: string): Promise<Record<string, string>> {
+  let dict: Record<string, string>;
+  if (await fs.pathExists(userDataPath)) {
+    const dictContent = await fs.readFile(userDataPath, "UTF-8");
+    dict = deserializeDict(dictContent);
+  } else {
+    dict = {};
```
SLdragon, line 107: Move the `dict = {}` to declare sentence?
```
let dict: Record<string, string>= {}
```

KennethBWSong: Fix. Thank you for pointing that out. @jayzhang You may update the same code in other middleware.

packages/fx-core/src/core/middleware/projectUpgrader.ts
KennethBWSong: Fixed. Thank you for pointing that out. @jayzhang You may update the same code in other middleware.

blackchoey (approved):

packages/fx-core/src/core/middleware/projectUpgrader.ts
```diff
@@ -0,0 +1,233 @@
...
+      // Can quit directly.
+      return;
+    }
+
+    // Some keys updated.
+    // Save the updated context and send log.
+    await saveContext(contextPath, context);
+    const core = ctx.self as FxCore;
+    core?.tools?.logProvider?.info(
+      "[core]: context version is too low. Will update context and move some config from env to userdata."
```
SLdragon, line 82: "context version is too low" is confusing for users, how about use "template" instead of "context"?

packages/fx-core/src/core/middleware/projectUpgrader.ts
```diff
@@ -0,0 +1,233 @@
...
+  plugin: "solution",
+  relatedKeys: ["localDebugTeamsAppId", "teamsAppTenantId"],
+};
+
+export const ProjectUpgraderMW: Middleware = async (ctx: CoreHookContext, next: NextFunction) => {
+  await upgradeContext(ctx);
+  await next();
+};
+
+export async function upgradeContext(ctx: CoreHookContext): Promise<void> {
```
SLdragon, line 45: You can add comments of the update flow

tecton (approved):
Encryption part is good for me.

jayzhang (approved):

packages/fx-core/src/core/middleware/projectUpgrader.ts
```diff
@@ -0,0 +1,234 @@
...
+      // Can quit directly.
+      return;
+    }
+
+    // Some keys updated.
+    // Save the updated context and send log.
+    await saveContext(contextPath, context);
+    const core = ctx.self as FxCore;
+    core?.tools?.logProvider?.info(
+      "[core]: template version is too low. Will update context and move some configs from env to userdata."
```
adashen, line 83: context is already updated in the above "saveContext" right? So it is not will....

packages/fx-core/src/core/middleware/projectUpgrader.ts
```diff
@@ -0,0 +1,234 @@
...
+  if (projectId) {
+    const cryptoProvider = new LocalCrypto(projectId);
+    for (const secretKey of Object.keys(dict)) {
+      if (!dataNeedEncryption(secretKey)) {
+        continue;
+      }
+      const secretValue = dict[secretKey];
+      const plaintext = cryptoProvider.decrypt(secretValue);
+      if (plaintext.isErr()) {
+        throw plaintext.error;
```
adashen, line 119: decrypt one field error will cause the whole load fail?

packages/fx-core/src/core/middleware/projectUpgrader.ts
```diff
@@ -0,0 +1,234 @@
...
+    mergeKeysToUserDate(userData, updatedKeys);
+
+    // Save UserData
+    await saveUserData(userDataPath, userData, projectSettings.projectId);
+  } catch (error) {
+    ctx.result = err(ContextUpgradeError(error));
+  }
+}
+
+async function readUserData(
```
adashen, line 100: will this api be provided in core level in stead of projectUpgrader? I thought lots of place will need read user data right?

packages/fx-core/src/core/middleware/projectUpgrader.ts
KennethBWSong: Following logic from @tecton.
If decrypt failed, userdata can not be load correctly, thus I also think we should fail the process here.

packages/fx-core/src/core/middleware/projectUpgrader.ts
KennethBWSong: Not quite sure about this. Maybe @jayzhang can answer this.

packages/fx-core/src/core/middleware/projectUpgrader.ts
```diff
@@ -0,0 +1,234 @@
...
+      if (!userData[userDataKey] && updatedKeys.has(userDataKey)) {
+        moved = true;
+        userData[userDataKey] = updatedKeys.get(userDataKey);
+      }
+    }
+  }
+
+  // If any key moved, means at least one secret exists.
+  // Move solution context.
+  if (moved) {
```
adashen, line 214: if not moved solution keys don't need to be updatd?

adashen (approved):

SLdragon (approved):

PR #1594, opened 2021-07-13
test: App Studio Plugin API combination and Unit Test Addition

yiqing-zhao:
Combine `getConfigForCreatingManifest()`, `getDevAppDefinition()`, `convertToAppDefinition()`, and `updateApp()`

Add unit test for app studio plugin. The coverage of app studio plugin is 70.38%.


packages/fx-core/tests/plugins/resource/appstudio/unit/createManifestForRemote.test.ts
```diff
@@ -0,0 +1,155 @@
...
+    BotPlugin.name = "fx-resource-bot";
+    BotPlugin.displayName = "Bot";
+    selectedPlugins = [BotPlugin];
+  });
+
+  afterEach(() => {
+    sandbox.restore();
+  });
+
+  it("maybeSelectedPlugins error", async () => {
```
1yefuwang1, line 78: BDD style tests are expected to be written in a way that a test can be read out. 
Something like it("should return error")

yiqing-zhao: Thanks for the reminder!  I have already change the testcase names.

1yefuwang1 (approved):

packages/fx-core/src/plugins/resource/appstudio/index.ts
```diff
@@ -77,23 +77,17 @@ export class AppStudioPlugin implements Plugin {
...
+  public async getAppDefinitionAndUpdate(
+    ctx: PluginContext,
     type: "localDebug" | "remote",
-    createIfNotExist: boolean,
-    teamsAppId?: string,
-    appStudioToken?: string,
-    logProvider?: LogProvider,
-    projectRoot?: string
+    manifest: TeamsAppManifest,
+    appStudioToken?: string
```
nliu-ms, line 84: appStudioToken can also be removed from parameter, you can get this token from context.

yiqing-zhao: Thanks! I have removed the appStudioToken from the parameter and get the token from the plugin Context.

nliu-ms (approved):

PR #1599, opened 2021-07-13
fix: cli login auto close timeout

xiaolang124:
1. Fix CLI login problem when successfully login but command hangs for a while.


packages/cli/src/constants.ts
```diff
@@ -76,5 +76,7 @@ export enum CLILogLevel {
 export const sqlPasswordQustionName = "sql-password";
 
 export const sqlPasswordConfirmQuestionName = "sql-confirm-password";
-
 export const deployPluginNodeName = "deploy-plugin";
+
+export const connection = "connection";
```
swatDong, line 81: "connection" and "close" are built-in event type and is defined by node-http APIs, ok to directly use without defining constants... just for code simplify.

``` node
// net.d.ts
on(event: "close", listener: () => void): this;
on(event: "connection", listener: (socket: Socket) => void): this;
```

packages/cli/src/commonlib/azureLogin.ts
```diff
@@ -256,6 +256,7 @@ export class AzureAccountManager extends login implements AzureAccountProvider {
   }
 
   private async doGetAccountCredentialAsync(): Promise<TokenCredentialsBase | undefined> {
+    AzureAccountManager.codeFlowInstance.destroySockets();
```
swatDong, line 259: call `destroySockets` when `server.on("close")`

packages/cli/src/constants.ts
xiaolang124: Remove inappropriate constants.

packages/cli/src/commonlib/azureLogin.ts
xiaolang124: Use ```server.on("close")``` listener.

swatDong (approved):

PR #1607, opened 2021-07-14
fix: teamsfx config get will show all configs' values

Alive-Fish:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10348493

### DIFF
--- a/packages/cli/src/cmds/config.ts
+++ b/packages/cli/src/cmds/config.ts
@@ -11,14 +11,14 @@ import CLILogProvider from "../commonlib/log";
 
 export class ConfigGet extends YargsCommand {
   public readonly commandHead = `get`;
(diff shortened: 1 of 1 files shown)


packages/cli/src/cmds/config.ts
```diff
@@ -11,14 +11,14 @@ import CLILogProvider from "../commonlib/log";
 
 export class ConfigGet extends YargsCommand {
   public readonly commandHead = `get`;
-  public readonly command = `${this.commandHead} [option...]`;
+  public readonly command = `${this.commandHead} <option>`;
```
Alive-Fish, line 14: Only changes this line. Others are all formatted.

jiayi-ruan (approved):

PR #1608, opened 2021-07-14
chore: add templates to lerna

wenytang-ms:


templates/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "templates",
+  "version": "0.1.3",
+  "private": "true",
```
qinezh, line 4: why the license is MIT while private is true?

.github/workflows/cd.yml
```diff
@@ -110,18 +81,18 @@ jobs:
       - name: release alpha npm packages to npmjs.org
         if: ${{ github.event_name == 'schedule' || (github.event_name == 'workflow_dispatch' && github.ref == 'refs/heads/dev' && github.event.inputs.preid == 'alpha')}}
         run: |
-          npx lerna version --no-private prerelease --preid=alpha.$(git rev-parse --short HEAD) --no-push --allow-branch dev --yes
+          npx lerna version prerelease --preid=alpha.$(git rev-parse --short HEAD) --no-push --allow-branch dev --yes
```
adashen, line 84: why remove --no-private?

.github/scripts/template-zip-autogen.sh
```diff
@@ -1,6 +1,15 @@
 #!/usr/bin/env bash
 set -x
 
+if git diff HEAD^ ./templates/package.json | grep version
+    then
```
henzhang-ms, line 5: Please reformat, indent is incorrect.

BTW, to maintain this script's function focus on zip generation. Can we move this section for version check in another place, like cd.yml?

.github/workflows/cd.yml
```diff
@@ -155,6 +126,49 @@ jobs:
...
+        uses: marvinpinto/action-automatic-releases@latest
+        with:
+          repo_token: ${{ secrets.CD_PAT }}
+          prerelease: true
+          automatic_release_tag: steps.zip-template.outputs.VERSION
+          files: |
+            ./*.zip
+
+      - name: Create Templates Stable Release
+        if: ${{ steps.zip-template.outputs.VERSION != '0' && github.event.inputs.preid == 'stable' }}
```
henzhang-ms, line 147: how does this action get version?

.github/workflows/cd.yml
wenytang-ms: private in package.json set true, will skip publish this package, but version is still work.
version with --no-private will skip version this package.
now templates add in lerna, lerna will version templates, but no need to publish it.

.github/scripts/template-zip-autogen.sh
wenytang-ms: sure, deleted.

.github/workflows/cd.yml
wenytang-ms: add package in templates folder, lerna will manage to bump up templates version.

henzhang-ms: I believe so. But, for this action, how does it know the release tag for templates? According to the doc, `tag: An optional tag for the release. If this is omitted the git ref will be used (if it is a tag)`, if the tag parameter is omitted, it uses git ref as tag. Does this workflow has that ref for it?

wenytang-ms: template version will use steps.version-change.outputs.CHANGED template as the true version
these varibles work success on https://github.com/wenytang-ms-123/testavc/actions/workflows/cd.yml

.github/workflows/cd.yml
```diff
@@ -160,6 +131,53 @@ jobs:
...
+          artifacts: ./*.zip
+          name: 'Release for $(tr ' ' '\n' <<< '${{ steps.version-change.outputs.CHANGED }}' | grep 'templates@')'
+          token: ${{ secrets.github_token }}
+          tag: $(tr ' ' '\n' <<< '${{ steps.version-change.outputs.CHANGED }}' | grep 'templates@')
+          allowUpdates: true
+
+      - name: Generate Tag List
+        if: ${{ contains(steps.version-change.outputs.CHANGED, 'templates@') && github.event.inputs.preid == 'stable' }}
+        run: |
+          tr ' ' '\n' <<< '${{ steps.version-change.outputs.CHANGED }}' | grep 'templates@' > ./template-tags.txt
```
henzhang-ms, line 165: The ./template-tags.txt should contains all template versions instead of the newest one.

.github/workflows/cd.yml
```diff
@@ -160,6 +131,53 @@ jobs:
...
+
+      - name: Update Template Tag list Release
+        if: ${{ contains(steps.version-change.outputs.CHANGED, 'templates@') && github.event.inputs.preid == 'stable' }}
```
henzhang-ms, line 168: please update the tag list even it is not stable, in case we have other prerelease suffix.

.github/workflows/cd.yml
wenytang-ms: thanks

.github/workflows/cd.yml
wenytang-ms: but this action only release stable template can trigger it

.github/workflows/cd.yml
```diff
@@ -160,6 +131,53 @@ jobs:
...
+        if: ${{ contains(steps.version-change.outputs.CHANGED, 'templates@') && github.event.inputs.preid != 'stable' }}
+        uses: marvinpinto/action-automatic-releases@latest
+        with:
+          repo_token: ${{ secrets.CD_PAT }}
+          prerelease: true
+          automatic_release_tag: "templates@0.0.0-${{github.event.inputs.preid}}"
+          files: |
+            ./*.zip
+
+      - name: Create Templates Stable Release
```
wenytang-ms, line 152: following is the stable template release actions

.github/workflows/cd.yml
```diff
@@ -160,6 +131,53 @@ jobs:
...
+        if: ${{ contains(steps.version-change.outputs.CHANGED, 'templates@') }}
+        run: .github/scripts/template-zip-autogen.sh
+
+      - name: release templates' prerelease version to github
+        if: ${{ contains(steps.version-change.outputs.CHANGED, 'templates@') && github.event.inputs.preid != 'stable' }}
+        uses: marvinpinto/action-automatic-releases@latest
+        with:
+          repo_token: ${{ secrets.CD_PAT }}
+          prerelease: true
+          automatic_release_tag: "templates@0.0.0-${{github.event.inputs.preid}}"
```
wenytang-ms, line 148: alpha and rc version still keep 0.0.0-alpha or 0.0.0-rc version

.github/workflows/cd.yml
henzhang-ms: Yes, please help to remove the limits. Because it is not hard-coded with rc/alpha anymore, it is 0.0.0-{github.event.inputs.preid}, so we'd better to consider the extension.

wenytang-ms: branch model only allows alpha, rc, stable release, others is forbidden

henzhang-ms (approved):
Approve for template release. Please reserve template-release.yml for other prerelease version except rc/alpha/stable. We may have a dev prerelease when bump minor version before daily release.

wenytang-ms:
> ove for template release. Please reserve template-release.yml for other prerel

restore template-release.yml for special release

.github/workflows/cd.yml
```diff
@@ -110,23 +81,29 @@ jobs:
...
       - name: version change
         id: version-change
         run: |
           echo "::set-output name=CHANGED::$(git tag --points-at HEAD | xargs)"
+          echo "::set-output name=TEMPLATE_VERSION::$(git tag --points-at HEAD | grep templates)"
+          VAR=$(git tag --points-at HEAD | grep templates | grep rc)
+          if [ ! -z "$VAR" ]
+          then
+              git push -d origin $VAR
+          fi
```
wenytang-ms, line 106: because templates RC version always be released with 0.0.0-rc, so the lerna version tag need to delete.

LongOddCode (approved):

adashen (approved):

PR #1614, opened 2021-07-14
refactor(aad): switch to graph api

KennethBWSong:
1. switch to graph api;
2. add error code classification, will add new error code later;

Telemetry:
image

Display:
image

Press "Get Help" will redirect to:
image

### DIFF
--- a/packages/fx-core/src/plugins/resource/aad/constants.ts
+++ b/packages/fx-core/src/plugins/resource/aad/constants.ts
@@ -49,6 +49,13 @@ export class Constants {
       },
     ],
   };
+
+  static errorInfo = {
+    type: "type",
+    user: "user",
+    system: "system",
+    helpLink: "helpLink",
+  };
 }
 
 export class Telemetry {
--- a/packages/fx-core/src/plugins/resource/aad/constants.ts
+++ b/packages/fx-core/src/plugins/resource/aad/constants.ts
@@ -49,13 +49,6 @@ export class Constants {
       },
     ],
   };
-
-  static errorInfo = {
-    type: "type",
-    user: "user",
-    system: "system",
-    helpLink: "helpLink",
-  };
 }
 
 export class Telemetry {
--- a/packages/fx-core/src/plugins/resource/aad/errors.ts
+++ b/packages/fx-core/src/plugins/resource/aad/errors.ts
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
-import { ConfigKeys, Plugins } from "./constants";
+import { ConfigKeys, Constants, Plugins } from "./constants";
 
 const referHelpLink = "Please refer to the help link for further steps.";
 const aadHelpLink = "https://aka.ms/teamsfx-aad-help";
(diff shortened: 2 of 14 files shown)


packages/fx-core/src/plugins/resource/aad/constants.ts
```diff
@@ -49,6 +49,13 @@ export class Constants {
       },
     ],
   };
+
+  static errorInfo = {
```
blackchoey, line 53: Can we define a class or interface for the error info?

KennethBWSong: This part is for saving constants used in error.ts and aadAppClient.ts.

packages/fx-core/src/plugins/resource/aad/index.ts
```diff
@@ -89,6 +89,8 @@ export class AadAppForTeamsPlugin implements Plugin {
         errorMessage += ` Detailed error: ${e.innerError.message}.`;
         if (e.innerError.response?.data?.errorMessage) {
           errorMessage += ` Reason: ${e.innerError.response?.data?.errorMessage}`;
+        } else if (e.innerError.response?.data?.error?.message) {
```
blackchoey, line 92: add some comments to tell where does the error come from? E.g. this line reads error returned by Graph API
This could help others understand the logic, since we get the error information from multiple sources

packages/fx-core/src/plugins/resource/aad/errors.ts
```diff
@@ -149,3 +149,28 @@ export class GraphClientErrorMessage {
   static readonly AppObjectIdIsNull = "Missing Object ID.";
   static readonly EmptyResponse = "Missing response.";
 }
+
+export const GraphErrorCodes = new Map([
+  [
```
blackchoey, line 154: Could we store the content into a resource file? This might be easier to maintain. You can decide which way to use based on the pros/cons comparison.

packages/fx-core/src/plugins/resource/aad/index.ts
KennethBWSong: Fixed. Thank you for pointing that out.

packages/fx-core/src/plugins/resource/aad/errors.ts
KennethBWSong: Fixed. Actually I thought it may be better to put errors in one file. For better maintenance, it should be better to put them in a file.

adashen (approved):

packages/fx-core/src/plugins/resource/aad/errors.ts
blackchoey: I means put graph error codes into a xxx.json file, and then load it into memory in our code. I raise this comment because we seems need to add a lot of error codes here.

blackchoey (approved):

packages/fx-core/src/plugins/resource/aad/errors.ts
KennethBWSong: Will update this when adding error code here.

PR #1618, opened 2021-07-15
chore(core): add encryption of secrets in userdata

tecton:
Use symmetric encryption to protect secret data. Since keytar and DPAPI both have some blocking issues, we choose a simplified way that selects project Id + fixed suffix as encryption key.


HuihuiWu-Microsoft (approved):

packages/fx-core/src/core/middleware/contextLoader.ts
```diff
@@ -1,107 +1,147 @@
...
+    const cryptoProvider = new LocalCrypto(projectSettings.projectId);
+    if (!projectIdMissing) {
+      for (const secretKey of Object.keys(dict)) {
+        if (!dataNeedEncryption(secretKey)) {
+          continue;
+        }
+        const secretValue = dict[secretKey];
+        const plaintext = cryptoProvider.decrypt(secretValue);
+        if (plaintext.isErr()) {
+          throw plaintext.error;
```
1yefuwang1, line 105: Maybe return err instead.

tecton: Thanks, updated.

1yefuwang1 (approved):

PR #1619, opened 2021-07-15
fix: add vscode selectSub

xiaolang124:
1. Add vscode select sub feature.


packages/vscode-extension/src/commonlib/azureLogin.ts
```diff
@@ -302,6 +309,18 @@ export class AzureAccountManager extends login implements AzureAccountProvider {
         if (item.subscription.subscriptionId == subscriptionId) {
           AzureAccountManager.tenantId = item.session.tenantId;
           AzureAccountManager.subscriptionId = subscriptionId;
+          TreeViewManagerInstance.getTreeView("teamsfx-accounts")!.refresh([
```
swatDong, line 312: Curious about the further plan of UI update - will the `statusMap` be retired after the componentize change?
If yes, ok to directly put UI operation in `azureLogin`.
Otherwise, keep `statusMap` and move UI operation out of `azureLogin`.

packages/vscode-extension/src/commonlib/azureLogin.ts
```diff
@@ -370,9 +389,16 @@ export class AzureAccountManager extends login implements AzureAccountProvider {
...
     const azureAccount = this.getAzureAccount();
+    if (triggerUI) {
+      if (azureAccount.status != loggedIn) {
+        await this.login(true);
+      }
+      if (azureAccount.status === loggedIn && !AzureAccountManager.subscriptionId) {
+        this.selectSubscription();
+      }
+    }
     if (azureAccount.status === loggedIn) {
```
swatDong, line 402: Is it possible that account is logged in, but no subs selected?

xiaolang124: Yes, login through treeView and have not selected subscription.

swatDong: If so, return undefined (not empty `SubscriptionInfo`) since no subs is selected.

packages/vscode-extension/src/commonlib/azureLogin.ts
xiaolang124: For now,  ```statusMap``` only notify login and logout message. TreeView subscription information can only be updated by the direct call command. I think later we can add more information to ```notify logic```.

packages/vscode-extension/src/commonlib/azureLogin.ts
xiaolang124: For this method, if a user logins in and cancels select subscription, we will throw an error.

xiaolang124: Changed to this:
```
if login and select sub:
    return SubscriptionInfo;
else if login and only have one sub:
    return SubscriptionInfo;
else 
    return undefined;
```

```
if trigger === true and (cancel login or select sub):
    throw error;
else
    return SubscriptionInfo;
```

packages/vscode-extension/src/commonlib/azureLogin.ts
```diff
@@ -370,10 +389,24 @@ export class AzureAccountManager extends login implements AzureAccountProvider {
     }
   }
 
-  // TODO add login and select subscription logic later
-  getSelectedSubscription(triggerUI=false): Promise<SubscriptionInfo | undefined> {
+  async getSelectedSubscription(triggerUI = false): Promise<SubscriptionInfo | undefined> {
     const azureAccount = this.getAzureAccount();
-    if (azureAccount.status === loggedIn) {
+    if (triggerUI) {
+      if (azureAccount.status != loggedIn) {
```
swatDong, line 395: !==

packages/vscode-extension/src/commonlib/azureLogin.ts
```diff
@@ -370,10 +389,24 @@ export class AzureAccountManager extends login implements AzureAccountProvider {
...
+        await this.login(true);
+      }
+      if (azureAccount.status === loggedIn && !AzureAccountManager.subscriptionId) {
+        await this.selectSubscription();
+      }
+    }
+    // no need to select sub when user only have one
+    if (azureAccount.status === loggedIn && !AzureAccountManager.subscriptionId && !triggerUI) {
```
swatDong, line 403: what's `!triggerUI` for? Could this single subs check always be there?
Or, move this part to `else` (if (triggerUI) ... **else ...**)

packages/vscode-extension/src/commonlib/azureLogin.ts
```diff
@@ -370,10 +389,24 @@ export class AzureAccountManager extends login implements AzureAccountProvider {
...
+      const subscriptionList = await this.listSubscriptions();
+      if (subscriptionList && subscriptionList.length == 1) {
+        await this.setSubscription(subscriptionList[0].subscriptionId);
+      }
+    }
+    if (azureAccount.status === loggedIn && AzureAccountManager.subscriptionId) {
```
swatDong, line 409: So the line 415 is redundant?

packages/vscode-extension/src/commonlib/azureLogin.ts
```diff
@@ -390,6 +423,40 @@ export class AzureAccountManager extends login implements AzureAccountProvider {
       return Promise.resolve(undefined);
```
swatDong, line 417: This async function already contains `await`. Remove all `Promise.resolve`/`Promise.reject` and just return result/throw error.

packages/vscode-extension/src/commonlib/azureLogin.ts
xiaolang124: Yes, already changed.

packages/vscode-extension/src/commonlib/azureLogin.ts
xiaolang124: Fixed.

packages/vscode-extension/src/commonlib/azureLogin.ts
xiaolang124: Fixed.

packages/vscode-extension/src/commonlib/azureLogin.ts
xiaolang124: Fixed.

swatDong (approved):
The main logic looks good to me.
One more question and ok to improve later - can we get subscription name from AzureAccountManager?

xiaolang124:
> The main logic looks good to me.
> One more question and ok to improve later - can we get subscription name from AzureAccountManager?

Ok, will be updated at later pr.

1openwindow (approved):

PR #1622, opened 2021-07-15
chore: fix pack bug

wenytang-ms:
before publish npm package, there will be a update AI key commit, so So it turns out that there will be errors by judging ^HEAD package.json version.
It's better to judge the version change by last commits with local tags, if changed, means lerna version tags on last commits, so there need to release visx to GitHub

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -126,7 +126,7 @@ jobs:
       - name: version change
         id: version-change
         run: |
-          echo "::set-output name=CHANGED::$(git tag --points-at HEAD)"
+          echo "::set-output name=CHANGED::$(git tag --points-at HEAD | xargs)"
 
       - name: update cli ai key
         if: ${{ github.event_name == 'workflow_dispatch' && github.ref == 'refs/heads/main' && (github.event.inputs.preid == 'stable'||github.event.inputs.preid == 'rc') }}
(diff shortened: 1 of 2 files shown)


LongOddCode (approved):

wenytang-ms:
before publish npm package, there will be a update ai key commit, so So it turns out that there will be errors by judging ^HEAD package.json version.
It's better to judge the version change by last commits with local tags, if changed, there need to release visx to github

adashen:
> before publish npm package, there will be a update AI key commit, so So it turns out that there will be errors by judging ^HEAD package.json version.
> It's better to judge the version change by last commits with local tags, if changed, means lerna version tags on last commits, so there need to release visx to GitHub

Does that mean we always update cli/vscode extension AI key no matter whether they need to be updated?

wenytang-ms:
> > before publish npm package, there will be a update AI key commit, so So it turns out that there will be errors by judging ^HEAD package.json version.
> > It's better to judge the version change by last commits with local tags, if changed, means lerna version tags on last commits, so there need to release visx to GitHub
> 
> Does that mean we always update cli/vscode extension AI key no matter whether they need to be updated?

Now we are doing this, This value is updated on the virtual machine locally, the update does not mean that it will be published, unless the version changes

If the judgment of the version change is added, the output can be used to determine whether to update AI key

adashen (approved):

PR #1623, opened 2021-07-15
fix: cli azure select sub

xiaolang124:
1. Add CLI azure select sub logic.
image


packages/cli/src/commonlib/azureLogin.ts
```diff
@@ -465,24 +459,57 @@ export class AzureAccountManager extends login implements AzureAccountProvider {
     }
   }
 
-  // TODO add login and select subscription logic later
-  getSelectedSubscription(triggerUI=false): Promise<SubscriptionInfo | undefined> {
-    if (AzureAccountManager.codeFlowInstance.account) {
+  async getSelectedSubscription(triggerUI = false): Promise<SubscriptionInfo | undefined> {
+    if (triggerUI) {
+      if (!AzureAccountManager.codeFlowInstance.account) {
+        await this.login(false);
```
swatDong, line 465: Why different from vscode, `true` or `false`?

packages/cli/src/commonlib/azureLogin.ts
```diff
@@ -465,24 +459,57 @@ export class AzureAccountManager extends login implements AzureAccountProvider {
...
+      }
+      if (AzureAccountManager.codeFlowInstance.account && !AzureAccountManager.subscriptionId) {
+        const subscriptionList = await this.listSubscriptions();
+        if (!subscriptionList || subscriptionList.length == 0) {
```
swatDong, line 469: `===`
Please fix all in this file...

packages/cli/src/commonlib/azureLogin.ts
xiaolang124: Now CLI will not ask user to confirm login or show learn more button.

packages/cli/src/commonlib/azureLogin.ts
xiaolang124: Fixed.

swatDong (approved):

PR #1624, opened 2021-07-15
feat(vscode): add CodeLens to support editing encrypted secrets

tecton:
In VS Code, CodeLens is used to provide a better experience.

VS Code Preview (gif):
vsc-crypto


packages/vscode-extension/src/telemetry/extTelemetryEvents.ts
```diff
@@ -79,6 +79,8 @@ export enum TelemetryEvent {
   DebugNpmInstall = "debug-npm-install",
 
   Survey = "survey",
+
+  EditSecret = "edit-secret",
```
HuihuiWu-Microsoft, line 83: Do you need another event to indicate whether edit secret is successful or not? You can see from the event file that the common pattern is event-name-start for start and event-name with property success(yes/no) to indicate the result

tecton: Thanks, updated.

packages/vscode-extension/src/handlers.ts
```diff
@@ -791,6 +791,44 @@ export async function cmpAccountsHandler() {
   quickPick.show();
 }
 
+export async function decryptSecret(cipher: string, selection: vscode.Range): Promise<void> {
+  ExtTelemetry.sendTelemetryEvent(TelemetryEvent.EditSecretStart, {
+    [TelemetryProperty.TriggerFrom]: TelemetryTiggerFrom.Other,
+  });
+  const editor = vscode.window.activeTextEditor;
+  if (!editor) {
+    ExtTelemetry.sendTelemetryEvent(TelemetryEvent.EditSecret, {
```
HuihuiWu-Microsoft, line 800: Use sendTelemetryErrorEvent() instead

tecton: Updated.

HuihuiWu-Microsoft (approved):

1openwindow (approved):

PR #1625, opened 2021-07-15
chore: add feature flag support

blackchoey:
Add feature flag support.
How to enable feature flag:
Set a given environment variable (decided by each feature owner) to `1` or `true` (accepts different upper/lower case combinations) to enable the feature.
Delete the environment variable or set to other values to disable the feature.

Usage:
```
if (isFeatureFlagEnabled("XXX_FEATURE_FLAG")) {
  // Feature flag enabled logic
} else {
  // Feature flag disabled logic
}
```


packages/fx-core/tests/core/other.test.ts
```diff
@@ -3,62 +3,82 @@
 
 import "mocha";
 import { FuncValidation, Inputs, Platform, Stage } from "@microsoft/teamsfx-api";
-import {QuestionAppName} from "../../src/core/question";
-import { assert } from "chai"; 
+import { QuestionAppName } from "../../src/core/question";
```
blackchoey, line 6: Some content is formatted automatically, you can only focus on the `tools: isFeatureFlagEnabled ...` test cases at the end of this file

packages/fx-core/tests/core/other.test.ts
```diff
@@ -93,4 +113,51 @@ describe("Other test case", () => {
...
+      [featureFlagName]: "true",
+    });
+    assert.isTrue(isFeatureFlagEnabled(featureFlagName));
+
+    restore = mockedEnv({
+      [featureFlagName]: "TruE", // should allow some characters be upper case
+    });
+    assert.isTrue(isFeatureFlagEnabled(featureFlagName));
+
+    restore();
```
wenytang-ms, line 134: seems called restore() will restore featureFlagName as "true", is this with expectations?
maybe restore need to called before mockedEnv?

xzf0587:
As other flags may be set in env.default.json file, seems the flag only be set in env. 
Should we consider collaboration scenario which the env setting will not work?

packages/fx-core/tests/core/other.test.ts
blackchoey: Thanks, seems I missed something when experimenting the behavior of mockedEnv. Will double confirm and update the test cases accordingly.

blackchoey:
> As other flags may be set in env.default.json file, seems the flag only be set in env.
> Should we consider collaboration scenario which the env setting will not work?

@xzf0587 the purpose of feature flag is to disable some features that we don't want to release right now, and also enable us to try the features internally. So we do not expect end users to set the environment variables.

packages/fx-core/tests/core/other.test.ts
blackchoey: resolved

packages/fx-core/package.json
```diff
@@ -65,6 +65,7 @@
     "klaw": "^3.0.0",
     "md5": "^2.3.0",
     "mime": "^2.5.2",
+    "mocked-env": "^1.3.5",
```
dilin-MS, line 68: This package should be in devDependencies instead of dependencies.

1yefuwang1 (approved):

packages/fx-core/src/common/tools.ts
```diff
@@ -466,3 +466,13 @@ export async function askSubscription(
   }
   return ok(resultSub);
 }
+
+// Determine whether feature flag is enabled based on environment variable setting
+export function isFeatureFlagEnabled(featureFlagName: string): boolean {
+  const flag = process.env[featureFlagName];
+  // can enable feature flag by set environment variable value to "1" or "true"
```
dilin-MS, line 473: Why do we need to set flag to "true" or "1"? Can we use one principle?

packages/fx-core/src/common/tools.ts
```diff
@@ -466,3 +466,13 @@ export async function askSubscription(
...
+  if (flag && (flag === "1" || flag.toLowerCase() === "true")) {
+    return true;
+  }
+  return false;
+}
```
dilin-MS, line 478: Seems can be simplified to 
```typescript
return (flag && (flag === "1" || flag.toLowerCase() === "true"));
```

packages/fx-core/src/common/tools.ts
blackchoey: Provide more ways to ensure our stakeholders can set the environment variable correctly if they didn't go through our instructions carefully.

packages/fx-core/src/common/tools.ts
blackchoey: This statement will also generate undefined result, so cannot return directly.

packages/fx-core/package.json
blackchoey: thanks, updated

packages/fx-core/tests/core/other.test.ts
```diff
@@ -3,62 +3,82 @@
...
+    );
+    assert.isTrue(validRes === undefined);
   });
 
   it("loader: DefaultSolutionLoader", async () => {
-    const inputs:Inputs = {platform: Platform.VSCode};
+    const inputs: Inputs = { platform: Platform.VSCode };
     const solution = await defaultSolutionLoader.loadSolution(inputs);
-    assert.isTrue(solution.name  === "fx-solution-azure");
+    assert.isTrue(solution.name === "fx-solution-azure");
```
wenytang-ms, line 71: maybe assert.strictEqual is better :-)

packages/fx-core/src/common/tools.ts
dilin-MS: Or
`return (flag && (flag === "1" || flag.toLowerCase() === "true"))? true : false;`

blackchoey: let me keep current implementation... it seems not a bad code style, and roughly equal to the one line code

packages/fx-core/tests/core/other.test.ts
blackchoey: leave to test case owner to make the decision... I don't know why this file is not formatted before.

adashen (approved):

PR #1630, opened 2021-07-16
feat(cli): update config command to support secret editing

tecton:
In CLI, we can use `teamsfx config get [option]` to view and `teamsfx config set <option> <value>` command to update.

CLI Preview (gif):
cli-crypto


packages/cli/src/cmds/config.ts
```diff
@@ -4,37 +4,118 @@
...
+            return projectResult;
+          }
+        } else {
+          CLILogProvider.necessaryLog(
+            LogLevel.Warning,
+            `You can change to teamsfx project folder or use --folder to specify.`
+          );
+          const error = NonTeamsFxProjectFolder();
+          CliTelemetry.sendTelemetryErrorEvent(TelemetryEvent.ConfigGet, error);
+          return ok(null);
```
Alive-Fish, line 100: this may be return error

packages/cli/src/cmds/config.ts
```diff
@@ -4,37 +4,118 @@
...
+    const rootFolder = path.resolve((args.folder as string) || "./");
+    const inProject = (await readConfigs(rootFolder)).isOk();
+
+    if (args.option === undefined) {
+      // print all
+      const globalResult = await this.printGlobalConfig();
+      if (globalResult.isErr()) {
+        return globalResult;
+      }
+      if (!args.global && inProject) {
```
Alive-Fish, line 67: global may be always `undefined` because it is not defined in the `builder`.

packages/cli/src/cmds/config.ts
```diff
@@ -4,37 +4,118 @@
...
+import {
+  TelemetryEvent,
+  TelemetryProperty,
+  TelemetrySuccess,
+} from "../telemetry/cliTelemetryEvents";
+import activate from "../activate";
+import { NonTeamsFxProjectFolder, ConfigNameNotFound } from "../error";
+import * as constants from "../constants";
+
+const CryptoDataMatchers = new Set([
```
Alive-Fish, line 29: Are these also used in the vscode? If yes, I think you can maintain these in the high level.

packages/cli/src/cmds/config.ts
tecton: Thanks, fixed.

packages/cli/src/cmds/config.ts
tecton: This is a good point, I updated codes to reuse the exported function in fx-core. VS Code doesn't share the same way because it uses regex to check now.

Alive-Fish (approved):
look good to me

PR #1632, opened 2021-07-16
chore(bot): update bot builder version for templates

YitongFeng-git:
related bug https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10334443

### DIFF
--- a/templates/bot/js/default/package.json
+++ b/templates/bot/js/default/package.json
@@ -15,8 +15,8 @@
   "dependencies": {
     "@microsoft/teamsfx": "^0.1.0",
     "@microsoft/microsoft-graph-client": "^2.2.1",
-    "botbuilder": "~4.11.0",
-    "botbuilder-dialogs": "~4.11.0",
+    "botbuilder": "~4.14.0",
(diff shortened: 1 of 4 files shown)


templates/bot/js/default/package.json
```diff
@@ -15,8 +15,8 @@
   "dependencies": {
     "@microsoft/teamsfx": "^0.1.0",
     "@microsoft/microsoft-graph-client": "^2.2.1",
-    "botbuilder": "~4.11.0",
-    "botbuilder-dialogs": "~4.11.0",
+    "botbuilder": "~4.14.0",
```
eriolchan, line 18: I am a little curious on these templates using different versions, and good to see they are aligned again. Since we upgrade the version for different templates, we need to test all of them to check functionality and SSO work correctly.

eriolchan (approved):

templates/bot/js/default/package.json
YitongFeng-git: I of course have tested them all.

PR #1634, opened 2021-07-16
fix(bot sdk): fix teams auto sign in wrong account issue

dilin-MS:
add login hint in bot auth start page to fix teams auto sign in wrong account issue

[Issue: "show profile" of Bot stuck when click continue using web client, and could not login using teams client using dev365 account](https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10230876)


packages/sdk/src/bot/teamsBotSsoPrompt.ts
```diff
@@ -284,7 +286,13 @@ export class TeamsBotSsoPrompt extends Dialog {
   private async sendOAuthCardAsync(context: TurnContext): Promise<void> {
     internalLogger.verbose("Send OAuth card to get SSO token");
 
-    const signInResource = this.getSignInResource();
+    const account: TeamsChannelAccount = await TeamsInfo.getMember(
+      context,
+      context.activity.from.id
+    );
+    internalLogger.verbose("Get Teams member account email address: " + account.email);
```
blackchoey, line 293: Some tenant may not use email as login account, can you see whether there's other property like user principle name?
https://docs.microsoft.com/en-us/azure/active-directory/hybrid/plan-connect-userprincipalname

templates/bot/js/default/public/auth-start.html
```diff
@@ -18,6 +18,7 @@
         var currentURL = new URL(window.location);
         var clientId = currentURL.searchParams.get("clientId");
         var tenantId = currentURL.searchParams.get("tenantId");
+        var email = currentURL.searchParams.get("email");
```
blackchoey, line 21: maybe use `loginHint` as parameter?

packages/sdk/src/bot/teamsBotSsoPrompt.ts
dilin-MS: Changed to userPrincipalName.

templates/bot/js/default/public/auth-start.html
dilin-MS: Fixed.

packages/sdk/src/bot/teamsBotSsoPrompt.ts
```diff
@@ -348,7 +350,7 @@ export class TeamsBotSsoPrompt extends Dialog {
       this.settings.scopes.join(" ")
     )}&clientId=${config.authentication!.clientId}&tenantId=${
       config.authentication!.tenantId
-    }&email=${email}`;
+    }&loginHint=${loginHint}`;
```
blackchoey, line 353: what will happen if some implementation does not pass loginHint? By the way do we need to make loginHint optional? Since it's a private function.

dilin-MS: If the loginHint is empty, the sign in page will let user choose account to sign in. So makes loginHint optional in case btobuilder TeamsInfo SDK fails to get member's userPrincipalName.

blackchoey (approved):

SLdragon (approved):

templates/bot/js/default/public/auth-start.html
```diff
@@ -18,6 +18,7 @@
         var currentURL = new URL(window.location);
         var clientId = currentURL.searchParams.get("clientId");
         var tenantId = currentURL.searchParams.get("tenantId");
+        var loginHint = currentURL.searchParams.get("loginHint");
```
IvanJobs, line 21: FYI: There're two `templates` folder:
1. TeamsFx/templates
2. TeamsFx/packages/fx-core/templates (fallback templates)

and under each `templates` folder, there're multiple public/auth-start.hml to be modified.

IvanJobs (approved):
lgtm

PR #1635, opened 2021-07-16
refactor: remove manifest input in validate and build

yiqing-zhao:


jayzhang (approved):

packages/fx-core/src/plugins/solution/fx-solution/solution.ts
```diff
@@ -1785,94 +1785,15 @@ export class TeamsAppSolution implements Solution {
...
-                  SolutionError.GetRemoteConfigError
-                )
-              );
-            } else {
-              return err(manifest.error);
-            }
-          }
-        }
-        return await appStudioPlugin.validateManifest(pluginCtx, manifestString);
+        const maybeSelectedPlugins = this.getSelectedPlugins(ctx);
```
nliu-ms, line 1788: this line can be removed?

packages/fx-core/src/plugins/solution/fx-solution/solution.ts
```diff
@@ -1785,94 +1785,15 @@ export class TeamsAppSolution implements Solution {
...
-                  "Solution",
-                  SolutionError.GetRemoteConfigError
-                )
-              );
-            } else {
-              return err(manifest.error);
-            }
-          }
-        }
+        const maybeSelectedPlugins = this.getSelectedPlugins(ctx);
```
nliu-ms, line 1793: as above

packages/fx-core/src/plugins/solution/fx-solution/solution.ts
yiqing-zhao: Thanks! The line is removed now!

packages/fx-core/src/plugins/solution/fx-solution/solution.ts
yiqing-zhao: Thanks! I have removed the line.

PR #1636, opened 2021-07-16
fix: add vscode azure save sub

xiaolang124:
1. Add save, read subscriptionInfo feature.
2. Add support for old projects.

### DIFF
--- a/packages/vscode-extension/src/commonlib/azureLogin.ts
+++ b/packages/vscode-extension/src/commonlib/azureLogin.ts
@@ -28,11 +28,7 @@ import {
   signedIn,
   signedOut,
   signingIn,
-  solution,
-  subscriptionIdString,
   subscriptionInfoFile,
-  subscriptionNameString,
-  tenantIdString,
 } from "./common/constant";
 import { login, LoginStatus } from "./common/login";
 import * as StringResources from "../resources/Strings.json";
@@ -275,6 +271,7 @@ export class AzureAccountManager extends login implements AzureAccountProvider {
       await vscode.commands.executeCommand("azure-account.logout");
       AzureAccountManager.tenantId = undefined;
       AzureAccountManager.subscriptionId = undefined;
+      this.clearSubscription();
       ExtTelemetry.sendTelemetryEvent(TelemetryEvent.SignOut, {
         [TelemetryProperty.AccountType]: AccountType.Azure,
         [TelemetryProperty.Success]: TelemetrySuccess.Yes,
@@ -506,6 +503,15 @@ export class AzureAccountManager extends login implements AzureAccountProvider {
     }
   }
 
+  async clearSubscription(): Promise<void> {
+    const subscriptionFilePath = await this.getSubscriptionInfoPath();
+    if (!subscriptionFilePath) {
+      return;
+    } else {
+      await fs.writeFile(subscriptionFilePath, JSON.stringify("", null, 4));
+    }
+  }
+
   async readSubscription(): Promise<SubscriptionInfo | undefined> {
     const subscriptionFilePath = await this.getSubscriptionInfoPath();
     if (!subscriptionFilePath || !fs.existsSync(subscriptionFilePath)) {
@@ -519,9 +525,9 @@ export class AzureAccountManager extends login implements AzureAccountProvider {
       const content = (await fs.readFile(subscriptionFilePath)).toString();
       const subcriptionJson = JSON.parse(content);
       return {
-        subscriptionId: subcriptionJson[subscriptionIdString],
-        tenantId: subcriptionJson[tenantIdString],
-        subscriptionName: subcriptionJson[subscriptionNameString],
+        subscriptionId: subcriptionJson.subscriptionId,
+        tenantId: subcriptionJson.tenantId,
+        subscriptionName: subcriptionJson.subscriptionName,
       };
     }
   }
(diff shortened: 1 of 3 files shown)


packages/vscode-extension/src/commonlib/azureLogin.ts
```diff
@@ -360,9 +379,21 @@ export class AzureAccountManager extends login implements AzureAccountProvider {
     const azureAccount: AzureAccount =
       vscode.extensions.getExtension<AzureAccount>("ms-vscode.azure-account")!.exports;
     AzureAccountManager.currentStatus = azureAccount.status;
+    if (AzureAccountManager.currentStatus === "LoggedIn") {
+      const subscriptioninfo = await this.readSubscription();
+      if (subscriptioninfo) {
+        this.setSubscription(subscriptioninfo.subscriptionId);
```
swatDong, line 385: What if the logged in account has no access to the selected subs?
Better to keep all the behaviors of current toolkit.
- When no user sign in but subs was selected
- When user sign in but no subs was selected
- When user sign in but selected subs mismatched

packages/vscode-extension/src/commonlib/azureLogin.ts
```diff
@@ -410,7 +441,7 @@ export class AzureAccountManager extends login implements AzureAccountProvider {
       const selectedSub: SubscriptionInfo = {
         subscriptionId: AzureAccountManager.subscriptionId,
         tenantId: AzureAccountManager.tenantId!,
-        subscriptionName: "",
+        subscriptionName: AzureAccountManager.subscriptionName ?? "",
```
swatDong, line 444: In which case the subsId exists but name not?

packages/vscode-extension/src/commonlib/azureLogin.ts
```diff
@@ -447,10 +483,97 @@ export class AzureAccountManager extends login implements AzureAccountProvider {
...
+    } else {
+      return undefined;
+    }
+  }
+
+  async getSubscriptionInfoFromEnv(): Promise<SubscriptionInfo | undefined> {
+    if (vscode.workspace.workspaceFolders) {
+      const workspaceFolder: vscode.WorkspaceFolder = vscode.workspace.workspaceFolders[0];
+      const workspacePath: string = workspaceFolder.uri.fsPath;
+      if (!(await commonUtils.isFxProject(workspacePath))) {
```
swatDong, line 551: Why `commonUtils` is from *debug/* folder... is there a more common utility?

packages/vscode-extension/src/commonlib/common/constant.ts
```diff
@@ -9,3 +9,10 @@ export const signingIn = "SigningIn";
 export const loggedOut = "LoggedOut";
 export const loggedIn = "LoggedIn";
 export const loggingIn = "LoggingIn";
+
+export const subscriptionInfoFile = "subscriptionInfo.json";
+export const envDefaultJsonFile = "env.default.json";
+export const solution = "solution";
+export const subscriptionIdString = "subscriptionId";
```
swatDong, line 16: No need to define constants for object property name. Instead of:

``` typescript
const xxxString = "xxx";
const x = object[xxxString];
```

Just use:
``` typescript
const x = object.xxx;
```

packages/vscode-extension/src/commonlib/azureLogin.ts
xiaolang124: When no user sign in but subs was selected
    No login means we can show nothing and do nothing.
When user sign in but no subs was selected
    We only show login information, if the user has only one sub we will select that and update subscription information file. If the user has more than 1 subs, we will show no sub selected.
When user sign in but selected subs mismatched
    For this scenario, solution will update env file when sign in or sign out. So i add signout clear file logic.

packages/vscode-extension/src/commonlib/azureLogin.ts
xiaolang124: Old project solution only save tenantid and subscriptionId.

packages/vscode-extension/src/commonlib/azureLogin.ts
xiaolang124: I find only env checker and debug have this feature.

packages/vscode-extension/src/commonlib/common/constant.ts
xiaolang124: Fixed.

packages/vscode-extension/src/commonlib/azureLogin.ts
swatDong: `solution will update env file when sign in or sign out` interesting ... this means even the project is deployed, subs info could still be changed. That could be something to improve later.

packages/vscode-extension/src/extension.ts
```diff
@@ -31,10 +32,9 @@ export async function activate(context: vscode.ExtensionContext) {
   await exp.initialize(context);
 
   // 1.1 Register the creating command.
-  const createCmd = vscode.commands.registerCommand(
-    "fx-extension.create",
-    handlers.createNewProjectHandler
-  );
+  const createCmd = vscode.commands.registerCommand("fx-extension.create", async () => {
+    console.log(await AzureAccountManager.readSubscription());
```
swatDong, line 36: revert this local test change

xiaolang124: My bad, fixed.

HuihuiWu-Microsoft:
Will you add unit test?

swatDong (approved):

xiaolang124:
> Will you add unit test?

Yes, will add later.

HuihuiWu-Microsoft (approved):
LGTM for extension part

PR #1637, opened 2021-07-16
fix(env checker): add dotnet-install-script in cli

meifans:


Alive-Fish (approved):
approve for cli webpack

qinezh (approved):

kuojianlu (changes requested):
Update .gitignore to skip ignoring these two scripts.

kuojianlu (approved):

PR #1642, opened 2021-07-19
chore: add interfaces for ARM support

blackchoey:
WIP: this PR adds required interfaces for ARM support implementation. It's a temporary solution to unblock ARM feature development, so I didn't evolve the folder structure for fx-core to provide a better place for them. The interfaces will be moved to the V2 teamsfx-api in the future.

### DIFF
--- a/packages/fx-core/src/plugins/solution/fx-solution/arm.ts
+++ b/packages/fx-core/src/plugins/solution/fx-solution/arm.ts
@@ -1,0 +1,37 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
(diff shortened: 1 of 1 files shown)


wenytang-ms:
will there be any API update in docs files?
if so, please add them in this pr, thanks, otherwise, CD will run fail

1yefuwang1 (approved):

blackchoey:
> will there be any API update in docs files?
> if so, please add them in this pr, thanks, otherwise, CD will run fail

fx-core does not have API doc, so no need to update doc files

wenytang-ms (approved):

PR #1645, opened 2021-07-19
fix: cli azure save sub

xiaolang124:
1. Add Azure write, read subscription information feature.
2. Only update subscription information when use ```--set``` or really ```provision, deploy```.
3. Try to load subscription from env when opening old project.


packages/cli/src/commonlib/azureLogin.ts
```diff
@@ -505,13 +519,95 @@ export class AzureAccountManager extends login implements AzureAccountProvider {
...
+  async getSubscriptionInfoPath(): Promise<string | undefined> {
+    if (AzureAccountManager.rootPath) {
+      if (isWorkspaceSupported(AzureAccountManager.rootPath)) {
+        const subscriptionFile = path.join(
+          AzureAccountManager.rootPath,
+          `.${ConfigFolderName}`,
+          subscriptionInfoFile
+        );
+        return subscriptionFile;
+      } else {
```
chagong, line 575: all the 'return undefined' can be simplified?

chagong (approved):

packages/cli/src/activate.ts
```diff
@@ -18,6 +18,11 @@ import CLIUIInstance from "./userInteraction";
 
 export default async function activate(rootPath?: string): Promise<Result<FxCore, FxError>> {
   if (rootPath) {
+    AzureAccountManager.setRootPath(rootPath);
+    const subscriptionInfo = await AzureAccountManager.readSubscription();
+    if (subscriptionInfo) {
+      await AzureAccountManager.setSubscription(subscriptionInfo.subscriptionId);
+    }
     const subscription = await getSubscriptionIdFromEnvFile(rootPath);
```
swatDong, line 26: Is this line still needed?

xiaolang124: Have removed old read sub logic, ```getSubscriptionIdFromEnvFile``` only read subscriptionId, ```getSubscriptionInfoFromEnv``` read subscription and tenant information.

swatDong (approved):

PR #1647, opened 2021-07-20
fix(env checker): add sentinel for func-core-tools

meifans:
* add sentinel for func-core-tools
* fix helplink for linux user

### DIFF
--- a/packages/cli/src/cmds/preview/depsChecker/funcToolChecker.ts
+++ b/packages/cli/src/cmds/preview/depsChecker/funcToolChecker.ts
@@ -167,7 +167,7 @@ export class FuncToolChecker implements IDepsChecker {
   }
 
   private static getSentinelPath(): string {
-    return path.join(os.homedir(), `.${ConfigFolderName}`, "sentinel");
+    return path.join(os.homedir(), `.${ConfigFolderName}`, "func-sentinel");
   }
 
   private static getPortableFuncExecPath(): string {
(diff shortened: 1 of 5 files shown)


kuojianlu (approved):

packages/vscode-extension/src/debug/depsChecker/funcToolChecker.ts
```diff
@@ -131,10 +145,13 @@ export class FuncToolChecker implements IDepsChecker {
       );
     }
 
-    if (!isInstallationValid) {
-      this._telemetry.sendEvent(DepsCheckerEvent.funcValidationError);
+    if (!isVersionSupported || !hasSentinel) {
+      this._telemetry.sendEvent(DepsCheckerEvent.funcValidationError, {
+        "func-v": String(isVersionSupported),
```
a1exwang, line 150: Is it better to use `"func-v": String(portableFunc)` since it suits better to the name?

packages/cli/src/cmds/preview/depsChecker/funcToolChecker.ts
```diff
@@ -149,6 +166,10 @@ export class FuncToolChecker implements IDepsChecker {
     return path.join(os.homedir(), `.${ConfigFolderName}`, "bin", "func");
   }
 
+  private static getSentinelPath(): string {
+    return path.join(os.homedir(), `.${ConfigFolderName}`, "sentinel");
```
a1exwang, line 170: The name `sentinel` is too general and it is put directly in the `.fx` folder. Better change to `func-sentinel` or `env-checker-func-sentinel` to avoid confusion or conflicting with future files.

a1exwang (approved):

packages/cli/src/cmds/preview/depsChecker/funcToolChecker.ts
meifans: Good idea, added

PR #1650, opened 2021-07-20
fix(env checker): filter out global sdk for event to dotnet installation

meifans:


packages/cli/src/cmds/preview/depsChecker/dotnetChecker.ts
```diff
@@ -90,7 +90,10 @@ export class DotnetChecker implements IDepsChecker {
 
     await this._logger.debug(`[start] check dotnet version`);
     if (dotnetPath !== null && (await this.isDotnetInstalledCorrectly())) {
-      this._telemetry.sendEvent(DepsCheckerEvent.dotnetInstallCompleted);
+      // filter out global sdk
+      if (dotnetPath.includes(".fx")) {
```
a1exwang, line 94: Better to check whether `dotnetPath` begins with `DotnetChecker.getDefaultInstallPath()`?

a1exwang (approved):

PR #1651, opened 2021-07-20
test: enable vsc extension ut on ci

HuihuiWu-Microsoft:


packages/vscode-extension/coverconfig.json
```diff
@@ -2,7 +2,7 @@
   "enabled": true,
   "relativeSourcePath": "../../src",
   "relativeCoverageDir": "../../../coverage",
-  "ignorePatterns": ["**/node_modules/**"],
+  "ignorePatterns": ["**/node_modules/**", "debug/depsChecker/**.*"],
```
LongOddCode, line 5: should this be "**/*"?

.github/workflows/unit-test.yml
```diff
@@ -51,7 +51,7 @@ jobs:
         id: unit-test
         shell: bash
         run: |
-          npx lerna run test:unit
+          xvfb-run -a npx lerna run test:unit
```
LongOddCode, line 54: should we install xvfb-run at first?

LongOddCode (approved):

PR #1652, opened 2021-07-20
fix: provision teams app again if it has been removed by the user

nliu-ms:
1) Add provision and postProvision stage for app studio plugin
2) Get provision results and pass it to onLifeCycleFinished, set it to solution config
3) AppStudio.provision will be triggered each time user click `provision`
4) Update the warning
image


packages/fx-core/src/plugins/solution/fx-solution/executor.ts
```diff
@@ -118,7 +120,7 @@ export async function executeLifecycles(
   lifecycles: LifecyclesWithContext[],
   postLifecycles: LifecyclesWithContext[],
   onPreLifecycleFinished?: () => Promise<Result<any, FxError>>,
-  onLifecycleFinished?: () => Promise<Result<any, FxError>>,
+  onLifecycleFinished?: (result?: any[]) => Promise<Result<any, FxError>>,
```
1yefuwang1, line 123: How about adding result to each `onXXXFinshed()`'s parameter for consistency. And reorder parameters to preLifecycles, onPreLifecycleFinished, lifecycles, onLifecycleFinished, postLifecycles, onPostLifecycleFinished

packages/fx-core/src/plugins/solution/fx-solution/executor.ts
```diff
@@ -81,26 +81,28 @@ export async function executeConcurrently(
 
   const results = await Promise.all(promises);
   if (logger) logger?.info(`${("[Solution] Execute " + step + "Task summary").padEnd(64, "-")}`);
-  let res: Result<any, FxError> = ok(undefined);
+  const res: Result<any, FxError>[] = [];
```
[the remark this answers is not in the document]
nliu-ms: ok, thanks!

packages/fx-core/src/plugins/solution/fx-solution/executor.ts
nliu-ms: As onXXXFinished are optional, I will not change the order. But I can add parameters for them.

1yefuwang1 (approved):

jayzhang (approved):

PR #1654, opened 2021-07-20
feat: prepare predefined ymls/scripts

IvanJobs:
1. prepare predefined ymls for github action, example scripts for other platforms.
2. add readme.md placehoder for PM.

### DIFF
--- a/docs/cicd/README.md
+++ b/docs/cicd/README.md
@@ -1,33 +1,30 @@
 # CI/CD Support for Teams Application Developers
 
-If you're developing teams applications inside a company and considering to build your workflows/pipelines, you've come to the right place. This document is to describe what are provided to help you to build workflows/pipelines, and also guide you to get started quickly.
+TeamsFx tooling provides support for you to automate, customize and execute your sfotware development workflows right in your repository when buidling a Teams applictaion. These documents describes the essential tools and predefined templates for you to quickly get started with CI/CD.
 
+|Tools and Templates|Description|
+|---|---|
+|[teamsfx-cli-action](https://github.com/OfficeDev/teamsfx-cli-action)|A ready-to-use GitHub Action and you can combineit in a completely customized workflow.|
+|[github-ci-template.yml](https://github.com/OfficeDev/TeamsFx/blob/ruhe/cicd_ymls/docs/cicd/github-ci-template.yml) and [github-cd-template.yml](https://github.com/OfficeDev/TeamsFx/blob/ruhe/cicd_ymls/docs/cicd/github-cd-template.yml)|GitHub workflow templates. It shows what a typical CI/CD workflow looks like for a Teams application.|
+|[script-ci-template.sh](https://github.com/OfficeDev/TeamsFx/blob/ruhe/cicd_ymls/docs/cicd/others-script-ci-template.sh) and [script-cd-template.sh](https://github.com/OfficeDev/TeamsFx/blob/ruhe/cicd_ymls/docs/cicd/others-script-cd-template.sh)|Script templates you can follow to build workflows outside GitHub. It shows what a typical CI/CD workflow looks like for a Teams application.|
 
-## What are provided?
-1. First-party published GitHub Action: [teamsfx-cli-action](https://github.com/OfficeDev/teamsfx-cli-action).
+## CI/CD Workflow Teamplates in GitHub
 
-1. Predefined example workflows: [github-ci-template.yml](https://github.com/OfficeDev/TeamsFx/blob/ruhe/cicd_ymls/docs/cicd/github-ci-template.yml), [github-cd-template.yml](https://github.com/OfficeDev/TeamsFx/blob/ruhe/cicd_ymls/docs/cicd/github-cd-template.yml).
+Follow these example [github-ci-template.yml](https://github.com/OfficeDev/TeamsFx/blob/ruhe/cicd_ymls/docs/cicd/github-ci-template.yml) and  [github-cd-template.yml](https://github.com/OfficeDev/TeamsFx/blob/ruhe/cicd_ymls/docs/cicd/github-cd-template.yml) if you want to create your own by copying-n-pasting the file into your repository under the folder `.github/workflows`. 
 
-1. Predefined example scripts: [others-script-ci-template.sh](https://github.com/OfficeDev/TeamsFx/blob/ruhe/cicd_ymls/docs/cicd/others-script-ci-template.sh), [others-script-cd-template.sh](https://github.com/OfficeDev/TeamsFx/blob/ruhe/cicd_ymls/docs/cicd/others-script-cd-template.sh).
+### Customize CI Workflow
+There are some potential changes you can make to adapt your project:
+1. Change the build scripts.
+1. Remove the test scripts if you don't want to set up the unit test framework for your project.
 
-## Getting started guide for GitHub
-
-### Continuous Integration
-If you want to create a GitHub workflow for continuous integration, this is simple. Just copy the predefined example workflow [github-ci-template.yml](https://github.com/OfficeDev/TeamsFx/blob/ruhe/cicd_ymls/docs/cicd/github-ci-template.yml) into your repository under the folder `.github/workflows` which follows GitHub's practise, and do modifications to meet your own needs.
-
-The expected modifications include but not limited to:
-1. Change the build scripts to suit your project.
-1. Set up the unit test framework for your project and run test by `npm run test`.
-
-### Continuous Deployment
-If you want to create a GitHub workflow for continuous deployment, you need to copy the predefined example workflow [github-cd-template.yml](https://github.com/OfficeDev/TeamsFx/blob/ruhe/cicd_ymls/docs/cicd/github-cd-template.yml) into your repository under the folder `.github/workflows` which follows GitHub's practise, and do modifications to meet your own needs.
-
-The expected modifications include but not limited to:
-1. Modify the trigger pattern to suit your own requirements.
-In the example workflow's definition, the workflow will be triggered when there's new commits pushed into the main branch. We don't force you to use this pattern, and you can change it by your case.
-
+### Customize CD Workflow
+1. Change how the CD flow is triggered. We default it to when new commits made to main branch.
 1. Create GitHub repository secrets by environment to hold Azure/M365 login credentials.
-The table below lists all the secrets you needs to create on GitHub, and for the detailed usage, please refer to the GitHub Action's [README.md](https://github.com/OfficeDev/teamsfx-cli-action/blob/main/README.md).
+The table below lists all the secrets you need to create on GitHub, and for the detailed usage, please refer to the GitHub Action's [README.md](https://github.com/OfficeDev/teamsfx-cli-action/blob/main/README.md).
+1. Update the environment variable `RUN_PROVISION` inside the workflow definition accordingly. Usually, the provision step will be performed for only once and subsequent runs of the workflow dosn't need to run provision again. So, there's environment variable `RUN_PROVISION` to control this. One thing needs to mentioned is, after provision, the configs will be commited into the repository, and the following deploy, publish operations will target for this committed configs.
+To be more specific, to run provision, set `RUN_PROVISION` to `true`, else set it to `false`.
+
+### Environment Variables
 
 |Name|Description|
 |---|---|
@@ -39,11 +36,15 @@ The table below lists all the secrets you needs to create on GitHub, and for the
 |M365_ACCOUNT_PASSWORD|The password of M365 account.|
 |M365_TENANT_ID|To identify the tenant in which the Teams App will be created/published. PS: if not provided, the default tenant id will be used.|
 
-1. Update the environment variable `RUN_PROVISION` inside the workflow definition accordingly.
-Usually, the provision step will be performed for only once and subsequent runs of the workflow dosn't need to run provision again. So, there's environment variable `RUN_PROVISION` to control this. One thing needs to mentioned is, after provision, the configs will be commited into the repository, and the following deploy, publish operations will target for this committed configs.
-To be more specific, to run provision, set `RUN_PROVISION` to `true`, else set it to `false`.
 
 ## Getting started guide for other platforms
-Predefined example scripts are provided to help you to build workflows/pipelines on other platforms, for example Jenkins, and the scripts are pretty straight forward. Although they are written in bash style, but you can easily convert them to any other platforms since [teamsfx-cli](https://www.npmjs.com/package/@microsoft/teamsfx-cli) itself is cross-platform and the scripts largely leverages it.
+You can follow the pre-defined example scripts to build and customize CI/CD pipelines on other platforms like Jenkins:
+* [CI Scripts](https://github.com/OfficeDev/TeamsFx/blob/ruhe/cicd_ymls/docs/cicd/others-script-ci-template.sh)
+* [CD Scripts](https://github.com/OfficeDev/TeamsFx/blob/ruhe/cicd_ymls/docs/cicd/others-script-cd-template.sh)
 
-One thing needs to be mentioned is, you needs to set Azure/M365 credentials in your environment variables safely.
+The scripts are based on a cross-platform TeamsFx command line tool [TeamsFx-CLI](https://www.npmjs.com/package/@microsoft/teamsfx-cli). You can install it with `npm install -g @microsoft/teamsfx-cli` and follow the [documentation](https://github.com/OfficeDev/TeamsFx/blob/dev/docs/cli/user-manual.md) to customize the scripts.
+
+Please keep in mind that you need to set Azure and M365 credentials in your environment variables safely.
+
+# Additional Notes
+* [Quick Start for GitHub Actions](https://docs.github.com/en/actions/quickstart#creating-your-first-workflow)
--- a/docs/cicd/README.md
+++ b/docs/cicd/README.md
@@ -28,6 +28,7 @@ In the example workflow's definition, the workflow will be triggered when there'
 
 1. Create GitHub repository secrets by environment to hold Azure/M365 login credentials.
 The table below lists all the secrets you needs to create on GitHub, and for the detailed usage, please refer to the GitHub Action's [README.md](https://github.com/OfficeDev/teamsfx-cli-action/blob/main/README.md).
+
 |Name|Description|
 |---|---|
 |AZURE_ACCOUNT_NAME|The account name of Azure which is used to provision resources.|
--- a/docs/cicd/README.md
+++ b/docs/cicd/README.md
@@ -1,0 +1,48 @@
+# CI/CD Support for Teams Application Developers
+
+If you're developing teams applications inside a company and considering to build your workflows/pipelines, you've come to the right place. This document is to describe what are provided to help you to build workflows/pipelines, and also guide you to get started quickly.
+
+
+## What are provided?
+1. First-party published GitHub Action: [teamsfx-cli-action](https://github.com/OfficeDev/teamsfx-cli-action).
+
+1. Predefined example workflows: [github-ci-template.yml](https://github.com/OfficeDev/TeamsFx/blob/ruhe/cicd_ymls/docs/cicd/github-ci-template.yml), [github-cd-template.yml](https://github.com/OfficeDev/TeamsFx/blob/ruhe/cicd_ymls/docs/cicd/github-cd-template.yml).
+
+1. Predefined example scripts: [others-script-ci-template.sh](https://github.com/OfficeDev/TeamsFx/blob/ruhe/cicd_ymls/docs/cicd/others-script-ci-template.sh), [others-script-cd-template.sh](https://github.com/OfficeDev/TeamsFx/blob/ruhe/cicd_ymls/docs/cicd/others-script-cd-template.sh).
+
+## Getting started guide for GitHub
+
+### Continuous Integration
+If you want to create a GitHub workflow for continuous integration, this is simple. Just copy the predefined example workflow [github-ci-template.yml](https://github.com/OfficeDev/TeamsFx/blob/ruhe/cicd_ymls/docs/cicd/github-ci-template.yml) into your repository under the folder `.github/workflows` which follows GitHub's practise, and do modifications to meet your own needs.
+
+The expected modifications include but not limited to:
+1. Change the build scripts to suit your project.
+1. Set up the unit test framework for your project and run test by `npm run test`.
+
+### Continuous Deployment
+If you want to create a GitHub workflow for continuous deployment, you need to copy the predefined example workflow [github-cd-template.yml](https://github.com/OfficeDev/TeamsFx/blob/ruhe/cicd_ymls/docs/cicd/github-cd-template.yml) into your repository under the folder `.github/workflows` which follows GitHub's practise, and do modifications to meet your own needs.
+
+The expected modifications include but not limited to:
+1. Modify the trigger pattern to suit your own requirements.
+In the example workflow's definition, the workflow will be triggered when there's new commits pushed into the main branch. We don't force you to use this pattern, and you can change it by your case.
+
+1. Create GitHub repository secrets by environment to hold Azure/M365 login credentials.
+The table below lists all the secrets you needs to create on GitHub, and for the detailed usage, please refer to the GitHub Action's [README.md](https://github.com/OfficeDev/teamsfx-cli-action/blob/main/README.md).
+|Name|Description|
+|---|---|
+|AZURE_ACCOUNT_NAME|The account name of Azure which is used to provision resources.|
+|AZURE_ACCOUNT_PASSWORD|The password of Azure account.|
+|AZURE_SUBSCRIPTION_ID|To identify the subscription in which the resources will be provisined.|
+|AZURE_TENANT_ID|To identify the tenant in which the subscription resides.|
+|M365_ACCOUNT_NAME|The M365 account for creating and publishing Teams App.|
+|M365_ACCOUNT_PASSWORD|The password of M365 account.|
+|M365_TENANT_ID|To identify the tenant in which the Teams App will be created/published. PS: if not provided, the default tenant id will be used.|
+
+1. Update the environment variable `RUN_PROVISION` inside the workflow definition accordingly.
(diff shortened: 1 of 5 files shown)


docs/cicd/github-cd-template.yml
```diff
@@ -0,0 +1,67 @@
...
+      AZURE_ACCOUNT_NAME: ${{secrets.AZURE_ACCOUNT_NAME}}
+      AZURE_ACCOUNT_PASSWORD: ${{secrets.AZURE_ACCOUNT_PASSWORD}}
+      AZURE_SUBSCRIPTION_ID: ${{secrets.AZURE_SUBSCRIPTION_ID}}
+      AZURE_TENANT_ID: ${{secrets.AZURE_TENANT_ID}}
+      M365_ACCOUNT_NAME: ${{secrets.M365_ACCOUNT_NAME}}
+      M365_ACCOUNT_PASSWORD: ${{secrets.M365_ACCOUNT_PASSWORD}}
+
+      # If the hosting environment is not provisioned, set this environment variable to true.
+      # or if it's provisioned and has not updates, set this environment variable to false.
+      RUN_PROVISION: true
```
zhenyasav, line 23: would be good to prefix our env variables with something to namespace them to our team: `TFX_RUN_PROVISON` for example?

zhenyasav: or `TEAMSFX_RUN_PROVISION`

docs/cicd/github-ci-template.yml
```diff
@@ -0,0 +1,31 @@
...
+          node-version: '14'
+
+      # Checkout the code.
+      - uses: actions/checkout@v2
+
+      # Build the project.
+      # The way to build the current project depends on how you scaffold it.
+      # Different folder structures require different commands set.
+      - name: Build the project
+        run: cd tabs && npm install && npm run build
```
zhenyasav, line 25: use `npm ci` instead of `install` in CI containers

docs/cicd/others-script-cd-template.sh
```diff
@@ -0,0 +1,52 @@
...
+# M365_ACCOUNT_PASSWORD
+
+# If the hosting environment is not provisioned, set this environment variable to false.
+# or if it's provisioned and has not updates, set this environment variable to true.     
+# RUN_PROVISION
+
+
+# Setup environment.
+sudo apt install -y nodejs npm curl git
+sudo npm install -g @microsoft/teamsfx-cli
```
zhenyasav, line 18: why do we need sudo?

docs/cicd/others-script-cd-template.sh
```diff
@@ -0,0 +1,52 @@
...
+# M365_ACCOUNT_NAME
+# M365_ACCOUNT_PASSWORD
+
+# If the hosting environment is not provisioned, set this environment variable to false.
+# or if it's provisioned and has not updates, set this environment variable to true.     
+# RUN_PROVISION
+
+
+# Setup environment.
+sudo apt install -y nodejs npm curl git
```
zhenyasav, line 17: why do we need sudo?

docs/cicd/others-script-cd-template.sh
zhenyasav: also shouldn't we use a local version of teamsfx and not a global one? perhaps aliased through npm scripts so it's easy to run? This way you could ensure you version lock your project to a specific teamsfx version

docs/cicd/others-script-cd-template.sh
```diff
@@ -0,0 +1,52 @@
...
+sudo npm install -g @microsoft/teamsfx-cli 
+
+# Checkout the code.
+git clone {RepositoryEndpoint}
+cd {FolderName}
+
+# Provision hosting environment.
+if [[ "${RUN_PROVISION}" = "true" ]]
+then
+    teamsfx provision --subscription ${AZURE_SUBSCRIPTION_ID}
```
zhenyasav, line 27: when using a local version, this would be aliased like `npm run teamsfx provision...`

docs/cicd/others-script-cd-template.sh
```diff
@@ -0,0 +1,52 @@
...
+    git add .fx
+    git commit -m "chore: commit provision configs"
+    git push
+fi
+
+# Deploy to hosting environment.
+teamsfx deploy
+      
+# Validate Teams App Manifest.
+teamsfx validate
```
zhenyasav, line 42: why are we doing this after deploy?

docs/cicd/others-script-cd-template.sh
```diff
@@ -0,0 +1,52 @@
...
+
+# Build Teams App's Package.
+teamsfx build
```
zhenyasav, line 45: why are we doing this after deploy?

docs/cicd/others-script-ci-template.sh
```diff
@@ -0,0 +1,22 @@
+# This is just an example workflow for continous integration.
+# You should customize it to meet your own requirements.
+
+# Setup environment.
+# The `apt install` command is supposed to run inside the latest ubuntu system.
+# If you're using other platforms, please customize the command to set up your environment.
+sudo apt install -y nodejs npm
```
zhenyasav, line 7: why sudo?

docs/cicd/others-script-ci-template.sh
```diff
@@ -0,0 +1,22 @@
...
+
+# Checkout the code.
+# Adapt the placeholders to meet your needs.
+git clone {RepositoryEndpoint}
+cd {FolderName}
+
+# Build the project.
+# The way to build the current project depends on how you scaffold it.
+# Different folder structures require different commands set.
+cd tabs && npm install && npm run build
```
zhenyasav, line 17: `npm ci` instead of `npm install` in CI containers

docs/cicd/others-script-ci-template.sh
```diff
@@ -0,0 +1,22 @@
...
+
+# Run unit test.
+# Currently, no opinioned solution for unit test provided during scaffolding, so,
```
zhenyasav, line 20: we should just drop in a default test framework like jest

zhenyasav (changes requested):
I will push more changes to the readme areas, otherwise there are a bunch of comments to look at. Thanks!

docs/cicd/others-script-ci-template.sh
zhenyasav: or mocha

docs/cicd/github-cd-template.yml
IvanJobs: Agree, I will change it to `TEAMSFX_RUN_PROVISION`.

docs/cicd/github-ci-template.yml
IvanJobs: Better, update it later.

docs/cicd/others-script-cd-template.sh
IvanJobs: > why do we need sudo?

Usually, when installing packages for the system, high level privileges will be required, adding `sudo` is a way for this purpose.

IvanJobs: > also shouldn't we use a local version of teamsfx and not a global one? perhaps aliased through npm scripts so it's easy to run? This way you could ensure you version lock your project to a specific teamsfx version

The idea of using local version of teamsfx-cli is better, will update it later.

zhenyasav: if high level privileges are required for installation of stuff like this, something is misconfigured. Just because we think something is "usual" doesn't mean we should make it a standard practice. Please take a look at how this can be done without sudo. Using sudo is not a best practice.

docs/cicd/others-script-cd-template.sh
IvanJobs: I'm going to use something like `npx -p @microsoft/teamsfx-cli@^0.2.2 teamsfx provision`. In this way, developers can specify the version of teamsfx-cli then want to use and it's a local version.

docs/cicd/github-ci-template.yml
zhenyasav: later like in another PR? why not now?

docs/cicd/others-script-cd-template.sh
zhenyasav: that's not better than using a local version. developers should be able to lock the version in their package.json and package-lock.json, instead of being forced to update their CI scripts whenever they update the toolchain

docs/cicd/others-script-cd-template.sh
IvanJobs: Actually there's no dependency between `teamsfx deploy` and `teamsfx validate`, the order doesn't matter too much.
In purpose, I put `teamsfx validate` before `teamsfx publish` and manifest should be valid before we publish the app.

docs/cicd/others-script-cd-template.sh
IvanJobs: `teamsfx build` is to build the Teams app's package zip which can be uploaded into teams client for installation.
The order here doesn't matter too much, and we can put it anywhere after `teamsfx provision`.

docs/cicd/others-script-cd-template.sh
zhenyasav: sounds like our lifecycle model is not implemented correctly. "validate" is the same as "test" and tests should run before any build.

docs/cicd/others-script-cd-template.sh
zhenyasav: I expect that to be able to deploy, you have valid build products to deploy. Build should have happened before or within "deploy" and shouldn't be happening again after deploy

zhenyasav: if the order doesn't matter it means our lifecycle model is broken, let's discuss more to help clarify

docs/cicd/github-ci-template.yml
IvanJobs: > later like in another PR? why not now?

No, in this PR, that means wip.

docs/cicd/others-script-cd-template.sh
IvanJobs: > that's not better than using a local version. developers should be able to lock the version in their package.json and package-lock.json, instead of being forced to update their CI scripts whenever they update the toolchain

To clarify, the example scripts are provided to help developers build pipelines in scripts. For example, in the case of Jenkins, they may write scripts using teamsfx-cli, this script file will give them a getting started example. So my point is they are not developing a npm package with package.json inside a project but only writing scripts under a specific cicd pipeline context, there's no package.json there. 
I'm not sure if I made myself clean enough, and how do you think about the point I described above?

docs/cicd/others-script-cd-template.sh
IvanJobs: I'm okay to remove `sudo`, but will add comments to remind users to prepare efficient permissions.

docs/cicd/others-script-cd-template.sh
IvanJobs: One thing is, it's kind of cli's sub-command naming issue, and we can discuss more about this.

docs/cicd/others-script-cd-template.sh
IvanJobs: `sudo` removed, and comments to remind users are added.

docs/cicd/others-script-cd-template.sh
IvanJobs: It seems teamsfx-cli needs to clarify concepts of these sub-commands to avoid ambiguity. Let me resolve this first.

docs/cicd/others-script-ci-template.sh
IvanJobs: I guess the expected change is only about the comments. Could you please help modify the comments to call out jest or mocha? thanks.

MuyangAmigo (approved):

docs/cicd/others-script-cd-template.sh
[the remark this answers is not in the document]
IvanJobs: Good reference, add it soon.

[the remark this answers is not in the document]
IvanJobs: since we use a local version of cli inside the project, the issue is out-of-date.

docs/cicd/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+
+## Getting started guide for other platforms
+You can follow the pre-defined example scripts to build and customize CI/CD pipelines on other platforms like Jenkins:
+* [CI Scripts](https://github.com/OfficeDev/TeamsFx/blob/main/docs/cicd/others-script-ci-template.sh)
+* [CD Scripts](https://github.com/OfficeDev/TeamsFx/blob/main/docs/cicd/others-script-cd-template.sh)
+
+The scripts are based on a cross-platform TeamsFx command line tool [TeamsFx-CLI](https://www.npmjs.com/package/@microsoft/teamsfx-cli). You can install it with `npm install -g @microsoft/teamsfx-cli` and follow the [documentation](https://github.com/OfficeDev/TeamsFx/blob/dev/docs/cli/user-manual.md) to customize the scripts.
+
+Please keep in mind that you need to set Azure and M365 credentials in your environment variables safely.
```
zhenyasav, line 51: what do we mean by "safely" here? I think we need to probably be using GitHub secrets and point people in the right direction with a link

docs/cicd/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+|Name|Description|
+|---|---|
+|AZURE_ACCOUNT_NAME|The account name of Azure which is used to provision resources.|
+|AZURE_ACCOUNT_PASSWORD|The password of Azure account.|
+|AZURE_SUBSCRIPTION_ID|To identify the subscription in which the resources will be provisined.|
+|AZURE_TENANT_ID|To identify the tenant in which the subscription resides.|
+|M365_ACCOUNT_NAME|The M365 account for creating and publishing the Teams App.|
+|M365_ACCOUNT_PASSWORD|The password of the M365 account.|
+|M365_TENANT_ID|To identify the tenant in which the Teams App will be created/published. PS: if not provided, the default tenant ID will be used.|
```
zhenyasav, line 41: what is the default Tenant ID? that needs some explanation.

docs/cicd/README.md
```diff
@@ -0,0 +1,54 @@
...
+1. Remove the test scripts if you don't have tests.
+
+> Note: The provision step is expected to run separately by hand or by workflow. Please remember to commit after provisioning (results of provisioning will be deposited inside the `.fx` folder) and save required secrets into GitHub secrets to generate file `default.userdata`.
+
+### Environment Variables
+
+|Name|Description|
+|---|---|
+|AZURE_ACCOUNT_NAME|The account name of Azure which is used to provision resources.|
+|AZURE_ACCOUNT_PASSWORD|The password of Azure account.|
```
zhenyasav, line 36: this is a little scary - we should be thinking about GitHub Secrets for this

docs/cicd/README.md
```diff
@@ -0,0 +1,54 @@
...
+|AZURE_SUBSCRIPTION_ID|To identify the subscription in which the resources will be provisined.|
+|AZURE_TENANT_ID|To identify the tenant in which the subscription resides.|
+|M365_ACCOUNT_NAME|The M365 account for creating and publishing the Teams App.|
+|M365_ACCOUNT_PASSWORD|The password of the M365 account.|
```
zhenyasav, line 40: this also - putting passwords in ENV is never a good idea anywhere

zhenyasav: I didn't realize we're recommending GH secrets above in line 25, apologies

docs/cicd/README.md
```diff
@@ -0,0 +1,54 @@
...
+There are some potential changes you can make to adapt your project:
+
+1. Change how the CI flow is triggered. We default it to when pull request is created targeting dev branch.
+1. Ensure you have an npm build script, or customize the way you build in the automation code.
+1. Ensure you have an npm test script which returns zero for success, and/or change the test commands.
+
+### Customize CD Workflow
+You may want to change:
+1. How the CD flow is triggered. By default it happens when new commits are made to the `main` branch.
+1. Create GitHub repository secrets by environment to hold Azure/M365 login credentials. The table below lists all the secrets you need to create on GitHub, and for detailed usage, please refer to the GitHub Actions [README.md](https://github.com/OfficeDev/teamsfx-cli-action/blob/main/README.md).
```
zhenyasav, line 25: we should link directly to the GH secrets documentation here as well

docs/cicd/github-cd-template.yml
```diff
@@ -0,0 +1,97 @@
...
+
+      # Run unit test.
+      # Currently, no opinioned solution for unit test provided during scaffolding, so,
+      # set up any unit test framework you prefer (for example, mocha or jest) and update the commands accordingly in below.
+      - name: Run Unit Test
+        run: cd tabs && npm run test
+
+      # We suggest to do the provision steps by case manually or in a separated workflow, so just comment the following steps for references.
+      # After provision, you should commit .fx/env.default.json into the repository for later use.
+      # You should pick required secrets from .fx/default.userdata, and save them in the repository secrets (https://docs.github.com/en/actions/reference/encrypted-secrets) which can be refered by the step with name 'Generate default.userdata'.
```
timngmsft, line 47: This means out of the box the action is not going to work right? That the user needs to do these things in order for deploy to work without provision?

timngmsft (changes requested):

docs/cicd/github-cd-template.yml
```diff
@@ -0,0 +1,97 @@
...
+        uses: OfficeDev/teamsfx-cli-action@v1
+        with:
+          commands: deploy
+
+      # This step is to pack the Teams App as zip file,
+      # which can be used to be uploaded onto Teams Client for installation.
+      - name: Build Teams App's package
+        uses: OfficeDev/teamsfx-cli-action@v1
+        with:
+          commands: build
```
zhenyasav, line 86: we need to fix this lifecycle, we should be doing build, then deploy, then publish. Building after deploy before publish breaks the lifecycle model

timngmsft, line 86: Is `teamsfx build` actually building the zip file and not building the project?

timngmsft (changes requested):

docs/cicd/others-script-cd-template.sh
```diff
@@ -0,0 +1,64 @@
+# This is just an example workflow for continuous deployment.
```
zhenyasav, line 1: bash files need to start with `#!/usr/bin/env bash` and then you need `set -euxo pipefail` or else the script won't halt execution if `npm run test` returns nonzero

zhenyasav: technically you don't need `set -x` but it's good practice because the logs will show what commands are being executed. WARNING when you have random secrets in ENV variables, they will get printed to stdout if they are being expanded in any variables like $VAR, this is another reason why passwords in ENV are a bad idea. (glad we are at least recommending encryption with GH secrets)

docs/cicd/others-script-cd-template.sh
```diff
@@ -0,0 +1,64 @@
...
+# git commit -m "chore: commit provision configs"
+# git push
+
+# Generate default.userdata
+[ ! -z "${USERDATA_TENANT_ID}" ] && echo "solution.teamsAppTenantId=${USERDATA_TENANT_ID}" >> .fx/default.userdata
+[ ! -z "${USERDATA_CLIENT_SECRET}" ] && echo "fx-resource-aad-app-for-teams.clientSecret=${USERDATA_CLIENT_SECRET}" >> .fx/default.userdata
+[ ! -z "${USERDATA_BOT_PASSWORD}" ] && echo "fx-resource-bot.botPassword=${USERDATA_BOT_PASSWORD}" >> .fx/default.userdata
+
+# Deploy to hosting environment.
+cd .. && npx teamsfx deploy
```
zhenyasav, line 52: this is ok as is, but we technically shouldn't have to refer to teamsfx here, this can be an npm script `"deploy" : "teamsfx deploy"` and this code here can become agnostic of teamsfx `npm run deploy` this enables more flexibility for developers where they can tweak their deploy/build scripts without having to change CI scripts. This is fine as is for now though.

docs/cicd/others-script-ci-template.sh
```diff
@@ -0,0 +1,25 @@
+# This is just an example workflow for continuous integration.
```
zhenyasav, line 1: `#!/usr/bin/env bash` and `set -euxo pipefail`

zhenyasav (changes requested):
A few minor changes to look at, I updated the readme myself, but there are a few questions left about tenant ID for example. Also take a look at the bash files and try to debug what happens when `npm run test` fails with nonzero codes, I strongly recommend `set -euxo pipefail` at the top of any bash code, as well as of course the hashbang comment ahead of that `#!/usr/bin/env bash`

docs/cicd/github-cd-template.yml
zhenyasav: i guess its ok to push this out as is and change later, i am just re-iterating my feedback from before

docs/cicd/README.md
zhenyasav: I think I added a link here, this comment might be out of date now

docs/cicd/README.md
IvanJobs: Yeah, I should put it more specifically here. `safely` is too general.

docs/cicd/README.md
IvanJobs: @xiaolang124 has more context about this. He may help with this explanation.

docs/cicd/README.md
IvanJobs: I've checked that the link is added and let me resolve this comment.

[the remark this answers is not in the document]
IvanJobs: The link is added by Zhenya, and I'm going to change the text to `Create GitHub Encrypted Secrets for your environment`.

docs/cicd/README.md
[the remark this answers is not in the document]
IvanJobs: Agree, nice to call out this here.

docs/cicd/github-cd-template.yml
IvanJobs: Yes, that's by design. Since provision is an operation which will be performed by case, we expect users to do it separately.

docs/cicd/github-cd-template.yml
IvanJobs: I agree with this. We will drive to refactor the naming of cli's sub-commands.

IvanJobs: Yeah, as the name told us, currently `teamsfx build` only builds the appPackage.zip.

docs/cicd/README.md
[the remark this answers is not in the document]
MuyangAmigo: I added the steps above to explain how to create an `environment` and `secrets`

docs/cicd/README.md
MuyangAmigo: Updated this.

docs/cicd/others-script-cd-template.sh
IvanJobs: updated

docs/cicd/others-script-ci-template.sh
IvanJobs: updated

docs/cicd/README.md
MuyangAmigo: I have updated this part to be more clear and included a doc link explaining how to find tenant id.

docs/cicd/others-script-cd-template.sh
```diff
@@ -0,0 +1,67 @@
...
+# AZURE_TENANT_ID
+# M365_ACCOUNT_NAME
+# M365_ACCOUNT_PASSWORD
+
+# Setup environment.
+# Sufficient permissions are required to run the commands below.
+apt install -y nodejs npm git
+
+# Checkout the code.
+git clone {RepositoryEndpoint}
```
zhenyasav, line 19: this doesn't look like a shell variable - are users expected to change this value? we should at least call it out in a comment or something

docs/cicd/others-script-cd-template.sh
```diff
@@ -0,0 +1,67 @@
...
+cd {FolderName}
```
zhenyasav, line 20: we should call this out in a comment so people know to change this

zhenyasav (changes requested):
two more issues I found: there is a spot in the bash scripts that uses `{FolderName}` and `{RepositoryEndpoint}` which don't look like valid bash code and should likely be explained in a comment in-place.

docs/cicd/github-cd-template.yml
timngmsft: Can we add that to the README? Like a section that is REQUIREMENTS or something, so that people can see. Otherwise the only way I would find out is to read the YML file.

docs/cicd/github-cd-template.yml
timngmsft: @MSFT-yiz or @eriolchan Can you track this? I'll resolve the comment below. Basically - `npm build` vs `teamsfx build` is confusing if `teamsfx build` just packages the app.

docs/cicd/github-cd-template.yml
IvanJobs: There exists one `Note:` section under `Customize CD Workflow` inside README to describe this. How about it?

docs/cicd/others-script-cd-template.sh
IvanJobs: Good ideas to decouple deployment details to a specific teamsfx command. Let's keep it as it is first.

docs/cicd/others-script-cd-template.sh
IvanJobs: It's better to call this out. Updated.

docs/cicd/others-script-cd-template.sh
IvanJobs: Updated.

IvanJobs:
> two more issues I found: there is a spot in the bash scripts that uses `{FolderName}` and `{RepositoryEndpoint}` which don't look like valid bash code and should likely be explained in a comment in-place.

hi @zhenyasav  @timngmsft , New comments are replied or updated, please help review, thx.

zhenyasav (approved):
LGTM, good work.

timngmsft (approved):

docs/cicd/others-script-cd-template.sh
IvanJobs: @qinezh FYI

PR #1658, opened 2021-07-20
fix: optimize login telemetry

xiaolang124:
1. Add error detail to telemetry.


packages/cli/src/commonlib/codeFlowLogin.ts
```diff
@@ -198,6 +199,18 @@ export class CodeFlowLogin {
 
       redirectPromise.then(cancelCodeTimer, cancelCodeTimer);
       accessToken = await redirectPromise;
+    } catch (e) {
+      CliTelemetry.sendTelemetryEvent(TelemetryEvent.AccountLogin, {
+        [TelemetryProperty.AccountType]: this.accountName,
+        [TelemetryProperty.Success]: TelemetrySuccess.No,
+        [TelemetryProperty.UserId]: "",
+        [TelemetryProperty.Internal]: "false",
```
swatDong, line 207: Could `Internal` just be empty? Since it has nothing to do with failed event.
The same question to all telemetries in this PR.

xiaolang124: Good idea, fixed.

swatDong (approved):

PR #1668, opened 2021-07-21
refactor: add scaffold stage to app studio plugins

yiqing-zhao:
Add scaffold stage to app studio plugins and modify related unit tests.


jayzhang (approved):

packages/fx-core/src/plugins/resource/appstudio/index.ts
```diff
@@ -157,6 +157,23 @@ export class AppStudioPlugin implements Plugin {
     return this.appStudioPluginImpl.createManifestForRemote(ctx, maybeSelectedPlugins, manifest);
   }
 
+  public async scaffold(ctx: PluginContext): Promise<Result<any, FxError>> {
+    TelemetryUtils.init(ctx);
+    TelemetryUtils.sendStartEvent(TelemetryEventName.scaffold);
+    try {
+      TelemetryUtils.sendSuccessEvent(TelemetryEventName.scaffold);
```
nliu-ms, line 164: success event should be sent after scaffold succeed.

yiqing-zhao: Thanks! The telemetry will now be sent after scaffold succeed.

PR #1669, opened 2021-07-21
docs(cli): add config command in user manual

tecton:


docs/cli/user-manual.md
```diff
@@ -246,4 +247,52 @@ teamsfx preview --remote
...
+
+### View all the configuration in project
+The secret will automatically decrypted.
+```bash
+teamsfx config get
+```
+
+### Update the secret configuration in project
+```bash
+teamsfx config set fx-resource-aad-app-for-teams.clientSecret
```
Alive-Fish, line 297: Does this need a value?

tecton: Thanks, updated

Alive-Fish (approved):

PR #1674, opened 2021-07-21
chore: add telemetry when user open & update teams app project

HuihuiWu-Microsoft:
ADO: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10385602


HuihuiWu-Microsoft:
According to my test, there's one thing need to notice: when user save file without any change, the notify will also fire. I guess it's not a big deal, we can regard user as engaged if he manually save the file even without modify.

chagong (approved):

1openwindow (approved):

PR #1678, opened 2021-07-21
fix: add query param to app studio calls

nliu-ms:
Add a query parameter to app studio calls, to let dev portal's telemetry know how many calls are from Teams toolkit.

Related bug: 10386504

image


KennethBWSong (approved):

jayzhang (approved):

IvanJobs (approved):
lgtm

PR #1698, opened 2021-07-22
fix(bot sdk): fix login hint as null

dilin-MS:
1. Hot fix bot sdk login hint issue: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10230876
2. update teamsfx sdk version in bot project to `^0.2.0`


blackchoey (approved):

wenytang-ms (approved):

eriolchan (approved):

packages/sdk/src/bot/teamsBotSsoPrompt.ts
```diff
@@ -294,7 +294,8 @@ export class TeamsBotSsoPrompt extends Dialog {
       "Get Teams member account user principal name: " + account.userPrincipalName
     );
 
-    const signInResource = this.getSignInResource(account.userPrincipalName);
+    const loginHint: string = account.userPrincipalName ? account.userPrincipalName : "";
```
IvanJobs, line 297: Nit: `const loginHint: string = account.userPrincipalName ?? ""` may work and be simpler.

YitongFeng-git (approved):

packages/sdk/src/bot/teamsBotSsoPrompt.ts
dilin-MS: This seems like a syntactic sugar and may not be as clear as the original one. Seems conditional operation is generic in many languages. I prefer to keep the original one. Thx.

IvanJobs (approved):
lgtm

PR #1700, opened 2021-07-22
chore: add project-id and correlation-id for CLI telemetry

chagong:
chore: add project-id and correlation-id for CLI telemetry
image

https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10160555

### DIFF
--- a/packages/cli/src/yargsCommand.ts
+++ b/packages/cli/src/yargsCommand.ts
@@ -14,6 +14,7 @@ import CliTelemetryInstance, { CliTelemetry } from "./telemetry/cliTelemetry";
 import { CliTelemetryReporter } from "./commonlib/telemetry";
 import { readFileSync } from "fs";
 import path from "path";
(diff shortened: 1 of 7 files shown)


packages/cli/src/yargsCommand.ts
```diff
@@ -14,6 +14,7 @@ import CliTelemetryInstance, { CliTelemetry } from "./telemetry/cliTelemetry";
 import { CliTelemetryReporter } from "./commonlib/telemetry";
 import { readFileSync } from "fs";
 import path from "path";
+import { Correlator } from "../../fx-core/build";
```
swatDong, line 17: from `@microsoft/teamsfx-core`

chagong: oh, yeah, thanks

Alive-Fish (approved):

swatDong (approved):

PR #1702, opened 2021-07-23
chore: add feature flag function for arm

blackchoey:
Add an extra `isArmSupportEnabled` function for ARM, so we can turn on/off ARM support by default in code easily.

### DIFF
--- a/packages/fx-core/src/common/tools.ts
+++ b/packages/fx-core/src/common/tools.ts
@@ -492,11 +492,7 @@ export function isFeatureFlagEnabled(featureFlagName: string, defaultValue = fal
   if (flag === undefined) {
     return defaultValue; // allows consumer to set a default value when environment variable not set
   } else {
-    if (flag === "1" || flag.toLowerCase() === "true") {
-      return true; // can enable feature flag by set environment variable value to "1" or "true"
-    } else {
-      return false;
-    }
+    return flag === "1" || flag.toLowerCase() === "true"; // can enable feature flag by set environment variable value to "1" or "true"
   }
 }
 
--- a/packages/fx-core/src/common/tools.ts
+++ b/packages/fx-core/src/common/tools.ts
@@ -487,13 +487,21 @@ export async function askSubscription(
 }
 
 // Determine whether feature flag is enabled based on environment variable setting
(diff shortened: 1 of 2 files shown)


packages/fx-core/src/common/tools.ts
```diff
@@ -487,13 +487,21 @@ export async function askSubscription(
...
+      return true; // can enable feature flag by set environment variable value to "1" or "true"
+    } else {
+      return false;
+    }
   }
-  return false;
+}
+
+export function isArmSupportEnabled(): boolean {
+  return isFeatureFlagEnabled("TEAMSFX_ARM_SUPPORT", false);
```
SLdragon, line 504: false can be omitted

packages/fx-core/src/common/tools.ts
```diff
@@ -487,13 +487,21 @@ export async function askSubscription(
...
-export function isFeatureFlagEnabled(featureFlagName: string): boolean {
+export function isFeatureFlagEnabled(featureFlagName: string, defaultValue = false): boolean {
   const flag = process.env[featureFlagName];
-  // can enable feature flag by set environment variable value to "1" or "true"
-  if (flag && (flag === "1" || flag.toLowerCase() === "true")) {
-    return true;
+  if (flag === undefined) {
+    return defaultValue; // allows consumer to set a default value when environment variable not set
+  } else {
+    if (flag === "1" || flag.toLowerCase() === "true") {
```
SLdragon, line 495: simplify to
```
return flag === "1" || flag.toLowerCase() === "true"
```

blackchoey: updated

packages/fx-core/src/common/tools.ts
blackchoey: yes, but here aims to turn off the feature flag explicitly. So I don't need to care what's the default value provided by `isFeatureFlagEnabled`. Imaging someone changed the default value of `isFeatureFlagEnabled` to true, but I always want this method to return false by default.

SLdragon: OK, I know

SLdragon (approved):

PR #1703, opened 2021-07-23
fix(localdebug): improve certificate flow

swatDong:
Update certificate properties and install locations for security requirement:
- Basic Constraints - Subject Type=End Entity
- For windows, use `certutil` to install to CurrentUser/TrustedPeople(Group Policy) store
- For mac, use `security add-trusted-cert` to install to $HOME/Library/Keychains/login.keychain-db


packages/fx-core/src/plugins/resource/localdebug/certificate.ts
```diff
@@ -89,11 +95,16 @@ export class LocalCertificateManager {
...
-  private async generateCertificate(certFile: string, keyFile: string): Promise<void> {
+  private async generateCertificate(certFile: string, keyFile: string): Promise<string> {
     // prepare attributes and extensions
     const now = new Date();
     const expiry = new Date();
-    expiry.setFullYear(expiry.getFullYear() + 1);
+    if (os.type() === "Windows_NT") {
+      expiry.setDate(expiry.getDate() + 7);
+    } else {
+      expiry.setFullYear(expiry.getFullYear() + 1);
```
kuojianlu, line 105: Just curious about why the expiration is still one year for macos.

kuojianlu (approved):

packages/fx-core/src/plugins/resource/localdebug/certificate.ts
swatDong: Still need user's credential on macos, so one week is a bit frequent...
I'll confirm the UX change with PM before merging this PR, so pending change may come in~

swatDong: Checked and update the UX:
- Linux & Mac: no change, keep the current UX
- Windows: do not show the confirm message, but popup a warning notification so that user can still view detailed documentation if wanted

PR #1705, opened 2021-07-23
fix: update plugin get sub

xiaolang124:
1. Change plugins get sub method.
2. Update test cases.


hund030 (approved):

KennethBWSong (approved):

jayzhang (approved):

packages/fx-core/src/plugins/resource/apim/factory.ts
```diff
@@ -168,11 +168,16 @@ export class Factory {
       "credential",
       await azureAccountProvider?.getAccountCredentialAsync()
     );
-    const apiManagementClient = new ApiManagementClient(credential, solutionConfig.subscriptionId);
+    const subscriptionInfo = await azureAccountProvider?.getSelectedSubscription();
+    AssertNotEmpty("subscriptionInfo", subscriptionInfo);
```
XiaofuHuang, line 172: Let `subscriptionInfo=AssertNotEmpty("subscriptionInfo", subscriptionInfo);`
so `subscriptionInfo.subscriptionId` can be used without `!`

xzf0587 (approved):

XiaofuHuang (approved):

packages/fx-core/src/plugins/resource/apim/factory.ts
xiaolang124: Good idea, but ts compile check does not recognize ```AssertNotEmpty```.

Alive-Fish (approved):

YitongFeng-git (approved):
lgtm for bot plugin

PR #1709, opened 2021-07-23
perf(arm support): add frontend hosting generate arm templates

dilin-MS:
1. [[resource][frontend-hosting] generateArmTemplate API unit test](https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10397150)
2. [[resource][frontend-hosting] generateArmTemplate API Implementation](https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10397148)
3. [bicep for front-hosting](https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10397164)


packages/fx-core/templates/plugins/resource/frontend/bicep/frontendHosting.bicep
```diff
@@ -0,0 +1,22 @@
+@minLength(3)
```
xzf0587, line 1: should we consider the bicep file as static file?
currently our static files are stored centralized under fx-core/templates directory since it can simplify webpack work.

packages/fx-core/templates/plugins/resource/frontend/bicep/module.template.bicep
```diff
@@ -0,0 +1,7 @@
+
+module frontendHostingProvision '{{fx-resource-frontend-hosting.modules.frontendHostingProvision.path}}' = {
+  name: 'frontendHostingDeploy'
```
hund030, line 3: I know that our provision step is the deployment for ARM, but this naming is still confusing for me.
You have a module named xxProvision and the module's name is xxDeploy.

packages/fx-core/tests/plugins/resource/frontend/unit/expectedBicepFiles/frontend_hosting_test.bicep
```diff
@@ -0,0 +1,22 @@
...
+    isHnsEnabled: false
+  }
+  sku: {
+    name: 'Standard_LRS'
+  }
+}
+
+output storageName string = frontendHostingStorage.name
+output endpoint string = frontendHostingStorage.properties.primaryEndpoints.web
+output domain string = replace(replace(frontendHostingStorage.properties.primaryEndpoints.web, 'https://', ''), '/', '')
```
hund030, line 22: Seems the endpoint looks like `https://domain.com/`? Currently we output the endpoint without the tailing slash. It may break those plugins who consume the endpoint if we have it. Please help confirm.

packages/fx-core/src/plugins/resource/frontend/index.ts
```diff
@@ -81,6 +81,13 @@ export class FrontendPlugin implements Plugin {
     );
   }
 
+  public async generateArmTemplates(ctx: PluginContext): Promise<TeamsFxResult> {
+    FrontendPlugin.setContext(ctx);
+    return this.runWithErrorHandling(ctx, TelemetryEvent.Deploy, () =>
```
blackchoey, line 86: The telemetry event should not be deploy?

packages/fx-core/tests/plugins/resource/frontend/unit/generateArmTemplates.test.ts
```diff
@@ -0,0 +1,87 @@
...
+      },
+    };
+    chai.assert.isTrue(result.isOk());
+    if (result.isOk()) {
+      const expectedResult = mockSolutionUpdateArmTemplates(
+        mockedSolutionDataContext,
+        result.value
+      );
+
+      const expectedBicepFileDirectory = path.join(__dirname, "expectedBicepFiles");
```
hund030, line 62: Seems the ut generate a series output files. Do we need to clean up those after ut done?

hund030: NIT: And I think the best practice is unit test should not touch file system. You could consider using [mock-fs](https://www.npmjs.com/package/mock-fs) though it significantly increases your work...

packages/fx-core/tests/plugins/resource/frontend/unit/generateArmTemplates.test.ts
```diff
@@ -0,0 +1,87 @@
...
+      "fx-resource-frontend-hosting": {
+        modules: {
+          frontendHostingProvision: {
+            path: `./${testModuleFileName}`,
+          },
+        },
+      },
+    };
+    chai.assert.isTrue(result.isOk());
+    if (result.isOk()) {
```
blackchoey, line 56: since there's already assert on `result.isOk()`, do we still need to check the result here?

packages/fx-core/tests/plugins/resource/util.ts
```diff
@@ -0,0 +1,32 @@
...
+export function mockSolutionUpdateArmTemplates(
+  mockedData: any,
+  template: ScaffoldArmTemplateResult
+): ScaffoldArmTemplateResult {
+  return {
+    Modules: template.Modules,
+    Orchestration: {
+      ParameterTemplate: {
+        Content: compileHandlebarsTemplateString(
+          template.Orchestration.ParameterTemplate!.Content,
```
blackchoey, line 13: maybe need to handle cases that resource plugin doesn't provide this content? We can also update this logic if we meet this case in the future.

blackchoey (commented):
Since our scaffolding template does not have copyright header, we may also not need to add copyright header to our bicep template. But let's confirm with @zyxiaoyuer

packages/fx-core/templates/plugins/resource/frontend/bicep/frontendHosting.bicep
dilin-MS: Yes. The file is placed under *fx-core/templates* directory.

packages/fx-core/src/plugins/resource/frontend/index.ts
dilin-MS: Nice catch. Fixed.

packages/fx-core/templates/plugins/resource/frontend/bicep/input_param.template.bicep
```diff
@@ -0,0 +1,5 @@
+
+@minLength(3)
+@maxLength(24)
+@description('Name of Storage Accounts for frontend hosting.')
+param frontendHosting_storageName string = 'frontendstg${uniqueString(resourceBaseName)}'
```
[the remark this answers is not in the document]
dilin-MS: This is an input parameter in the final main.bicep and will be generated by solution plugin.

packages/fx-core/templates/plugins/resource/frontend/bicep/module.template.bicep
dilin-MS: This should be changed to `frontendHostingProvision`. Thx. Fixed.

packages/fx-core/tests/plugins/resource/frontend/unit/expectedBicepFiles/frontend_hosting_test.bicep
dilin-MS: I have already trim the `https://` and the output is the same as current behavior. Thx.

packages/fx-core/src/plugins/resource/frontend/constants.ts
```diff
@@ -50,6 +50,11 @@ export class EnvironmentVariables {
 export class FrontendPathInfo {
   static WorkingDir = "tabs";
   static TemplateDir = path.join("plugins", "resource", "frontend");
+  static bicepTemplateFolderName = "bicep";
```
eriolchan, line 53: @hund030  What's our code style of static variable of Class? I see some are camelCase, some are PascalCase

packages/fx-core/src/plugins/resource/frontend/index.ts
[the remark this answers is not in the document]
dilin-MS: I suppose yes?

packages/fx-core/tests/plugins/resource/util.ts
dilin-MS: Nice catch. Fixed.

packages/fx-core/src/plugins/resource/frontend/plugin.ts
```diff
@@ -249,4 +253,55 @@ export class FrontendPluginImpl {
...
+    );
+    const outputOrchestrationFilePath = path.join(
+      bicepTemplateDir,
+      FrontendPathInfo.outputOrchestrationFileName
+    );
+
+    const result: ScaffoldArmTemplateResult = {
+      Modules: {
+        simpleAuthProvision: {
+          Content: await fs.readFile(moduleFilePath, "utf-8"),
```
eriolchan, line 284: I guess there is a constant for "utf-8", if not, please help create one.

packages/fx-core/tests/plugins/resource/frontend/unit/expectedBicepFiles/frontend_hosting_test.bicep
eriolchan: It seems this test bicep file is the same content as packages/fx-core/templates/plugins/resource/frontend/bicep/frontendHosting.bicep. Is it possible to use the same file in the unit test so that we can keep single source of truth?

packages/fx-core/src/plugins/resource/frontend/index.ts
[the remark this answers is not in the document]
eriolchan: In which stage will generateArmTemplates() be called, provision or scaffolding or a new stage? I don't think it is in deployment stage

packages/fx-core/tests/plugins/resource/frontend/unit/generateArmTemplates.test.ts
dilin-MS: This is not checking whether the result is ok. It is to avoid compilation error.

packages/fx-core/src/plugins/resource/frontend/constants.ts
hund030: Let me create a PR to make it consistent.

packages/fx-core/src/plugins/resource/frontend/plugin.ts
hund030: Same as above one.

dilin-MS: Zhijie will help to make it a constant string in another PR.

packages/fx-core/src/plugins/resource/frontend/plugin.ts
```diff
@@ -249,4 +253,55 @@ export class FrontendPluginImpl {
...
+      FrontendPathInfo.moduleOrchestrationFileName
+    );
+    const outputOrchestrationFilePath = path.join(
+      bicepTemplateDir,
+      FrontendPathInfo.outputOrchestrationFileName
+    );
+
+    const result: ScaffoldArmTemplateResult = {
+      Modules: {
+        simpleAuthProvision: {
```
blackchoey, line 283: We should name it as frontendHostingProvision

hund030 (approved):
frontend part looks good to me.

packages/fx-core/src/plugins/resource/frontend/index.ts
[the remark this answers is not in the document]
dilin-MS: generateArmTemplates() is called during provision stage. Do you think it the telemetryEvent here should be `Scaffold` or `GenerateArmTemplates`?

packages/fx-core/src/plugins/resource/frontend/plugin.ts
dilin-MS: Fixed.

packages/fx-core/tests/plugins/resource/frontend/unit/expectedBicepFiles/frontend_hosting_test.bicep
dilin-MS: @eriolchan This file is to check frontend hosting plugin successfully generate bicep template files which can be correct bicep files after solution plugin compile it. Though some of the bicep template files have no changes during the process, we will keep the file here as a new file to show what the final wanted bicep file looks like.

packages/fx-core/src/plugins/resource/frontend/index.ts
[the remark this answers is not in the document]
hund030: I think it's called in scaffold stage. Is it typo?

blackchoey (approved):

packages/fx-core/src/plugins/resource/frontend/index.ts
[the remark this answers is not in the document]
dilin-MS: @eriolchan , This telemetry event name is changed to `TelemetryEvent.GenerateArmTemplates`. Using a new event name will be better since this function can be called in different stages. Thx.

KennethBWSong (approved):

PR #1711, opened 2021-07-23
fix(sql): broaden firewall and improve error message

xzf0587:
1. broaden the ip range of the local firewall rule from class C(a.b.c.0-a.b.c.255) to class B(a.b.0.0-a.b.255.255)
2. improve the error message about adding firewall rule.
image


packages/fx-core/src/plugins/resource/sql/sqlClient.ts
```diff
@@ -26,18 +27,22 @@ export class SqlClient {
         return false;
       }
     } catch (error) {
-      this.ctx.logProvider?.error(
-        ErrorMessage.SqlCheckDBUserError.message(this.config.identity, error.message)
-      );
-      if (error?.message?.includes(ErrorMessage.AccessMessage)) {
+      if (error?.message?.includes(ErrorMessage.AccessMessage) && extractIp(error.message)) {
+        const ip = extractIp(error.message);
```
adashen, line 31: better to cache the extracted value in the previous if condition.

xzf0587: it is fixed. Thanks.

adashen (approved):

packages/fx-core/src/plugins/resource/sql/errors.ts
```diff
@@ -91,7 +91,8 @@ export class ErrorMessage {
 
   public static readonly SqlAccessError = {
     name: "SqlAccessError",
-    message: (sqlName: string, detail = "") => `Failed to access server '${sqlName}'. ${detail}`,
+    message: (sqlName: string, ip: string, detail = "") =>
+      `Failed to access server '${sqlName}', please refer the help to add '${ip}' into the server firewall rule since your public ip may occasionally change. ${detail}`,
```
timngmsft, line 95: I don't think this needs to change. The '${detail}' contains enough information.

timngmsft: The change adds redundant information.

timngmsft (changes requested):

packages/fx-core/src/plugins/resource/sql/errors.ts
xzf0587: It is fixed. Thanks.

timngmsft (approved):

PR #1716, opened 2021-07-23
refactor: remove app manifest in Context

jayzhang:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10420874


packages/fx-core/src/plugins/resource/appstudio/plugin.ts
```diff
@@ -502,7 +502,9 @@ export class AppStudioPluginImpl {
       }
     } else {
       appDirectory = `${ctx.root}/.${ConfigFolderName}`;
-      manifestString = JSON.stringify(ctx.app);
+      manifestString = (
+        await fs.readFile(`${ctx.root}/.${ConfigFolderName}/${REMOTE_MANIFEST}`)
```
nliu-ms, line 506: for azure project, we need to generate remote manifest, not just read manifest file.

packages/fx-core/tests/plugins/resource/function/unit/scaffold.test.ts
```diff
@@ -43,6 +43,16 @@ const context: any = {
   root: path.join(__dirname, "ut"),
 };
 
+context.projectSettings = {
```
hund030, line 46: Seems this setting isn't used by function ut. Could you help confirm and remove this if we don't need?

KennethBWSong (approved):

YitongFeng-git (approved):

1yefuwang1 (approved):

hund030 (approved):

packages/fx-core/src/plugins/resource/appstudio/plugin.ts
```diff
@@ -328,27 +329,27 @@ export class AppStudioPluginImpl {
...
+    // if (maybeSelectedPlugins.isErr()) {
+    //   return err(maybeSelectedPlugins.error);
+    // }
+    // const selectedPlugins = maybeSelectedPlugins.value;
+    // if (selectedPlugins.some((plugin) => plugin.name === "fx-resource-bot")) {
+    const capabilities = (ctx.projectSettings?.solutionSettings as AzureSolutionSettings)
+      .capabilities;
+    const hasBot = capabilities?.includes(BotOptionItem.id);
+    const hasMsgExt = capabilities?.includes(MessageExtensionItem.id);
+    if (!hasBot && !hasMsgExt) {
```
nliu-ms, line 344: The original logic throw error when bot plugin is selected without bot/me capability.

But here, it throws error when the app is not bot/me app.

packages/fx-core/src/plugins/resource/appstudio/plugin.ts
```diff
@@ -1137,28 +1151,8 @@ export class AppStudioPluginImpl {
...
-            AppStudioError.PluginNotFound.name,
-            AppStudioError.PluginNotFound.message(name)
-          )
-        );
-      }
-      results.push(plugin);
-    }
-    return ok(results);
+    const plugins = getActivatedResourcePlugins(azureSettings);
+    azureSettings.activeResourcePlugins = plugins.map((p) => p.name);
```
nliu-ms, line 1155: seems the assignment to azureSettings is not used anymore.

getSelectedPlugins equals getActivatedResourcePlugins here.

jayzhang: Yes, just copy from solution. This method can be deleted

nliu-ms (approved):

PR #1731, opened 2021-07-26
fix(bot): scaffolding bring last project's context

YitongFeng-git:
fix bug https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10430876

### DIFF
--- a/packages/fx-core/src/plugins/resource/bot/plugin.ts
+++ b/packages/fx-core/src/plugins/resource/bot/plugin.ts
@@ -83,7 +83,7 @@ export class TeamsBotImpl {
 
   public async preScaffold(context: PluginContext): Promise<FxResult> {
     this.ctx = context;
-    await this.config.restoreConfigFromContext(context);
+    this.config = new TeamsBotConfig();
     Logger.info(Messages.PreScaffoldingBot);
 
     const rawWay = this.ctx.answers![QuestionNames.WAY_TO_REGISTER_BOT];


packages/fx-core/src/plugins/resource/bot/plugin.ts
```diff
@@ -83,7 +83,7 @@ export class TeamsBotImpl {
 
   public async preScaffold(context: PluginContext): Promise<FxResult> {
     this.ctx = context;
-    await this.config.restoreConfigFromContext(context);
+    this.config = new TeamsBotConfig();
```
henzhang-ms, line 86: does solution passes the config of current project to preScaffold of new project?

YitongFeng-git: It is not related with any restore method, after we run local debug, the settings will store in TeamsBotImpl.localDebugConfig until we reopen a folder or restart vscode. Another solution is let the TeamsBotImpl not be global.
image

eriolchan: @henzhang-ms @hund030 will this issue also happen to Frontend and Function plugin?

According to the code of [teamsBotConfig.ts](https://github.com/OfficeDev/TeamsFx/blob/a0787e0cf142f1cb1c9ee58aa6592dbdea48aeed/packages/fx-core/src/plugins/resource/bot/configs/teamsBotConfig.ts#L26), why do we always restore configs of all stages (scaffold, provision, deploy, local-debug) even when toolkit is executing only one stage of them? @IvanJobs do you have more information and comments about the fix?

henzhang-ms: Should no for function, function all the config as context given, that's why it is called `syncConfigFromContext`. I think the root cause of bot is that it only sync from context if the context value is not empty. But I am not sure whether it has some by-design concern.

```
    const objectIdValue: ConfigValue = context.config.get(PluginBot.OBJECT_ID);
    if (objectIdValue) {
      this.objectId = objectIdValue as string;
    }
```

hund030: No for frontend, either. Frontend plugin is totally stateless now. It only accepts input from context.

PR #1742, opened 2021-07-27
perf(simple auth): check and register resource provider

KennethBWSong:


adashen (changes requested):
It should not be a feature. Please update the commit message. which will impact the conventional commits

packages/fx-core/src/plugins/resource/simpleauth/webAppClient.ts
```diff
@@ -46,13 +50,31 @@ export class WebAppClient {
...
     try {
+      // Check and register resource provider
+      try {
+        DialogUtils.progressBar?.next(Constants.ProgressBar.provision.registerResourceProvider);
+        await ResourceManagementClient.ensureResourceProviders(
+          this.resourceManagementClient,
+          Constants.RequiredResourceProviders
+        );
+      } catch (error) {
+        throw ResultFactory.UserError(
```
adashen, line 71: error will block the following steps?

KennethBWSong: Yes. As user does not have permission to register the RP, we think we should block provision and ask user to manually register the RP.

adashen: The scenario is the RP is already registered although user don't have the permission to register. And the user has permission to provision.

KennethBWSong: Here we will check whether the RP is registered. If registered, we will not register again. So in this scenario, it will not throw error here.

adashen: If the user don't have permission to check whether registered?

KennethBWSong: Seems we can not find a permission on checking the register status.
image

Shall we separate the check and registering, and only throw error when fail to register?

adashen: I think register the RP is just tries to do something. If cannot sucessfully run, it should not block provision.

KennethBWSong: Do you mean we just tries to register the RP instead of checking the status first, and no matter whether it success, we will go on and try to provision app services?

KennethBWSong: Fixed. Thanks!

packages/fx-core/src/plugins/resource/simpleauth/plugin.ts
```diff
@@ -99,7 +99,7 @@ export class SimpleAuthPluginImpl {
       ctx
     );
 
-    DialogUtils.progressBar = ctx.ui?.createProgressBar(Constants.ProgressBar.provision.title, 3);
+    DialogUtils.progressBar = ctx.ui?.createProgressBar(Constants.ProgressBar.provision.title, 4);
```
adashen, line 102: Can we get the number auto? If hard code does that mean each new added step needs to increase the number?

KennethBWSong: Fixed. Thanks!

adashen (approved):

PR #1743, opened 2021-07-27
fix(bot): sync TeamsBotConfig from context

YitongFeng-git:
fix bug https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10430876

### DIFF
--- a/packages/fx-core/src/plugins/resource/bot/configs/teamsBotConfig.ts
+++ b/packages/fx-core/src/plugins/resource/bot/configs/teamsBotConfig.ts
@@ -29,33 +29,21 @@ export class TeamsBotConfig {
     await this.localDebug.restoreConfigFromContext(context);
     await this.deploy.restoreConfigFromContext(context);
 
-    const clientIdValue: ConfigValue = context.configOfOtherPlugins
+    this.teamsAppClientId = context.configOfOtherPlugins
       .get(PluginAAD.PLUGIN_NAME)
-      ?.get(PluginAAD.CLIENT_ID);
-    if (clientIdValue) {
-      this.teamsAppClientId = clientIdValue as string;
-    }
+      ?.get(PluginAAD.CLIENT_ID) as string;
 
-    const clientSecretValue: ConfigValue = context.configOfOtherPlugins
+    this.teamsAppClientSecret = context.configOfOtherPlugins
       .get(PluginAAD.PLUGIN_NAME)
-      ?.get(PluginAAD.CLIENT_SECRET);
-    if (clientSecretValue) {
-      this.teamsAppClientSecret = clientSecretValue as string;
-    }
+      ?.get(PluginAAD.CLIENT_SECRET) as string;
 
-    const tenantIdValue: ConfigValue = context.configOfOtherPlugins
+    this.teamsAppTenant = context.configOfOtherPlugins
       .get(PluginSolution.PLUGIN_NAME)
-      ?.get(PluginSolution.M365_TENANT_ID);
-    if (tenantIdValue) {
-      this.teamsAppTenant = tenantIdValue as string;
-    }
+      ?.get(PluginSolution.M365_TENANT_ID) as string;
 
-    const applicationIdUrisValue: ConfigValue = context.configOfOtherPlugins
+    this.applicationIdUris = context.configOfOtherPlugins
       .get(PluginAAD.PLUGIN_NAME)
-      ?.get(PluginAAD.APPLICATION_ID_URIS);
-    if (applicationIdUrisValue) {
-      this.applicationIdUris = applicationIdUrisValue as string;
-    }
+      ?.get(PluginAAD.APPLICATION_ID_URIS) as string;
 
     const capabilities = (context.projectSettings?.solutionSettings as AzureSolutionSettings)
       .capabilities;
@@ -89,4 +77,5 @@ export class TeamsBotConfig {
   public toString(): string {
     return JSON.stringify(this);
   }
+
 }
--- a/packages/fx-core/src/plugins/resource/bot/configs/deployConfig.ts
+++ b/packages/fx-core/src/plugins/resource/bot/configs/deployConfig.ts
@@ -10,10 +10,7 @@ export class DeployConfig {
   public unPackFlag = "true";
 
   public async restoreConfigFromContext(context: PluginContext): Promise<void> {
-    const unPackFlagValue: ConfigValue = context.config.get(PluginBot.UNPACK_FLAG);
-    if (unPackFlagValue) {
-      this.unPackFlag = unPackFlagValue as string;
-    }
+    this.unPackFlag= context.config.get(PluginBot.UNPACK_FLAG) as string;
   }
 
   public saveConfigIntoContext(context: PluginContext): void {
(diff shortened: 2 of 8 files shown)


packages/fx-core/src/plugins/resource/bot/configs/teamsBotConfig.ts
```diff
@@ -23,60 +23,60 @@ export class TeamsBotConfig {
...
-    await this.deploy.restoreConfigFromContext(context);
+  private constructor() {
+    this.scaffold = new ScaffoldConfig();
+    this.provision = new ProvisionConfig();
+    this.localDebug = new LocalDebugConfig();
+    this.deploy = new DeployConfig();
+  }
+
+  public static async restoreConfigFromContext(context: PluginContext): Promise<TeamsBotConfig> {
+    let teamsBotConfig = new TeamsBotConfig();
```
henzhang-ms, line 34: Nit: could be `const`. When we say `const` in JS, it means the reference is `const` instead of the object.

YitongFeng-git: ok

henzhang-ms: BTW, if it reset config in every restoreConfigFromContext. Would it drop any  in-memory information that's shared among the command, e.g. if preprovision wants to pass some value to provision or postprovision. Please check your design.

packages/fx-core/src/plugins/resource/bot/configs/teamsBotConfig.ts
```diff
@@ -23,60 +23,60 @@ export class TeamsBotConfig {
...
-    }
 
     const capabilities = (context.projectSettings?.solutionSettings as AzureSolutionSettings)
       .capabilities;
 
-    if (capabilities?.includes(PluginActRoles.Bot) && !this.actRoles.includes(PluginActRoles.Bot)) {
-      this.actRoles.push(PluginActRoles.Bot);
+    if (
+      capabilities?.includes(PluginActRoles.Bot) &&
+      !teamsBotConfig.actRoles.includes(PluginActRoles.Bot)
```
henzhang-ms, line 61: maybe the `!teamsBotConfig.actRoles.includes(PluginActRoles.Bot)` is not required anymore

henzhang-ms: or just use
```
teamsBotConfig.actRoles =
  capabilities?.filter(ar =>
     [PluginActRoles.Bot, PluginActRoles.MessageExtension].includes(ar)
  ) ?? [];
```

packages/fx-core/src/plugins/resource/bot/plugin.ts
```diff
@@ -83,7 +82,7 @@ export class TeamsBotImpl {
 
   public async preScaffold(context: PluginContext): Promise<FxResult> {
     this.ctx = context;
-    await this.config.restoreConfigFromContext(context);
+    let config = await TeamsBotConfig.restoreConfigFromContext(context);
```
henzhang-ms, line 85: Nit: better to user `const` to align our code pattern.

packages/fx-core/src/plugins/resource/bot/configs/teamsBotConfig.ts
```diff
@@ -23,60 +23,60 @@ export class TeamsBotConfig {
...
-    if (applicationIdUrisValue) {
-      this.applicationIdUris = applicationIdUrisValue as string;
-    }
 
     const capabilities = (context.projectSettings?.solutionSettings as AzureSolutionSettings)
       .capabilities;
 
-    if (capabilities?.includes(PluginActRoles.Bot) && !this.actRoles.includes(PluginActRoles.Bot)) {
-      this.actRoles.push(PluginActRoles.Bot);
+    if (
```
eriolchan, line 59: why don't use set instead of array so that it can simplify the logic of checking existence?

packages/fx-core/src/plugins/resource/bot/configs/teamsBotConfig.ts
henzhang-ms: There are too may replace work in this PR. Ask Ivan to check the design first. I think even we want to reset these field, we don't need to involve such large change.

YitongFeng-git: @henzhang-ms there is also a step "saveToContext", will sync botconfig to context, and next step it will restore config from context.

packages/fx-core/src/plugins/resource/bot/configs/teamsBotConfig.ts
```diff
@@ -23,60 +23,60 @@ export class TeamsBotConfig {
...
+
+  public static async restoreConfigFromContext(context: PluginContext): Promise<TeamsBotConfig> {
+    let teamsBotConfig = new TeamsBotConfig();
+    await teamsBotConfig.scaffold.restoreConfigFromContext(context);
+    await teamsBotConfig.provision.restoreConfigFromContext(context);
+    await teamsBotConfig.localDebug.restoreConfigFromContext(context);
+    await teamsBotConfig.deploy.restoreConfigFromContext(context);
 
-    const clientIdValue: ConfigValue = context.configOfOtherPlugins
+    teamsBotConfig.teamsAppClientId = context.configOfOtherPlugins
```
IvanJobs, line 40: I believe the change is slightly huge, but the key point is to make TeamsBotConfig stateless, that is, make it as a complete mirror of the context since solution will pass a fresh context whenever a new project is created.

So keep this target in mind, one thing to do is:
change code like:
` if (context.value is not empty/null) config.value = context.value` 
to
` config.value = context.value`

IvanJobs: Extra changes are kind of coding style changes, it's not related too much to fix the bug.

packages/fx-core/src/plugins/resource/bot/configs/teamsBotConfig.ts
henzhang-ms: The design has a drawback, once we need to shared some internal variable, we needs to be write it into *.env.json and expose to user. It is fine to me, just for notice.

packages/fx-core/src/plugins/resource/bot/configs/teamsBotConfig.ts
IvanJobs: Also, ut cases need to be updated accordingly.

IvanJobs (approved):
Approved since it works, but I don't prefer to write it like this to introduce those extra changes.

packages/fx-core/src/plugins/resource/bot/configs/teamsBotConfig.ts
YitongFeng-git: resolved

packages/fx-core/src/plugins/resource/bot/configs/teamsBotConfig.ts
YitongFeng-git: Will keep it for now

packages/fx-core/src/plugins/resource/bot/configs/teamsBotConfig.ts
YitongFeng-git: Will keep it for now

PR #1748, opened 2021-07-27
fix: add call destroy sockets

xiaolang124:
1. Because ```server.close``` only stops the server from accepting new connections and keeps existing connections. So I can not close old connections by ```server.on('close')```.
2. I add a direct call to ```destroySockets``` after get token.


swatDong:
What's the exiting connection(s)? Suppose the only connect is the result page and should be closed after getting token...

xiaolang124:
> What's the exiting connection(s)? Suppose the only connect is the result page and should be closed after getting token...
> 
> Noticed that `codeFlowLogin.sendFile` calls `readFile` with callback, which does not return immediately... does that cause the connection not closed?

I think `sendFIle` may have some influence. Some machines will never close the old connection even after success page shows and closes it.

swatDong:
> > What's the exiting connection(s)? Suppose the only connect is the result page and should be closed after getting token...
> > Noticed that `codeFlowLogin.sendFile` calls `readFile` with callback, which does not return immediately... does that cause the connection not closed?
> 
> I think `sendFIle` may have some influence. Some machines will never close the old connection even after success page shows and closes it.

Still need to fix `sendFile` as well. Otherwise `destroySockets` will force abort `sendFile` and user will see a broken page.
- Make `sendFile` an async call and wait for it in `codeFlowLogin`
- Destroy all sockets after getting token (at this time the `sendFile` operation itself is done)

xiaolang124:
> > > What's the exiting connection(s)? Suppose the only connect is the result page and should be closed after getting token...
> > > Noticed that `codeFlowLogin.sendFile` calls `readFile` with callback, which does not return immediately... does that cause the connection not closed?
> > 
> > 
> > I think `sendFIle` may have some influence. Some machines will never close the old connection even after success page shows and closes it.
> 
> Still need to fix `sendFile` as well. Otherwise `destroySockets` will force abort `sendFile` and user will see a broken page.
> 
> * Make `sendFile` an async call and wait for it in `codeFlowLogin`
> * Destroy all sockets after getting token (at this time the `sendFile` operation itself is done)

Good idea, fixed.

packages/cli/src/commonlib/codeFlowLogin.ts
```diff
@@ -378,12 +377,12 @@ export class CodeFlowLogin {
...
 
-function sendFile(
+async function sendFile(
   res: http.ServerResponse,
   filepath: string,
   contentType: string,
   accountName: string
-) {
+): Promise<void> {
   fs.readFile(filepath, (err, body) => {
```
swatDong, line 386: `await fs.readFile` or call `resolve` in callback, otherwise it's not a real async function...

xiaolang124: Fixed.

packages/cli/src/commonlib/codeFlowLogin.ts
```diff
@@ -378,26 +377,26 @@ export class CodeFlowLogin {
...
+  try {
+    let body = await fs.readFile(filepath);
+    let data = body.toString();
+    data = data.replace(/\${accountName}/g, accountName == "azure" ? "Azure" : "M365");
+    body = Buffer.from(data, UTF8);
+    res.writeHead(200, {
+      "Content-Length": body.length,
+      "Content-Type": contentType,
+    });
+    res.end(body);
```
swatDong, line 395: Is `res.end` sync/async/callback?

xiaolang124: ```end(chunk: any, cb?: () => void): void;```  is a sync method. And we don't use callback parameter.

swatDong: So the return for `end()` does not close the response, but `cb` means the response is ended?

swatDong (approved):

PR #1750, opened 2021-07-27
chore: add generate arm template function

blackchoey:
Add generate arm template function
Some implementation is temporary since the interface is not in TeamsFx-API yet


packages/fx-core/.eslintignore
```diff
@@ -1,2 +1,4 @@
 templates/plugins/resource/spfx/*
-src/plugins/resource/spfx/authCode.ts
\ No newline at end of file
+src/plugins/resource/spfx/authCode.ts
+src/plugins/solution/fx-solution/arm.ts
```
blackchoey, line 3: temporary ignore the 2 files to avoid eslint error, which caused by required API not in TeamsFx-API plugin interface. In the future the API will be added to plugin interface and this change can be reverted.

blackchoey: @LongOddCode just help found a better way for my situation, I will try his approach and update this PR later.

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -0,0 +1,215 @@
...
+        if (result.value.Modules) {
+          for (const module of Object.entries(result.value.Modules)) {
+            const moduleFileName = module[0];
+            const moduleFileContent = module[1].Content;
+            moduleFiles.set(generateBicepModuleFilePath(moduleFileName), moduleFileContent);
+          }
+        }
+      } else {
+        const msg = format(getStrings().solution.ProvisionFailNotice, ctx.projectSettings?.appName);
+        ctx.logProvider?.error(msg);
```
adashen, line 58: the error in result is not output here?

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -0,0 +1,215 @@
...
+        return result;
+      }
+    }
+  }
+
+  // Write bicep content to project folder
+  if (bicepOrchestrationTemplate.needsGenerateTemplate()) {
```
adashen, line 65: all the write will over write the existing files right?

1yefuwang1 (approved):

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
blackchoey: I'm following the practice in existing solution logic. Methods (scaffold/provision/etc.) in solution plugin returns the error result to core and core returns to VS Code Extension / TeamsFx CLI. It should be VS Code extension / TeamsFx CLI that finally shows the error to user.

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
blackchoey: This function currently didn't handle situations that bicep files already exists. I will add extra logic here to backup existing files after the demo.

dilin-MS: We can add a TODO comment here.

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -0,0 +1,215 @@
...
+        bicepOrchestrationTemplate.addTemplate(pluginWithArm.name, result.value);
+        if (result.value.Modules) {
+          for (const module of Object.entries(result.value.Modules)) {
+            const moduleFileName = module[0];
+            const moduleFileContent = module[1].Content;
+            moduleFiles.set(generateBicepModuleFilePath(moduleFileName), moduleFileContent);
+          }
+        }
+      } else {
+        const msg = format(getStrings().solution.ProvisionFailNotice, ctx.projectSettings?.appName);
```
SLdragon, line 55: Do we need send telemetry for the failure

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -0,0 +1,215 @@
...
+    if (pluginWithArm.scaffoldArmTemplate) {
+      // find method using method name
+      const pluginContext = getPluginContext(ctx, pluginWithArm.name);
+      const result = (await pluginWithArm.scaffoldArmTemplate(pluginContext)) as Result<
+        ScaffoldArmTemplateResult,
+        FxError
+      >;
+      if (result.isOk()) {
+        bicepOrchestrationTemplate.addTemplate(pluginWithArm.name, result.value);
+        if (result.value.Modules) {
```
SLdragon, line 47: should also return result when result.isOk()===true

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -0,0 +1,215 @@
...
+        pluginOutputContext.Outputs![outputKey] = outputValue;
+      }
+    }
+
+    this.PluginOutput[pluginName] = pluginOutputContext;
+  }
+}
+
+// Stores the bicep orchestration information for all resource plugins
+class BicepOrchestrationTemplate {
```
dilin-MS, line 132: This class has exactly the same name as the interface. Can we rename it to reduce confusion?

hund030 (approved):
Subtle changes in frontend plugin.

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -0,0 +1,215 @@
...
+      parameters: this.ParameterJsonTemplate,
+    };
+    return JSON.stringify(parameterObject, undefined, 2);
+  }
+
+  public needsGenerateTemplate(): boolean {
+    return this.TemplateAdded;
+  }
+
+  private normalizeTemplateSnippt(
```
dilin-MS, line 195: Typo here. Should be `normalizeTemplateSnippet`

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -0,0 +1,215 @@
...
+
+  constructor(pluginNames: string[]) {
+    this.RenderContenxt = new ArmTemplateRenderContext(pluginNames);
+  }
+
+  public addTemplate(pluginName: string, scaffoldResult: ScaffoldArmTemplateResult): void {
+    this.ParameterTemplate += this.normalizeTemplateSnippt(
+      scaffoldResult.Orchestration.ParameterTemplate?.Content
+    );
+    this.VariableTemplate += this.normalizeTemplateSnippt(
```
xzf0587, line 154: I found no variable template in current plugin templates.
what is the usage of it?

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
blackchoey: Thanks, we will review the telemetry strategy and add them later. Have created one task to track this.

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
dilin-MS: @xzf0587  not all plugins have variable template. It is an optional property.

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
blackchoey: thanks, renamed to BicepOrchestrationContent

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
blackchoey: Yes, bicep file contains four parts: param/variable/module/output. We want to allow resource plugins to add any of these content.

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
blackchoey: thanks, fixed

KennethBWSong (approved):

packages/fx-core/resource/strings.json
```diff
@@ -18,7 +18,8 @@
     "AddCapabilityNoticeForCli": "[Teams Toolkit] Capability \"%s\" have been successfully configured for your project, run 'teamsfx provision' will create the resource(s) in your Azure subscription.",
     "CancelProvision": "CancelProvision",
     "AlreadyProvisionNotice": "'%s' is already provisioned in the cloud, will create the teams app again if it has been removed.",
-    "NotProvisionedNotice": "'%s' is not provisioned in the cloud yet. Please provision first."
+    "NotProvisionedNotice": "'%s' is not provisioned in the cloud yet. Please provision first.",
+    "GenerateArmTemplateFailNotice": "Failed to create ARM template for required Azure services."
```
dilin-MS, line 22: Seems this constant string is not used.

blackchoey: thanks, now used in code..

PR #1752, opened 2021-07-27
refactor(core): get target env from user input instead of project settings

qinezh:
As multiple environments will be supported, the target env name should be specified by user as user input, instead of reading from the project setting. 

This PR is going to:
* Extract the logic to load env profile (including *.userdata) to `environment.ts`.
* Update the logic to load env profile from `projectSettings` to `Inputs`.
* Add unit test cases for the logic to load env profile.

Next steps:
* Extract the logic to persist the env profile from `configWriter.ts` to `environment.ts`.
* Clean up the `currentEnv` from `projectSettings`.
* ...


packages/fx-core/src/core/middleware/contextLoader.ts
```diff
@@ -90,35 +93,24 @@ export async function loadSolutionContext(
...
-          continue;
-        }
-        const secretValue = dict[secretKey];
-        const plaintext = cryptoProvider.decrypt(secretValue);
-        if (plaintext.isErr()) {
-          return err(plaintext.error);
-        }
-        dict[secretKey] = plaintext.value;
-      }
+    // ensure backforwards compatibility:
```
dooriya, line 98: >backforwards [](http://example.com/codeflow?start=14&length=12)

backforwards -> backwards?

qinezh: good catch!

packages/fx-core/src/core/environment.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    envName = envName ?? this.defaultEnvName;
+    const envFiles = this.getEnvFilesPath(envName, projectPath);
+    const userDataResult = await this.loadUserData(envFiles.userDataFile, cryptoProvider);
+    if (userDataResult.isErr()) {
+      return err(userDataResult.error);
+    }
+
+    const userData = userDataResult.value;
+    const data = await fs.readJson(envFiles.envProfile);
```
dooriya, line 47: >fs.readJson(envFiles.envProfile); [](http://example.com/codeflow?start=23&length=33)

We'd better check the `env.${envName}.json` file before reading it. We can have a pre-validation to check the envirronment files, and:
1. Implicitly create the file(s) if it is missing (may be deleted by user by accident);
**OR**
2. Pop up error if env files not found.

qinezh: For this stage(M1), I would prefer to keep the user behavior unchanged, but it's valuable to handle it gracefully. Let's add it as TODO comment and take care of it later.
Agree that we do need to return accurate error if the `env.${envName}.json` file doesn't exist for further error handling.

dooriya (approved):

dooriya (approved):

1yefuwang1 (commented):
It seems "targetEnvName" will always be undefined if this PR is merged, making the code in dev stops working.

1yefuwang1 (approved):

qinezh:
> It seems "targetEnvName" will always be undefined if this PR is merged, making the code in dev stop working.

The `targetEnvName` will be undefined until the user input for env has been enabled in UI layer. Before that, the `targetEnvName` will be set to the default value `default`, so it can keep consistent with current behavoir.

PR #1765, opened 2021-07-29
chore: load config from arm output

xzf0587:
under the isArmSupportEnabled flag:
1. add common tool to load config from arm output.
2. update simple auth config load.
3. move zip deploy to post-provision stage.
4. move config setting to post-provision stage.
5. remove appsettings.
6. disable provision stage.

### DIFF
--- a/packages/fx-core/src/plugins/resource/simpleauth/plugin.ts
+++ b/packages/fx-core/src/plugins/resource/simpleauth/plugin.ts
@@ -12,7 +12,7 @@ import * as path from "path";
 import * as fs from "fs-extra";
 import { getTemplatesFolder } from "../../..";
 import { ScaffoldArmTemplateResult } from "../../../common/armInterface";
-import { generateBicepFiles } from "../../../common";
+import { generateBicepFiles, getArmOutput, isArmSupportEnabled } from "../../../common";
 
 export class SimpleAuthPluginImpl {
   webAppClient!: WebAppClient;
@@ -135,7 +135,56 @@ export class SimpleAuthPluginImpl {
 
     const configs = Utils.getWebAppConfig(ctx, false);
 
+    if (isArmSupportEnabled()) {
+      const credentials = await ctx.azureAccountProvider!.getAccountCredentialAsync();
+      const subscriptionInfo = await ctx.azureAccountProvider?.getSelectedSubscription();
+      const subscriptionId = subscriptionInfo!.subscriptionId;
+      const resourceNameSuffix = Utils.getConfigValueWithValidation(
+        ctx,
+        Constants.SolutionPlugin.id,
+        Constants.SolutionPlugin.configKeys.resourceNameSuffix
+      ) as string;
+      const resourceGroupName = Utils.getConfigValueWithValidation(
+        ctx,
+        Constants.SolutionPlugin.id,
+        Constants.SolutionPlugin.configKeys.resourceGroupName
+      ) as string;
+      const location = Utils.getConfigValueWithValidation(
+        ctx,
+        Constants.SolutionPlugin.id,
+        Constants.SolutionPlugin.configKeys.location
+      ) as string;
+
+      const webAppName = Utils.generateResourceName(
+        ctx.projectSettings!.appName,
+        resourceNameSuffix
+      );
+      const appServicePlanName = webAppName;
+      this.webAppClient = new WebAppClient(
+        credentials!,
+        subscriptionId,
+        resourceGroupName,
+        appServicePlanName,
+        webAppName,
+        location,
+        ctx
+      );
+    }
     await this.webAppClient.configWebApp(configs);
+
+    if (isArmSupportEnabled()) {
+      const simpleAuthFilePath = Utils.getSimpleAuthFilePath();
+      await Utils.downloadZip(simpleAuthFilePath);
+      await this.webAppClient.zipDeploy(simpleAuthFilePath);
+
+      const endpoint = getArmOutput(ctx, Constants.ArmOutput.simpleAuthEndporint) as string;
+      ctx.config.set(Constants.SimpleAuthPlugin.configKeys.endpoint, endpoint);
+
+      const sku = getArmOutput(ctx, Constants.ArmOutput.simpleAuthSkuName) as string;
+      if (sku) {
+        ctx.config.set(Constants.SimpleAuthPlugin.configKeys.skuName, sku);
+      }
+    }
 
     await DialogUtils.progressBar?.end();
 
--- a/packages/fx-core/src/plugins/resource/simpleauth/plugin.ts
+++ b/packages/fx-core/src/plugins/resource/simpleauth/plugin.ts
@@ -169,10 +169,7 @@ export class SimpleAuthPluginImpl {
         location,
         ctx
       );
-    }
-    await this.webAppClient.configWebApp(configs);
 
-    if (isArmSupportEnabled()) {
       const simpleAuthFilePath = Utils.getSimpleAuthFilePath();
       await Utils.downloadZip(simpleAuthFilePath);
       await this.webAppClient.zipDeploy(simpleAuthFilePath);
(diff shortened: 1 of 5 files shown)


packages/fx-core/src/plugins/resource/simpleauth/utils/common.ts
```diff
@@ -107,13 +107,24 @@ export class Utils {
...
     ) as string;
-    const endpoint = Utils.getConfigValueWithValidation(
-      ctx,
-      isLocalDebug ? Constants.LocalDebugPlugin.id : Constants.FrontendPlugin.id,
-      isLocalDebug
-        ? Constants.LocalDebugPlugin.configKeys.endpoint
-        : Constants.FrontendPlugin.configKeys.endpoint
-    ) as string;
+
+    // const endpoint = Utils.getConfigValueWithValidation(
```
KennethBWSong, line 111: delete useless code?

packages/fx-core/src/plugins/resource/simpleauth/utils/common.ts
```diff
@@ -107,13 +107,24 @@ export class Utils {
...
+    // ) as string;
+    let endpoint: string;
+    if (isLocalDebug) {
+      endpoint = Utils.getConfigValueWithValidation(
+        ctx,
+        Constants.LocalDebugPlugin.id,
+        Constants.LocalDebugPlugin.configKeys.endpoint
+      ) as string;
+    } else {
+      endpoint = getArmOutput(ctx, "frontendHosting_endpoint") as string;
```
blackchoey, line 126: we need to check the feature flag first. If feature flag is disabled, we should still use original logic.

KennethBWSong, line 126: we will transfer to use arm now in provision?

packages/fx-core/src/plugins/resource/simpleauth/utils/common.ts
xzf0587: It is fixed. Thanks.

packages/fx-core/src/plugins/resource/simpleauth/utils/common.ts
xzf0587: Added the check. Thanks.

xzf0587: Added an arm env flag to control the flow. 
arm is used for inner intergaration.

blackchoey:
We do not need to provision simple auth Azure resources in `provision()` function if ARM support enabled.

packages/fx-core/src/plugins/resource/simpleauth/utils/common.ts
```diff
@@ -107,13 +107,27 @@ export class Utils {
...
-    const endpoint = Utils.getConfigValueWithValidation(
-      ctx,
-      isLocalDebug ? Constants.LocalDebugPlugin.id : Constants.FrontendPlugin.id,
-      isLocalDebug
-        ? Constants.LocalDebugPlugin.configKeys.endpoint
-        : Constants.FrontendPlugin.configKeys.endpoint
-    ) as string;
+
+    let endpoint: string;
+    if (!isArmSupportEnabled()) {
```
KennethBWSong, line 112: Seems some kind of duplicated here? Maybe better to check isLocalDebug first?

packages/fx-core/src/plugins/resource/simpleauth/plugin.ts
```diff
@@ -135,7 +135,55 @@ export class SimpleAuthPluginImpl {
 
     const configs = Utils.getWebAppConfig(ctx, false);
 
-    await this.webAppClient.configWebApp(configs);
+    if (isArmSupportEnabled()) {
+      const credentials = await ctx.azureAccountProvider!.getAccountCredentialAsync();
```
KennethBWSong, line 139: Check credential exists?

packages/fx-core/src/plugins/resource/simpleauth/utils/common.ts
xzf0587: I think using the isArmSupportEnabled first can make the logic clear since we will remove the no arm support case in the future refactor.

packages/fx-core/src/plugins/resource/simpleauth/plugin.ts
```diff
@@ -135,7 +135,55 @@ export class SimpleAuthPluginImpl {
...
+        Constants.SolutionPlugin.id,
+        Constants.SolutionPlugin.configKeys.location
+      ) as string;
+
+      const webAppName = Utils.generateResourceName(
+        ctx.projectSettings!.appName,
+        resourceNameSuffix
+      );
+      const appServicePlanName = webAppName;
+      this.webAppClient = new WebAppClient(
```
KennethBWSong, line 163: Seems duplicated as above. Can we extract them in a method?

packages/fx-core/src/plugins/resource/simpleauth/utils/common.ts
KennethBWSong: Agree. Good point.

packages/fx-core/src/plugins/resource/simpleauth/plugin.ts
```diff
@@ -135,7 +135,55 @@ export class SimpleAuthPluginImpl {
...
+        Constants.SolutionPlugin.id,
+        Constants.SolutionPlugin.configKeys.resourceGroupName
+      ) as string;
+      const location = Utils.getConfigValueWithValidation(
+        ctx,
+        Constants.SolutionPlugin.id,
+        Constants.SolutionPlugin.configKeys.location
+      ) as string;
+
+      const webAppName = Utils.generateResourceName(
```
blackchoey, line 158: get from ARM output

packages/fx-core/src/plugins/resource/simpleauth/plugin.ts
```diff
@@ -135,7 +135,55 @@ export class SimpleAuthPluginImpl {
...
+        ctx.projectSettings!.appName,
+        resourceNameSuffix
+      );
+      const appServicePlanName = webAppName;
```
blackchoey, line 162: get from ARM output

packages/fx-core/src/plugins/resource/simpleauth/plugin.ts
blackchoey: let's use `simpleAuth_webAppName` as key, we can add the missing piece in bicep template in integration

packages/fx-core/src/plugins/resource/simpleauth/plugin.ts
blackchoey: let's use `simpleAuth_appServicePlanName` as key

packages/fx-core/src/plugins/resource/simpleauth/plugin.ts
xzf0587: it is fixed. thanks

packages/fx-core/src/plugins/resource/simpleauth/plugin.ts
xzf0587: it is fixed. thanks

packages/fx-core/src/plugins/resource/simpleauth/plugin.ts
xzf0587: it is fixed. thanks

packages/fx-core/src/plugins/resource/simpleauth/plugin.ts
xzf0587: it is fixed. thanks

KennethBWSong (approved):

packages/fx-core/src/plugins/resource/simpleauth/utils/common.ts
```diff
@@ -107,13 +107,27 @@ export class Utils {
...
+      ) as string;
+    } else {
+      if (isLocalDebug) {
+        endpoint = Utils.getConfigValueWithValidation(
+          ctx,
+          Constants.LocalDebugPlugin.id,
+          Constants.LocalDebugPlugin.configKeys.endpoint
+        ) as string;
+      } else {
+        endpoint = getArmOutput(ctx, Constants.ArmOutput.frontendEndpoint) as string;
```
blackchoey, line 128: ARM deployment already set these settings, so no need to config the web app any more. We can keep this logic for now, and update related logic in next week.

blackchoey (approved):

PR #1766, opened 2021-07-29
fix: add cli app studio test

xiaolang124:
1. Add cli app studio test case.


packages/cli/tests/unit/commonlib/appStudioLogin.tests.ts
```diff
@@ -0,0 +1,45 @@
...
+  beforeEach(() => {});
+
+  it("GetAccessToken", async () => {
+    const result = await AppStudioTokenProvider.getAccessToken();
+    expect(result);
+  });
+
+  it("GetJsonObject", async () => {
+    const result = await AppStudioTokenProvider.getJsonObject();
+    expect(result);
```
swatDong, line 33: Verify the result value as well. (The same comment to all results)

xiaolang124: Fixed.

swatDong (approved):

PR #1771, opened 2021-07-29
chore: update aad provision logic to work with ARM

blackchoey:


KennethBWSong (approved):

packages/fx-core/src/common/tools.ts
```diff
@@ -529,3 +530,9 @@ export function compileHandlebarsTemplateString(templateString: string, context:
   const template = Handlebars.compile(templateString);
   return template(context);
 }
+
+export function getArmOutput(ctx: PluginContext, key: string): string | undefined {
```
blackchoey, line 534: logic copied from Zhaofeng

PR #1774, opened 2021-07-30
chore(collaborate): add error message in telemetry and log

KennethBWSong:


packages/fx-core/src/core/middleware/projectUpgrader.ts
```diff
@@ -115,12 +115,18 @@ export async function upgradeContext(ctx: CoreHookContext): Promise<void> {
     });
   } catch (error) {
     const errorObject = ContextUpgradeError(error);
+    core.tools.logProvider.info(
```
xzf0587, line 118: image
need to add ? to check undefine case.

KennethBWSong: Fixed. Thanks!

xzf0587 (approved):

SLdragon (approved):

jayzhang (approved):

PR #1776, opened 2021-07-30
chore: add arm for frontend

xzf0587:
under the isArmSupportEnabled flag:
1. update frontend config load.
2. move enable static web to post-provision.
3. disable provision stage.
4. move config setting to post-provision stage.
5. fix bug for arm support in simple auth.

### DIFF
--- a/packages/fx-core/src/plugins/resource/frontend/configs.ts
+++ b/packages/fx-core/src/plugins/resource/frontend/configs.ts
@@ -48,12 +48,7 @@ export class FrontendConfig {
       throw new UnauthenticatedError();
     }
 
-    let appName: string;
-    if (isArmSupportEnabled()) {
-      appName = getArmOutput(ctx, ArmOutput.frontendWebAppName) as string;
-    } else {
-      appName = ctx.projectSettings!.appName;
-    }
+    const appName = ctx.projectSettings!.appName;
     const solutionConfigs = ctx.configOfOtherPlugins.get(DependentPluginInfo.SolutionPluginName);
 
     const subscriptionInfo = await ctx.azureAccountProvider?.getSelectedSubscription();
--- a/packages/fx-core/src/plugins/resource/frontend/configs.ts
+++ b/packages/fx-core/src/plugins/resource/frontend/configs.ts
@@ -71,7 +71,7 @@ export class FrontendConfig {
 
     let storageName: string | undefined;
     if (isArmSupportEnabled()) {
-      storageName = getArmOutput(ctx, ArmOutput.frontendStorageName) as string;
+      storageName = getArmOutput(ctx, ArmOutput.FrontendStorageName) as string;
     } else {
       storageName = ctx.config.getString(FrontendConfigInfo.StorageName);
     }
--- a/packages/fx-core/src/plugins/resource/frontend/configs.ts
+++ b/packages/fx-core/src/plugins/resource/frontend/configs.ts
@@ -3,13 +3,20 @@
 import { PluginContext, ReadonlyPluginConfig } from "@microsoft/teamsfx-api";
 import { TokenCredentialsBase } from "@azure/ms-rest-nodeauth";
 
-import { Constants, DependentPluginInfo, FrontendConfigInfo, RegularExpr } from "./constants";
+import {
+  ArmOutput,
+  Constants,
+  DependentPluginInfo,
+  FrontendConfigInfo,
+  RegularExpr,
+} from "./constants";
 import {
   InvalidConfigError,
   InvalidStorageNameError,
   UnauthenticatedError,
 } from "./resources/errors";
 import { Utils } from "./utils";
+import { getArmOutput, isArmSupportEnabled } from "../../..";
 
 export class FrontendConfig {
   subscriptionId: string;
@@ -41,7 +48,12 @@ export class FrontendConfig {
       throw new UnauthenticatedError();
     }
 
-    const appName = ctx.projectSettings!.appName;
+    let appName: string;
+    if (isArmSupportEnabled()) {
+      appName = getArmOutput(ctx, ArmOutput.frontendWebAppName) as string;
+    } else {
+      appName = ctx.projectSettings!.appName;
+    }
     const solutionConfigs = ctx.configOfOtherPlugins.get(DependentPluginInfo.SolutionPluginName);
 
     const subscriptionInfo = await ctx.azureAccountProvider?.getSelectedSubscription();
@@ -62,7 +74,12 @@ export class FrontendConfig {
       solutionConfigs
     );
 
-    let storageName = ctx.config.getString(FrontendConfigInfo.StorageName);
+    let storageName: string | undefined;
+    if (isArmSupportEnabled()) {
+      storageName = getArmOutput(ctx, ArmOutput.frontendStorageName) as string;
+    } else {
+      storageName = ctx.config.getString(FrontendConfigInfo.StorageName);
+    }
     if (!storageName) {
       storageName = Utils.generateStorageAccountName(
         appName,
@@ -70,6 +87,7 @@ export class FrontendConfig {
         Constants.FrontendSuffix
       );
     }
+
     if (!RegularExpr.FrontendStorageNamePattern.test(storageName)) {
       throw new InvalidStorageNameError();
     }
--- a/packages/fx-core/src/plugins/resource/frontend/constants.ts
+++ b/packages/fx-core/src/plugins/resource/frontend/constants.ts
@@ -102,7 +102,6 @@ export class ArmOutput {
   static readonly frontendEndpoint = "frontendHosting_endpoint";
   static readonly frontendDomain = "frontendHosting_domain";
   static readonly frontendStorageName = "frontendHosting_storageName";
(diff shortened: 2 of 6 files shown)


packages/fx-core/src/plugins/resource/frontend/configs.ts
```diff
@@ -41,7 +48,12 @@ export class FrontendConfig {
       throw new UnauthenticatedError();
     }
 
-    const appName = ctx.projectSettings!.appName;
+    let appName: string;
+    if (isArmSupportEnabled()) {
+      appName = getArmOutput(ctx, ArmOutput.frontendWebAppName) as string;
```
blackchoey, line 53: according to the original code, the app name should be the Teams App's name, not the frontend hosting's resource name, we do not need to do extra things here.

packages/fx-core/src/plugins/resource/frontend/plugin.ts
```diff
@@ -173,12 +174,22 @@ export class FrontendPluginImpl {
       };
     }
 
-    const authPlugin = ctx.configOfOtherPlugins.get(DependentPluginInfo.RuntimePluginName);
-    if (authPlugin) {
-      runtimeEnv = {
-        endpoint: authPlugin.get(DependentPluginInfo.RuntimeEndpoint) as string,
-        startLoginPageUrl: DependentPluginInfo.StartLoginPageURL,
-      };
+    if (isArmSupportEnabled()) {
```
blackchoey, line 177: we need to enable static web app feature in post provision, when arm support enabled

packages/fx-core/src/plugins/resource/frontend/plugin.ts
```diff
@@ -197,6 +208,19 @@ export class FrontendPluginImpl {
       );
     }
 
+    if (isArmSupportEnabled()) {
+      const config = await FrontendConfig.fromPluginContext(ctx);
+      config.endpoint = getArmOutput(ctx, ArmOutput.frontendEndpoint) as string;
```
blackchoey, line 213: may also need to set the storage account name?

packages/fx-core/src/plugins/resource/frontend/configs.ts
```diff
@@ -62,14 +74,20 @@ export class FrontendConfig {
       solutionConfigs
     );
 
-    let storageName = ctx.config.getString(FrontendConfigInfo.StorageName);
+    let storageName: string | undefined;
+    if (isArmSupportEnabled()) {
+      storageName = getArmOutput(ctx, ArmOutput.frontendStorageName) as string;
```
hund030, line 79: Seems there are two places store the frontend Storage info, Arm output and config. When we shift to ARM, do we still need to write these info to configs? e.g. StorageName, StorageEndpoint...

packages/fx-core/src/plugins/resource/frontend/constants.ts
```diff
@@ -97,6 +97,14 @@ export class DependentPluginInfo {
   static readonly LocalTabEndpoint = "localTabEndpoint";
 }
 
+export class ArmOutput {
+  static readonly simpleAuthEndpoint = "simpleAuth_endpoint";
```
hund030, line 101: We use PascalCase for constants in frontend plugin.

packages/fx-core/src/plugins/resource/frontend/plugin.ts
xzf0587: storage name is set in FrontendConfig.fromPluginContext.

packages/fx-core/src/plugins/resource/frontend/index.ts
```diff
@@ -55,10 +57,14 @@ export class FrontendPlugin implements Plugin {
   }
 
   public async provision(ctx: PluginContext): Promise<TeamsFxResult> {
-    FrontendPlugin.setContext(ctx);
-    return this.runWithErrorHandling(ctx, TelemetryEvent.Provision, () =>
-      this.frontendPluginImpl.provision(ctx)
-    );
+    if (isArmSupportEnabled()) {
```
hund030, line 60: What if `isArmSupportEnabled` throws error? Do we have readable error message? Do we send telemetry?

packages/fx-core/src/plugins/resource/frontend/plugin.ts
hund030, line 177: I assume after we support ARM provision for function, there will be a snippet like:
```
if (isArmSupportEnabled()) {
   // get function's output
}
```
Then I think it would be better to have an `originPostProvision` and a `postProvisioin` for ARM. And put the condition to index.ts. So that we can simply remove the `originPostProvision` when we decide to totally shift to ARM.

packages/fx-core/src/plugins/resource/frontend/constants.ts
```diff
@@ -97,6 +97,14 @@ export class DependentPluginInfo {
...
+  static readonly frontendEndpoint = "frontendHosting_endpoint";
+  static readonly frontendDomain = "frontendHosting_domain";
+  static readonly frontendStorageName = "frontendHosting_storageName";
+  static readonly frontendWebAppName = "frontendHosting_webAppName";
```
hund030, line 105: What's this? I think current frontend hosting solution doesn't introduce web app? Is it for simple auth?

packages/fx-core/src/plugins/resource/frontend/configs.ts
xzf0587: it is fixed. thanks.

packages/fx-core/src/plugins/resource/frontend/configs.ts
xzf0587: if we shift to ARM, info about azure resource can be loaded from arm output.
But some info is out of arm scope, we may use config as well.
@blackchoey  may help add more info about this.

packages/fx-core/src/plugins/resource/frontend/constants.ts
xzf0587: this key is defined in arm output. I follow it.
@blackchoey  in the future, can we provide a whole arm output key names in common?

packages/fx-core/src/plugins/resource/frontend/index.ts
xzf0587: isArmSupportEnabled  is an env flag. 
Actually, we just skip provision step, and there will be no error.
If we use arm, we will not do provision in plugin.

packages/fx-core/src/plugins/resource/frontend/constants.ts
xzf0587: the frontendHosting_webAppName is removed. Thanks.

packages/fx-core/src/plugins/resource/frontend/index.ts
hund030: I think the provision step in frontend plugin will be removed eventually. The env flag is temporary solution. Is it right?

packages/fx-core/src/plugins/resource/frontend/plugin.ts
xzf0587: It is a good suggestion. 
What we did now is to make a  prototype for arm support. It is a quick change.
After the prototype check, we can fork the arm support and origin flow.

packages/fx-core/src/plugins/resource/frontend/index.ts
xzf0587: Yes.

packages/fx-core/src/plugins/resource/frontend/configs.ts
blackchoey: yes, all Azure related information can be retrieved from ARM output. For other information, config is still needed.

packages/fx-core/src/plugins/resource/frontend/constants.ts
blackchoey: I'm not sure whether @hund030 means the variable name `simpleAuthEndpoint` or the value `simpleAuth_endpoint`. If it's the value, do you think it's appropriate for us update the name in next week? So we can reduce some integration work.

hund030: I mean the variable name `simpleAuthEndpoint`.

blackchoey (approved):

packages/fx-core/src/plugins/resource/frontend/plugin.ts
hund030: Ok for me.

packages/fx-core/src/plugins/resource/frontend/constants.ts
xzf0587: OK. I fixed the names. Thanks.

packages/fx-core/src/plugins/resource/frontend/configs.ts
hund030: Then if we totally shift to ARM, we don't need to write Azure related information to config any more. Just to confirm.

hund030 (approved):

KennethBWSong (approved):

PR #1779, opened 2021-07-30
feat(collaboration): add api

KennethBWSong:


packages/fx-core/src/plugins/solution/fx-solution/solution.ts
```diff
@@ -1104,6 +1104,7 @@ export class TeamsAppSolution implements Solution {
 
   @hooks([ErrorHandlerMW])
   async grantPermission(ctx: SolutionContext): Promise<Result<any, FxError>> {
+    console.log(1);
```
SLdragon, line 1107: This can be removed

SLdragon (approved):

PR #1781, opened 2021-07-30
perf(arm support): add progress bar when deploying arm templates

dilin-MS:


packages/fx-core/src/plugins/solution/fx-solution/constants.ts
```diff
@@ -168,3 +169,17 @@ export enum SolutionTelemetrySuccess {
 }
 
 export const SolutionTelemetryComponentName = "solution";
+
+export class Messages {
+  // Progress bar messages
+  static readonly DeployArmTemplatesProgressTitle = "Deploying Arm Templates";
```
jayzhang, line 175: Suggest: it's better to place constant strings in `fx-core/resource/strings.json` file

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -143,7 +150,7 @@ export async function deployArmTemplates(ctx: SolutionContext): Promise<Result<v
   );
   const armTemplateJsonFilePath = path.join(azureInfraDir, templateFolder, armTemplateJsonFileName);
   await compileBicepToJson(orchestrationFilePath, armTemplateJsonFilePath);
-  ctx.logProvider?.info("[Solution] Successfully compile bicep files to JSON.");
+  ctx.logProvider?.info("Successfully compile bicep files to JSON.");
```
blackchoey, line 153: Why we removed the prefix?

packages/fx-core/src/plugins/solution/fx-solution/utils/progressHelper.ts
```diff
@@ -0,0 +1,28 @@
+import { IProgressHandler, PluginContext } from "../../../../../../api/build";
```
blackchoey, line 1: maybe we need to reference from teamsfx-api package?

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
dilin-MS: Only the start and the end message will prefix "solution". Or do we have a consistent practice? @1yefuwang1  @jayzhang

packages/fx-core/src/plugins/solution/fx-solution/utils/progressHelper.ts
dilin-MS: Fixed.

packages/fx-core/src/plugins/solution/fx-solution/constants.ts
dilin-MS: Moved.

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
dilin-MS: Will add the prefix.

jayzhang (approved):

PR #1784, opened 2021-07-30
chore: template deps versions sync up with sdk

wenytang-ms:


templates/scripts/sync-up-version.js
```diff
@@ -0,0 +1,57 @@
+const path = require("path");
+const writePkg = require("write-pkg");
+const semver = require("semver");
+const fse = require("fs-extra");
+
+const sdkVersion = require(path.join(__dirname, "../../packages/sdk/package.json")).version;
+const sdkName = require(path.join(__dirname, "../../packages/sdk/package.json")).name;
+
+function listFile(dir, list = []) {
```
dilin-MS, line 9: Name like `recursivelyListPackageJsonFilePath()` may be more clear.

templates/scripts/sync-up-version.js
```diff
@@ -0,0 +1,57 @@
...
+    change = true;
+    pkg_.dependencies = dep;
+    writePkg(file, pkg_);
+  }
+}
+
+let needBumpUp = process.argv[2] === "yes" ? true : false;
+if (change && needBumpUp) {
+  let file = path.join(template_dir, "package.json");
+  let pkg_ = loadJsonFile.sync(file);
```
hund030, line 47: maybe `fse.readJsonSync(file)`. By the way, usually we use `fs` as alias no matter we actually require "fs" or "fs-extra" or others.

packages/sdk/tools/sync-up-version.js
```diff
@@ -0,0 +1,67 @@
...
+}
+
+// only alpha and stable release bump up version
+let needBumpUp = process.argv[2] === "yes" ? true : false;
+if (change && needBumpUp) {
+    let file = path.join(template_dir, "package.json");
+    let pkg_ = fse.readJsonSync(file);
+    let ver = pkg_.version;
+    if(semver.prerelease(sdkVersion)) {
+        ver = semver.inc(ver, "prerelease", "alpha");
```
henzhang-ms, line 51: can we pass prerelease from arg instead of hard coded?

wenytang-ms: this sync-up-version run in npm lifecycle, not lerna scripts, so it's hard to get current release type.
Are there any concerns?

packages/sdk/tools/sync-up-version.js
```diff
@@ -0,0 +1,67 @@
...
+    } else {
+        ver = semver.inc(ver, "patch");
+    }
+
+    pkg_.version = ver;
+    writePkg(file, pkg_);
+
+    file = path.join(template_dir, "package-lock.json");
```
henzhang-ms, line 59: Does `template_dir` have such dependency? Not quite understand the reason.

henzhang-ms (approved):
looks good to me on the entire logic and js script. I am not familiar with the lerna so has no review on that part.

packages/sdk/tools/sync-up-version.js
henzhang-ms: No, just feel not good about one ad-hoc naming customizes multiple scripts. But if we have no better choice, just keep it.

packages/sdk/tools/sync-up-version.js
wenytang-ms: yes, since templates has it's own package.json and devdeps, so when we do npm run setup, there will be an lock file, and if the lock file not sync up, there may be uncommits stash when npm run setup again

packages/sdk/tools/sync-up-version.js
wenytang-ms: in the npm and lerna lifecycle, semver handle the release type and version
alpha version just a temp file, cause it will not check in, just as a mark.

tecton (approved):
Looks good to me. lerna bumps version at preversion step and sync version script put that in templates.

.github/workflows/cd.yml
```diff
@@ -104,6 +104,12 @@ jobs:
           then
               git push -d origin $(git tag --points-at HEAD | grep templates | grep rc)
           fi
+            
+      - name: sync up templates deps with sdk version for RC
+        if: ${{ contains(steps.version-change.outputs.CHANGED, '@microsoft/teamsfx') && !contains(steps.version-change.outputs.CHANGED, 'templates@') && github.event_name == 'workflow_dispatch' && github.ref == 'refs/heads/main' && github.event.inputs.preid == 'rc' }}
+        run: |
+          node ./tools/sync-up-version.js
+        working-directory: ./packages/sdk
```
adashen, line 112: The script should not be put under sdk folder.

wenytang-ms: move to .github/scripts folder

LongOddCode (approved):

PR #1786, opened 2021-07-30
chore(sql): register resource provider

xzf0587:
1. register sql resource provider.
2. refactor sql management client creating and sql query client creating.


packages/fx-core/src/plugins/resource/sql/utils/dialogUtils.ts
```diff
@@ -16,12 +16,13 @@ export class DialogUtils {
 
 export class ProgressTitle {
   static readonly Provision = "Provisioning SQL";
-  static readonly ProvisionSteps = 2;
+  static readonly ProvisionSteps = 3;
```
KennethBWSong, line 19: Better to get the number of steps according to the messages so that we don't need to update the number every time we add/remove a step.

blackchoey (commented):
Looks good to me, please resolve other comments before merge

blackchoey (approved):

packages/fx-core/src/plugins/resource/sql/utils/dialogUtils.ts
xzf0587: It is fixed. Thanks.

KennethBWSong (approved):

PR #1789, opened 2021-07-30
feat(collaboration): add grant permission

KennethBWSong:


packages/fx-core/src/plugins/solution/fx-solution/solution.ts
```diff
@@ -1898,4 +1959,47 @@ export class TeamsAppSolution implements Solution {
...
+    }
+    const tenantId = currentUser["tid"] as string;
+
+    if (email) {
+      const graphToken = await ctx.graphTokenProvider?.getAccessToken();
+      const instance = axios.create({
+        baseURL: "https://graph.microsoft.com/v1.0",
+      });
+      instance.defaults.headers.common["Authorization"] = `Bearer ${graphToken}`;
+      const res = await instance.get(`/users`);
```
SLdragon, line 1976: This request may only return first 100 users, you need to user filter to get user

KennethBWSong: Fixed. Thx!

packages/fx-core/src/plugins/solution/fx-solution/question.ts
```diff
@@ -224,3 +225,9 @@ export const ProgrammingLanguageQuestion: SingleSelectQuestion = {
     return "Select a programming language.";
   },
 };
+
+export const GrantPermissionQuestion: TextInputQuestion = {
```
SLdragon, line 229: We should also consider to ask user to select environment

KennethBWSong: Fixed. Thanks.

SLdragon (approved):

PR #1792, opened 2021-07-30
chore: support provision tab app resources using arm

blackchoey:
With this PR, we can provision tab app resources using ARM when related feature flag is turned on


packages/fx-core/src/common/tools.ts
```diff
@@ -533,6 +533,6 @@ export function compileHandlebarsTemplateString(templateString: string, context:
 
 export function getArmOutput(ctx: PluginContext, key: string): string | undefined {
   const solutionConfig = ctx.configOfOtherPlugins.get("solution");
-  const output = solutionConfig?.get("armTemplate");
+  const output = solutionConfig?.get("armTemplateOutput");
```
dilin-MS, line 536: Please use the constant value `ARM_TEMPLATE_OUTPUT` in *packages\fx-core\src\plugins\solution\fx-solution\constants.ts*.

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -122,19 +122,30 @@ export async function deployArmTemplates(ctx: SolutionContext): Promise<Result<v
...
-  const parameterJson = JSON.parse(expandParameterPlaceholders(ctx, parameterTemplate));
   const parameterDefaultFilePath = path.join(
     azureInfraDir,
     parameterFolder,
     parameterDefaultFileName
   );
-  await fs.writeFile(parameterDefaultFilePath, parameterJson);
+  let parameterTemplate, parameterJsonString;
+  try {
+    await fs.stat(parameterDefaultFilePath);
```
dilin-MS, line 135: What about check whether the `parameterDefaultFilePath` exists first. If yes, read the file. Else, read the parameterTemplateFile.

dilin-MS: What if fail to read parameter from these two files?

packages/fx-core/src/plugins/resource/frontend/configs.ts
```diff
@@ -72,6 +72,9 @@ export class FrontendConfig {
     let storageName: string | undefined;
     if (isArmSupportEnabled()) {
       storageName = getArmOutput(ctx, ArmOutput.FrontendStorageName) as string;
+      if (!storageName) {
```
hund030, line 75: The logic is quiet complex. Ok for now but we better to improve it.

hund030 (approved):

packages/fx-core/src/plugins/resource/frontend/configs.ts
blackchoey: yes, I think we need to do some refractor to split the provision and deploy related logic here. Let's do this later.

1yefuwang1 (approved):

jayzhang (approved):

packages/fx-core/src/plugins/resource/frontend/configs.ts
dilin-MS: Suggest add "todo" comment to avoid being forgotten.

KennethBWSong (approved):

PR #1794, opened 2021-07-30
fix(arm support): fix deploy arm template polling status

dilin-MS:


blackchoey (commented):
Please also target this PR to dev branch

PR #1798, opened 2021-08-02
refactor(core): extract the logic to write env profiles

qinezh:
This is the following PR after https://github.com/OfficeDev/TeamsFx/pull/1752: Extract the logic to persist the env profile from `configWriter.ts` to `environment.ts`, and add a few test cases for the new added API.

### DIFF
--- a/packages/fx-core/src/core/environment.ts
+++ b/packages/fx-core/src/core/environment.ts
@@ -6,17 +6,27 @@ import {
   CryptoProvider,
   err,
   FxError,
-  Json,
   ok,
   Result,
+  Void,
 } from "@microsoft/teamsfx-api";
 import path from "path";
 import fs from "fs-extra";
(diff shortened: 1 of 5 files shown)


dooriya (approved):

1yefuwang1 (approved):

packages/fx-core/src/core/environment.ts
```diff
@@ -6,17 +6,27 @@ import {
...
+  sperateSecretData,
+  WriteFileError,
+  mapToJson,
+  objectToMap,
+} from "..";
 
 export interface EnvInfo {
   envName: string;
-  data: Json;
+  data: Map<string, any>;
```
jayzhang, line 29: why to Map to replace Json, Map is somehow typeless.

jayzhang (approved):

PR #1801, opened 2021-08-02
chore(identity): add resource provider check

xzf0587:
add resource provider check for identity

### DIFF
--- a/packages/fx-core/src/plugins/resource/identity/index.ts
+++ b/packages/fx-core/src/plugins/resource/identity/index.ts
@@ -22,6 +22,8 @@ import { getTemplatesFolder } from "../../..";
 import { AzureResourceSQL } from "../../solution/fx-solution/question";
 import { Service } from "typedi";
 import { ResourcePlugins } from "../../solution/fx-solution/ResourcePluginContainer";
+import { Providers, ResourceManagementClientContext } from "@azure/arm-resources";
+
 @Service(ResourcePlugins.IdentityPlugin)
 export class IdentityPlugin implements Plugin {
   name = "fx-resource-identity";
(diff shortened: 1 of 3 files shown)


packages/fx-core/src/plugins/resource/identity/index.ts
```diff
@@ -65,6 +67,17 @@ export class IdentityPlugin implements Plugin {
     );
     this.config.location = ContextUtils.getConfigString(Constants.solution, Constants.location);
 
+    try {
+      ctx.logProvider?.info(Message.checkProvider);
```
KennethBWSong, line 71: Add progress bar?

KennethBWSong (approved):

packages/fx-core/src/plugins/resource/identity/utils/messages.ts
```diff
@@ -5,7 +5,11 @@ export class Message {
   public static readonly endProvision = `[${Constants.pluginName}] end provision`;
   public static readonly provisionIdentity = `[${Constants.pluginName}] provision identity`;
   public static readonly getIdentityId = `[${Constants.pluginName}] get identity id`;
+  public static readonly checkProvider = `[${Constants.pluginName}] check SQL resource provider`;
```
blackchoey, line 8: check identity resource provide?

packages/fx-core/src/plugins/resource/identity/index.ts
```diff
@@ -65,6 +67,17 @@ export class IdentityPlugin implements Plugin {
...
+      const credentials = await ctx.azureAccountProvider!.getAccountCredentialAsync();
+      const resourceManagementClient = new Providers(
+        new ResourceManagementClientContext(credentials!, this.config.azureSubscriptionId)
+      );
+      await resourceManagementClient.register(Constants.resourceProvider);
```
blackchoey, line 76: Just realized we will try to register resource provider every time? If user does not have permission to register resource provider, but the resource provider is already registered, will this operation fail?

packages/fx-core/src/plugins/resource/identity/utils/messages.ts
xzf0587: it is fixed. Thanks.

packages/fx-core/src/plugins/resource/identity/index.ts
xzf0587: We try to register the resource every time. If the register failed, we continue to provide the resource. So the permission to register resource will not interrupt the provision.

PR #1807, opened 2021-08-02
chore(sdk): include doc update in build and remain error class name

tecton:
1. Error class name was minified in terser plugin.
2. Doc updates need to be included in PR, otherwise CI will fail.


wenytang-ms:
seems there has uncommits files after npm run setup

jiayi-ruan (approved):

wenytang-ms:
still has uncommit file

tecton:
> still has uncommit file
> or check in all the md files

The generated files are ended with CRLF but committed ones are LF, it causes the difference in `git status` and we have to generate markdown files with LF.

PR #1808, opened 2021-08-02
perf(arm support): fix end deploy arm templates progress bar

dilin-MS:
1. Make fx-solution log consistent: string start with `[Teams Toolkit]`, which is the displayName of solution plugin.
2. Fix "polling arm templates deployment progress bar never ends when meet error"
3. Move solution related constant string to *packages\fx-core\resource\strings.json* as it is the recommended practice from @jayzhang


packages/fx-core/resource/strings.json
```diff
@@ -19,13 +19,31 @@
...
     "GenerateArmTemplateFailNotice": "'%s' Failed to create ARM template for required Azure services.",
-    "StartDeployArmTemplateNotice": "[%s] Start Deploying Arm templates.",
-    "SucessfullyCompileBicepNotice": "[%s] Successfully compile bicep files to JSON.",
-    "SucessfullyDeployArmTemplateNotice": "[%s] Successfully deploy Arm templates. Resource group name: %s. Deployment name: %s",
-    "FailedToDeployArmTemplateNotice": "[%s] Failed to deploy Arm templates. Resource group name: %s. Deployment name: %s",
-    "PollDeploymentStatusNotice": "[%s] Polling deployment status...",
-    "EndDeployArmTemplateNotice": "[%s] Successfully deployed Arm templates."
+
+    "DeployArmTemplates": {
+      "StartNotice": "[%s] Start deploying Arm templates.",
```
blackchoey, line 26: use `ARM` instead of `Arm` ?

packages/fx-core/resource/strings.json
```diff
@@ -19,13 +19,31 @@
...
+      "CompileBicepSuccessNotice": "[%s] Successfully compile bicep files to JSON.",
+      "SuccessNotice": "[%s] Successfully deploy Arm templates. Resource group name: %s. Deployment name: %s",
+      "FailNotice": "[%s] Failed to deploy Arm templates. Resource group name: %s. Deployment name: %s",
+      "PollDeploymentStatusNotice": "[%s] Polling deployment status...",
+      "EndNotice": "[%s] End Arm templates deployment.",
```
blackchoey, line 31: What's the different between "end" and “success"/"fail"?

packages/fx-core/resource/strings.json
```diff
@@ -19,13 +19,31 @@
...
     "NotProvisionedNotice": "'%s' is not provisioned in the cloud yet. Please provision first.",
+
     "GenerateArmTemplateFailNotice": "'%s' Failed to create ARM template for required Azure services.",
-    "StartDeployArmTemplateNotice": "[%s] Start Deploying Arm templates.",
-    "SucessfullyCompileBicepNotice": "[%s] Successfully compile bicep files to JSON.",
-    "SucessfullyDeployArmTemplateNotice": "[%s] Successfully deploy Arm templates. Resource group name: %s. Deployment name: %s",
-    "FailedToDeployArmTemplateNotice": "[%s] Failed to deploy Arm templates. Resource group name: %s. Deployment name: %s",
-    "PollDeploymentStatusNotice": "[%s] Polling deployment status...",
-    "EndDeployArmTemplateNotice": "[%s] Successfully deployed Arm templates."
+
```
blackchoey, line 24: cc @MuyangAmigo to help review the wording here. @MuyangAmigo you may focus on this file only.

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -113,75 +111,39 @@ export async function generateArmTemplate(ctx: SolutionContext): Promise<Result<
...
+  const parameterJson = await getParameterJson(ctx);
 
   const resourceGroupName = ctx.config.get(GLOBAL_CONFIG)?.getString(RESOURCE_GROUP_NAME);
   if (!resourceGroupName) {
-    throw returnSystemError(
-      new Error("Failed to get resource group from project solution settings."),
-      PluginNames.SOLUTION,
-      SolutionError.NoResourceGroupFound
-    );
+    throw new Error("Failed to get resource group from project solution settings.");
```
jayzhang, line 127: Why not throw FxError?

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -299,13 +312,10 @@ async function compileBicepToJson(
   const command = `bicep build ${bicepOrchestrationFilePath} --outfile ${jsonFilePath}`;
   const { stdout, stderr } = await execAsync(command);
   if (stderr) {
-    throw returnSystemError(
-      new Error(`Failed to compile bicep files to Json arm templates file: ${stderr}`),
-      PluginNames.SOLUTION,
-      SolutionError.FailedToCompileBicepFiles
-    );
+    throw new Error(`Failed to compile bicep files to Json arm templates file: ${stderr}`);
```
jayzhang, line 315: Why not throw FxError?

jayzhang (approved):

packages/fx-core/resource/strings.json
dilin-MS: Fixed.

packages/fx-core/resource/strings.json
dilin-MS: No matter success or fail, we will hit "end" event.
- Start
- - Success / fail
- End

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
dilin-MS: Since all error thrown from `doDeployArmDeployment()` method is of the same type `SolutionError.FailedToDeployArmTemplatesToAzure` (`FxError`), in `DeployArmDeployment()` we will wrap all native Error thrown from `doDeployArmDeployment()`.

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
dilin-MS: Same as above.

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -267,14 +219,75 @@ export async function deployArmTemplates(ctx: SolutionContext): Promise<Result<v
...
+async function getParameterJson(ctx: SolutionContext) {
+  const parameterDir = path.join(ctx.root, baseFolder, parameterFolder);
+  const parameterDefaultFilePath = path.join(parameterDir, parameterDefaultFileName);
+  let parameterJsonString;
+  let parameterFilePath = parameterDefaultFilePath;
+  try {
+    parameterJsonString = await getExpandedParameter(ctx, parameterFilePath);
+  } catch {
+    const parameterTemplateFilePath = path.join(parameterDir, parameterTemplateFileName);
+    ctx.logProvider?.warning(
```
blackchoey, line 255: we do not expect the parameter.default.json file always exists, so treat this message as information. By the way, if file already exists and some unexpected error happened (like handlebars error), we should not try to use the template file to generate a new default file.

The integration code is rough, let's polish the details.

packages/fx-core/src/plugins/solution/fx-solution/utils/progressHelper.ts
```diff
@@ -1,10 +1,9 @@
 import { IProgressHandler, PluginContext } from "@microsoft/teamsfx-api";
-import { Messages } from "../constants";
-
-export const DeployArmTemplatesSteps = {
-  DeployArmTemplates: Messages.ProgressDeployArmTemplates,
-};
+import { getStrings } from "../../../../common";
 
+export enum DeployArmTemplatesSteps {
+  ExecuteDeployment = "Deploying solution Arm templates to Azure. This could take several minutes.",
```
blackchoey, line 5: ARM template

packages/fx-core/resource/strings.json
blackchoey: Since success/fail already expresses the progress has ended, maybe we no longer need to add an end log unless it brings extra information.

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
dilin-MS: Fixed.

packages/fx-core/src/plugins/solution/fx-solution/utils/progressHelper.ts
dilin-MS: This is one thing I want to make consistent. From [Docs](https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/overview), `ARM templates` and `ARM template` are both valid proper noun.

dilin-MS: I prefer to use `ARM template` since template files are usually multiple.

blackchoey: The key point is `Arm` -> `ARM`... I don't have any preference on the plurality. Sorry for the confusing...
For whether use plurality, I saw you're using `ARM templates` in other places, may be you can just align with other strings if there's no special reason to change it. Of curse you can change it to better align with current context.

packages/fx-core/resource/strings.json
dilin-MS: Removed end notice.

blackchoey (approved):

PR #1812, opened 2021-08-02
chore(api): api v2

1yefuwang1:
This PR contains the next version of plugin API.
Highlights include: 1. ARM support 2. more meaningful type signature


packages/api/src/v2/resourcePlugin.ts
```diff
@@ -0,0 +1,235 @@
...
+    tokenProvider: TokenProvider
+  ) => Promise<Result<LocalProvisionOutput, FxError>>;
+
+  getQuestionsForLifecycleTask?: (
+    ctx: Context,
+    stage: Stage,
+    inputs: Inputs
+  ) => Promise<Result<QTreeNode | undefined, FxError>>;
+
+  getQuestionsForUserTask?: (
```
jayzhang, line 227: getQuestionsForLifecycleTask: change to getQuestionForScaffold()
getQuestionsForUserTask: deleted
Just as we discussed, only scaffold questions can be extensible.

packages/api/src/v2/types.ts
```diff
@@ -0,0 +1,73 @@
...
+  };
+  solutionSetting: SolutionSetting;
+}
+
+export interface SolutionSetting extends Json {
+  resourcePlugins: string[];
+}
+
+export interface Inputs extends Json {
+  stage: Stage;
```
jayzhang, line 44: For Core's API, the only input is:
`createProject: (systemInputs: Inputs) => Promise<Result<string, FxError>>;`
Suggest projectPath and platform move to Inputs

jayzhang (approved):

packages/api/src/v2/types.ts
1yefuwang1: Makes sense. Done

PR #1815, opened 2021-08-03
feat(multi-env): add localSettings in context and implement provider for init/load/save

dooriya:
**Changes in this PR**:
1. Add `localSettings` API contract in solution & plugin context.
2. Define interface `LocalSettings` for the schema of `localSettings.json` file.
3. Provide a `LocalSetingsProvider` to init/local/save `localSettings.json` file.

**Changes in upcoming PRs**:
1. Update resource plugins to read/write local debug related configs from `localSettings.json` file.
2. Update local debug component to generated env variables from `localSettings.json` file.
3. Update VS Code extension & CLI to integrate with the new `localSettings.json` file. 
4. Add feature flag to enable/disable using the new local debug settings.
5. Clean up legacy code for `local.env`.
6. ...

### DIFF
--- a/packages/fx-core/src/common/localSettingsProvider.ts
+++ b/packages/fx-core/src/common/localSettingsProvider.ts
@@ -1,0 +1,111 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+"use strict";
+
+import * as fs from "fs-extra";
+import * as os from "os";
+import { ConfigFolderName, ConfigMap, LocalSettings } from "@microsoft/teamsfx-api";
+import {
+  LocalSettingsAuthKeys,
+  LocalSettingsBackendKeys,
+  LocalSettingsBotKeys,
+  LocalSettingsFrontendKeys,
+  LocalSettingsTeamsAppKeys,
+} from "./localSettingsConstants";
+
+export const localSettingsFileName = "localSettings.json";
+
+export class LocalSettingsProvider {
+  public readonly localSettingsFilePath: string;
+  constructor(workspaceFolder: string) {
+    this.localSettingsFilePath = `${workspaceFolder}/.${ConfigFolderName}/${localSettingsFileName}`;
+  }
+
+  public init(
+    includeFrontend: boolean,
+    includeBackend: boolean,
+    includeBot: boolean
+  ): LocalSettings {
+    // initialize Teams app related config for local debug.
+    const teamsAppLocalConfig = new ConfigMap();
+    teamsAppLocalConfig.set(LocalSettingsTeamsAppKeys.TenantId, "");
+    teamsAppLocalConfig.set(LocalSettingsTeamsAppKeys.TeamsAppId, "");
+
+    const localSettings: LocalSettings = {
+      teamsApp: teamsAppLocalConfig,
+    };
+
+    let keys: string[];
+
+    // initialize frontend and simplee auth local settings.
+    if (includeFrontend) {
+      const frontendLocalConfig = new ConfigMap();
+      frontendLocalConfig.set(LocalSettingsFrontendKeys.Browser, "none");
+      frontendLocalConfig.set(LocalSettingsFrontendKeys.Https, true);
+      frontendLocalConfig.set(LocalSettingsFrontendKeys.TrustDevCert, true);
+      frontendLocalConfig.set(LocalSettingsFrontendKeys.SslCertFile, "");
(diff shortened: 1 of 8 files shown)


packages/fx-core/src/common/localSettingsProvider.ts
```diff
@@ -0,0 +1,111 @@
...
+    teamsAppLocalConfig.set(LocalSettingsTeamsAppKeys.TenantId, "");
+    teamsAppLocalConfig.set(LocalSettingsTeamsAppKeys.TeamsAppId, "");
+
+    const localSettings: LocalSettings = {
+      teamsApp: teamsAppLocalConfig,
+    };
+
+    let keys: string[];
+
+    // initialize frontend and simplee auth local settings.
```
swatDong, line 40: `simple`

packages/fx-core/src/plugins/solution/fx-solution/solution.ts
```diff
@@ -1106,6 +1126,9 @@ export class TeamsAppSolution implements Solution {
       ctx.config.get(GLOBAL_CONFIG)?.set(LOCAL_DEBUG_TEAMS_APP_ID, maybeTeamsAppId.value);
     }
 
+    // persistent localSettings.json after local debug.
+    localSettingsProvider.save(ctx.localSettings!);
```
swatDong, line 1130: Now *localSettings.json* becomes the only single source of truth of all local configurations, so it's not enough to only save it after local debug. If *localDebug* step breaks by any error, local configurations are lost.

Make sure *localSettings* is persisted after both *localdebug* and *postlocaldebug* steps, no matter the step itself is succeeded or failed.

packages/fx-core/src/common/localSettingsProvider.ts
dooriya: Good catch.

packages/fx-core/src/plugins/solution/fx-solution/solution.ts
dooriya: Thanks Qianhao. You also remind me of **not to refresh** the previous local debug settings provided by resource plugins when add new resource / capability. I will handle these parts later.

swatDong (approved):

packages/fx-core/src/plugins/solution/fx-solution/solution.ts
```diff
@@ -374,8 +376,22 @@ export class TeamsAppSolution implements Solution {
           await fs.copy(readme, `${ctx.root}/README.md`);
         }
       }
-    } else {
-      return res;
+
+      const azureResources = (ctx.projectSettings?.solutionSettings as AzureSolutionSettings)
+        ?.azureResources;
+      const hasBackend = azureResources?.includes(AzureResourceFunction.id);
```
qinezh, line 382: do we need to check if `azureResources` is null here?

packages/fx-core/src/plugins/solution/fx-solution/solution.ts
```diff
@@ -1062,12 +1078,35 @@ export class TeamsAppSolution implements Solution {
...
+    const hasFrontend = selectedPlugins?.some((plugin) => plugin.name === PluginNames.FE);
+    const hasBackend = selectedPlugins?.some((plugin) => plugin.name === PluginNames.FUNC);
+    const hasBot = selectedPlugins?.some((plugin) => plugin.name === PluginNames.BOT);
+
+    // load localSettings into context before local debug.
+    const localSettingsProvider = new LocalSettingsProvider(ctx.root);
+    if (await fs.pathExists(localSettingsProvider.localSettingsFilePath)) {
+      ctx.localSettings = await localSettingsProvider.load();
+    } else {
+      ctx.localSettings = await localSettingsProvider.init(hasFrontend, hasBackend, hasBot);
```
qinezh, line 1098: no need `await` here, since `localSettingsProvider.init` is a sync method?

packages/fx-core/src/plugins/solution/fx-solution/solution.ts
dooriya: Yes we do. Already checked.

packages/fx-core/src/plugins/solution/fx-solution/solution.ts
dooriya: Good catch. Will remove the redundant `await`.

qinezh (approved):

1yefuwang1 (approved):

PR #1818, opened 2021-08-03
fix(apim): register apim rp to subscription

XiaofuHuang:
Register apim rp to subscription

Detail: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10404485


packages/fx-core/src/plugins/resource/apim/managers/apimManager.ts
```diff
@@ -39,6 +39,9 @@ export class ApimManager {
     appName: string
   ): Promise<void> {
     const apimService: ApimService = await this.lazyApimService.getValue();
+
+    await apimService.ensureResourceProvider();
```
dooriya, line 43: Any notification/log to user when we help them to add APIM RP to the subscription?

XiaofuHuang: The logs are in ApimService#execute() func.

If the user do not have permission to add resource provider, the error message will be like:
image

dooriya (approved):

PR #1820, opened 2021-08-03
chore: add login click event and correlation id for login/logout

HuihuiWu-Microsoft:
1. add login click event when user click login on UI. The signout event for user clicking on UI already exists as signout-start
2. add correlation id for login and logout

### DIFF
--- a/packages/vscode-extension/src/commandsTreeViewProvider.ts
+++ b/packages/vscode-extension/src/commandsTreeViewProvider.ts
@@ -8,6 +8,7 @@ import { TreeItem, TreeCategory, Result, FxError, ok } from "@microsoft/teamsfx-
 import * as StringResources from "./resources/Strings.json";
 import { exp } from "./exp/index";
 import { TreatmentVariables } from "./exp/treatmentVariables";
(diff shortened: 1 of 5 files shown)


packages/vscode-extension/src/commandsTreeViewProvider.ts
```diff
@@ -313,7 +314,9 @@ export class CommandsTreeViewProvider implements vscode.TreeDataProvider<TreeVie
         continue;
       }
 
-      const disposable = vscode.commands.registerCommand(treeItem.commandId, treeItem.callback!);
+      const disposable = vscode.commands.registerCommand(treeItem.commandId, (...args) =>
+        Correlator.run(treeItem.callback!, args)
```
1openwindow, line 318: 1. why need correlator here? 
2. looks like the add function is not used any more?

HuihuiWu-Microsoft: 1. The correlator here is for adding correlation id for sign in
2. sign in is still using this API to register command

1openwindow (approved):

1openwindow (approved):

PR #1825, opened 2021-08-04
perf(arm support): print detailed error message

dilin-MS:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10512760
Detailed error message
image


blackchoey (approved):

blackchoey:
cc @MuyangAmigo you can find the sample error message in PR description. We may discuss further on the UX things.

jayzhang (approved):

PR #1832, opened 2021-08-04
refactor(multi-env): move programming language to project settings

dooriya:
Previously the `programmingLanguage` settings is saved in `env.default.json` file. As this is a project-level config, we'll move it to `settings.json` file to better support multi-env.

@YitongFeng-git As we modified bot & function plugin to read `programmingLanguage` from project settings instead of solution plugin config, please also help to review this PR. Thanks in advance.

### DIFF
--- a/packages/api/review/teamsfx-api.api.md
+++ b/packages/api/review/teamsfx-api.api.md
@@ -682,6 +682,8 @@ export interface ProjectSettings {
     // (undocumented)
     appName: string;
     // (undocumented)
+    programmingLanguage?: string;
+    // (undocumented)
     projectId: string;
     // (undocumented)
     solutionSettings?: SolutionSettings;
--- a/packages/api/src/types.ts
+++ b/packages/api/src/types.ts
@@ -126,6 +126,7 @@ export type EnvConfig = Dict<string>;
 export interface ProjectSettings {
   appName: string;
   projectId: string;
+  programmingLanguage?: string;
   solutionSettings?: SolutionSettings;
 }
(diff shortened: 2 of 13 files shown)


1yefuwang1 (approved):

hund030 (approved):

qinezh (approved):

IvanJobs (approved):
for bot part, looks good to me.

packages/api/review/teamsfx-api.api.md
```diff
@@ -682,6 +682,8 @@ export interface ProjectSettings {
     // (undocumented)
     appName: string;
     // (undocumented)
+    programmingLanguage?: string;
```
IvanJobs, line 685: Nit:  one thing comes to my mind is, how about extract this programming language as enum, and all stakeholders share the same standard of the programming language?
Currently, bot plugin will validate this programming language to check if it's in list of ['javascript', 'typescript']. If the enum definition is in a central place, it would be better.

dooriya: Making the `programmingLanguage` as enum sounds good to me.
As the purpose of this PR is to move this setting from `env.default.json` to `settings.json`, I would prefer keep the logic of consuming the `programmingLanguage` not change too much.

Since mainly bot & function plugin use the `progammingLanguage` setting, as the plugin owners, you can add a enum (e,g. in `api` package) to facilitate using this setting.

kuojianlu (approved):

PR #1838, opened 2021-08-05
refactor(core): extract the env info loader/writer from context loader/writer middleware

qinezh:
Separate the original `ContextLoad` and `ConfigWriter` middleware into two parts:
- Project settings loader/writer
- Env profile loader/writer

Reason to do it:
- Some lifecycle or operation doesn't need to load env profile, such like local debug.
- Env profile loader would require user input for the target env name if there're multiple environments, while project settings loader doesn't need.

### DIFF
--- a/packages/fx-core/src/core/middleware/contextLoader.ts
+++ b/packages/fx-core/src/core/middleware/contextLoader.ts
@@ -10,13 +10,12 @@ import {
   PluginConfig,
   ProjectSettings,
   Result,
-  SolutionConfig,
   SolutionContext,
   Stage,
   StaticPlatforms,
   Tools,
 } from "@microsoft/teamsfx-api";
-import { CoreHookContext, FxCore, objectToMap } from "../..";
+import { CoreHookContext } from "../..";
 import {
   InvalidProjectError,
   NoProjectOpenedError,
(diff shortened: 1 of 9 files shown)


packages/fx-core/src/core/middleware/contextLoader.ts
```diff
@@ -26,15 +25,14 @@ import {
...
 import * as uuid from "uuid";
 import { LocalCrypto } from "../crypto";
 import {
   GLOBAL_CONFIG,
   PluginNames,
   PROGRAMMING_LANGUAGE,
 } from "../../plugins/solution/fx-solution/constants";
-import { environmentManager } from "../environment";
 
 export const ContextLoaderMW: Middleware = async (ctx: CoreHookContext, next: NextFunction) => {
```
dooriya, line 37: minor: to be clear, we may need to also update the naming for `ContextLoaderMW` and `ConfigWriterMW` as their scope is changed.

dooriya (approved):

packages/fx-core/src/core/middleware/contextLoader.ts
qinezh: Yes, it will be in next PR. I didn't cover it in this PR because I would like to make the code diff be clearer to be reviewed😄

jayzhang (approved):

PR #1847, opened 2021-08-05
fix(localdebug): revert cert behavior for windows non-admin users

swatDong:
Fix ADO bug-10533295 - Revert the trust cert behavior for windows non-admin users.


swatDong:
Close this since PR from forked repo failed.

PR #1848, opened 2021-08-05
refactor(local-debug): update aad plugin to use localSettings for local debug

dooriya:
If multi-env feature flag is enabled, AAD plugin will use `localSettings.json` to read/write local debug related configuration. Otherwise, the local debug config logic remains the same as original.

### DIFF
--- a/packages/fx-core/src/common/localSettingsConstants.ts
+++ b/packages/fx-core/src/common/localSettingsConstants.ts
@@ -8,11 +8,11 @@ export const LocalSettingsTeamsAppKeys = Object.freeze({
 });
 
 export const LocalSettingsAuthKeys = Object.freeze({
-  AadClientId: "aadClientId",
-  AadClientSecret: "aadClientSecret",
-  AadObjectId: "aadObjectId",
-  AadOauth2PermissionScopeId: "aadOauth2PermissionScopeId",
-  AadApplicationIdUris: "aadApplicationIdUris",
+  AadClientId: "clientId",
+  AadClientSecret: "clientSecret",
+  AadObjectId: "objectId",
+  AadOauth2PermissionScopeId: "oauth2PermissionScopeId",
+  AadApplicationIdUris: "applicationIdUris",
   SimpleAuthFilePath: "simpleAuthFilePath",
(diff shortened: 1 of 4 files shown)


packages/fx-core/src/common/localSettingsConstants.ts
```diff
@@ -8,11 +8,11 @@ export const LocalSettingsTeamsAppKeys = Object.freeze({
 });
 
 export const LocalSettingsAuthKeys = Object.freeze({
-  AadClientId: "aadClientId",
-  AadClientSecret: "aadClientSecret",
-  AadObjectId: "aadObjectId",
-  AadOauth2PermissionScopeId: "aadOauth2PermissionScopeId",
-  AadApplicationIdUris: "aadApplicationIdUris",
+  AadClientId: "clientId",
```
swatDong, line 11: Just better style, align the key and value `ClientId: "clientId"`.

packages/fx-core/src/plugins/resource/aad/utils/configs.ts
```diff
@@ -8,14 +8,84 @@ import { format, Formats } from "./format";
...
+import { getArmOutput, isArmSupportEnabled, isMultiEnvEnabled } from "../../../../common";
+import {
+  LocalSettingsBotKeys,
+  LocalSettingsFrontendKeys,
+} from "../../../../common/localSettingsConstants";
 
-function checkAndSaveConfig(context: PluginContext, key: string, value: ConfigValue) {
-  if (!value) {
-    return;
+// function checkAndSaveConfig(context: PluginContext, key: string, value: ConfigValue) {
```
swatDong, line 17: Why comment out but not remove?

packages/fx-core/src/common/localSettingsConstants.ts
dooriya: Agree. updated in the latest commit.

swatDong (approved):

packages/fx-core/src/plugins/resource/aad/utils/configs.ts
dooriya: Just forgot to remove it . Already clean up.

KennethBWSong (approved):

qinezh (approved):

PR #1881, opened 2021-08-09
refactor(localdebug): read config from localSettings for AppStudio local debug

dooriya:


qinezh (approved):

packages/fx-core/src/plugins/resource/appstudio/plugin.ts
```diff
@@ -1006,6 +994,107 @@ export class AppStudioPluginImpl {
     return ok({ tabEndpoint, tabDomain, aadId, botDomain, botId, webApplicationInfoResource });
   }
 
+  private getTabEndpoint(ctx: PluginContext, isLocalDebug: boolean) {
```
nliu-ms, line 997: it's better to announce the return type?

dooriya: Good catch! Let's me fix it.

nliu-ms (approved):

PR #1887, opened 2021-08-10
chore(arm): fix bug of reading arm output in appstudio

xzf0587:
Since the validate manifest need to read arm output, remove the code to delete arm output in config.
@blackchoey, it seems you added the deleting code deliberately.  Please help to review it.


blackchoey:
when will validate manifest consume ARM output? if it's not part of provision, maybe consuming the resource plugin output can provide more unified experience (which should be same logic in existing code)?
In my opinion, we may need to avoid cases that a feature needs to consume ARM output and plugin output at the same time. This may make resource plugin developer confused: when should I use ARM output and when should I use plugin output?
In the meantime, our customers may modify the ARM template so ARM output is not reliable and we may record sensitive info by accident.

xzf0587:
> Thank you for helping make this change. Can you help me understand when will validate manifest be used? if it's not part of provision, maybe consuming the resource plugin output can provide more unified experience (which should be same logic in existing code)? if it's part of provision, we may need to fix related bug.
> In my opinion, we may need to avoid cases that a feature needs to consume ARM output and plugin output at the same time. This may make resource plugin developer confused: when should I use ARM output and when should I use plugin output?
> In the meantime, our customers may modify the ARM template so ARM output is not reliable and we may record sensitive info by accident.

The validate manifest is a command in vscode-extension the toolkit provided. Actually, the case is that developers run the provision command and then run the validate manifest command. Validate manifest will read arm output from solution context which is deleted in the end of provision command. So it will fail.

In addition, if we reload vscode after provision, the validate manifest command will succeed as the solution will read the env.default.json and get the arm output in context ready.

blackchoey:
If the validate command can success after reload VS Code after provision, this seems to be an issue around solution context management: since ARM output already been deleted after provision. Can we fix related issue from solution context management perspective? Not sure if anything is missed when we update related logic.

xzf0587:
After discussion with @dilin-MS  and @adashen, we agreed that the time of using arm output  is only **provision stage**. Plugins could set necessary arm output into context config in post-provision.
As validate manifest command is out of  provision stage, it should read setting from context config.
And the problem that the context config persist arm output will be tracked as well.

nliu-ms (approved):

PR #1892, opened 2021-08-10
refactor: update check permission messages part 1

SLdragon:


packages/fx-core/src/plugins/solution/fx-solution/solution.ts
```diff
@@ -1335,30 +1351,51 @@ export class TeamsAppSolution implements Solution {
...
-              }
-            }
-          }
-        }
+      ctx.logProvider?.info("=== M365 account Permissions ===");
+      if (m365PermissionCheckError) {
+        ctx.logProvider?.error(`${errorMsg}: ${m365PermissionCheckError}`);
+      } else {
+        ctx.logProvider?.info(`Your current signed M365 account is: ${userInfo.userPrincipalName}`);
+        this.printPermissionLogs("Azure", results, ctx);
```
KennethBWSong, line 1359: Why here is azure?

packages/fx-core/src/plugins/solution/fx-solution/solution.ts
```diff
@@ -1335,30 +1351,51 @@ export class TeamsAppSolution implements Solution {
...
+      }
+
+      ctx.logProvider?.info("=== Azure Resource Permissions ===");
+      if (azurePermissionCheckError) {
+        ctx.logProvider?.error(`${errorMsg}: ${azurePermissionCheckError}`);
+      } else {
+        const accountInfo = await ctx.azureAccountProvider?.getAccountInfo();
+        ctx.logProvider?.info(`Your current signed Azure account is: ${accountInfo!.unique_name}`);
+        this.printPermissionLogs("M365", results, ctx);
```
KennethBWSong, line 1368: same as above

packages/fx-core/src/plugins/solution/fx-solution/solution.ts
SLdragon: Oh, sorry for mistake

packages/fx-core/src/plugins/solution/fx-solution/solution.ts
SLdragon: Updated

packages/fx-core/src/plugins/solution/fx-solution/solution.ts
SLdragon: Updated

KennethBWSong (approved):

PR #1899, opened 2021-08-11
fix(frontend): launch cmd with uppercase drive letter in windows

hund030:


packages/fx-core/src/plugins/resource/frontend/utils.ts
```diff
@@ -21,6 +22,10 @@ export class Utils {
     return raw.replace(RegularExpr.AllCharToBeSkippedInName, Constants.EmptyString).toLowerCase();
   }
 
+  static capitalizeFirstLetter([first, ...rest]: Iterable<string>): string {
+    return [first?.toUpperCase(), ...rest].join("");
```
henzhang-ms, line 26: Nit: what if string is empty? I am not familiar with this style. Want a test result for it.

henzhang-ms (approved):

packages/fx-core/src/plugins/resource/frontend/utils.ts
hund030: If string is empty, the `first` element would be undefined and the method return empty.
```
> (([first, ...rest]) => {
      return {first: first, rest: rest, value: [first?.toUpperCase(), ...rest].join("")};
  }) ("aaaa")
{first: "a", rest: Array(3), value: "Aaaa"}
> (([first, ...rest]) => {
      return {first: first, rest: rest, value: [first?.toUpperCase(), ...rest].join("")};
  }) ("")
{first: undefined, rest: Array(0), value: ""}
```

henzhang-ms: cool

packages/fx-core/src/plugins/resource/frontend/utils.ts
```diff
@@ -66,6 +71,11 @@ export class Utils {
     return new Promise((resolve, reject) => {
       Logger.info(`Start to run command: "${command}".`);
 
+      // Drive letter should be uppercase, otherwise when we run webpack in exec, it fails to resolve nested dependencies.
+      if (os.platform() === "win32") {
+        workingDir = this.capitalizeFirstLetter(path.resolve(workingDir ?? Constants.EmptyString));
```
hund030, line 76: `path.resolve("")` return current working directory.

henzhang-ms: You want to regard it as an error? What cwd do we want to pass to exec if we give it ""?

hund030: If we pass "" or undefined, then I assume the command can be run in any dir. It should be ok to use cwd.
Do you worry we may get empty path from vscode?

henzhang-ms: No, I just wonder the cwd of exec if the path is relative.
If path is relative and the cwd of subprocess starts with lower-case drive letter, then we'd better also upper case the drive letter for it.
For me, if cwd is `""`, then it means `"."`. I think exec also regard `""` as `"."`. (need verify)

henzhang-ms (approved):

packages/fx-core/src/plugins/resource/frontend/utils.ts
hund030: exec indeed treats `""` as `"."`. 
As for the drive letter, it depends on the exact path starts with lowercase or uppercase. 
For example, if I start a PowerShell in `c:\` and run exec(""), its path is `c:\`.
Anyway `path.resolve` always return absolute path, I think we don't need to worry about this.

henzhang-ms: good.

packages/fx-core/src/plugins/resource/frontend/utils.ts
hund030: The array indexed solution can only deal with UTF-16 string, although I think it's enough for path.
I just think the destructuring assignment looks fancier.

PR #1901, opened 2021-08-11
fix(collaborate): fix local collaborate errors

KennethBWSong:


packages/fx-core/src/core/middleware/projectUpgrader.ts
```diff
@@ -94,41 +117,47 @@ export async function upgradeContext(ctx: CoreHookContext): Promise<void> {
...
     sendTelemetryErrorEvent(
       core?.tools?.telemetryReporter,
       inputs,
       TelemetryEvent.ProjectUpgrade,
-      errorObject,
-      {
-        [TelemetryProperty.Success]: TelemetrySuccess.No,
-        [TelemetryProperty.ErrorMessage]: error?.message,
-      }
+      errorObject
```
SLdragon, line 148: why this part remove:
```
  {
        [TelemetryProperty.Success]: TelemetrySuccess.No,
        [TelemetryProperty.ErrorMessage]: error?.message,
      }
```

KennethBWSong: It is included in sendTelemetryErrorEvent:
image

packages/fx-core/src/core/error.ts
```diff
@@ -122,11 +122,20 @@ export function FunctionRouterError(func: Func) {
...
-    CoreSource,
-    new Error().stack
-  );
+export function ContextUpgradeError(error: any, isUserError = false): FxError {
+  if (isUserError) {
+    return new UserError(
+      "ContextUpgradeError",
+      `Failed to update context: ${error.message}`,
+      CoreSource,
+      new Error().stack
```
SLdragon, line 131: Why create a new Error to get stack, how about error.stack?

KennethBWSong: Following the code structure above. Not quite sure about this. Maybe @jayzhang know more about this.

SLdragon: OK

SLdragon (approved):

packages/fx-core/src/core/error.ts
jayzhang: This is just make the stack is not null for some cases when `error.stack` is `undefined`.
This can be improved by checking existence of `error.stack`.

jayzhang (approved):

PR #1905, opened 2021-08-11
chore(multi-env): support user input for target env

qinezh:
If `TEAMSFX_MULTI_ENV` feature flag is enabled, user will be asked to input the target environment when provision, deploy and publish.

Here's the gif for the new UX:
Animation

Please note:
* the UX design is not finalized and above user behavoir might be changed in future.
* only **VS Code extension** is supported, while **CLI** is not supported at the moment.
* only projects contain **Tab** or **Function** is supported, while **Bot/SQL** is not support.

### DIFF
--- a/packages/fx-core/src/core/middleware/envInfoLoader.ts
+++ b/packages/fx-core/src/core/middleware/envInfoLoader.ts
@@ -7,52 +7,76 @@ import {
   Inputs,
   ok,
   ProjectSettings,
+  QTreeNode,
   Result,
   SolutionConfig,
   SolutionContext,
   Tools,
+  traverse,
 } from "@microsoft/teamsfx-api";
 import { CoreHookContext, FxCore } from "../..";
-import { NoProjectOpenedError } from "../error";
+import { NoProjectOpenedError, ProjectSettingsUndefinedError } from "../error";
 import { Middleware, NextFunction } from "@feathersjs/hooks/lib";
 import { LocalCrypto } from "../crypto";
 import { environmentManager } from "../environment";
 import { GLOBAL_CONFIG, PROGRAMMING_LANGUAGE } from "../../plugins/solution/fx-solution/constants";
+import { QuestionSelectTargetEnvironment, QuestionNewTargetEnvironmentName } from "../question";
+import { desensitize } from "./questionModel";
+import { shouldIgnored } from "./projectSettingsLoader";
 
-export const EnvInfoLoaderMW: Middleware = async (ctx: CoreHookContext, next: NextFunction) => {
-  if (ctx.projectSettings) {
+const newTargetEnvNameOption = "+ new environment";
+
+export function EnvInfoLoaderMW(
+  isMultiEnvEnabled: boolean,
+  allowCreateNewEnv: boolean
+): Middleware {
+  return async (ctx: CoreHookContext, next: NextFunction) => {
     const inputs = ctx.arguments[ctx.arguments.length - 1] as Inputs;
-    const core = ctx.self as FxCore;
-
-    const result = await loadSolutionContext(
-      core.tools,
-      inputs,
-      ctx.projectSettings,
-      ctx.projectIdMissing
-    );
-    if (result.isErr()) {
-      ctx.result = err(result.error);
+    if (shouldIgnored(ctx)) {
+      await next();
       return;
     }
 
-    ctx.solutionContext = result.value;
-  }
+    if (!ctx.projectSettings) {
+      ctx.result = err(ProjectSettingsUndefinedError());
+      return;
+    }
 
-  await next();
-};
+    const core = ctx.self as FxCore;
+    const targetEnvName = isMultiEnvEnabled
+      ? await askTargetEnvironment(ctx, inputs, allowCreateNewEnv)
+      : environmentManager.defaultEnvName;
+    if (targetEnvName) {
+      const result = await loadSolutionContext(
+        core.tools,
+        inputs,
+        ctx.projectSettings,
+        ctx.projectIdMissing,
+        targetEnvName
+      );
+      if (result.isErr()) {
+        ctx.result = err(result.error);
+        return;
+      }
+
+      ctx.solutionContext = result.value;
+
+      await next();
+    }
+  };
+}
 
 export async function loadSolutionContext(
   tools: Tools,
   inputs: Inputs,
   projectSettings: ProjectSettings,
-  projectIdMissing?: boolean
+  projectIdMissing?: boolean,
+  targetEnvName?: string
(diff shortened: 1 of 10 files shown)


packages/fx-core/src/plugins/solution/fx-solution/solution.ts
```diff
@@ -484,6 +484,11 @@ export class TeamsAppSolution implements Solution {
     if (canProvision.isErr()) {
       return canProvision;
     }
+
+    if (!ctx.config.get(GLOBAL_CONFIG)) {
+      ctx.config.set(GLOBAL_CONFIG, new ConfigMap());
+    }
```
dooriya, line 490: Suppose the solution config is checked and loaded in the context loader. How to about move it to the `envInfoLoader`?

qinezh: updated and move the logic while loading the env profile.

packages/fx-core/src/core/question.ts
```diff
@@ -55,6 +58,33 @@ export const QuestionRootFolder: FolderQuestion = {
   title: "Workspace folder",
 };
 
+export const QuestionSelectTargetEnvironment: SingleSelectQuestion = {
+  type: "singleSelect",
+  name: CoreQuestionNames.TargetEnvName,
+  title: "Select the target environment",
+  staticOptions: [],
+  skipSingleOption: true,
+  forgetLastValue: true,
```
dooriya, line 67: Why not make the `forgetLastValue` to be `false` to remember the last input value?

packages/fx-core/src/core/question.ts
```diff
@@ -55,6 +58,33 @@ export const QuestionRootFolder: FolderQuestion = {
   title: "Workspace folder",
 };
 
+export const QuestionSelectTargetEnvironment: SingleSelectQuestion = {
+  type: "singleSelect",
+  name: CoreQuestionNames.TargetEnvName,
+  title: "Select the target environment",
```
dooriya, line 64: wording: the -> to

packages/fx-core/src/core/question.ts
```diff
@@ -55,6 +58,33 @@ export const QuestionRootFolder: FolderQuestion = {
...
+  staticOptions: [],
+  skipSingleOption: true,
+  forgetLastValue: true,
+};
+
+export const QuestionNewTargetEnvironmentName: TextInputQuestion = {
+  type: "text",
+  name: CoreQuestionNames.NewTargetEnvName,
+  title: "New target environment name",
```
dooriya, line 73: nit: I would prefer to remove "target" for the new environment scenario, just "New environment name".

packages/fx-core/src/core/question.ts
qinezh: If not, the display order of environment list might be changed and `+ new environment` won't be displayed as the 1st place, which would be weird.

packages/fx-core/src/core/question.ts
qinezh: Thanks, updated.

packages/fx-core/src/core/question.ts
qinezh: Yeah that looks better. Thanks for the suggestion.

And just would like to clarify that, the user experience will be reviewed with UX designer, and the current behavior is for demo purpose and not finalized. So it might be changed in future.

packages/fx-core/src/core/middleware/envInfoLoader.ts
```diff
@@ -91,3 +115,76 @@ export async function loadSolutionContext(
 
   return ok(solutionContext);
 }
+
+async function askTargetEnvironment(
```
dooriya, line 119: Since we already have a [`QuestionModel middle`](https://github.com/OfficeDev/TeamsFx/blob/dev/packages/fx-core/src/core/middleware/questionModel.ts) to handle the user input for the lifecycle operations (such as provision, deploy, publish), can we also handle the "ask target environment" question, to keep single responsibility in the `envInfoLoader`?

qinezh: It's due to the logic in `QuestionModelMW` requires `solutionContext`, which needs to load env info before entering this middleware. In another word, `EnvInfoLoader` needs to be processed before `QuestionModelMW`, but `EnvInfoLoader` needs to ask user input before loading the env info.....

dooriya: I see, Thanks Qinen for the explanation.

packages/fx-core/src/core/question.ts
```diff
@@ -55,6 +58,33 @@ export const QuestionRootFolder: FolderQuestion = {
...
+export const QuestionNewTargetEnvironmentName: TextInputQuestion = {
+  type: "text",
+  name: CoreQuestionNames.NewTargetEnvName,
+  title: "New environment name",
+  validation: {
+    validFunc: async (input: string): Promise<string | undefined> => {
+      const targetEnvName = input as string;
+      const match = targetEnvName.match(environmentManager.envNameRegex);
+      if (!match) {
+        return "Environment name can only contain letters, digits, _ and -.";
```
dooriya, line 79: If we will use the env name as part of the app name/azure source name, then we may need to make the rule more restrict, like the current app name input validation.

jayzhang (approved):

packages/fx-core/src/core/question.ts
qinezh: Not sure if we have such requirements, as current pattern is `appName` + `suffix` for the resource naming. Maybe we could update it when we have this requirement?

dooriya: I can imagine the app name would be different for different environment (please correct me if I'm wrong), like current teams toolkit, the remote app for the default env use `{app_name}`, while local debug app use `{app_name}_localdebug`.

Maybe this is overthinking, just a remind that if that case happen, we may have to consider backwards compatibility for the old env name with not-allowed characters.

qinezh: Got it, so the question is "will the env name be exposed to user to reference, such like `app name`?". If so, I would prefer to discuss it with PM and take care of it as a standalone task and not cover it in this PR, since the `app name` is a example and there might be other restrictions if the env name is supported be referenced.

qinezh: Since we don't know where the env name will be used, we don't know all the restriction. (user might use it out of our control). So another point is that, it's the consumer's responsibility to format the naming/user input.

dooriya: Sure, let's discuss this with PM first.

PR #1906, opened 2021-08-11
refactor: update check permission messages

SLdragon:
image


packages/fx-core/src/plugins/resource/aad/plugin.ts
```diff
@@ -347,21 +349,24 @@ export class AadAppForTeamsImpl {
...
     }
 
-    return ResultFactory.Success([
+    const result: ResourcePermission[] = [
       {
         name: Constants.permissions.name,
         type: Constants.permissions.type,
         roles: isAadOwner ? [Constants.permissions.owner] : [Constants.permissions.noPermission],
         error: checkAadPermissionError,
+        resourceId: userObjectId,
```
KennethBWSong, line 365: Here id should be aad app object id?

KennethBWSong (approved):

PR #1910, opened 2021-08-11
chore(arm): use target env name from context

qinezh:
Replace the hardcoded `default` env name with the target env name from context. 

- When the feature flag `TEAMSFX_MULTI_ENV` is not enabled, the target env equals to `default`.
- When the feature flag `TEAMSFX_MULTI_ENV` enabled, the target env is specified by user and would be accessed from context.

### DIFF
--- a/packages/fx-core/src/plugins/solution/fx-solution/arm.ts
+++ b/packages/fx-core/src/plugins/solution/fx-solution/arm.ts
@@ -36,7 +36,7 @@ const parameterFolder = "parameters";
 const bicepOrchestrationFileName = "main.bicep";
 const armTemplateJsonFileName = "main.json";
 const parameterTemplateFileName = "parameters.template.json";
-const parameterDefaultFileName = "parameters.default.json";
+const parameterFileNameTemplate = "parameters.@envName.json";
 const solutionLevelParameters = `param resourceBaseName string\n`;
 const solutionLevelParameterObject = {
   resourceBaseName: {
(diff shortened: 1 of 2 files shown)


packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -260,8 +260,13 @@ export async function deployArmTemplates(ctx: SolutionContext): Promise<Result<v
 }
 
 async function getParameterJson(ctx: SolutionContext) {
+  if (!ctx.targetEnvName) {
+    throw new Error("Failed to get target environment name from solution context.");
+  }
+
+  const parameterFileName = parameterFileNameTemplate.replace("@envName", ctx.targetEnvName);
```
dilin-MS, line 267: Where does the default value `default` configured into `ctx.targetEnvName`?

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -260,8 +260,13 @@ export async function deployArmTemplates(ctx: SolutionContext): Promise<Result<v
 }
 
 async function getParameterJson(ctx: SolutionContext) {
+  if (!ctx.targetEnvName) {
```
dilin-MS, line 263: If `ctx.targetEnvName` not set, should we use `default` as the default value?

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
qinezh: It's in the middleware `envInfoLoader.ts`: https://github.com/OfficeDev/TeamsFx/blob/dev/packages/fx-core/src/core/middleware/envInfoLoader.ts#L55

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
qinezh: I would prefer to manage the default value in one place, otherwise the fallback login will be hard to maintain.

dilin-MS (approved):

dooriya (approved):

1yefuwang1 (approved):

PR #1911, opened 2021-08-11
fix: fill manifest id with teams app id

nliu-ms:
1) Fill manifest with teams app id, rather than aad app id.
2) Replace `createManifestForRemote`, `getConfigAndAppDefinition` and `getDevAppDefinition`,  with `getAppDefinitionAndManifest` method.


1yefuwang1 (commented):
Maybe a test case can be added to enforce the property that manifest id equals teams app id.

jayzhang (approved):

PR #1912, opened 2021-08-11
fix: make sure SystemError and UserError has stack

jayzhang:
When creating UserError or SystemError, the stack fields will be created automatically, user don't need to input.
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10577820

### DIFF
--- a/packages/api/src/error.ts
+++ b/packages/api/src/error.ts
@@ -19,7 +19,7 @@ export interface FxError extends Error {
 /**
  * Users can recover by themselves, e.g., users input invalid app names.
  */
-export class UserError implements FxError {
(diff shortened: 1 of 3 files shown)


packages/api/src/error.ts
```diff
@@ -57,11 +45,11 @@ export class UserError implements FxError {
     helpLink?: string,
     innerError?: any
   ) {
+    super(message);
```
jayzhang, line 48: stack will be automatically created, user don't need to input

LongOddCode (approved):

meifans (approved):
LGTM

PR #1914, opened 2021-08-12
chore(function): add bicep support

xzf0587:
1. add function bicep scaffold
2. update function post-provision 
3. update relative frontend reading arm output
4. add unit test
5. walk around appstudio reading arm output bug


packages/fx-core/src/plugins/resource/function/plugin.ts
```diff
@@ -636,6 +641,64 @@ export class FunctionPluginImpl {
...
+    const result: ScaffoldArmTemplateResult = {
+      Modules: {
+        functionProvision: {
+          Content: await fs.readFile(moduleTemplateFilePath, Bicep.uft8Encoding),
+        },
+      },
+      Orchestration: {
+        ParameterTemplate: {
+          Content: await fs.readFile(parameterTemplateFilePath, Bicep.uft8Encoding),
+          ParameterJson: JSON.parse(await fs.readFile(parameterFilePath, Bicep.uft8Encoding)),
```
dilin-MS, line 684: Please use the constant defined in packages\fx-core\src\common\constants.ts `ConstantString.UTF8Encoding`

packages/fx-core/src/plugins/resource/function/plugin.ts
```diff
@@ -636,6 +641,64 @@ export class FunctionPluginImpl {
...
+      "bicep"
+    );
+
+    const moduleTemplateFilePath = path.join(bicepTemplateDirectory, Bicep.moduleTemplateFileName);
+
+    const parameterTemplateFilePath = path.join(
+      bicepTemplateDirectory,
+      Bicep.inputParameterOrchestrationFileName
+    );
+    const resourceTemplateFilePath = path.join(
```
dilin-MS, line 665: Please rename the variable as `moduleOrchestrationFilePath` to make it consistent. (It is `module` instead of `resource`)

packages/fx-core/templates/plugins/resource/function/bicep/function.template.bicep
```diff
@@ -0,0 +1,125 @@
+param functionServerfarmsName string
```
dilin-MS, line 1: Please also add tests case for generating function bicep files. Or test coverage will reduce.

packages/fx-core/templates/plugins/resource/function/bicep/function.template.bicep
```diff
@@ -0,0 +1,125 @@
...
+@maxLength(24)
+@description('Name of Storage Accounts for function backend.')
+param functionStorageName string
+param AADClientId string
+@secure()
+param AADClientSecret string
+param tenantId string
+param applicationIdUri string
+
+{{#contains 'frontend_hosting' pluginTypes}}
```
dilin-MS, line 14: `pluginTypes` should be `plugins`.

packages/fx-core/templates/plugins/resource/function/bicep/function.template.bicep
```diff
@@ -0,0 +1,125 @@
...
+}
+
+resource functionStorage 'Microsoft.Storage/storageAccounts@2021-04-01' = {
+  kind: 'StorageV2'
+  location: resourceGroup().location
+  name: functionStorageName
+  properties: {
+    accessTier: 'Hot'
+    supportsHttpsTrafficOnly: true
+    isHnsEnabled: true
```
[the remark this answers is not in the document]
xzf0587: If Linux app service plan true, false otherwise.
I will remove it to make code clean. Thanks.

packages/fx-core/templates/plugins/resource/function/bicep/function.template.bicep
```diff
@@ -0,0 +1,125 @@
...
+
+resource functionServerfarms 'Microsoft.Web/serverfarms@2020-06-01' = {
+  name: functionServerfarmsName
+  location: resourceGroup().location
+  sku: {
+    name: 'Y1'
+  }
+  kind: 'functionapp'
+  properties: {
+    reserved: false
```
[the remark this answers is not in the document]
xzf0587: If Linux app service plan true, false otherwise.
I will remove it to make code clean. Thanks.

packages/fx-core/templates/plugins/resource/function/bicep/function.template.bicep
```diff
@@ -0,0 +1,125 @@
...
+    serverFarmId: functionServerfarms.id
+    siteConfig: {
+      cors: {
+        allowedOrigins: [
+          frontendHostingStorageEndpoint
+        ]
+      }
+      alwaysOn: false
```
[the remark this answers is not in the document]
xzf0587: This is a default setting. I will remove it to keep clean.
image

packages/fx-core/templates/plugins/resource/function/bicep/function.template.bicep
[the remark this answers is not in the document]
xzf0587: It is removed. Thanks.

packages/fx-core/src/plugins/resource/appstudio/plugin.ts
```diff
@@ -761,12 +761,15 @@ export class AppStudioPluginImpl {
   > {
     let tabEndpoint, tabDomain;
     if (isArmSupportEnabled()) {
-      if (localDebug) {
+      // getConfigForCreatingManifest is called in post-provision and validate manifest
+      // only in post stage, we find the value from arm output.
+      // Here is a walk-around way, try to get from arm output first and then get from ctx config.
```
dilin-MS, line 766: Why do we `try to get from arm output first and then get from ctx config`? Should the value always can be got from arm output?

packages/fx-core/src/plugins/resource/appstudio/plugin.ts
```diff
@@ -761,12 +761,15 @@ export class AppStudioPluginImpl {
   > {
     let tabEndpoint, tabDomain;
     if (isArmSupportEnabled()) {
-      if (localDebug) {
```
dilin-MS, line 764: What's the difference between local-debug and remote?

packages/fx-core/src/plugins/resource/frontend/plugin.ts
```diff
@@ -169,23 +170,33 @@ export class FrontendPluginImpl {
...
-      functionEnv = {
-        defaultName: functionPlugin.get(DependentPluginInfo.FunctionDefaultName) as string,
-        endpoint: functionPlugin.get(DependentPluginInfo.FunctionEndpoint) as string,
-      };
-    }
-
     if (isArmSupportEnabled()) {
-      const endpoint = getArmOutput(ctx, ArmOutput.SimpleAuthEndpoint) as string;
-      if (endpoint) {
+      const functionPlugin = ctx.configOfOtherPlugins.get(DependentPluginInfo.FunctionPluginName);
```
dilin-MS, line 174: This line can be moved out of the if-else clause.

packages/fx-core/src/plugins/resource/appstudio/plugin.ts
xzf0587: Because this method will be called both in provision stage and no-provision stage.
arm output only exists in provision stage.
Here is a walk-around way to handle it. 
Ideally we should withdraw the code about provision stage.
I think we can refactor it when our design about arm output and ctx config is done.

packages/fx-core/src/plugins/resource/frontend/plugin.ts
```diff
@@ -169,23 +170,33 @@ export class FrontendPluginImpl {
...
+      if (functionPlugin) {
+        const functionEndpoint = getArmOutput(ctx, FunctionArmOutput.Endpoint) as string;
+        functionEnv = {
+          defaultName: functionPlugin.get(DependentPluginInfo.FunctionDefaultName) as string,
+          endpoint: `https://${functionEndpoint}`,
```
dilin-MS, line 179: Looks we only need to calculate endpoint inside the if-else clause and then construct the functionEnv outside so we can reduce duplicate code.

packages/fx-core/src/plugins/resource/appstudio/plugin.ts
xzf0587: local debug is out of provision stage. It will use ctx config other than arm output.
Actually, this method will be called in provision, local debug and validate manifest command.
It uses this condition to separate local debug from other scenarios.

packages/fx-core/src/plugins/resource/frontend/plugin.ts
```diff
@@ -169,23 +170,33 @@ export class FrontendPluginImpl {
...
-    }
-
     if (isArmSupportEnabled()) {
-      const endpoint = getArmOutput(ctx, ArmOutput.SimpleAuthEndpoint) as string;
-      if (endpoint) {
+      const functionPlugin = ctx.configOfOtherPlugins.get(DependentPluginInfo.FunctionPluginName);
+      if (functionPlugin) {
+        const functionEndpoint = getArmOutput(ctx, FunctionArmOutput.Endpoint) as string;
+        functionEnv = {
+          defaultName: functionPlugin.get(DependentPluginInfo.FunctionDefaultName) as string,
```
hund030, line 178: @dooriya is working on moving the `defaultFunctionName` from function config to project setting. Please decide which one we use in this PR. https://github.com/OfficeDev/TeamsFx/pull/1944

packages/fx-core/src/plugins/resource/frontend/plugin.ts
```diff
@@ -169,23 +170,33 @@ export class FrontendPluginImpl {
...
+          endpoint: `https://${functionEndpoint}`,
+        };
+      }
+
+      const authPlugin = ctx.configOfOtherPlugins.get(DependentPluginInfo.RuntimePluginName);
+      if (authPlugin) {
```
dilin-MS, line 184: Same here.

packages/fx-core/src/plugins/resource/function/constants.ts
```diff
@@ -181,3 +181,17 @@ export class Commands {
 export class QuestionValidationFunc {
   public static readonly validateFunctionName: string = "validateFunctionName";
 }
+
+export class BicepSnippet {
+  static readonly moduleTemplateFileName: string = "function.template.bicep";
```
dilin-MS, line 186: This is a file name variable, not like a bicep snippet.

packages/fx-core/src/plugins/resource/function/plugin.ts
```diff
@@ -522,7 +530,7 @@ export class FunctionPluginImpl {
       this.config.subscriptionId,
       FunctionConfigKey.subscriptionId
     );
-    const functionAppName = this.checkAndGet(
+    let functionAppName = this.checkAndGet(
```
dilin-MS, line 533: If this value is used only when not arm supported, what about move it to the `if (!isArmSupported)` block?

packages/fx-core/src/plugins/resource/frontend/plugin.ts
xzf0587: I think we can avoid sharing logic code between arm support and no arm support because they will not coexist if we switch to arm support. The current style is more readable. We can improve the whole code easily when we switch it.

packages/fx-core/src/plugins/resource/function/plugin.ts
xzf0587: The value will be updated by arm output in arm support scenario.

packages/fx-core/src/plugins/resource/function/resources/errors.ts
```diff
@@ -243,6 +243,12 @@ export class PublishCredentialError extends FunctionPluginError {
   }
 }
 
+export class FindAppError extends FunctionPluginError {
```
dilin-MS, line 246: What about `FunctionAppNotFoundError`?

packages/fx-core/src/plugins/resource/function/plugin.ts
```diff
@@ -540,15 +543,37 @@ export class FunctionPluginImpl {
...
-    const site = this.checkAndGet(this.config.site, FunctionConfigKey.site);
-    this.config.site = undefined;
-
     const webSiteManagementClient: WebSiteManagementClient = await runWithErrorCatchAndThrow(
       new InitAzureSDKError(),
       () => AzureClientFactory.getWebSiteManagementClient(credential, subscriptionId)
     );
 
+    let site: Site;
+    if (isArmSupportEnabled()) {
```
hund030, line 552: I think the post provision flow is becoming more and more complex. Both in frontend and function plugin. I really suggest to split the arm flow and origin flow to make things easier.

packages/fx-core/templates/plugins/resource/function/bicep/function.template.bicep
```diff
@@ -0,0 +1,120 @@
...
+{{/contains}}
+{{#contains 'fx-resource-azure-sql' Plugins}}
+param sqlDatabaseName string
+param sqlEndpoint string
+{{/contains}}
+{{#contains 'fx-resource-identity' Plugins}}
+param identityId string
+{{/contains}}
+
+var oauthAuthorityHost = environment().authentication.loginEndpoint
```
dilin-MS, line 24: `oauthAuthorityHost` has been moved to be an input parameter which is passed in by aad plugin.

packages/fx-core/tests/plugins/resource/function/unit/expectedBicepFiles/function_test.bicep
```diff
@@ -0,0 +1,104 @@
...
+param functionStorageName string
+param aadClientId string
+@secure()
+param aadClientSecret string
+param m365TenantId string
+param applicationIdUri string
+
+param frontendHostingStorageEndpoint string
+
+var oauthAuthorityHost = environment().authentication.loginEndpoint
```
dilin-MS, line 15: Same here.

packages/fx-core/tests/plugins/resource/function/unit/generateArmTemplates.test.ts
```diff
@@ -0,0 +1,116 @@
...
+        },
+      },
+      projectSettings: {
+        appName: "ut",
+        programmingLanguage: "javascript",
+      },
+      config: new Map(),
+      root: path.join(__dirname, "ut"),
+    };
+  });
```
dilin-MS, line 49: This context setup is exactly the same as code in *packages\fx-core\tests\plugins\resource\function\unit\scaffold.test.ts*. Can we move it to utils.ts for test?

packages/fx-core/src/plugins/resource/function/resources/errors.ts
xzf0587: In post-provision, the plugin will get the function resource and prepare to update it.
This error means finding function failed.

packages/fx-core/tests/plugins/resource/function/unit/expectedBicepFiles/function_test.bicep
```diff
@@ -0,0 +1,104 @@
...
+var teamsAadIds = '1fec8e78-bce4-4aaf-ab1b-5451cc387264;5e3ce6c0-2b1f-4285-8d4b-75ee78787346'
```
dilin-MS, line 16: Please refer to *packages\fx-core\templates\plugins\resource\simpleauth\bicep\simpleAuth.template.bicep* and use readable variables to refer to this two client id.

packages/fx-core/src/plugins/resource/function/plugin.ts
xzf0587: OK. I split them. Thanks.

packages/fx-core/src/plugins/resource/frontend/plugin.ts
xzf0587: merged the update of defaultFunctionName. Thanks.

packages/fx-core/src/plugins/resource/frontend/plugin.ts
xzf0587: I split the code to different methods. Thanks.

packages/fx-core/src/plugins/resource/function/resources/errors.ts
```diff
@@ -243,6 +243,12 @@ export class PublishCredentialError extends FunctionPluginError {
...
+  constructor() {
+    super(ErrorType.User, "FindAppError", "Failed to find the function app.", [tips.doProvision]);
```
hund030, line 248: I think this could be system error. We just successfully created the function app in provision stage, and should be able to find it in post-provision stage, otherwise there may be a bug.

hund030 (approved):
Overall look good to me. Thanks!

packages/fx-core/src/plugins/resource/frontend/plugin.ts
```diff
@@ -222,19 +202,59 @@ export class FrontendPluginImpl {
...
-      config.domain = getArmOutput(ctx, ArmOutput.FrontendDomain) as string;
-      config.syncToPluginContext(ctx);
+    return ok(undefined);
+  }
 
-      const client = new AzureStorageClient(config);
-      await runWithErrorCatchAndThrow(
-        new EnableStaticWebsiteError(),
-        async () => await client.enableStaticWebsite()
+  public async postProvisionWithArm(ctx: PluginContext): Promise<TeamsFxResult> {
```
dilin-MS, line 208: NOT suggest to copy the same logic function into two functions since it will highly increase maintenance effort. 

Suggest to extract the get-config logic as a sub-function so we can move the isArmSupportEnabled  judgement to the sub-function to make the logic more clear.

hund030: As for frontend part. Since I'm going to update the whole post provision method, it's ok to keep this and I will do some refactor before my changes.

packages/fx-core/src/plugins/resource/simpleauth/plugin.ts
```diff
@@ -142,8 +142,8 @@ export class SimpleAuthPluginImpl {
     const selectedPlugins = (ctx.projectSettings?.solutionSettings as AzureSolutionSettings)
       .activeResourcePlugins;
     const context = {
-      plugins: selectedPlugins,
-    };
+      Plugins: selectedPlugins,
+    } as BicepPluginsContext;
```
blackchoey, line 146: the context here should be determined by each resource plugin, instead of asking solution plugin define an interface, because we do not need to limit what context can be used to render bicep files in resource plugin. Any data (e.g. some answers from users) can be used by resource plugin to generate the bicep file, this is also why we ask each resource plugin render their own bicep module files in resource plugin's logic, instead of render them in solution plugin's logic.

packages/fx-core/templates/plugins/resource/function/bicep/function.template.bicep
```diff
@@ -0,0 +1,122 @@
...
+var authorizedClientApplicationIds = '${teamsMobileOrDesktopAppClientId};${teamsWebAppClientId}'
+
+resource functionServerfarms 'Microsoft.Web/serverfarms@2020-06-01' = {
+  name: functionServerfarmsName
+  location: resourceGroup().location
+  sku: {
+    name: 'Y1'
+  }
+  kind: 'functionapp'
+  properties: {
```
blackchoey, line 36: can we remove the empty properties?

packages/fx-core/templates/plugins/resource/function/bicep/function.template.bicep
```diff
@@ -0,0 +1,122 @@
...
+  kind: 'functionapp'
+  name: functionAppName
+  location: resourceGroup().location
+  properties: {
+    reserved: false
+    serverFarmId: functionServerfarms.id
+    siteConfig: {
+      cors: {
+        allowedOrigins: [
+          frontendHostingStorageEndpoint
```
blackchoey, line 50: may also need to see whether the frontend hosting plugin is enabled here

packages/fx-core/templates/plugins/resource/function/bicep/function.template.bicep
```diff
@@ -0,0 +1,122 @@
...
+        ]
+      }
+      numberOfWorkers: 1
+    }
+  }
+}
+
+resource functionStorage 'Microsoft.Storage/storageAccounts@2021-04-01' = {
+  kind: 'StorageV2'
```
blackchoey, line 59: for `name`, `kind`, `location`, we'd better align the sequence of them for all the resources.

packages/fx-core/templates/plugins/resource/function/bicep/function.template.bicep
```diff
@@ -0,0 +1,122 @@
...
+
+var oauthAuthority = uri(oauthAuthorityHost, m365TenantId)
+
+resource functionAppAppSettings 'Microsoft.Web/sites/config@2018-02-01' = {
+  parent: functionApp
+  name: 'appsettings'
+  properties: {
+    API_ENDPOINT: 'https://${functionApp.properties.hostNames[0]}'
+    ALLOWED_APP_IDS: authorizedClientApplicationIds
+    AzureWebJobsDashboard: 'DefaultEndpointsProtocol=https;AccountName=${functionStorage.name};AccountKey=${listKeys(resourceId(resourceGroup().name, 'Microsoft.Storage/storageAccounts', functionStorage.name), '2019-04-01').keys[0].value};EndpointSuffix=${environment().suffixes.storage}'
```
blackchoey, line 79: `AccountKey=${listKeys(storageAccount.id, storageAccount.apiVersion).keys[0].value}` to avoid hardcode the resource group name and api version

packages/fx-core/templates/plugins/resource/function/bicep/function.template.bicep
```diff
@@ -0,0 +1,122 @@
...
+    M365_AUTHORITY_HOST: oauthAuthorityHost
+    {{#contains 'fx-resource-azure-sql' Plugins}}
+    SQL_DATABASE_NAME: sqlDatabaseName
+    SQL_ENDPOINT: sqlEndpoint
+    {{/contains}}
+    WEBSITE_CONTENTAZUREFILECONNECTIONSTRING: 'DefaultEndpointsProtocol=https;AccountName=${functionStorage.name};AccountKey=${listKeys(resourceId(resourceGroup().name, 'Microsoft.Storage/storageAccounts', functionStorage.name), '2019-04-01').keys[0].value};EndpointSuffix=${environment().suffixes.storage}'
+    WEBSITE_RUN_FROM_PACKAGE: '1'
+    WEBSITE_CONTENTSHARE: toLower(functionAppName)
+  }
+  dependsOn: [
```
blackchoey, line 99: maybe no need to set explicit dependencies, bicep can find the implicit dependency

packages/fx-core/templates/plugins/resource/function/bicep/input_param.template.bicep
```diff
@@ -0,0 +1,7 @@
+
+param function_serverfarmsName string = '${resourceBaseName}-function-serverfarms'
+param function_webappName string = '${resourceBaseName}-function-webapp'
+@minLength(3)
+@maxLength(24)
+@description('Name of Storage Accounts for function backend.')
```
blackchoey, line 6: What's the rule to add constraints and descriptions for parameters?

packages/fx-core/templates/plugins/resource/function/bicep/module.template.bicep
```diff
@@ -0,0 +1,24 @@
...
+  name: 'functionProvision'
+  params: {
+    functionAppName: function_webappName
+    functionServerfarmsName: function_serverfarmsName
+    functionStorageName: function_storageName
+    aadClientId: aadClientId
+    aadClientSecret: aadClientSecret
+    m365TenantId: m365TenantId
+    applicationIdUri: applicationIdUri
+    oauthAuthorityHost: m365OauthAuthorityHost
```
blackchoey, line 12: also add `m365` to the beginning of parameter?

packages/fx-core/templates/plugins/resource/function/bicep/function.template.bicep
```diff
@@ -0,0 +1,122 @@
...
+    M365_CLIENT_SECRET: aadClientSecret
+    M365_TENANT_ID: m365TenantId
+    M365_AUTHORITY_HOST: oauthAuthorityHost
+    {{#contains 'fx-resource-azure-sql' Plugins}}
+    SQL_DATABASE_NAME: sqlDatabaseName
+    SQL_ENDPOINT: sqlEndpoint
+    {{/contains}}
+    WEBSITE_CONTENTAZUREFILECONNECTIONSTRING: 'DefaultEndpointsProtocol=https;AccountName=${functionStorage.name};AccountKey=${listKeys(resourceId(resourceGroup().name, 'Microsoft.Storage/storageAccounts', functionStorage.name), '2019-04-01').keys[0].value};EndpointSuffix=${environment().suffixes.storage}'
+    WEBSITE_RUN_FROM_PACKAGE: '1'
+    WEBSITE_CONTENTSHARE: toLower(functionAppName)
```
blackchoey, line 97: There's many app settings here, can we categorize the app settings? For example: put app settings that every Function App have together, put app settings that every TeamFx project have together, then put app settings that from other resource plugins.

packages/fx-core/tests/plugins/resource/function/unit/generateArmTemplates.test.ts
```diff
@@ -0,0 +1,95 @@
...
+import { ConstantString, mockSolutionUpdateArmTemplates, ResourcePlugins } from "../../util";
+import {
+  DependentPluginInfo,
+  FunctionPluginInfo,
+} from "../../../../../src/plugins/resource/function/constants";
+import { MockContext } from "../helper";
+
+chai.use(chaiAsPromised);
+
+describe("FrontendGenerateArmTemplates", () => {
```
blackchoey, line 21: The "Frontend" should be "Function"?

packages/fx-core/src/plugins/resource/function/resources/errors.ts
xzf0587: It is fixed. Thanks.

packages/fx-core/src/common/armInterface.ts
```diff
@@ -37,3 +37,7 @@ export interface ScaffoldArmTemplateResult extends Record<string, unknown> {
   Modules?: { [moduleFileName: string]: BicepModule };
   Orchestration: BicepOrchestration;
 }
+
+export interface BicepPluginsContext {
```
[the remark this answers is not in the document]
xzf0587: This interface is to keep the keys like in bicep script the same with the  handlebars context. It could be better for all plugins' bicep script keeping the same key style.

packages/fx-core/src/plugins/resource/function/plugin.ts
```diff
@@ -607,6 +615,85 @@ export class FunctionPluginImpl {
...
+    this.collectFunctionAppSettings(ctx, site);
+
+    await runWithErrorCatchAndThrow(
+      new ConfigFunctionAppError(),
+      async () =>
+        await step(
+          StepGroup.PostProvisionStepGroup,
+          PostProvisionSteps.updateFunctionSettings,
+          async () =>
+            await webSiteManagementClient.webApps.update(resourceGroupName, functionAppName, site)
```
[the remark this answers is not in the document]
xzf0587: As some plugins do not support arm like apim, we need to keep this code.
All the setting written in bicep are removed in arm support condition.

packages/fx-core/templates/plugins/resource/function/bicep/function.template.bicep
xzf0587: sure.

packages/fx-core/templates/plugins/resource/function/bicep/function.template.bicep
xzf0587: sure

packages/fx-core/tests/plugins/resource/function/unit/generateArmTemplates.test.ts
xzf0587: sure

packages/fx-core/templates/plugins/resource/function/bicep/input_param.template.bicep
xzf0587: I will remove them to keep all parameters aligned.

packages/fx-core/templates/plugins/resource/function/bicep/function.template.bicep
xzf0587: sure

packages/fx-core/templates/plugins/resource/function/bicep/function.template.bicep
xzf0587: Yes, we enabled frontend hosting in frontend plugin.

packages/fx-core/templates/plugins/resource/function/bicep/module.template.bicep
xzf0587: sure

packages/fx-core/templates/plugins/resource/function/bicep/function.template.bicep
xzf0587: sure

KennethBWSong (approved):

PR #1922, opened 2021-08-12
feat(localdebug): add migrate v1 tab app logic in core and solution

XiaofuHuang:
Add migrate v1 tab app logic in core and solution
Include migrate V1 project questions and context settings
Disable simple auth and frontend plugins for V1 projects.

Change app studio plugin to generate manifest.source.json for V1 in another PR

Questions:
Users will be asked about these questions, like creating new project flow: 
What capability does your project include? (Single Choice: Tab, Bot, or Messaging Extension) 
Optionally provide project name when validation does not pass. 

Automatically detect the programming language 

settings.json:
image

Detail: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10576872


packages/api/review/teamsfx-api.api.md
```diff
@@ -47,6 +47,8 @@ export interface AzureSolutionSettings extends SolutionSettings {
     capabilities: string[];
     // (undocumented)
     hostType: string;
+    // (undocumented)
+    projectVersion?: string;
```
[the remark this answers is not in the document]
XiaofuHuang: Here I try to keep its behavior the same as host type and programming language.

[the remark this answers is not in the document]
kuojianlu: Make sense.

packages/fx-core/src/core/error.ts
```diff
@@ -15,6 +15,24 @@ export function ProjectFolderExistError(path: string) {
...
+    "ProjectFolderNotExistError",
+    `Path ${path} does not exist. Select a different folder.`,
+    CoreSource,
+    new Error().stack
+  );
+}
+
+export function MigrateNotImplementError(path: string) {
+  return new SystemError(
+    "MigrateNotImplement",
```
kuojianlu, line 29: MigrateNotImplement**ed** may be better?

XiaofuHuang: Thx, it has been changed.

hund030 (approved):
Approve for frontend part.

KennethBWSong (approved):

packages/fx-core/src/common/constants.ts
```diff
@@ -20,3 +20,7 @@ export class PluginDisplayName {
 export class FeatureFlagName {
   static readonly MultiEnv = "TEAMSFX_MULTI_ENV";
 }
+
+export class ProjectVersion {
+  static readonly V1 = "v1";
```
dooriya, line 25: Is the project version config only added for v1 project. For v2, shall we need to also add this config for consistency?

packages/fx-core/src/plugins/solution/fx-solution/solution.ts
```diff
@@ -165,7 +167,12 @@ export class TeamsAppSolution implements Solution {
     return ok({});
   }
 
-  fillInSolutionSettings(ctx: SolutionContext): Result<AzureSolutionSettings, FxError> {
+  assertContextNotEmpty(
+    ctx: SolutionContext
+  ): Result<
+    { answers: Inputs; projectSettings: ProjectSettings; solutionSettings: SolutionSettings },
```
dooriya, line 173: To keep consistent with the input parameter and method name, why not simply return the `SolutionContext` instead of another object? All these property won't be modified and still can be accessed from `SolutionContext`.

swatDong (commented):
The code logic looks good to me. Have some questions on the e2e flow:
1. Check with core/solution on the new setting field, to ensure no conflict.
2. Suppose user needs following steps. This PR covers the migration step itself. Better to confirm other steps to avoid redundant implementation.
  a. Trigger migration
  b. Reload migrated project
  c. F5 / local debug
  d. (?) Provision...

packages/fx-core/src/common/constants.ts
XiaofuHuang: The `projectVersion` is similar to the property `version` in `SolutionSettings`.
Rename it to `migrateFromV1`.

PS: `Version` in `SolutionSettings` means the solution plugin version.

packages/fx-core/src/plugins/solution/fx-solution/solution.ts
XiaofuHuang: Just do not want to use the non-null assertion operator. Split this function to three smaller assert functions.

packages/fx-core/src/plugins/solution/fx-solution/solution.ts
```diff
@@ -196,23 +201,26 @@ export class TeamsAppSolution implements Solution {
         )
       );
     }
-
-    return ok({
-      answers: answers,
-      projectSettings: projectSettings,
-      solutionSettings: projectSettings.solutionSettings,
-    });
+    return ok(solutionSettings);
```
hund030, line 204: Just suggestion, feel free to resolve.
```
 assertSettingsNotEmpty<T>(
    settings: T | undefined,
    key: string
  ): Result<T, FxError> {
    if (setting) {
      return err(
        returnSystemError(
          new Error(`${key} is undefined`),
          "Solution",
          SolutionError.InternelError
        )
      );
    }
    return ok(setting);
}
this.assertSettingsNotEmpty<ProjectSettings>(ctx.projectSettings, "projectSettings")
```

packages/fx-core/src/plugins/solution/fx-solution/solution.ts
dooriya: Cool.

packages/fx-core/src/plugins/solution/fx-solution/solution.ts
XiaofuHuang: Good idea! Thx

jayzhang (approved):

packages/api/src/types.ts
```diff
@@ -136,6 +136,7 @@ export interface AzureSolutionSettings extends SolutionSettings {
   capabilities: string[];
   azureResources: string[];
   activeResourcePlugins: string[];
+  projectVersion?: string;
```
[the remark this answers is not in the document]
XiaofuHuang: The` projectVersion` is similar to the property `version` in SolutionSettings.
Rename it to `migrateFromV1`.
`activate` function cannot read  `ProjectSettings`, so we use a new property.

swatDong (approved):

Issue #1882, opened 2021-08-09
Error :Could not attach to main target


sehulugh:
**Describe the bug**
I get the error 'Could not attach to main target' when I try to run the debugger in Vs Code, I typically try about 2 more times before it opens up the browser and presents the login page

**To Reproduce**
1. Update tenant information in the **env.default.json** file
2. Press F5 to start debugging
3. error message in screenshot below appears
4. Try again 1 or 2 more times and it works

**Expected behavior**
Teams App should open up in the browser after successful login

**Screenshots**
![image](https://user-images.githubusercontent.com/29524814/128744900-6815b441-c259-4488-ba8c-9858644a5f29.png)

**VS Code Extension Information (please complete the following information):**
 - OS: Windows
 - Version 2.3.2

**CLI Information (please complete the following information):**
 - OS: Windows
 - Version 2.3.2

**Additional context**

kuojianlu:
@sehulugh Thanks for reporting it. I suppose it may be an issue from vscode-js-debug. It is similar to https://github.com/microsoft/vscode-js-debug/issues/974. We will deep into it.
What you can do now:
- Add `"timeout": 50000` to the debug configuration. For example, if you F5 with **Debug (Edge)**, then add `"timeout": 50000` to `Start and Attach to Frontend (Edge)` in `.vscode/launch.json`; if you F5 with **Debug (Chrome)**, then add `"timeout": 50000` to `Start and Attach to Frontend (Chrome)`. You can make the timeout longer (here 50000 means 50 seconds).
- [Optional] add `"trace": true` to the debug configuration. In case you encounter this issue again, you can collect the log file from debug console and report to us. 
![image](https://user-images.githubusercontent.com/37978464/128966359-105e0097-d258-42dc-aec1-aee9302d75e7.png)

sehulugh:
Experiencing the issue on both Chrome and edge.
Using a Windows OS. Thanks

kuojianlu:
@sehulugh Thanks for your information. This issue has been fixed in #1926. This fix will be released next week. The fix will also take effect on the existing project.

sehulugh:
@kuojianlu Thank you

PR #1926, opened 2021-08-13
fix(local-debug): add timeout to launch browser configuration

kuojianlu:
- add timeout to launch browser configuration and set it to 20 seconds (default is 10 seconds) to mitigate "could not attach to main target" error

To fix: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10576081 or #1882

### DIFF
--- a/packages/fx-core/src/plugins/resource/localdebug/launch.ts
+++ b/packages/fx-core/src/plugins/resource/localdebug/launch.ts
@@ -27,6 +27,7 @@ export function generateConfigurations(
         group: "remote",
         order: edgeOrder,
       },
+      timeout: 20000,
     },
     {
       name: "Launch Remote (Chrome)",
@@ -37,6 +38,7 @@ export function generateConfigurations(
         group: "remote",
         order: chromeOrder,
       },
+      timeout: 20000,
     },
   ];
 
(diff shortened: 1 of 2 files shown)


packages/fx-core/src/plugins/resource/localdebug/launch.ts
```diff
@@ -27,6 +27,7 @@ export function generateConfigurations(
         group: "remote",
         order: edgeOrder,
       },
+      timeout: 20000,
```
swatDong, line 30: Could this timeout be set via debug provider?

kuojianlu: Yes, just like resolving the url in debug provider. But if so, user will not modify the timeout if it is not enough.

swatDong: Emm, so is it possible that:
- if user defines timeout in tasks.json, do nothing
- else, add `timeout: 20000` as default value in debug provider

This will affect all existing projects without any user-side modification.

kuojianlu: Nice idea. I did not consider the existing projects. Will update it later.

swatDong (approved):

PR #1930, opened 2021-08-13
fix: fix tab app cannot run correctly when deploy project without .env file

SLdragon:


packages/fx-core/src/plugins/resource/frontend/utils/environment-utils.ts
```diff
@@ -20,4 +22,50 @@ export class EnvironmentUtils {
...
+  ): Promise<void> {
+    let envs = await EnvironmentUtils.readEnvironments(envFilePath);
+
+    if (!envs) {
+      envs = {};
+    }
+
+    if (functionEnv) {
+      if (!envs[EnvironmentVariables.FuncName]) {
+        envs[EnvironmentVariables.FuncName] = functionEnv.defaultName;
```
henzhang-ms, line 40: Do we need to add these env var even .env file exists?

packages/fx-core/src/plugins/resource/frontend/plugin.ts
```diff
@@ -229,6 +233,57 @@ export class FrontendPluginImpl {
...
+    const functionPlugin = ctx.configOfOtherPlugins.get(DependentPluginInfo.FunctionPluginName);
+    if (functionPlugin) {
+      functionEnv = {
+        defaultName: functionPlugin.get(DependentPluginInfo.FunctionDefaultName) as string,
+        endpoint: functionPlugin.get(DependentPluginInfo.FunctionEndpoint) as string,
+      };
+    }
+
+    if (isArmSupportEnabled()) {
+      const endpoint = getArmOutput(ctx, ArmOutput.SimpleAuthEndpoint) as string;
```
adashen, line 250: Please confirm with Chaoyi or Di ,whether ArmOutput could be used here.

packages/fx-core/src/plugins/resource/frontend/utils/environment-utils.ts
```diff
@@ -20,4 +22,50 @@ export class EnvironmentUtils {
...
+      }
+
+      if (!envs[EnvironmentVariables.FuncEndpoint]) {
```
henzhang-ms, line 43: Nit: this kind of check-update can be wrapped into a function so we don't need these `if`s
e.g.,
```
  const addEnvOnEmpty = (key, v) => {
    if (v !== undefined && env[key] === undefined) {
      env[key] = v;
    }
  }

  addEnvOnEmpty(EnvironmentVariables.FuncName, functionEnv.defaultName);
  addEnvOnEmpty(EnvironmentVariables.FuncEndpoint, functionEnv.endpoint);
  ...
  
```

SLdragon: OK, thanks!

packages/fx-core/src/plugins/resource/frontend/utils/environment-utils.ts
SLdragon: If .env file exists, we will add needed variable that not exist, to avoid runtime error for Tab App

packages/fx-core/src/plugins/resource/frontend/plugin.ts
SLdragon: Sure

packages/fx-core/src/plugins/resource/frontend/utils/environment-utils.ts
SLdragon: Updated

packages/fx-core/src/plugins/resource/frontend/utils/environment-utils.ts
```diff
@@ -20,4 +22,43 @@ export class EnvironmentUtils {
...
     await fs.writeFile(envFile, envs);
   }
+
+  static async updateEnvironment(
+    envFilePath: string,
+    runtimeEnv: RuntimeEnvironment,
+    aadEnv: AADEnvironment,
+    functionEnv?: FunctionEnvironment
+  ): Promise<void> {
+    let envs = await EnvironmentUtils.readEnvironments(envFilePath);
```
henzhang-ms, line 32: Nit: The code looks good to me. Here is my preference style. The code looks good to me. But please ask @hund030 to confirm the entire logic because he is the owner for this plugin.
```
  const envs = (await EnvironmentUtils.readEnvironments(envFilePath)) ?? {};
  const addEnv = (key: string, v: string) => {
    if (v !== undefined && envs[key] === undefined) {
      envs[key] = v;
    }
  };

  addEnv(EnvironmentVariables.FuncName, functionEnv?.defaultName);
  addEnv(EnvironmentVariables.FuncEndpoint, functionEnv?.endpoint);
  addEnv(EnvironmentVariables.RuntimeEndpoint, runtimeEnv.endpoint);
  addEnv(EnvironmentVariables.StartLoginPage, runtimeEnv.startLoginPageUrl);
  addEnv(EnvironmentVariables.ClientID, aadEnv.clientId);

  await EnvironmentUtils.writeEnvironments(envFilePath, envs);
```

SLdragon: Thanks for your tip

packages/fx-core/src/plugins/resource/frontend/plugin.ts
dilin-MS: Please use `ctx.config.getString(xxx)` to get config.
`getArmOutput()` should only be used during provision. After provision, the arm output will be deleted and cannot be accessed. @xzf0587 is working on making `getArmOutput()` only available during provision on code level.

SLdragon: OK, thanks!

KennethBWSong (approved):

hund030 (approved):

packages/fx-core/src/plugins/resource/frontend/plugin.ts
SLdragon: Updated

hund030:
I still think the solution is somehow tricky. We set the environment variables just for react building. Perhaps we can pass these variables during react-scripts build, so that we don't need to maintain a dotenv file. Let's merge this PR for now and I will follow up this issue.

SLdragon:
> I still think the solution is somehow tricky. We set the environment variables just for react building. Perhaps we can pass these variables during react-scripts build, so that we don't need to maintain a dotenv file. Let's merge this PR for now and I will follow up this issue.

OK, however, from my perspective, pass variables during build maybe not the best solution, because this would like a black box for user, if user want to change the variable or do not use our tool to build the Tab App, user should know the details inside our toolkit.

PR #1940, opened 2021-08-16
fix: add cli env support

xiaolang124:
1. Add provision, deploy, build, publish env parameter.
2. Add user set env feature.

### DIFF
--- a/packages/cli/src/cmds/preview/preview.ts
+++ b/packages/cli/src/cmds/preview/preview.ts
@@ -86,7 +86,7 @@ export default class Preview extends YargsCommand {
       string: true,
     });
     yargs.option("env", {
-      description: "select an existed env for the project",
+      description: "select an existing env for the project",
       string: true,
     });
 
--- a/packages/cli/src/cmds/preview/preview.ts
+++ b/packages/cli/src/cmds/preview/preview.ts
@@ -85,6 +85,10 @@ export default class Preview extends YargsCommand {
       array: false,
       string: true,
     });
+    yargs.option("env", {
+      description: "select an existed env for the project",
+      string: true,
+    });
 
     return yargs.version(false);
   }
@@ -136,7 +140,7 @@ export default class Preview extends YargsCommand {
       const result =
         previewType === "local"
           ? await this.localPreview(workspaceFolder, browser)
-          : await this.remotePreview(workspaceFolder, browser);
+          : await this.remotePreview(workspaceFolder, browser, args.env as any);
       if (result.isErr()) {
         throw result.error;
       }
@@ -491,7 +495,8 @@ export default class Preview extends YargsCommand {
 
   private async remotePreview(
     workspaceFolder: string,
-    browser: constants.Browser
+    browser: constants.Browser,
+    env: string | undefined
   ): Promise<Result<null, FxError>> {
     /* === get remote teams app id === */
     const coreResult = await activate();
@@ -503,6 +508,7 @@ export default class Preview extends YargsCommand {
     const inputs: Inputs = {
       projectPath: workspaceFolder,
       platform: Platform.CLI,
+      env: env,
     };
 
     const configResult = await core.getProjectConfig(inputs);
--- a/packages/cli/src/cmds/preview/preview.ts
+++ b/packages/cli/src/cmds/preview/preview.ts
@@ -86,7 +86,7 @@ export default class Preview extends YargsCommand {
       string: true,
     });
     yargs.option("env", {
-      description: "select an existing env for the project",
+      description: "Select an existing env for the project",
       string: true,
     });
 
@@ -169,6 +169,7 @@ export default class Preview extends YargsCommand {
     const inputs: Inputs = {
       projectPath: workspaceFolder,
       platform: Platform.CLI,
+      previewType: "local",
     };
 
     let configResult = await core.getProjectConfig(inputs);
--- a/packages/cli/src/constants.ts
+++ b/packages/cli/src/constants.ts
@@ -16,6 +16,12 @@ export const RootFolderNode = new QTreeNode({
   default: "./",
 });
 
+export const EnvNode = new QTreeNode({
+  type: "text",
+  name: "env",
+  title: "Select or create an env for the project",
+});
+
 export const SubscriptionNode = new QTreeNode({
   type: "text",
   name: "subscription",
--- a/packages/cli/src/constants.ts
+++ b/packages/cli/src/constants.ts
@@ -19,13 +19,13 @@ export const RootFolderNode = new QTreeNode({
 export const EnvNode = new QTreeNode({
   type: "text",
   name: "env",
-  title: "Select or create an env for the project",
+  title: "Select or create an environment for the project",
 });
 
 export const EnvNodeNoCreate = new QTreeNode({
   type: "text",
   name: "env",
-  title: "Select an existing env for the project",
+  title: "Select an existing environment for the project",
 });
 
 export const SubscriptionNode = new QTreeNode({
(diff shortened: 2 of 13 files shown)


xiaolang124:
image
image

packages/cli/src/constants.ts
```diff
@@ -16,6 +16,12 @@ export const RootFolderNode = new QTreeNode({
   default: "./",
 });
 
+export const EnvNode = new QTreeNode({
+  type: "text",
+  name: "env",
+  title: "Select or create an env for the project",
```
swatDong, line 22: Do we allow creating new env for deploy stage?
In other words, do we need to distinguish `select only` with `select or create`?

swatDong:
Curious on multi-env for cli - do we have feature flag for cli args?

packages/cli/src/constants.ts
xiaolang124: For now, if user uses a non-existent environment, CLI will create an env and pop up an error. @qinezh what's your opinion?
image

xiaolang124:
> Curious on multi-env for cli - do we have feature flag for cli args?

Have added check env feature to cli help command.

packages/cli/src/constants.ts
qinezh: for this demo, how about throwing a friendly user error like `the specified env xxx does not exist.`

xiaolang124: Fixed.

packages/cli/src/cmds/preview/preview.ts
```diff
@@ -85,6 +85,10 @@ export default class Preview extends YargsCommand {
       array: false,
       string: true,
     });
+    yargs.option("env", {
+      description: "select an existed env for the project",
+      string: true,
+    });
```
kuojianlu, line 91: Any default value for `env`?

packages/cli/src/cmds/preview/preview.ts
```diff
@@ -136,7 +140,7 @@ export default class Preview extends YargsCommand {
       const result =
         previewType === "local"
           ? await this.localPreview(workspaceFolder, browser)
-          : await this.remotePreview(workspaceFolder, browser);
+          : await this.remotePreview(workspaceFolder, browser, args.env as any);
```
kuojianlu, line 143: What if `env` is undefined?

packages/cli/src/cmds/preview/preview.ts
xiaolang124: No default value, if the user does not specify an env he need to choose one.
image

packages/cli/src/cmds/preview/preview.ts
xiaolang124: The same as above, the user needs to choose one.

packages/cli/src/cmds/preview/preview.ts
kuojianlu: Make sense. For local preview, `env` will be ignored, right?

packages/cli/src/cmds/preview/preview.ts
```diff
@@ -85,6 +85,10 @@ export default class Preview extends YargsCommand {
       array: false,
       string: true,
     });
+    yargs.option("env", {
+      description: "select an existed env for the project",
```
kuojianlu, line 89:
```suggestion
      description: "Select an existing env for the project",
```

packages/cli/src/cmds/preview/preview.ts
xiaolang124: Yes.

packages/cli/src/cmds/preview/preview.ts
xiaolang124: Good idea. Fixed.

kuojianlu:
Could `env` be sent as telemetry? If so, you may add `env` to CLI preview telemetry property.

qinezh:
> Could `env` be sent as telemetry? If so, you may add `env` to CLI preview telemetry property.

As this is for Milestone 1 stage, let's take care of the telemetry related logic in later PR as the env name might need to be hashed.

packages/cli/src/cmds/preview/preview.ts
```diff
@@ -86,7 +86,7 @@ export default class Preview extends YargsCommand {
       string: true,
     });
     yargs.option("env", {
-      description: "select an existed env for the project",
+      description: "select an existing env for the project",
```
qinezh, line 89: Upper case for the first character to keep consistent?

packages/cli/src/helpParamGenerator.ts
```diff
@@ -19,21 +19,27 @@ import {
...
+  RootFolderNode,
+  sqlPasswordConfirmQuestionName,
+} from "./constants";
 import { NoInitializedHelpGenerator } from "./error";
 
 export class HelpParamGenerator {
   private core: FxCore;
   private questionsMap: Map<string, QTreeNode> = new Map<string, QTreeNode>();
   private initialized = false;
+  private static showEnvStage: string[] = [Stage.build, Stage.publish, Stage.deploy];
```
qinezh, line 42: we also need to support validate the application of one specified environment.

packages/cli/src/constants.ts
```diff
@@ -16,6 +16,18 @@ export const RootFolderNode = new QTreeNode({
...
```
dooriya, line 22: better use full name "environment" in the user facing string.

packages/cli/src/cmds/preview/preview.ts
kuojianlu: `env` -> `environment`?

packages/fx-core/src/core/error.ts
```diff
@@ -157,3 +157,12 @@ export function ProjectSettingsUndefinedError(): FxError {
     new Error().stack
   );
 }
+
+export function ProjectEnvNotExistError(env: string) {
+  return new UserError(
+    "ProjectEnvNotExistError",
+    `The specified env ${env} does not exist. Select an exited env.`,
```
dooriya, line 164: >exited [](http://example.com/codeflow?start=56&length=6)

typo: exited -> existing

packages/cli/src/helpParamGenerator.ts
xiaolang124: Will add later.

packages/cli/src/cmds/preview/preview.ts
xiaolang124: Fixed.

packages/cli/src/constants.ts
xiaolang124: Fixed.

packages/fx-core/src/core/error.ts
xiaolang124: Fixed.

kuojianlu (approved):

packages/fx-core/src/core/middleware/envInfoLoader.ts
```diff
@@ -200,6 +211,24 @@ async function askTargetEnvironment(
...
+  inputs: Inputs,
+  allowCreateNewEnv: boolean
+): Promise<string | undefined> {
+  const checkEnv = await environmentManager.checkEnvExist(inputs.projectPath!, inputs.env);
+  if (checkEnv.isErr()) {
+    ctx.result = checkEnv.error;
+    return undefined;
+  }
+  if (checkEnv.value || allowCreateNewEnv) {
+    return inputs.env;
```
qinezh, line 225: we might need to validate the user input env name here, as it might contain invalid characters and cannot be saved into a env profile file.

qinezh (approved):

packages/fx-core/src/core/middleware/envInfoLoader.ts
xiaolang124: Will add later.

LongOddCode (approved):

Alive-Fish (approved):
When the feature flag (multi-env) enables, the CLI e2e may be blocked, so please pay some attention to it when enable the flag later. Approved first.

PR #1944, opened 2021-08-16
refactor: move the defaultFunctionName config to project settings

dooriya:
Currently, the `defaultFunctionName` config is an output for function plugin scaffolding:
```json
    "fx-resource-function": {
        "defaultFunctionName": "getUserProfile"
    },
```

As this config is set in `env.default.json` but it is more like a project-level setting which can be shared by all environments, so we decide to move it to the project settings (in settings.json).


packages/fx-core/src/plugins/resource/frontend/plugin.ts
```diff
@@ -169,7 +169,7 @@ export class FrontendPluginImpl {
     const functionPlugin = ctx.configOfOtherPlugins.get(DependentPluginInfo.FunctionPluginName);
     if (functionPlugin) {
       functionEnv = {
-        defaultName: functionPlugin.get(DependentPluginInfo.FunctionDefaultName) as string,
+        defaultName: ctx.projectSettings?.defaultFunctionName as string,
```
hund030, line 172: I think in latest dev branch, `defaultFunctionName` is also used in frontend's pre-deploy method. Please rebase and also change that one. Thanks!

dooriya: Thanks for reminder. 
I've pulled the latest dev branch and updated the frontend plugin pre-deploy in the latest commit.

packages/fx-core/src/core/middleware/envInfoLoader.ts
```diff
@@ -125,6 +124,37 @@ export async function loadSolutionContext(
...
+  }
+
+  return [solutionConfig, projectSettings];
+}
+
+export function upgradeDefaultFunctionName(
+  solutionConfig: SolutionConfig,
+  projectSettings: ProjectSettings
+) {
+  // upgrade defaultFuncttionName if exists.
```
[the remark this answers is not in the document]
dooriya: Good catch. Will fix it.

kuojianlu (approved):

qinezh (approved):

packages/fx-core/src/plugins/resource/function/plugin.ts
```diff
@@ -226,6 +230,13 @@ export class FunctionPluginImpl {
     });
 
     if (func.method === "addResource") {
+      const functionNameQuestion: TextInputQuestion = {
+        name: QuestionKey.functionName,
+        title: InfoMessages.askFunctionName,
+        type: "text",
+        default: ctx.projectSettings?.defaultFunctionName ?? DefaultValues.functionName,
```
hund030, line 237: I didn't get the purpose of this change. If we set `default` as `projectSettings.defaultFunctionName`, toolkit always throws error telling Function already exists.
image

1yefuwang1 (approved):

packages/fx-core/src/plugins/resource/function/plugin.ts
dooriya: Toolkit will also throw error if use constant value "getUserProfile" when user already has a function named `getUSerProfile`.
I think both the constant value "getUserProfile" as well as the config value for `defaultFunctionName` doesn't make much sense in the get question name on add resource. Let's me change back to the original logic.

hund030 (approved):
Good to see this change. Thanks!

PR #1945, opened 2021-08-16
fix: auth failed in local debug and skip local debug config persistent

dooriya:
Fix in this PR:
- When local debug, retrieve the `tabDomain` and `tabEndpoint` from `localSettings` if `TEAMSFX_ARM_SUPPORT` is enabled.
- Do not persist the local debug plugin config in `env.default.json`,

### DIFF
--- a/packages/fx-core/src/plugins/resource/aad/utils/configs.ts
+++ b/packages/fx-core/src/plugins/resource/aad/utils/configs.ts
@@ -198,12 +198,15 @@ export class SetApplicationInContextConfig {
         ConfigKeysOfOtherPlugin.localDebugTabDomain
       );
     } else {
-      if (isArmSupportEnabled()) {
-        frontendDomain = getArmOutput(ctx, ConfigKeysOfOtherPlugin.frontendHostingDomainArm);
-      } else {
-        frontendDomain = ctx.configOfOtherPlugins
-          .get(Plugins.frontendHosting)
-          ?.get(ConfigKeysOfOtherPlugin.frontendHostingDomain);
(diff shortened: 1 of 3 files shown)


packages/fx-core/src/plugins/resource/aad/utils/configs.ts
```diff
@@ -191,27 +191,26 @@ export class SetApplicationInContextConfig {
   }
 
   public restoreConfigFromContext(ctx: PluginContext): void {
-    let frontendDomain: ConfigValue = ctx.config.get(ConfigKeys.domain);
-    if (frontendDomain) {
-      this.frontendDomain = format(frontendDomain as string, Formats.Domain);
+    let frontendDomain: ConfigValue;
```
KennethBWSong, line 194: Previously aad plugin allows user to add "domain" key in context since fe plugin did not allow user to customize frontend endpoint, thus aad plugin will read domain first like this:
`frontendDomain = ctx.config.get(ConfigKeys.domain)`
Since now fe plugin has fixed this problem, @adashen shall we remove this part?

1yefuwang1 (approved):

packages/fx-core/src/plugins/resource/aad/utils/configs.ts
dooriya: As local debug will no longer consume values in env.default.json, changed to the following behavior:
- if is local debug:  then read from `localSettings.json`
- if is remote: then first check the "domain" config key

KennethBWSong (approved):

PR #1956, opened 2021-08-17
perf(collaboration): add permission check in solution

SLdragon:


packages/fx-core/src/plugins/solution/fx-solution/solution.ts
```diff
@@ -1177,7 +1183,124 @@ export class TeamsAppSolution implements Solution {
 
   @hooks([ErrorHandlerMW])
   async checkPermission(ctx: SolutionContext): Promise<Result<any, FxError>> {
-    return ok(Void);
+    const canCheckPermission = this.checkWhetherSolutionIsIdle();
+    if (canCheckPermission.isErr()) {
+      return canCheckPermission;
+    }
+
+    const provisioned = this.checkWetherProvisionSucceeded(ctx.config);
```
adashen, line 1191: we still check azure permission?

packages/fx-core/src/common/permissionInterface.ts
```diff
@@ -0,0 +1,12 @@
+export interface ResourcePermission {
+  name: string;
+  type: string;
+  error: any;
+  resourceId: string | undefined;
+  roles: string[] | undefined;
+
+  // Only Azure resource contains following properties
```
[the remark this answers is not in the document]
SLdragon: updated

packages/fx-core/src/plugins/solution/fx-solution/solution.ts
SLdragon: We still need to check whether it has provisioned, because only after provision Teams App and AAD would be created

KennethBWSong (approved):

jayzhang (approved):

PR #1957, opened 2021-08-17
refactor: move icons logic from solution to appstudio

yiqing-zhao:


1yefuwang1 (commented):
How about adding a test for appstudio plugin.

yiqing-zhao:
> How about adding a test for appstudio plugin.

@1yefuwang1 Thanks. A check for the icons is added in the unit test.

nliu-ms (approved):

1yefuwang1 (approved):

PR #1961, opened 2021-08-17
refactor(identity): v2 identity plugin adapter

jayzhang:


packages/fx-core/src/plugins/resource/identity/v2/index.ts
```diff
@@ -0,0 +1,49 @@
...
+import { IdentityPlugin } from "..";
+import {
+  ResourcePlugins,
+  ResourcePluginsV2,
+} from "../../../solution/fx-solution/ResourcePluginContainer";
+import { provisionResourceAdapter } from "../../utils4v2";
+
+@Service(ResourcePluginsV2.IdentityPlugin)
+export class IdentityPluginV2 implements ResourcePlugin {
+  name = "fx-resource-identity";
```
xzf0587, line 27: this name can refer to constants class.

xzf0587 (approved):

PR #1964, opened 2021-08-18
perf(collaboration): add check permission for aad plugin

KennethBWSong:


packages/fx-core/src/plugins/resource/aad/constants.ts
```diff
@@ -20,6 +20,16 @@ export class Constants {
   static statusCodeUserError = 400;
   static statusCodeServerError = 500;
 
+  static createOwnerDuplicatedMessage =
+    "One or more added object references already exist for the following modified properties: 'owners'.";
```
SLdragon, line 24: This error message is very technical, we'd better use another words to general users such "xxx has already be added to the owner list"

KennethBWSong: This error message is from Graph, and is used to match the scenario that we add a user who is already owner. Should be added when adding grant api. I will remove this line.

packages/fx-core/src/plugins/resource/aad/graph.ts
```diff
@@ -111,6 +111,34 @@ export namespace GraphClient {
...
+    }
+
+    if (!userObjectId) {
+      throw new Error(
+        `${GraphClientErrorMessage.CheckPermissionFailed}: ${GraphClientErrorMessage.UserObjectIdIsNull}.`
+      );
+    }
+
+    const instance = initAxiosInstance(graphToken);
+    const response = await instance.get(
```
SLdragon, line 132: Do we need to add try catch block in case of network error and other unknown errors?

KennethBWSong: All the errors here will be caught in AadAppClient.ts and throw as user/system error. We don't need try catch here.

packages/fx-core/src/plugins/resource/aad/plugin.ts
```diff
@@ -321,6 +326,34 @@ export class AadAppForTeamsImpl {
...
+
+    const result = [
+      {
+        name: Constants.permissions.name,
+        type: Constants.permissions.type,
+        roles: isAadOwner ? [Constants.permissions.owner] : [Constants.permissions.noPermission],
+        resourceId: config.objectId!,
+      },
+    ];
+    Utils.addLogAndTelemetry(ctx.logProvider, Messages.EndCheckPermission);
```
SLdragon, line 353: where we send the error telemetry?

packages/fx-core/src/plugins/resource/aad/graph.ts
SLdragon: OK

packages/fx-core/src/plugins/resource/aad/constants.ts
SLdragon: Sure

packages/fx-core/src/plugins/resource/aad/plugin.ts
KennethBWSong: Error telemetry will be sent in index.ts. 
For other info that needs to add to telemetry, will update necessary info in telemetry in another pr after we have confirmed required info with yanliang.

SLdragon (approved):

adashen (approved):

PR #1977, opened 2021-08-19
chore: check local cache for treatment variable

HuihuiWu-Microsoft:


HuihuiWu-Microsoft:
Will create another PR

PR #1978, opened 2021-08-19
feat(localdebug): add migrate v1 project command into vsc

XiaofuHuang:
Add migrate v1 project command into vsc

### DIFF
--- a/packages/vscode-extension/src/handlers.ts
+++ b/packages/vscode-extension/src/handlers.ts
@@ -40,6 +40,7 @@ import {
   globalStateGet,
   Correlator,
   getAppDirectory,
+  isV1Project,
 } from "@microsoft/teamsfx-core";
 import GraphManagerInstance from "./commonlib/graphLogin";
 import AzureAccountManager from "./commonlib/azureLogin";
@@ -84,6 +85,7 @@ import * as path from "path";
 import { exp } from "./exp/index";
 import { TreatmentVariables } from "./exp/treatmentVariables";
 import { StringContext } from "./utils/stringContext";
+import { ext } from "./extensionVariables";
 
 export let core: FxCore;
 export let tools: Tools;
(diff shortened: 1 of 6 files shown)


packages/vscode-extension/src/handlers.ts
```diff
@@ -176,6 +176,14 @@ export async function createNewProjectHandler(args?: any[]): Promise<Result<null
   return await runCommand(Stage.create);
 }
 
+export async function migrateV1ProjectHandler(args?: any[]): Promise<Result<null, FxError>> {
+  ExtTelemetry.sendTelemetryEvent(
+    TelemetryEvent.MigrateV1ProjectStart,
+    getTriggerFromProperty(args)
+  );
+  return await runCommand(Stage.migrateV1);
```
kuojianlu, line 184: What if user trigger this command from a V2 project?

swatDong: An option may be - make this command a hidden one and can only be triggered by our detection.

XiaofuHuang: This command may be also used in a react app. 
We need to prohibit users from migrating v2 projects in both cli and vsc.
Would you like to throw an error message when the app is a v2 app (exist a `.fx` folder)?  It can be implemented in core.

kuojianlu: I think it is fine. Maybe need to confirm with Junjie.

1openwindow: we may can add a "custom when clause context" for this command. this command only available when open v1 project.

XiaofuHuang: The command is hidden in vscode if it is not a v1 project.

The same v1 project validation will be added into core in another PR.

kuojianlu (approved):

LongOddCode (approved):

swatDong (approved):

PR #1979, opened 2021-08-19
perf(collaboration): add check permission for app studio

KennethBWSong:


packages/fx-core/src/plugins/resource/appstudio/index.ts
```diff
@@ -278,6 +278,25 @@ export class AppStudioPlugin implements Plugin {
     return ok(localTeamsAppId);
   }
 
+  public async checkPermission(ctx: PluginContext): Promise<Result<any, FxError>> {
+    TelemetryUtils.init(ctx);
+    TelemetryUtils.sendStartEvent(TelemetryEventName.checkPermission);
+
+    try {
+      const checkPermissionResult = await this.appStudioPluginImpl.checkPermission(ctx);
+      TelemetryUtils.sendSuccessEvent(TelemetryEventName.checkPermission);
```
SLdragon, line 287: Do we need to add some properties for the telemetry?

KennethBWSong: As commented in aad plugin, telemetry will be added after we have conclusion with yanliang in another pr. Here I only add least properties.

packages/fx-core/src/plugins/resource/appstudio/appStudio.ts
```diff
@@ -135,9 +136,13 @@ export namespace AppStudioClient {
           `Cannot get the app definition with app ID ${teamsAppId}, due to ${e.name}: ${e.message}`
         );
       }
-      throw new Error(
+      const err = new Error(
         `Cannot get the app definition with app ID ${teamsAppId}, due to ${e.name}: ${e.message}`
       );
+      if (e.response?.data?.status) {
+        err.name = e.response?.data?.status;
```
SLdragon, line 143: Use name is confusing, how about use statusCode since error is any type?

KennethBWSong: Error type only have name and message property. I am also considering whether it is better to add a new interface here.

nliu-ms (approved):

adashen (approved):

PR #1982, opened 2021-08-19
fix: save teams app id when provisioning failed for other resources

nliu-ms:
Save remote teams app id into config file, even if other resources provision failed. (like app service plan quota)

Otherwise, the toolkit will create multiple teams app when provision again.


1yefuwang1 (commented):
Please add a UT case covering this change.

jayzhang (approved):

PR #1991, opened 2021-08-19
fix(frontend): write tabs/.env only in deploy stage

hund030:
In multi environment scenario, tabs/.env may use the dirty value from other environment. Overwrite it every time in deploy stage.


henzhang-ms (approved):
lgtm

PR #2002, opened 2021-08-20
feat(appstudio): add migrate v1 to app studio plugin

XiaofuHuang:
Generate `manifest.source.json` when migrate v1 project.

If v1 project has `manifest.json`, add some placeholder to generate `manifest.source.json`.
If v1 project does not have `manifest.json`,  scaffold `manifest.source.json`.


packages/fx-core/src/common/tools.ts
```diff
@@ -451,3 +451,19 @@ export function getAppStudioEndpoint(): string {
...
+  excludeFileNames?: string[]
+): Promise<void> {
+  await fs.ensureDir(distPath);
+
+  const fileNames = await fs.readdir(srcPath);
+  for (const fileName of fileNames) {
+    if (excludeFileNames && excludeFileNames.length > 0 && excludeFileNames.includes(fileName)) {
+      continue;
+    }
+    await fs.move(path.join(srcPath, fileName), path.join(distPath, fileName), { overwrite: true });
```
kuojianlu, line 467: `fs.copy`?

packages/fx-core/src/plugins/resource/appstudio/plugin.ts
```diff
@@ -309,6 +351,22 @@ export class AppStudioPluginImpl {
...
   }
 
+  public async migrateV1Project(ctx: PluginContext): Promise<any> {
+    let manifest: TeamsAppManifest | undefined;
+    const archiveAppPackageFolder = path.join(ctx.root, ArchiveFolderName, AppPackageFolderName);
+    const archiveManifestPath = path.join(archiveAppPackageFolder, V1_MANIFEST);
+    const newAppPackageFolder = path.join(ctx.root, AppPackageFolderName);
+    await fs.ensureDir(newAppPackageFolder);
+    if (await this.checkFileExist(archiveManifestPath)) {
+      await copyFiles(archiveAppPackageFolder, newAppPackageFolder, [V1_MANIFEST]);
```
kuojianlu, line 361: - Whose responsibility to copy V1 `appPackage` folder to `.archive`?
- There may be `appPackage.zip` and `README.md` under V1 `appPackage` folder, is it our expectation to also copy these two files?

packages/api/src/constants.ts
```diff
@@ -4,6 +4,8 @@
 
 export const ConfigFolderName = "fx";
 export const AppPackageFolderName = "appPackage";
+export const ArchiveFolderName = ".archive";
+export const ArchiveLogFileName = ".archive.log";
```
kuojianlu, line 8: What is this log file for? Seems not used in this PR.

packages/fx-core/src/common/tools.ts
XiaofuHuang: Sorry, fixed.

packages/api/src/constants.ts
XiaofuHuang: Sorry, the file is used by two prs at the same time. The log file used in the archive step.

packages/fx-core/src/plugins/resource/appstudio/plugin.ts
XiaofuHuang: Good idea! We should exclude more files in `appPackage` folder. Maybe allowlist is a better way. I'm only worried about the changes of the run icon file name.


PS: There is another PR to archive the existing files in `core`. Here only contains the app studio part.

XiaofuHuang: Or we can exclude files according to the file extension

kuojianlu: I think we can simply exclude `appPackage.zip` and `README.md` as well since they are generated by V1. If user added some files in V1 `appPackage` folder, these files are out of our control, just let them here.

kuojianlu (approved):

packages/fx-core/src/plugins/resource/appstudio/plugin.ts
```diff
@@ -187,16 +188,51 @@ export class AppStudioPluginImpl {
       if (solutionSettings.capabilities.includes(MessageExtensionItem.id)) {
         manifest.composeExtensions = COMPOSE_EXTENSIONS_TPL;
       }
+
+      if (settings?.solutionSettings?.migrateFromV1) {
```
swatDong, line 192: How is it possible a migrated project calls `createManifest`? Since appstudio plugin is inactive for migrated project.

nliu-ms (approved):

nliu-ms:
placeholders may be updated to mustache template later.

packages/fx-core/src/plugins/resource/appstudio/plugin.ts
XiaofuHuang: App studio plugin, local debug plugin and frontend hosting plugin will be active for a migrated project.

XiaofuHuang: For the v1 project created before v1.2.0, there is no manifest.json file in the project. so we will the same logic to create a new one.

swatDong (approved):

PR #2003, opened 2021-08-20
perf(collaboration): add grant permission in solution

SLdragon:


packages/fx-core/src/plugins/solution/fx-solution/solution.ts
```diff
@@ -1331,6 +1303,68 @@ export class TeamsAppSolution implements Solution {
...
+        ([plugin, context]) => {
+          return [plugin?.grantPermission?.bind(plugin), context, plugin.name];
+        }
+      );
+
+      const results = await executeConcurrently("", grantPermissionWithCtx);
+      const permissions: ResourcePermission[] = [];
+      for (const result of results) {
+        if (result.isErr()) {
+          return result;
```
KennethBWSong, line 1316: Here shall we just return? According to yanliang's doc, we need to print some info.

KennethBWSong (approved):

PR #2004, opened 2021-08-20
refactor(function): adapt v2 api

hund030:
V2 api treats `addResource` as customizedTask of plugin. Function plugin needs to implement the `executeUserTask` for addResource.

### DIFF
--- a/packages/fx-core/src/plugins/resource/function/index.ts
+++ b/packages/fx-core/src/plugins/resource/function/index.ts
@@ -82,7 +82,7 @@ export class FunctionPlugin implements Plugin {
   public async executeUserTask(func: Func, ctx: PluginContext): Promise<FxResult> {
     this.setContext(ctx);
     const res = await this.runWithErrorWrapper(ctx, FunctionEvent.executeUserTask, () =>
-      Promise.resolve(this.functionPluginImpl.executeUserTask(func, ctx))
+      this.functionPluginImpl.executeUserTask(func, ctx)
     );
     return res;
   }
--- a/packages/fx-core/src/plugins/resource/function/index.ts
+++ b/packages/fx-core/src/plugins/resource/function/index.ts
@@ -37,6 +37,7 @@ import {
 import { Service } from "typedi";
 import { ResourcePlugins } from "../../solution/fx-solution/ResourcePluginContainer";
 import { isArmSupportEnabled } from "../../..";
+
 @Service(ResourcePlugins.FunctionPlugin)
 export class FunctionPlugin implements Plugin {
   name = "fx-resource-function";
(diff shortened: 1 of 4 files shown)


packages/fx-core/src/plugins/resource/function/index.ts
```diff
@@ -78,6 +79,14 @@ export class FunctionPlugin implements Plugin {
     return res;
   }
 
+  public async executeUserTask(func: Func, ctx: PluginContext): Promise<FxResult> {
+    this.setContext(ctx);
+    const res = await this.runWithErrorWrapper(ctx, FunctionEvent.executeUserTask, () =>
+      Promise.resolve(this.functionPluginImpl.executeUserTask(func, ctx))
```
hund030, line 85: Send telemetry here. We can get the operation result by union telemetry of  `executeUserTask` and `addResource`.

jayzhang (approved):

packages/fx-core/src/plugins/resource/function/index.ts
henzhang-ms: why add `Promise.resolve` here

hund030: Removed.

packages/fx-core/src/plugins/resource/function/plugin.ts
```diff
@@ -273,6 +274,16 @@ export class FunctionPluginImpl {
     return ResultFactory.Success(res);
   }
 
+  public async executeUserTask(func: Func, ctx: PluginContext): Promise<FxResult> {
+    if (func.method === CustomizedTask.addResource) {
+      TelemetryHelper.sendGeneralEvent(FunctionEvent.addResource);
+      await this.preScaffold(ctx);
```
henzhang-ms, line 280: We'd better to check the ret value according to our design. I know it is redundant for current logic, but we need to code for principle.

hund030: Agree. Updated.

packages/fx-core/src/plugins/resource/function/plugin.ts
```diff
@@ -273,6 +274,18 @@ export class FunctionPluginImpl {
     return ResultFactory.Success(res);
   }
 
+  public async executeUserTask(func: Func, ctx: PluginContext): Promise<FxResult> {
+    let result = ResultFactory.Success();
+
+    if (func.method === CustomizedTask.addResource) {
+      TelemetryHelper.sendGeneralEvent(FunctionEvent.addResource);
+      const _res = await this.preScaffold(ctx);
```
henzhang-ms, line 282: Nit: we can use  a unified pattern to simulate monad.
```
result = result.isOK() ? (await this.preScaffold(ctx)) : result;
result = result.isOK() ? (await this.scaffold(ctx)) : result;

```

henzhang-ms (approved):

henzhang-ms (approved):
good job.

PR #2011, opened 2021-08-20
chore: fix daily vs-code-extension tag name without @ char

wenytang-ms:


wenytang-ms:
fix vscode-extension tag 
alpha should not contains @ in tag
rc and stable no problem

LongOddCode (approved):

PR #2012, opened 2021-08-20
chore: check cache before fetch treatment variable and update string

HuihuiWu-Microsoft:


packages/vscode-extension/src/resources/Strings.json
```diff
@@ -41,7 +41,7 @@
       "buildPackageTitleNew": "Zip Teams metadata package",
       "buildPackageDescription": "Build your Teams app into a package for publishing",
       "provisionTitle": "Provision in the Cloud",
-      "provisionTitleNew": "Provision in the cloud",
+      "provisionTitleNew": "Provision in the Cloud",
```
HuihuiWu-Microsoft, line 44: I found the legacy version is 'Cloud'

1openwindow (approved):

PR #2015, opened 2021-08-23
feat(core): add file operations into core and solution

XiaofuHuang:
1. Reject v2 projects or projects that have been archived to run the migrate command
2. Archive the existing v1 project to `.archive` folder 
3. Use `executeUserTask` instead of `scaffold` in plugins.
4. Generate the basic fold structure


packages/fx-core/src/core/index.ts
```diff
@@ -277,6 +281,40 @@ export class FxCore implements Core {
...
+      await fs.appendFile(
+        path.join(projectPath, ArchiveLogFileName),
+        `[${new Date().toISOString()}] ${message}\n`
+      );
+    };
+
+    await archiveLog(projectPath, `Start to move files into '${ArchiveFolderName}' folder.`);
+    for (const fileName of fileNames) {
+      if (fileName === ArchiveFolderName || fileName === ArchiveLogFileName) {
+        continue;
```
kuojianlu, line 299: Exclude `node_modules`?

packages/fx-core/src/core/index.ts
```diff
@@ -277,6 +281,40 @@ export class FxCore implements Core {
...
+      }
+
+      try {
+        await fs.move(path.join(projectPath, fileName), path.join(archiveFolderPath, fileName), {
+          overwrite: true,
+        });
+      } catch (e: any) {
+        await archiveLog(projectPath, `Failed to move '${fileName}'. ${e.message}`);
+        throw e;
```
kuojianlu, line 308: Just a consideration, should we make archiving an atomic operation?

kuojianlu:
If user selects a project with auth to migrate, what is our expected behavior?

packages/fx-core/src/core/index.ts
XiaofuHuang: We move `node_modules` into `.archive` folder, but we do not copy `node_modules` into the new project

packages/fx-core/src/core/index.ts
XiaofuHuang: The first design is to copy all the files into `.archive` folder, then delete the files in the root folder. In this way the operation can be atomic (using rollback).
But copy operation takes a lot of time(node_modules more than 5 minutes), and move operation is fast (less than 1 second).
So we use a log file to let the user known what happened.

kuojianlu: Make sense.

packages/fx-core/src/core/index.ts
kuojianlu: Not sure if `node_modules` should be moved to `.archive` folder, but I think it is OK since `move` operation is very fast.

XiaofuHuang:
> If user selects a project with auth to migrate, what is our expected behavior?

The behavior is same to the tab app without SSO. User should manually start the auth service and config the AAD.

packages/fx-core/src/core/index.ts
```diff
@@ -730,7 +768,7 @@ export class FxCore implements Core {
       );
       await fs.writeFile(
         path.join(inputs.projectPath!, `.gitignore`),
-        `node_modules\n/.${ConfigFolderName}/*.env\n/.${ConfigFolderName}/*.userdata\n.DS_Store`
+        `node_modules\n/.${ConfigFolderName}/*.env\n/.${ConfigFolderName}/*.userdata\n.DS_Store\n${ArchiveFolderName}`
```
kuojianlu, line 771: `ArchiveLogFileName` as well?

swatDong (approved):

packages/fx-core/src/core/index.ts
XiaofuHuang: Good idea! Thx!

jayzhang (approved):

PR #2016, opened 2021-08-23
perf(collaboration): list collaborator interface, and internal API changes

SLdragon:


KennethBWSong (approved):

LongOddCode (approved):

packages/fx-core/src/plugins/resource/appstudio/plugin.ts
```diff
@@ -592,6 +592,10 @@ export class AppStudioPluginImpl {
     return result;
   }
 
+  public async listCollaborator(ctx: PluginContext): Promise<Result<TeamsAppAdmin[], FxError>> {
+    return ok([]);
```
nliu-ms, line 596: will you update this later?

KennethBWSong: I will implement this API later after design doc is ready. We expect to finish the implement in this week.

SLdragon: Yes, the detailed implementation will in another pr

nliu-ms (approved):

jayzhang (approved):

PR #2019, opened 2021-08-23
perf(collaboration): add grant permission in aad plugin

KennethBWSong:


packages/fx-core/src/plugins/resource/aad/aadAppClient.ts
```diff
@@ -228,6 +229,29 @@ export class AadAppClient {
...
+    ctx: PluginContext,
+    stage: string,
+    objectId: string,
+    userObjectId: string
+  ): Promise<void> {
+    try {
+      await GraphClient.grantPermission(TokenProvider.token as string, objectId, userObjectId);
+    } catch (error) {
+      if (error?.response?.data?.error.message == Constants.createOwnerDuplicatedMessage) {
+        return;
```
SLdragon, line 242: We can add logs here to inform user that the owner has already been added

packages/fx-core/src/plugins/resource/aad/errors.ts
```diff
@@ -160,6 +165,7 @@ export class GraphClientErrorMessage {
     "Failed to create an application secret in Azure Active Directory.";
   static readonly GetFailed = "Failed to retrieve Azure Active Directory application registration.";
   static readonly CheckPermissionFailed = "Failed to check permission in Azure Active Directory.";
+  static readonly GrantPermissionFailed = "Failed to check permission in Azure Active Directory.";
```
SLdragon, line 168: typo: to **grant** permission

packages/fx-core/src/plugins/resource/aad/aadAppClient.ts
KennethBWSong: Fixed. Thx!

packages/fx-core/src/plugins/resource/aad/errors.ts
KennethBWSong: Fixed. Thx!

SLdragon (approved):

PR #2021, opened 2021-08-23
perf(collaboration): add permission grant in app studio

KennethBWSong:


nliu-ms (approved):

packages/fx-core/src/plugins/resource/appstudio/appStudio.ts
```diff
@@ -411,4 +411,26 @@ export namespace AppStudioClient {
...
+    newUser: IUserList
+  ): Promise<void> {
+    let app;
+    try {
+      app = await getApp(teamsAppId, appStudioToken);
+    } catch (error) {
+      throw error;
+    }
+
+    const findUser = app.userList?.findIndex((user: any) => user["aadId"] === newUser.aadId);
```
SLdragon, line 427: user :any can be user: IUserList

KennethBWSong: Fixed. Thx!

SLdragon (approved):

PR #2025, opened 2021-08-24
fix: add env ui 1

xiaolang124:
1. Add a basic env UI component.
2. Only show env UI when the user has set environment variable set.
3. TODO: core update, UI update.

image
image


packages/vscode-extension/src/commandsTreeViewProvider.ts
```diff
@@ -29,6 +29,11 @@ class TreeViewManager {
     const accountProvider = new CommandsTreeViewProvider([]);
     disposables.push(vscode.window.registerTreeDataProvider("teamsfx-accounts", accountProvider));
 
+    const environmentProvider = new CommandsTreeViewProvider([]);
```
HuihuiWu-Microsoft, line 32: Does this view show for non-teamsfx project?

qinezh:
Please don't use `feat` in the PR title, as we're not going to release it in this sprint.

packages/vscode-extension/src/commandsTreeViewProvider.ts
xiaolang124: No, already add check teamsfx project check logic, only show env in teamsfx project.

xiaolang124:
> Please don't use `feat` in the PR title, as we're not going to release it in this sprint.

Change to fix.

packages/vscode-extension/src/utils/commonUtils.ts
```diff
@@ -134,3 +137,16 @@ export function anonymizeFilePaths(stack?: string): string {
 
   return updatedStack;
 }
+
+export async function isTeamsfx(): Promise<boolean> {
+  if (workspace.workspaceFolders && workspace.workspaceFolders.length > 0) {
+    const workspaceFolder = workspace.workspaceFolders[0];
+    const fxPath = await commonUtils.getProjectRoot(workspaceFolder.uri.fsPath, ".fx");
+    if (fxPath && fs.pathExistsSync(fxPath)) {
```
qinezh, line 145: better to use the async method to check if path exists?

packages/vscode-extension/src/envTree.ts
```diff
@@ -0,0 +1,31 @@
+import { FxError, Result, err, ok, Void, TreeCategory } from "@microsoft/teamsfx-api";
```
qinezh, line 1: add copyright info?

packages/vscode-extension/src/utils/commonUtils.ts
```diff
@@ -134,3 +137,16 @@ export function anonymizeFilePaths(stack?: string): string {
 
   return updatedStack;
 }
+
+export async function isTeamsfx(): Promise<boolean> {
```
qinezh, line 141: Looks `debug/commonUtils` already has similar method `isFxProject`, can we reuse it?

xiaolang124: ```debug/commonUtils``` needs path as a parameter. If I use this method, also need to add a method to handle path logic.

packages/vscode-extension/src/envTree.ts
xiaolang124: Fixed.

packages/vscode-extension/src/utils/commonUtils.ts
xiaolang124: Fixed.

packages/vscode-extension/src/utils/commonUtils.ts
```diff
@@ -135,3 +137,16 @@ export function anonymizeFilePaths(stack?: string): string {
...
+  if (workspace.workspaceFolders && workspace.workspaceFolders.length > 0) {
+    const workspaceFolder = workspace.workspaceFolders[0];
+    const fxPath = await commonUtils.getProjectRoot(workspaceFolder.uri.fsPath, ".fx");
```
qinezh, line 144: Keep consistent to use the const variable instead of hard code here?
```suggestion
    const fxPath = await commonUtils.getProjectRoot(workspaceFolder.uri.fsPath, `.${ConfigFolderName}`);
```

qinezh: looks we can reuse `isFxProject` here like: 

```javascript
return await isFxProject(workspaceFolder.uri.fsPath);
```

xiaolang124: Fixed.

qinezh (approved):

HuihuiWu-Microsoft (approved):

LongOddCode (approved):

PR #2026, opened 2021-08-24
fix: users can re-provision resource

nliu-ms:
1) Provision resources
2) click provision again, will check resources and show success message
3) Delete resource group
4) click provision again, resources will be re-provisioned.

### DIFF
--- a/packages/fx-core/src/plugins/solution/fx-solution/solution.ts
+++ b/packages/fx-core/src/plugins/solution/fx-solution/solution.ts
@@ -583,22 +583,6 @@ export class TeamsAppSolution implements Solution {
       return canProvision;
     }
 
-    const provisioned = this.checkWetherProvisionSucceeded(ctx.config);
-    if (provisioned) {
-      const msg = util.format(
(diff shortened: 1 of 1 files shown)


packages/fx-core/src/plugins/solution/fx-solution/solution.ts
```diff
@@ -583,22 +583,6 @@ export class TeamsAppSolution implements Solution {
       return canProvision;
     }
 
-    const provisioned = this.checkWetherProvisionSucceeded(ctx.config);
```
1yefuwang1, line 586: Simply removing state checking here doesn't seem to deliver "click provision again, will check resources and show success message"

1yefuwang1 (approved):

packages/fx-core/src/plugins/solution/fx-solution/solution.ts
nliu-ms: we will have a provision success message instead.

PR #2027, opened 2021-08-24
feat(vsc): hide login in case of spfx project

1openwindow:
ADO: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10629499
hidelogin

### DIFF
--- a/packages/vscode-extension/src/accountTree.ts
+++ b/packages/vscode-extension/src/accountTree.ts
@@ -2,7 +2,6 @@
 // Licensed under the MIT license.
 
 import {
-  AzureSolutionSettings,
   err,
   FxError,
   ok,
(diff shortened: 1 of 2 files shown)


packages/vscode-extension/src/accountTree.ts
```diff
@@ -316,41 +316,49 @@ export async function registerAccountTreeHandler(): Promise<Result<Void, FxError
...
+
+  const specifySubscriptionTreeItem: TreeItem = {
+    commandId: "fx-extension.specifySubscription",
+    label: StringResources.vsc.accountTree.specifySubscription,
+    callback: selectSubscriptionCallback,
+    parent: undefined,
+  };
+
+  const solutionSettings = await getAzureSolutionSettings();
+  if (solutionSettings && "Azure" === solutionSettings.hostType) {
```
HuihuiWu-Microsoft, line 357: I think you might need to add this logic to setStatusChangeMap to only listen to the azure account change for azure project

HuihuiWu-Microsoft (approved):

PR #2033, opened 2021-08-24
perf(arm support): bot post provision arm support

dilin-MS:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10645220

1. Update bot plugin post-provision for arm support scenario.
2. Fix error in bot bicep files.


packages/fx-core/src/plugins/resource/bot/plugin.ts
```diff
@@ -220,15 +224,21 @@ export class TeamsBotImpl {
     // 1. Do bot registration.
     if (this.config.scaffold.wayToRegisterBot === WayToRegisterBot.CreateNew) {
       await handler?.next(ProgressBarConstants.PROVISION_STEP_BOT_REG);
-      await this.createNewBotRegistrationOnAzure();
+      const botAuthCreds = await this.createOrGetBotAppRegistration();
+
+      if (!isArmSupportEnabled()) {
```
xzf0587, line 229: the provision function will be skipped when arm enabled. There should be no arm enabled condition in it.

packages/fx-core/src/plugins/resource/bot/plugin.ts
```diff
@@ -220,15 +224,21 @@ export class TeamsBotImpl {
...
+        await this.provisionBotServiceOnAzure(botAuthCreds);
+      }
     }
 
-    await handler?.next(ProgressBarConstants.PROVISION_STEP_WEB_APP);
-    // 2. Provision azure web app for hosting bot project.
-    await this.provisionWebApp();
+    if (!isArmSupportEnabled()) {
```
xzf0587, line 234: same as above

packages/fx-core/templates/plugins/resource/bot/bicep/param.template.bicep
```diff
@@ -8,4 +8,4 @@ param bot_webAppSKU string = 'F1'
 param bot_serviceSKU string = 'F1'
 param bot_sitesName string = '${resourceBaseName}-bot-sites'
 param bot_displayName string = '${resourceBaseName}-bot-displayname'
-param authLoginUriSuffix string = 'auth-start.html'
+param authLoginUriSuffix string = 'public'
```
blackchoey, line 11: why we change this value? based on existing bicep file the login url is `var initiateLoginEndpoint = uri(botEndpoint, authLoginUriSuffix)`, seems not the expected one after changed the value?

packages/fx-core/src/plugins/resource/bot/plugin.ts
dilin-MS: Why? The provision function will not be skipped.

packages/fx-core/src/plugins/resource/bot/plugin.ts
dilin-MS: Pls see the above reply.

packages/fx-core/templates/plugins/resource/bot/bicep/param.template.bicep
dilin-MS: The value is supposed to be 'public' instead of 'auth-start.html' here so fix the bot bicep error here.
See the non-arm-support implementation [here](https://github.com/OfficeDev/TeamsFx/blob/dev/packages/fx-core/src/plugins/resource/bot/plugin.ts#L349).
The expected redirect value is something like `"redirectUri": "https://tabandbotbt47702d.azurewebsites.net/public"`.

blackchoey: but it misses the auth-start.html part... can we see what's the magics behind this?

packages/fx-core/src/plugins/resource/bot/plugin.ts
xzf0587: Provision function is logically used to provision azure resource and handle related work based on the returned result of  provision. 
When arm is enabled, it will not provision azure resource. Related work also has no returned provision result to go on.

dilin-MS: Bot plugin needs to provision bot aad app so we need to save the aad registration logic in provision stage.

packages/fx-core/src/plugins/resource/bot/plugin.ts
dilin-MS: Same as above.

packages/fx-core/templates/plugins/resource/bot/bicep/param.template.bicep
dilin-MS: Fixed.

xzf0587 (approved):

packages/fx-core/src/plugins/resource/bot/plugin.ts
```diff
@@ -220,15 +224,21 @@ export class TeamsBotImpl {
...
+      await handler?.next(ProgressBarConstants.PROVISION_STEP_WEB_APP);
```
YitongFeng-git, line 235: Can this if 234 combined with L229's if

packages/fx-core/src/plugins/resource/bot/plugin.ts
```diff
@@ -174,20 +176,23 @@ export class TeamsBotImpl {
...
-    CheckThrowSomethingMissing(CommonStrings.SHORT_APP_NAME, this.ctx.projectSettings?.appName);
 
-    if (!this.config.provision.siteName) {
-      this.config.provision.siteName = ResourceNameFactory.createCommonName(
-        this.config.resourceNameSuffix,
-        this.ctx.projectSettings?.appName,
-        MaxLengths.WEB_APP_SITE_NAME
-      );
-      Logger.debug(`Site name generated to use is ${this.config.provision.siteName}.`);
+    if (!isArmSupportEnabled()) {
```
YitongFeng-git, line 180: Nit: If the if( ! isArmSupportEnabled()) is the main logic, it's better to use if(isArmSupportEnabled()) when needed.

packages/fx-core/templates/plugins/resource/bot/bicep/bot.template.bicep
```diff
@@ -122,5 +122,4 @@ output botDomain string = botWebAppHostname // validDomain
 output appServicePlanName string = botServerfarmsName // appServicePlan
 output botServiceName string = botServiceName // botChannelReg
 output botWebAppEndpoint string = botEndpoint // siteEndpoint
-output initiateLoginEndpoint string = initiateLoginEndpoint // redirectUri
```
YitongFeng-git, line 125: why deleted?

packages/fx-core/src/plugins/resource/bot/plugin.ts
dilin-MS: The scenario is that when provisioning bot plugin azure resources. 
1. arm support enabled: 
    - provision bot aad app.
2. no arm support enabled:
    - provision bot aad app.
    - provision azure resources (bot service, web app, etc.)
    
 Since both scenario need to do the "provision bot aad app" step and only non-arm situation needs to provision azure resources, we have the `if( ! isArmSupportEnabled())` to see whether provision azure resources here.

packages/fx-core/src/plugins/resource/bot/plugin.ts
dilin-MS: No. You can see the previous if-clause is inside the `if (this.config.scaffold.wayToRegisterBot === WayToRegisterBot.CreateNew)`

packages/fx-core/templates/plugins/resource/bot/bicep/bot.template.bicep
dilin-MS: This is the value for the redirect uri. Since we do not need this value, remove the output.

YitongFeng-git (approved):

PR #2034, opened 2021-08-24
fix(bot): unhandled error when scaffold

YitongFeng-git:


packages/fx-core/src/plugins/resource/bot/plugin.ts
```diff
@@ -155,8 +156,12 @@ export class TeamsBotImpl {
       group_name
     );
 
-    await handler?.next(ProgressBarConstants.SCAFFOLD_STEP_UNZIP);
-    zipContent.extractAllTo(this.config.scaffold.workingDir!, true);
+    try {
+      await handler?.next(ProgressBarConstants.SCAFFOLD_STEP_UNZIP);
+      zipContent.extractAllTo(this.config.scaffold.workingDir!, true);
+    } catch (err) {
+      throw new InvalidPathError(this.config.scaffold.workingDir!, err);
```
hund030, line 163: We can't say that all errors are caused by invalid path, maybe just ExtractTemplateError is enough. And we can suggest user to check whether path valid.

packages/fx-core/src/plugins/resource/bot/plugin.ts
```diff
@@ -155,8 +156,12 @@ export class TeamsBotImpl {
       group_name
     );
 
-    await handler?.next(ProgressBarConstants.SCAFFOLD_STEP_UNZIP);
-    zipContent.extractAllTo(this.config.scaffold.workingDir!, true);
+    try {
+      await handler?.next(ProgressBarConstants.SCAFFOLD_STEP_UNZIP);
```
henzhang-ms, line 160: move `handler?.next` outside of the try block. The block should not cover its error.

packages/fx-core/src/plugins/resource/bot/plugin.ts
YitongFeng-git: OK

henzhang-ms (approved):

henzhang-ms:
I remember we will refine the error message to tell customer to check project path.

packages/fx-core/src/plugins/resource/bot/errors.ts
```diff
@@ -264,3 +264,15 @@ export class RegisterResourceProviderError extends PluginError {
...
   }
 }
+
+export class ExtractZipError extends PluginError {
+  constructor(path: string, innerError?: Error) {
+    super(
+      ErrorType.User,
+      "ExtractZipError",
+      `Failed to extract zip, please check the input: ${path}`,
+      [Messages.CheckOutputLogAndTryToFix],
```
henzhang-ms, line 274: Refine the suggestions.

PR #2036, opened 2021-08-24
feat(localdebug): scaffold and localdebug project migrated from Team Toolkit V1 in localdebug plugin

XiaofuHuang:
Add `migrateV1Project` method in `executeUserTask`.
Localdebug project migrates from Team Toolkit V1.

### DIFF
--- a/packages/fx-core/src/plugins/resource/localdebug/tasks.ts
+++ b/packages/fx-core/src/plugins/resource/localdebug/tasks.ts
@@ -9,6 +9,7 @@ export function generateTasks(
   includeFrontend: boolean,
   includeBackend: boolean,
   includeBot: boolean,
+  includeAuth: boolean,
   isMigrateFromV1: boolean,
   programmingLanguage: string
 ): Record<string, unknown>[] {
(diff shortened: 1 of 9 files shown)


packages/fx-core/src/plugins/resource/localdebug/tasks.ts
```diff
@@ -9,6 +9,8 @@ export function generateTasks(
   includeFrontend: boolean,
   includeBackend: boolean,
   includeBot: boolean,
+  includeAuth: boolean,
```
swatDong, line 12: Is `includeAuth = !(isMigrateFromV1)`?

XiaofuHuang: Yes, now includeAuth = !(isMigrateFromV1).
Would you like to split includeAuth and isMigrateFromV1?
Maybe there will be no simple auth tab app in the future.

swatDong: Both ok to me, just to confirm.

swatDong (approved):

nliu-ms (approved):

PR #2079, opened 2021-08-27
fix(migrate-v1-bot): replace botId in manifest with placeholder

kuojianlu:


packages/fx-core/src/plugins/resource/appstudio/plugin.ts
```diff
@@ -74,6 +74,7 @@ import {
   SOLUTION,
   MANIFEST_TEMPLATE,
   TEAMS_APP_MANIFEST_TEMPLATE_FOR_MULTI_ENV,
+  BOT_PLUGIN_NAME,
```
nliu-ms, line 77: import {PluginNames} from "../../solution/fx-solution/constants"

avoid duplication.

nliu-ms (approved):

packages/fx-core/src/plugins/resource/appstudio/plugin.ts
kuojianlu: Nice catch. Updated.

PR #2094, opened 2021-08-30
fix(solution): fix migrate v1 project programming language detection

XiaofuHuang:
The origin project has been moved to the archive folder, change the language detection file path


swatDong (approved):
The fix logic looks good for the issue.
But considering more, could the sequence be "prepare all required config -> move content -> generate config", in case any error on the preparation step and user's project will not be changed.

XiaofuHuang:
> The fix logic looks good for the issue.
> But considering more, could the sequence be "prepare all required config -> move content -> generate config", in case any error on the preparation step and user's project will not be changed.

Now the create basic folder structure logic is in `core`. So the archive logic is in `core` too. The archive will be executed after all the `core` configuration is prepared but before any solution and resource plugins migration logic.

packages/fx-core/src/common/fileUtils.ts
```diff
@@ -34,3 +34,12 @@ export async function readJson(filePath: string): Promise<any> {
   sendTelemetryErrorEvent(Component.core, TelemetryEvent.ReadJson, fxError);
   throw fxError;
 }
+
+export async function checkFileExist(filePath: string): Promise<boolean> {
```
1yefuwang1, line 38: Does fs.pathExists meet your requirement?

1yefuwang1 (commented):
Could you please add a ut case?

XiaofuHuang:
> Could you please add a ut case?

1yefuwang1 (approved):

PR #2102, opened 2021-08-30
fix(migratev1): disable the migration of the projects created before V1.2.0

XiaofuHuang:


kuojianlu (approved):

packages/fx-core/src/core/tools.ts
```diff
@@ -111,26 +117,53 @@ export function isValidProject(workspacePath?: string): boolean {
...
+    return `Cannot read 'package.json'. ${error?.message}`;
+  }
+
+  if (!packageSettings?.msteams) {
+    return "Teams Toolkit V1 settings cannot be found in 'package.json'.";
+  }
+
+  const manifestPath = path.resolve(workspacePath, AppPackageFolderName, V1ManifestFileName);
+  if (!(await fs.pathExists(manifestPath))) {
+    return "The project should be created after version 1.2.0";
```
swatDong, line 147: If any user error on migration, is there any documentation to explain the detailed prerequisites or limitations?

swatDong (approved):

1openwindow (approved):

packages/fx-core/src/core/tools.ts
XiaofuHuang: Good idea, I will confirm with PM about the documents on the docs site.

1yefuwang1 (approved):

PR #2108, opened 2021-08-30
perf(arm support): bot using existing bot

dilin-MS:
If user choose to create bot project using existing bot channel registration, we should remove bot service resource code from bot bicep files.

[Bug Item](https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10719110)


packages/fx-core/templates/plugins/resource/bot/bicep/module.template.bicep
```diff
@@ -1,27 +1,31 @@
...
+    botAadClientId: bot_aadClientId
+    botDisplayName: bot_displayName
+    {{/ifCond}}
     botServiceSKU: bot_serviceSKU
     botWebAppName: bot_sitesName
     botWebAppSKU: bot_webAppSKU
-    {{#contains 'fx-resource-identity' Plugins}}
-    identityName: {{../PluginOutput.fx-resource-identity.Outputs.identityName}}
-    {{/contains}}
+    \{{#contains 'fx-resource-identity' Plugins}}
```
dilin-MS, line 13: The `\` symbot is used to bypass the handlebars syntax during handlebars compilation. This part of handlebars syntax will be compiled by solution plugin rather than bot resource plugin.

YitongFeng-git (approved):

packages/fx-core/templates/plugins/resource/bot/bicep/botConfiguration.template.bicep
```diff
@@ -1,4 +1,6 @@
+{{#ifCond createNewBotService true}}
```
blackchoey, line 1: can we see whether `#if` helper from handlebars works? https://handlebarsjs.com/guide/builtin-helpers.html#if

packages/fx-core/templates/plugins/resource/bot/bicep/module.template.bicep
blackchoey: You can just generate the final content based on the available plugins, no need to escape it here which makes the template complicated.

dilin-MS: Yes. I worked. Let me fix it in next PR.

packages/fx-core/templates/plugins/resource/bot/bicep/botConfiguration.template.bicep
dilin-MS: Yes. I worked. Let me fix it in next PR.

PR #2109, opened 2021-08-30
chore(env checker): add bicep cli support - 1

meifans:
add depsChecker files to solution


a1exwang (commented):
what is the empty file named `placeholder` for?

packages/fx-core/src/plugins/solution/fx-solution/utils/depsChecker/bicepChecker.ts
```diff
@@ -0,0 +1,356 @@
...
+    }
+    return "bicep";
+  }
+
+  private getBicepBitSuffixName(): string {
+    if (isWindows()) {
+      return "bicep-win-x64.exe";
+    }
+    if (isMacOS()) {
+      return "bicep-osx-x64";
```
a1exwang, line 270: Does bicep support M1? If not, maybe detect M1 Mac and explicitly tell it to the user

meifans: > what is the empty file named `placeholder` for?

removed

meifans: > Does bicep support M1? If not, maybe detect M1 Mac and explicitly tell it to the user

`bicep` work well on my mac m1, but it available for all mac m1 is to verify.
If bicep not support mac m1, the version of ARM need to have a workaround for m1 user.

a1exwang (approved):

PR #2115, opened 2021-08-31
fix(arm support): fix spfx project provision error

dilin-MS:
[[Bug Bash] Provision for SPFx project will fail](https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10719592)


LongOddCode (approved):

packages/fx-core/src/plugins/solution/fx-solution/solution.ts
```diff
@@ -744,7 +743,7 @@ export class TeamsAppSolution implements Solution {
           util.format(getStrings().solution.ProvisionFinishNotice, PluginDisplayName.Solution)
         );
 
-        if (isArmSupportEnabled()) {
+        if (isArmSupportEnabled() && this.isAzureProject(ctx)) {
```
blackchoey, line 746: can we add a test case for the bug?

dilin-MS: Fixed.

blackchoey (approved):

PR #2121, opened 2021-08-31
chore(env): use command from bicep env checker

meifans:
Ensure bicep cli is installed before compiling bicep to json.


packages/fx-core/src/plugins/solution/fx-solution/utils/depsChecker/bicepChecker.ts
```diff
@@ -41,8 +41,8 @@ const source = "bicep-envchecker";
 export async function ensureBicep(ctx: SolutionContext): Promise<string> {
   const bicepChecker = new BicepChecker(ctx.logProvider, ctx.telemetryReporter);
   try {
-    if (bicepChecker.isEnabled() && !bicepChecker.isInstalled()) {
-      bicepChecker.install();
+    if ((await bicepChecker.isEnabled()) && !(await bicepChecker.isInstalled())) {
```
dilin-MS, line 44: I checked and found that the BicepEnvCheckerEnable feature flag default value is false. Should we turn it as true by default?

packages/fx-core/src/plugins/solution/fx-solution/utils/depsChecker/bicepChecker.ts
```diff
@@ -41,8 +41,8 @@ const source = "bicep-envchecker";
...
+      await bicepChecker.install();
```
dilin-MS, line 45: Should we notify user that we will install bicep binary for them or ask them for permission?

packages/fx-core/src/plugins/solution/fx-solution/utils/depsChecker/bicepChecker.ts
meifans: It's not ready to release and need more testing.

packages/fx-core/src/plugins/solution/fx-solution/utils/depsChecker/bicepChecker.ts
meifans: We notify user by logging like `Dotnet SDK` and `Azure Function Core Tools.
There's a popup with a help link if fail.
It's not a global install, so we don't ask user for permission.(install dir: ~/.fx/bin/bicep)

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -383,11 +386,12 @@ async function getResourceManagementClientForArmDeployment(
 }
 
 async function compileBicepToJson(
+  bicepCommand: string,
   bicepOrchestrationFilePath: string,
   jsonFilePath: string
 ): Promise<void> {
   // TODO: ensure bicep cli is installed
```
dilin-MS, line 393: Remove this todo comment since the ensure bicep cli step is done. Thx

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -161,13 +162,15 @@ export async function doDeployArmTemplates(ctx: SolutionContext): Promise<Result
     throw new Error("Failed to get resource group from project solution settings.");
   }
 
+  const bicepCommand = await ensureBicep(ctx);
```
dilin-MS, line 165: Maybe move the bicep check step into the `compileBiceoToJson()` method so the method can self-comtain?

meifans: it depends `ctx`, and `compileBiceoToJson` don't need `ctx`.

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
meifans: Removed

dilin-MS (approved):

a1exwang (approved):

jayzhang (approved):

PR #2125, opened 2021-08-31
fix(appstudio): add v1 project sso telemetry during migration

XiaofuHuang:
1. Replace `V1_MANIFEST` to `V1ManifestFileName`
2. Add a telemetry property to track whether a v1 tab app includes SSO.

### DIFF
--- a/packages/fx-core/src/plugins/resource/appstudio/plugin.ts
+++ b/packages/fx-core/src/plugins/resource/appstudio/plugin.ts
@@ -385,7 +385,7 @@ export class AppStudioPluginImpl {
       if (await this.checkFileExist(archiveOutlineFile)) {
         await fs.copyFile(archiveOutlineFile, newOutlineFile);
       }
-      return { enableAuth: !!manifest.webApplicationInfo };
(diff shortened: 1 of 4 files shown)


nliu-ms (approved):

packages/fx-core/src/plugins/resource/appstudio/plugin.ts
```diff
@@ -387,8 +385,10 @@ export class AppStudioPluginImpl {
       if (await this.checkFileExist(archiveOutlineFile)) {
         await fs.copyFile(archiveOutlineFile, newOutlineFile);
       }
+      return { enableAuth: !!manifest.webApplicationInfo };
```
swatDong, line 388: For SSO, do we care empty `webApplicationInfo` or `webApplicationInfo.id` is required?

XiaofuHuang: Yes, webApplicationInfo.id is more accurate. 
Changed it to !!manifest?.webApplicationInfo?.id

swatDong (approved):

PR #2131, opened 2021-08-31
chore: add telemetries for ARM operations

blackchoey:


packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -56,93 +60,39 @@ const parameterFileNameTemplateNew = "azure.parameters.@envName.json";
...
-
-    // Output bicep module files from each resource plugin
-    const modulesFolderPath = isMultiEnvEnabled()
-      ? path.join(templateFolderPath, modulesFolder)
-      : templateFolderPath;
-    await fs.ensureDir(modulesFolderPath);
-    for (const module of moduleFiles) {
-      // module[0] contains relative path to template folder, e.g. "./modules/frontendHosting.bicep"
-      await fs.writeFile(path.join(templateFolderPath, module[0]), module[1]);
+  let result: Result<void, FxError>;
```
blackchoey, line 63: moved original logic to `doGenerateArmTemplate` function, this function is now a wrapper which takes care of telemetry and error handling

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -56,93 +61,39 @@ const parameterFileNameTemplateNew = "azure.parameters.@envName.json";
...
-    const gitignoreFilePath = isMultiEnvEnabled()
-      ? path.join(ctx.root, templateFolderNew, gitignoreFileName)
-      : path.join(ctx.root, baseFolder, gitignoreFileName);
-    if (!(await fs.pathExists(gitignoreFilePath))) {
-      await fs.writeFile(gitignoreFilePath, gitignoreContent);
-    }
   }
-
-  return ok(undefined); // Nothing to return when success
+  return result;
```
1yefuwang1, line 96: You can return sendErrorTelemetryThenReturnError(...) instead.

blackchoey: offline synced, the result is shared in between exception handling logic and regular logic, so will keep existing implementation for now.

1yefuwang1 (approved):

PR #2133, opened 2021-08-31
fix(multi-env): fix the issue that toolkit asking env when localdebug

a1exwang:
- set `ignoreEnvInfo` to true to prevent asking for env when local debug 
- for local debug, it also needs the default env's solution context

Tested local debug with multi env feature flags on

### DIFF
--- a/packages/fx-core/src/core/middleware/envInfoLoader.ts
+++ b/packages/fx-core/src/core/middleware/envInfoLoader.ts
@@ -65,7 +65,8 @@ export function EnvInfoLoaderMW(
     const core = ctx.self as FxCore;
 
     if (inputs.ignoreEnvInfo === true) {
-      const result = await loadSolutionContextWithoutEnv(core.tools, inputs, ctx.projectSettings);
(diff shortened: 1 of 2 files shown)


packages/fx-core/src/core/middleware/envInfoLoader.ts
```diff
@@ -136,6 +153,30 @@ export async function loadSolutionContext(
   return ok(solutionContext);
 }
 
+export async function loadSolutionContextWithoutEnv(
```
dooriya, line 156: Can we reuse the `loadSolutionContext`? For example, add a parameter to skip loading env config.

a1exwang: Discussed offline with @dooriya. For local debug, it needs to read from default environment's profile file for upgrading `programmingLanguage`, etc. So just reuse the `loadSolutionContext` to use default env

LongOddCode (approved):

qinezh (approved):

PR #2139, opened 2021-08-31
chore: bicep error handling

xzf0587:
1. update poll arm deployment status
2. update the returned error when deployment failed


packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -673,3 +743,12 @@ async function areFoldersEmpty(folderPaths: string[]): Promise<boolean> {
   }
   return isEmpty;
 }
+
+const FeatureFlagNewFolderStructure = "TEAMSFX_ARM_NEW_FOLDER_STRUCTURE";
```
blackchoey, line 747: this seems removed by @qinezh , can you confirm with him?

xzf0587:
here are the pop-up notification and the output in channel
image
image

packages/fx-core/src/common/constants.ts
```diff
@@ -34,3 +34,9 @@ export class FeatureFlagName {
   static readonly ArmSupport = "TEAMSFX_ARM_SUPPORT";
   static readonly BicepEnvCheckerEnable = "TEAMSFX_BICEP_ENV_CHECKER_ENABLE";
 }
+
+export class ArmDeploymentStatus {
```
dilin-MS, line 38: I don't suggest we define another status class for @azure/arm-resource ProvisioningState. We will have to maintain it once the lib upgrades. Why don't we use the type defined in the lib directly?

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -172,60 +268,34 @@ export async function doDeployArmTemplates(ctx: SolutionContext): Promise<Result
...
-        pollDeploymentStatus(client, resourceGroupName, deploymentStartTime);
-      }
-    }, waitingTimeSpan);
+        }
+      });
+      returnError = new Error(errorMessage);
+    } else {
+      returnError = new Error(`Failed to deploy arm templates to azure. Error: ${error.message}`);
+    }
+    return err(returnUserError(returnError, "Solution", "ArmDeploymentFailed"));
```
blackchoey, line 298: since we no longer throw an error, please also help update the telemetry part which determines the operation result based on the thrown error.

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -97,6 +103,80 @@ export async function generateArmTemplate(ctx: SolutionContext): Promise<Result<
...
+      );
+    }
+    if (status.get(ArmDeploymentStatus.Succeeded)?.length) {
+      deployCtx.ctx.logProvider?.info(
+        `[${PluginDisplayName.Solution}] Succedded deployments-> ${status
+          .get(ArmDeploymentStatus.Succeeded)
+          ?.join(",")}`
+      );
+    }
+    if (status.get(ArmDeploymentStatus.Failed)?.length) {
```
dilin-MS, line 141: Can we only do special treatment to failure situation and leave other status have some behavior? The code seems redundant here.

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
qinezh: Thanks @blackchoey to mention. 

Yeah, this feature flag has already been removed and replaced with the existing feature flag `TEAMFX_MULTI_ENV`.

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
xzf0587: Added it, thanks.

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
xzf0587: It is removed, thanks.

packages/fx-core/src/common/constants.ts
xzf0587: It is fixed. Thanks.

packages/fx-core/src/plugins/solution/fx-solution/arm.ts

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -309,10 +270,18 @@ export async function deployArmTemplates(ctx: SolutionContext): Promise<Result<v
...
+    } else {
+      sendErrorTelemetryThenReturnError(
+        SolutionTelemetryEvent.ArmDeployment,
+        result.error,
+        ctx.telemetryReporter
+      );
+    }
   } catch (error) {
     result = err(
       returnUserError(
```
blackchoey, line 287: L287 may need to be changed to SystemError, since user error ( ARM deployment error ) are handled in L279.

xzf0587: Sure, it is fixed, thanks.

blackchoey (approved):

1yefuwang1 (approved):

PR #2147, opened 2021-09-01
fix: improve sideloading check flow

swatDong:
Per suggestions, improve the sideloading check flow:

- Show nothing if signin user has sideloading permission or the check is failed (the same ui as current stable release)
image

- Click warning message on tree view will popup the warning notification
image
image


kuojianlu (approved):

HuihuiWu-Microsoft (approved):
LGTM

PR #2158, opened 2021-09-01
fix: remove vs check for app studio build/publish

nliu-ms:
As vs extension won't call teamsfx cli anymore, remove logics for vs platform.

### DIFF
--- a/packages/fx-core/src/plugins/resource/appstudio/index.ts
+++ b/packages/fx-core/src/plugins/resource/appstudio/index.ts
@@ -2,7 +2,6 @@
 // Licensed under the MIT license.
 
 import {
-  ConfigFolderName,
   FxError,
   ok,
   err,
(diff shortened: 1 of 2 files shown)


packages/fx-core/src/plugins/resource/appstudio/index.ts
```diff
@@ -51,22 +48,7 @@ export class AppStudioPlugin implements Plugin {
...
-        appStudioQuestions.addChild(appPath);
-
-        const remoteTeamsAppId = new QTreeNode({
-          type: "text",
-          name: Constants.REMOTE_TEAMS_APP_ID,
-          title: "Please input the teams app id in App Studio",
-        });
-        appStudioQuestions.addChild(remoteTeamsAppId);
-      } else if (ctx.answers?.platform === Platform.VSCode) {
+      if (ctx.answers?.platform === Platform.VSCode) {
```
tecton, line 51: Do we need to keep this platform check? Since the following codes don't have this check.
I think it's better to be consistent?
1. Keep the check for other platforms in future.
2. Remove the check if there's only one platform.

tecton (approved):

packages/fx-core/src/plugins/resource/appstudio/index.ts
nliu-ms: this check is kept because CLI platform has different behavior for publish command.

tecton: OK that's no problem.

PR #2161, opened 2021-09-01
fix(bot): copy adaptive cards to lib

YitongFeng-git:
To fix remote debug failed, for it runs index.js under lib/ when remote debug. Local debug works well.


IvanJobs (approved):
lgtm

PR #2162, opened 2021-09-01
fix(multi-env): fix remote debug for multi env

a1exwang:
- should not ignore env info for remote debug

Tested that local debug and remote debug works.


qinezh (approved):

packages/vscode-extension/src/debug/commonUtils.ts
```diff
@@ -133,7 +133,8 @@ async function executeLocalDebugUserTask(funcName: string, params?: unknown): Pr
     const inputs = getSystemInputs();
     inputs.ignoreLock = true;
     inputs.ignoreConfigPersist = true;
-    inputs.ignoreEnvInfo = true;
+    const isRemote = typeof params === "string" && params === "remote";
```
qinezh, line 136: Seems params === "remote" is enough?

a1exwang: Yes. It is enough.

PR #2164, opened 2021-09-01
fix(migrate-v1-bot): replace message extension botId in manifest with placeholder

kuojianlu:


nliu-ms (approved):

XiaofuHuang (approved):

packages/fx-core/src/plugins/resource/appstudio/plugin.ts
```diff
@@ -238,6 +238,9 @@ export class AppStudioPluginImpl {
       if (manifest.bots !== undefined && manifest.bots.length > 0) {
         manifest.bots[0].botId = `{${BOT_ID}}`;
       }
+      if (manifest.composeExtensions !== undefined && manifest.composeExtensions.length > 0) {
+        manifest.composeExtensions[0].botId = `{${BOT_ID}}`;
```
XiaofuHuang, line 242: Do we need to use replaceExistingValueToPlaceholder to replace all the origin bot id to the place holder?

kuojianlu: Good point. Replaced all.

PR #2173, opened 2021-09-02
fix: add rg location

xiaolang124:
1. Add resource group location list.

image

### DIFF
--- a/packages/fx-core/src/plugins/solution/fx-solution/commonQuestions.ts
+++ b/packages/fx-core/src/plugins/solution/fx-solution/commonQuestions.ts
@@ -218,7 +218,10 @@ export async function askResourceGroupInfo(
   }
 }
 
-async function getLocations(ctx: SolutionContext, rmClient: ResourceManagementClient) {
+async function getLocations(
+  ctx: SolutionContext,
+  rmClient: ResourceManagementClient
+): Promise<string[] | undefined> {
   const credential = await ctx.azureAccountProvider!.getAccountCredentialAsync();
   const subscriptionClient = new SubscriptionClient(credential as any);
   const askSubRes = await ctx.azureAccountProvider!.getSelectedSubscription(true);
--- a/packages/fx-core/src/plugins/solution/fx-solution/commonQuestions.ts
+++ b/packages/fx-core/src/plugins/solution/fx-solution/commonQuestions.ts
@@ -158,15 +158,6 @@ export async function askResourceGroupInfo(
     .map((item) => [item.name, item.location] as [string, string]);
 
   const locations = await getLocations(ctx, rmClient);
-  if (!locations || locations.length == 0) {
-    return err(
-      returnUserError(
-        new Error(`Failed to list resource group locations`),
-        "Solution",
-        SolutionError.FailedToListResourceGroup
-      )
-    );
-  }
 
   const node = await getQuestionsForResourceGroup(
     defaultResourceGroupName,
@@ -224,7 +215,16 @@ async function getLocations(
   rmClient: ResourceManagementClient
 ): Promise<string[] | undefined> {
   const credential = await ctx.azureAccountProvider!.getAccountCredentialAsync();
-  const subscriptionClient = new SubscriptionClient(credential as any);
+  let subscriptionClient = undefined;
+  if (credential) {
+    subscriptionClient = new SubscriptionClient(credential);
+  } else {
+    throw returnUserError(
+      new Error(`Failed to get azure credential`),
+      "Solution",
+      SolutionError.FailedToGetAzureCredential
+    );
+  }
   const askSubRes = await ctx.azureAccountProvider!.getSelectedSubscription(true);
   const listLocations = await subscriptionClient.subscriptions.listLocations(
     askSubRes!.subscriptionId
@@ -235,7 +235,15 @@ async function getLocations(
     (rt) => rt.resourceType?.toLowerCase() === ResourceGroups.toLowerCase()
   );
   const resourceLocations = resourceTypeData?.locations;
-  return resourceLocations?.filter((item) => locations.includes(item));
+  const rgLocations = resourceLocations?.filter((item) => locations.includes(item));
+  if (!rgLocations || rgLocations.length == 0) {
+    throw returnUserError(
+      new Error(`Failed to list resource group locations`),
+      "Solution",
+      SolutionError.FailedToListResourceGroupLocation
+    );
+  }
+  return rgLocations;
 }
 
 async function getResourceGroupInfoFromEnvConfig(
--- a/packages/fx-core/src/plugins/solution/fx-solution/commonQuestions.ts
+++ b/packages/fx-core/src/plugins/solution/fx-solution/commonQuestions.ts
@@ -22,6 +22,7 @@ import {
 import { GLOBAL_CONFIG, LOCATION, RESOURCE_GROUP_NAME, SolutionError } from "./constants";
 import { v4 as uuidv4 } from "uuid";
 import { ResourceManagementClient } from "@azure/arm-resources";
+import { SubscriptionClient } from "@azure/arm-subscriptions";
 import { PluginDisplayName } from "../../../common/constants";
 import { isMultiEnvEnabled } from "../../../common";
 import {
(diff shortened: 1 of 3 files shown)


packages/fx-core/src/plugins/solution/fx-solution/commonQuestions.ts
```diff
@@ -152,10 +156,21 @@ export async function askResourceGroupInfo(
...
+  );
+  const locations = listLocations.map((item) => item.displayName);
+  const providerData = await rmClient.providers.get(MsResources);
+  const resourceTypeData = providerData.resourceTypes?.find(
+    (rt) => rt.resourceType?.toLowerCase() === ResourceGroups.toLowerCase()
+  );
+  const resourceLocations = resourceTypeData?.locations;
+  resourceLocations?.filter((item) => locations.includes(item));
+
+  const availableLocations = resourceLocations ?? ["East US", "West US"];
```
qinezh, line 173: How about to extract above logic into a method here?

packages/fx-core/src/plugins/solution/fx-solution/commonQuestions.ts
```diff
@@ -152,10 +156,21 @@ export async function askResourceGroupInfo(
...
+  const listLocations = await subscriptionClient.subscriptions.listLocations(
+    askSubRes!.subscriptionId
+  );
+  const locations = listLocations.map((item) => item.displayName);
+  const providerData = await rmClient.providers.get(MsResources);
+  const resourceTypeData = providerData.resourceTypes?.find(
+    (rt) => rt.resourceType?.toLowerCase() === ResourceGroups.toLowerCase()
+  );
+  const resourceLocations = resourceTypeData?.locations;
+  resourceLocations?.filter((item) => locations.includes(item));
```
qinezh, line 171: Looks the usage is not correct, as it will return a new array instead of updating current array.

packages/fx-core/src/plugins/solution/fx-solution/commonQuestions.ts
a1exwang: Should we throw an error on failure to get locations?

packages/fx-core/src/plugins/solution/fx-solution/commonQuestions.ts
xiaolang124: Fixed.

packages/fx-core/src/plugins/solution/fx-solution/commonQuestions.ts
xiaolang124: Fixed.

packages/fx-core/src/plugins/solution/fx-solution/commonQuestions.ts
```diff
@@ -207,6 +218,22 @@ export async function askResourceGroupInfo(
   }
 }
 
+async function getLocations(ctx: SolutionContext, rmClient: ResourceManagementClient) {
```
qinezh, line 221: missing return type?

xiaolang124: Fixed.

packages/fx-core/src/plugins/solution/fx-solution/commonQuestions.ts
```diff
@@ -208,6 +219,25 @@ export async function askResourceGroupInfo(
   }
 }
 
+async function getLocations(
+  ctx: SolutionContext,
+  rmClient: ResourceManagementClient
+): Promise<string[] | undefined> {
+  const credential = await ctx.azureAccountProvider!.getAccountCredentialAsync();
+  const subscriptionClient = new SubscriptionClient(credential as any);
```
qinezh, line 227: `credential as any` looks weird here, any reason to do the type cast here?

packages/fx-core/src/plugins/solution/fx-solution/commonQuestions.ts
```diff
@@ -208,6 +219,25 @@ export async function askResourceGroupInfo(
...
+  const askSubRes = await ctx.azureAccountProvider!.getSelectedSubscription(true);
```
qinezh, line 228: shall we check if the result of `askSubRes` is undefined here? for example, the selection of subscription is cancelled by user.

packages/fx-core/src/plugins/solution/fx-solution/commonQuestions.ts
```diff
@@ -153,14 +157,21 @@ export async function askResourceGroupInfo(
...
+  const locations = await getLocations(ctx, rmClient);
+  if (!locations || locations.length == 0) {
+    return err(
+      returnUserError(
+        new Error(`Failed to list resource group locations`),
+        "Solution",
+        SolutionError.FailedToListResourceGroup
+      )
+    );
+  }
```
qinezh, line 169: how about wrapping the error handling logic to the method `getLocations` as well?

packages/fx-core/src/plugins/solution/fx-solution/commonQuestions.ts
xiaolang124: Add check credential logic, so now cast logic is not needed.

packages/fx-core/src/plugins/solution/fx-solution/commonQuestions.ts
xiaolang124: No, ```getSelectedSubscription``` already handles this part.

packages/fx-core/src/plugins/solution/fx-solution/commonQuestions.ts
xiaolang124: Fixed.

packages/fx-core/src/plugins/solution/fx-solution/commonQuestions.ts
qinezh: From the method description https://github.com/OfficeDev/TeamsFx/blob/dev/packages/api/src/utils/login.ts#L83, subscriptionId will be empty string if user cancel the selection. I don't think the case has been handled correctly.

xiaolang124: When ```triggerUI == true```, means users have to login and select subscription. Otherwise this method will throw user cancel error.

qinezh: Offline discussed with @xiaolang124 , the user error caused by cancel operation will be handled and won't be exposed to user.

packages/fx-core/src/plugins/solution/fx-solution/commonQuestions.ts
```diff
@@ -208,6 +210,42 @@ export async function askResourceGroupInfo(
...
+  );
+  const locations = listLocations.map((item) => item.displayName);
+  const providerData = await rmClient.providers.get(MsResources);
+  const resourceTypeData = providerData.resourceTypes?.find(
+    (rt) => rt.resourceType?.toLowerCase() === ResourceGroups.toLowerCase()
+  );
+  const resourceLocations = resourceTypeData?.locations;
+  const rgLocations = resourceLocations?.filter((item) => locations.includes(item));
+  if (!rgLocations || rgLocations.length == 0) {
+    throw returnUserError(
```
qinezh, line 240: Seems the common usage is to return something like `Result<Promise<string[] | undefined>, FxError>` instead of throwing it?

xiaolang124: Good idea. Fixed.

packages/fx-core/src/plugins/solution/fx-solution/commonQuestions.ts
```diff
@@ -208,6 +210,42 @@ export async function askResourceGroupInfo(
...
+      new Error(`Failed to list resource group locations`),
+      "Solution",
+      SolutionError.FailedToListResourceGroupLocation
+    );
+  }
+  return rgLocations;
```
a1exwang: Since you have checked `!rgLocations` so the return type can be `string[]` instead of `string[] | undefined`. This can simplify the type assertion afterwards.

xiaolang124: Make sense, fixed.

a1exwang (approved):

qinezh (approved):

LongOddCode (approved):

PR #2174, opened 2021-09-02
chore(sql): fix reprovision bug

xzf0587:
Since getQuestions will skip in reprovision if first provision is successful, the check status of sql existing is skipped as well.
Use the config prepareQuestions to keep the status of sql existing when the getQuestions is skipped.


xzf0587:
Here is the bug.
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10719968

blackchoey (approved):

adashen (approved):

PR #2185, opened 2021-09-03
fix(solution common question): fix error handling when unauthorized subscription

dilin-MS:
Fix [bug](https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10719249)

### DIFF
--- a/packages/fx-core/src/plugins/solution/fx-solution/commonQuestions.ts
+++ b/packages/fx-core/src/plugins/solution/fx-solution/commonQuestions.ts
@@ -321,19 +321,29 @@ async function askCommonQuestions(
     }
     resourceGroupInfo = res.value;
   } else if (resourceGroupNameFromProfile && resourceGroupLocationFromProfile) {
-    const checkRes = await rmClient.resourceGroups.checkExistence(resourceGroupNameFromProfile);
-    if (checkRes.body) {
-      resourceGroupInfo = {
-        createNewResourceGroup: false,
-        name: resourceGroupNameFromProfile,
-        location: resourceGroupLocationFromProfile,
-      };
-    } else {
-      resourceGroupInfo = {
-        createNewResourceGroup: true,
-        name: resourceGroupNameFromProfile,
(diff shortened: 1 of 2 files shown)


packages/fx-core/src/plugins/solution/fx-solution/commonQuestions.ts
```diff
@@ -321,19 +321,32 @@ async function askCommonQuestions(
...
-        location: resourceGroupLocationFromProfile,
-      };
-    } else {
-      resourceGroupInfo = {
-        createNewResourceGroup: true,
-        name: resourceGroupNameFromProfile,
-        location: resourceGroupLocationFromProfile,
-      };
+    try {
+      const checkRes = await rmClient.resourceGroups.checkExistence(resourceGroupNameFromProfile);
```
1yefuwang1, line 325: There is another checkExistence() call in this file. Is it deliberately left unchanged?

packages/vscode-extension/package-lock.json
```diff
@@ -4,55 +4,6 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
```
1yefuwang1, line 6: Why is package-lock.json changed?

packages/fx-core/src/plugins/solution/fx-solution/commonQuestions.ts
dilin-MS: I failed to find it. Could you please provide the code line? Thx

packages/vscode-extension/package-lock.json
dilin-MS: After updating the base branch `dev` and run `npm run setup`, this change is auto-done. Are we supposed to add the changes?

dilin-MS: Removed the changes.

1yefuwang1 (approved):

PR #2187, opened 2021-09-03
refactor(multi-env): plugin retrieves subscription from ctx instead of accountProvider

a1exwang:
Plugins will read sub from `ctx` (just like resource group) instead of `accountProvider`. This will clear the path to change the logic of reading subs in solution.

Tested local debug and remote happy path.

https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10736857

### DIFF
--- a/packages/fx-core/src/plugins/resource/frontend/configs.ts
+++ b/packages/fx-core/src/plugins/resource/frontend/configs.ts
@@ -52,11 +52,10 @@ export class FrontendConfig {
     const appName = ctx.projectSettings!.appName;
     const solutionConfigs = ctx.envInfo.profile.get(DependentPluginInfo.SolutionPluginName);
 
-    const subscriptionInfo = await ctx.azureAccountProvider?.getSelectedSubscription();
-    if (!subscriptionInfo) {
-      throw new InvalidConfigError(DependentPluginInfo.SubscriptionId);
-    }
-    const subscriptionId = subscriptionInfo.subscriptionId;
+    const subscriptionId = FrontendConfig.getConfig<string>(
+      DependentPluginInfo.SubscriptionId,
+      solutionConfigs
+    );
     const resourceNameSuffix = FrontendConfig.getConfig<string>(
       DependentPluginInfo.ResourceNameSuffix,
       solutionConfigs
--- a/packages/fx-core/src/plugins/solution/fx-solution/constants.ts
+++ b/packages/fx-core/src/plugins/solution/fx-solution/constants.ts
@@ -50,6 +50,16 @@ export const RESOURCE_GROUP_NAME = "resourceGroupName";
 export const LOCATION = "location";
 
 /**
+ * Config key whose value is the subscription ID of project.
+ */
+export const SUBSCRIPTION_ID = "subscriptionId";
+
+/**
+ * Config key whose value is the subscription ID of project.
+ */
+export const SUBSCRIPTION_NAME = "subscriptionName";
+
+/**
  * Config key whose value is the user info of collaborator
  */
 export const USER_INFO = "userInfo";
--- a/packages/fx-core/src/plugins/solution/fx-solution/constants.ts
+++ b/packages/fx-core/src/plugins/solution/fx-solution/constants.ts
@@ -55,7 +55,7 @@ export const LOCATION = "location";
 export const SUBSCRIPTION_ID = "subscriptionId";
 
 /**
- * Config key whose value is the subscription ID of project.
+ * Config key whose value is the subscription name of project.
  */
 export const SUBSCRIPTION_NAME = "subscriptionName";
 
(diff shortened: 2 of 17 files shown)


packages/fx-core/src/plugins/solution/fx-solution/solution.ts
```diff
@@ -641,10 +643,11 @@ export class TeamsAppSolution implements Solution {
 
       // Only Azure project requires this confirm dialog
       const username = (azureToken as any).username ? (azureToken as any).username : "";
-      const subscriptionInfo = await ctx.azureAccountProvider?.getSelectedSubscription();
+      const subscriptionId = ctx.envInfo.profile.get(GLOBAL_CONFIG)?.get(SUBSCRIPTION_ID) as string;
```
jayzhang, line 646: If user didn't config the sub id in config.env.json, can we use the one from account provider?

jayzhang (approved):

packages/fx-core/src/plugins/solution/fx-solution/solution.ts
dooriya, line 646: Where to set the subscription info before to reading it?

packages/fx-core/src/plugins/solution/fx-solution/constants.ts
```diff
@@ -49,6 +49,16 @@ export const RESOURCE_GROUP_NAME = "resourceGroupName";
  */
 export const LOCATION = "location";
 
+/**
+ * Config key whose value is the subscription ID of project.
+ */
+export const SUBSCRIPTION_ID = "subscriptionId";
+
+/**
+ * Config key whose value is the subscription ID of project.
```
qinezh, line 58:
```suggestion
 * Config key whose value is the subscription name of project.
```

packages/fx-core/src/plugins/solution/fx-solution/solution.ts
```diff
@@ -1861,6 +1864,7 @@ export class TeamsAppSolution implements Solution {
           AskSubscriptionQuestion.func = async (
             inputs: Inputs
           ): Promise<Result<SubscriptionInfo, FxError>> => {
+            ctx.envInfo.profile.get(GLOBAL_CONFIG)?.get(SUBSCRIPTION_ID);
```
qinezh, line 1867: seems this line is useless?

packages/fx-core/src/plugins/solution/fx-solution/solution.ts
a1exwang: Yes. This PR is only a refactor, it does not change business logic. In commonQuestions.ts, solution has already put `subscriptionId` into `ctx`, and the plugins only need to read from it.

a1exwang: Solution has already put subscription info into `ctx.envInfo.profile` before provision (similar to what we do with the resource group info). So this PR is just a refactor. Plugins can read directly from ctx.envInfo.profile because solution has already put subs in it.

packages/fx-core/src/plugins/solution/fx-solution/constants.ts
a1exwang: fixed

packages/fx-core/src/plugins/solution/fx-solution/solution.ts
a1exwang: fixed

packages/fx-core/src/plugins/solution/fx-solution/solution.ts
dooriya: I see. Thanks for the explanation.

dooriya (approved):

packages/fx-core/src/plugins/resource/frontend/configs.ts
```diff
@@ -52,11 +52,10 @@ export class FrontendConfig {
     const appName = ctx.projectSettings!.appName;
     const solutionConfigs = ctx.envInfo.profile.get(DependentPluginInfo.SolutionPluginName);
 
-    const subscriptionInfo = await ctx.azureAccountProvider?.getSelectedSubscription();
-    if (!subscriptionInfo) {
-      throw new InvalidConfigError(DependentPluginInfo.SubscriptionId);
-    }
-    const subscriptionId = subscriptionInfo.subscriptionId;
+    const subscriptionId = FrontendConfig.getConfig<string>(
```
hund030, line 55: Do we need null check?

hund030 (approved):

XiaofuHuang (approved):

KennethBWSong (approved):

xzf0587 (approved):

qinezh (approved):

IvanJobs (approved):
lgtm for bot part.

PR #2188, opened 2021-09-03
fix: remove cryptor in project upgrade

KennethBWSong:


packages/fx-core/src/core/middleware/projectUpgrader.ts
```diff
@@ -179,58 +179,18 @@ export async function upgradeContext(ctx: CoreHookContext): Promise<Result<undef
...
-            fxError.userData = `file: ${fileName}\n------------FILE START--------\n${dictContent}\n------------FILE END----------`;
-            sendTelemetryErrorEvent(Component.core, TelemetryEvent.DecryptUserdata, fxError);
-            throw plaintext.error;
-          }
-          dict[secretKey] = plaintext.value;
-        }
-        return dict;
-      }
-    }
+    dict = deserializeDict(dictContent);
```
adashen, line 186: do we check dictContent before deserialize?

packages/fx-core/src/core/middleware/projectUpgrader.ts
```diff
@@ -179,58 +179,18 @@ export async function upgradeContext(ctx: CoreHookContext): Promise<Result<undef
...
-        continue;
-      }
-
-      const encryptedSecret = cryptoProvider.encrypt(userData[secretKey]);
-      if (encryptedSecret.isOk()) {
-        userData[secretKey] = encryptedSecret.value;
-      }
-    }
-  }
+async function saveUserData(userDataPath: string, userData: Record<string, string>): Promise<void> {
```
adashen, line 193: will userData be null or empty?

packages/fx-core/src/core/middleware/projectUpgrader.ts
KennethBWSong: It will just return empty object.

packages/fx-core/src/core/middleware/projectUpgrader.ts
KennethBWSong: Will not throw error when userdata is empty.

adashen: serializeDict will handle the cipher work right?

KennethBWSong: Here we will not handle the cipher. What we do is just read and write and we don't care about whether it is encrypted.

packages/fx-core/src/core/middleware/projectUpgrader.ts
```diff
@@ -179,58 +179,18 @@ export async function upgradeContext(ctx: CoreHookContext): Promise<Result<undef
...
   await fs.writeFile(userDataPath, serializeDict(userData));
```
SLdragon, line 194: Any chance that serializeDict throw error, do we need to handle it?

KennethBWSong: It will be handled here:
image

adashen (approved):

SLdragon (approved):

PR #2192, opened 2021-09-03
feat(multi-env): support to read subs from input config

a1exwang:
- when multi-env is enabled, it will first read subs from input config (config.dev.json). If not found, it will pop up and ask user, same as the previous behavior)
- confirmed with @xiaolang124 that `getAccountCredentialAsync` will never return undefined

Tested remote happy path with and without feature flag

https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10736857

### DIFF
--- a/packages/fx-core/src/plugins/solution/fx-solution/commonQuestions.ts
+++ b/packages/fx-core/src/plugins/solution/fx-solution/commonQuestions.ts
@@ -27,7 +27,7 @@ import { GLOBAL_CONFIG, LOCATION, RESOURCE_GROUP_NAME, SolutionError } from "./c
 import { v4 as uuidv4 } from "uuid";
 import { ResourceManagementClient } from "@azure/arm-resources";
 import { PluginDisplayName } from "../../../common/constants";
-import { askSubscription, isMultiEnvEnabled } from "../../../common";
+import { isMultiEnvEnabled } from "../../../common";
 import {
   CoreQuestionNames,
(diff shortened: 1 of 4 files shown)


packages/fx-core/src/plugins/solution/fx-solution/commonQuestions.ts
```diff
@@ -77,8 +77,34 @@ export async function checkSubscription(
...
-  return ok(askSubRes!);
+
+  const subscriptionId = ctx.envInfo.config.azure.subscriptionId;
+  if (!isMultiEnvEnabled() || !subscriptionId) {
+    const askSubRes = await ctx.azureAccountProvider.getSelectedSubscription(true);
+    return ok(askSubRes!);
+  }
+
+  // make sure the user is logged in
+  await ctx.azureAccountProvider.getAccountCredentialAsync(true);
```
qinezh, line 88: Do we need to check if user cancel the login dialog?

a1exwang: The current behavior of `azureAccountProvider` is throwing exceptions on cancel or login failure, and core will handle exceptions. So the current implemention will work on these cases.

Although there is `undefined` in the method signature, there is no such cases when `undefined` is returned. Need to confirm with @xiaolang124.

qinezh (approved):

jayzhang (approved):

PR #2197, opened 2021-09-06
fix: update env ui

xiaolang124:
1. Show different icons for active environment and inactive environments.

image

### DIFF
--- a/packages/fx-core/src/plugins/solution/fx-solution/solution.ts
+++ b/packages/fx-core/src/plugins/solution/fx-solution/solution.ts
@@ -687,7 +687,11 @@ export class TeamsAppSolution implements Solution {
           const envName = await askTargetEnvironment(ctx as any, ctx.answers!);
           if (envName) {
             ctx.projectSettings!.activeEnvironment = envName;
-            ctx.ui?.showMessage("info", `[${envName}] is activated.`, false);
+            ctx.ui?.showMessage(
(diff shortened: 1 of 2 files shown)


packages/fx-core/src/plugins/solution/fx-solution/solution.ts
```diff
@@ -687,7 +687,11 @@ export class TeamsAppSolution implements Solution {
           const envName = await askTargetEnvironment(ctx as any, ctx.answers!);
           if (envName) {
             ctx.projectSettings!.activeEnvironment = envName;
-            ctx.ui?.showMessage("info", `[${envName}] is activated.`, false);
+            ctx.ui?.showMessage(
+              "info",
+              `[${envName}] is activated. Please try to do provision again`,
```
qinezh, line 692:
```suggestion
              `[${envName}] is activated. Please try to do provision again.`,
```

qinezh (approved):

dooriya (approved):

packages/fx-core/src/plugins/solution/fx-solution/solution.ts
xiaolang124: Fixed.

HuihuiWu-Microsoft (approved):

1yefuwang1 (approved):

PR #2201, opened 2021-09-06
fix: workaround to expand account child item

swatDong:
Per *vscode/issues/131955*, need to call `TreeView.reveal` to show a collapsed tree item.

This PR is a workaround to force expand the item by changing its label, and could be improved later when `reveal` is supported.

### DIFF
--- a/packages/vscode-extension/src/accountTree.ts
+++ b/packages/vscode-extension/src/accountTree.ts
@@ -247,19 +247,22 @@ export async function registerAccountTreeHandler(): Promise<Result<Void, FxError
     ) => {
       if (status === "SignedIn") {
         if (token !== undefined && accountInfo !== undefined) {
-          tools.treeProvider?.refresh([
-            {
-              commandId: "fx-extension.signinM365",
-              label: (accountInfo.upn as string) ? (accountInfo.upn as string) : "",
-              callback: signinM365Callback,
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/accountTree.ts
```diff
@@ -247,19 +247,22 @@ export async function registerAccountTreeHandler(): Promise<Result<Void, FxError
...
+            icon: "M365",
+          };
+          tools.treeProvider?.refresh([treeItem]);
           const subItem = await getSideloadingItem(token);
           if (subItem && subItem.length > 0) {
             tools.treeProvider?.add(subItem);
+
+            // this is a workaround to expand this child, to be improved when TreeView.reveal is supported
+            treeItem.label += " ";
+            tools.treeProvider?.refresh([treeItem]);
```
HuihuiWu-Microsoft, line 265: What if you refresh again without treeItem.label += " "?

swatDong: That's the workaround point, changing label somehow makes the item force refreshed so its child will be expanded. (Suppose changing label also changes its internal ID so it follows the same behavior as new item).

If refresh without changing label, nothing happens.

HuihuiWu-Microsoft (approved):

PR #2205, opened 2021-09-06
refactor: add manifest related properties to config file

nliu-ms:
image

### DIFF
--- a/packages/api/src/schemas/envConfig.json
+++ b/packages/api/src/schemas/envConfig.json
@@ -33,7 +33,41 @@
         },
         "values": {
           "type": "object",
-          "description": "Configs to customize the Teams app manifest."
+          "description": "Configs to customize the Teams app manifest.",
+          "properties": {
+            "appName": {
+              "type": "object",
+              "description": "Teams app name.",
+              "properties": {
+                "short": {
+                  "type": "string",
+                  "description": "A short display name for teams app.",
+                  "maxLength": 30
+                },
+                "full": {
+                  "type": "string",
+                  "description": "The full name for teams app.",
+                  "maxLength": 100
+                }
+              }
+            },
+            "description": {
+              "type": "object",
+              "description": "Teams app description.",
+              "properties": {
+                "short": {
+                  "type": "string",
+                  "description": "A short description of the app used when space is limited. Maximum length is 80 characters.",
+                  "maxLength": 80
+                },
+                "full": {
+                  "type": "string",
+                  "description": "The full description of the app. Maximum length is 4000 characters.",
+                  "maxLength": 4000
+                }
+              }
+            }
+          }
         }
       },
       "required": ["values"]
--- a/packages/api/src/schemas/envConfig.json
+++ b/packages/api/src/schemas/envConfig.json
@@ -42,32 +42,19 @@
                 "short": {
                   "type": "string",
                   "description": "A short display name for teams app.",
-                  "maxLength": 30
+                  "maxLength": 30,
+                  "minLength": 1
                 },
                 "full": {
                   "type": "string",
                   "description": "The full name for teams app.",
                   "maxLength": 100
                 }
-              }
-            },
-            "description": {
-              "type": "object",
-              "description": "Teams app description.",
-              "properties": {
-                "short": {
-                  "type": "string",
-                  "description": "A short description of the app used when space is limited. Maximum length is 80 characters.",
-                  "maxLength": 80
-                },
-                "full": {
-                  "type": "string",
-                  "description": "The full description of the app. Maximum length is 4000 characters.",
-                  "maxLength": 4000
-                }
-              }
+              },
+              "required": ["short"]
             }
-          }
+          },
+          "required": ["appName"]
         }
       },
       "required": ["values"]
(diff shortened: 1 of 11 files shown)


packages/api/src/schemas/envConfig.json
```diff
@@ -33,7 +33,41 @@
         },
         "values": {
           "type": "object",
-          "description": "Configs to customize the Teams app manifest."
+          "description": "Configs to customize the Teams app manifest.",
+          "properties": {
+            "appName": {
```
qinezh, line 38: Shall we mark the properties like `appName` as required? Otherwise, the mustache render would be failed for `{{config.manifest.values.appName.short}}`?

packages/api/src/schemas/envConfig.json
```diff
@@ -33,7 +33,41 @@
...
+              "type": "object",
+              "description": "Teams app name.",
+              "properties": {
+                "short": {
+                  "type": "string",
+                  "description": "A short display name for teams app.",
+                  "maxLength": 30
```
qinezh, line 45: is empty string valid for the short name?

packages/api/src/schemas/envConfig.json
dooriya: We can mark the `appName` as required, but we need to generate a default value for `appName` in `config.{env}.json` when we create a new environment. And user have the flexibility to customize this property.

packages/fx-core/src/plugins/resource/appstudio/constants.ts
```diff
@@ -101,12 +101,12 @@ export const TEAMS_APP_MANIFEST_TEMPLATE_FOR_MULTI_ENV = `{
...
   "name": {
-      "short": "{appName}",
-      "full": "This field is not used"
+      "short": "{{config.manifest.values.appName.short}}",
+      "full": "{{config.manifest.values.appName.full}}"
   },
   "description": {
-      "short": "Short description of {appName}.",
-      "full": "Full description of {appName}."
+      "short": "{{config.manifest.values.description.short}}",
```
dooriya, line 108: IMO, by default we only expose the `appName` as configurable property in the manifest template (as well as the `config.{env}.json`). Can we leave the `description` property to user to control whether they want to customize its value?

nliu-ms: ok

packages/api/src/schemas/envConfig.json
```diff
@@ -33,7 +33,41 @@
...
+                },
+                "full": {
+                  "type": "string",
+                  "description": "The full name for teams app.",
+                  "maxLength": 100
+                }
+              }
+            },
+            "description": {
```
dooriya, line 54: If we really want to expose the `description` config by default, App Studio plugin needs to also provide the initial value for this property in the config file at scaffold stage (same as the `appName` property).

packages/fx-core/src/core/tools.ts
```diff
@@ -190,7 +190,11 @@ export function newEnvInfo(
     config: config ?? {
       azure: {},
       manifest: {
-        values: {},
+        values: {
+          appName: {
+            short: "myApp",
```
dooriya, line 195: Will the `myApp` be the default value for the manifest app name? or AppStudio plugin will update it to use the true application name (like the current toolkit's behavior)?

qinezh, line 195: Shall we use the project name to keep consistent with previous behavior? So that it can avoid generating same app name for different project.

packages/fx-core/src/plugins/resource/appstudio/plugin.ts
```diff
@@ -1486,6 +1488,11 @@ export class AppStudioPluginImpl {
 
     let manifest = (await fs.readFile(await this.getManifestTemplatePath(ctx.root))).toString();
 
+    if (isMultiEnvEnabled()) {
```
qinezh, line 1491: Before multi-env is enabled, will the new manifest template be rendered correctly?

packages/fx-core/src/core/tools.ts
nliu-ms: This value will be updated in scaffolding and creating new environment.
image

qinezh (approved):

packages/fx-core/src/core/environment.ts
```diff
@@ -216,7 +221,7 @@ class EnvironmentManager {
     if (!isMultiEnvEnabled()) {
       return ok({
         azure: {},
-        manifest: { values: {} },
+        manifest: { values: { appName: { short: "myApp" } } },
```
dooriya, line 224: Can we also consistently to use the empty string `""` as default to create a new `EnvConfig`? In `tools.ts` we set the `appName:short` to `""`, while we set it to `"myApp"` here.

packages/fx-core/src/core/tools.ts
dooriya: Thanks for explanation~

packages/fx-core/src/core/environment.ts
nliu-ms: Sure, thanks!

dooriya (approved):

jayzhang (approved):

jayzhang (approved):

PR #2207, opened 2021-09-06
fix(tab): refine type annotation in ts error handling

hund030:
This PR is going to fix the compiler error:
```
Object is of type 'unknown'.  TS2571

    27 |   } catch (err) {
    28 |     let funcErrorMsg = "";
  > 29 |     if (err.response?.status === 404) {
       |         ^
    30 |       funcErrorMsg = `There may be a problem with the deployment of Azure Function App, please deploy Azure Function (Run command palette "TeamsFx - Deploy Package") first before running this App`;
    31 |     } else if (err.message === "Network Error") {
    32 |       funcErrorMsg =
```

Before TypeScript 4.4, the `catch` clause bindings were set to `any`. TypeScript 4.4 introduced a new compiler option [`useUnknownInCatchVariables`](https://www.typescriptlang.org/tsconfig#useUnknownInCatchVariables) which is `true` by default if you have `strict` option turned on (as we have done).
By turning on `useUnknownInCatchVariables` option, the `catch` clause bindings are set to 'unknown'.

### DIFF
--- a/templates/tab/ts/default/src/components/sample/AzureFunctions.tsx
+++ b/templates/tab/ts/default/src/components/sample/AzureFunctions.tsx
@@ -17,25 +17,30 @@ async function callFunction() {
       },
     });
     return response.data;
-  } catch (err) {
-    let funcErrorMsg = "";
-    if (err?.response?.status === 404) {
-      funcErrorMsg = `There may be a problem with the deployment of Azure Function App, please deploy Azure Function (Run command palette "TeamsFx - Deploy Package") first before running this App`;
-    } else if (err.message === "Network Error") {
-      funcErrorMsg =
-        "Cannot call Azure Function due to network error, please check your network connection status and ";
-      if (err.config.url.indexOf("localhost") >= 0) {
-        funcErrorMsg += `make sure to start Azure Function locally (Run "npm run start" command inside api folder from terminal) first before running this App`;
+  } catch (err: unknown) {
+    if (axios.default.isAxiosError(err)) {
+      let funcErrorMsg = "";
+
+      if (err?.response?.status === 404) {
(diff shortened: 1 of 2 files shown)


henzhang-ms (approved):

templates/tab/ts/default/src/components/sample/AzureFunctions.tsx
```diff
@@ -17,25 +17,30 @@ async function callFunction() {
...
-    let funcErrorMsg = "";
-    if (err?.response?.status === 404) {
-      funcErrorMsg = `There may be a problem with the deployment of Azure Function App, please deploy Azure Function (Run command palette "TeamsFx - Deploy Package") first before running this App`;
-    } else if (err.message === "Network Error") {
-      funcErrorMsg =
-        "Cannot call Azure Function due to network error, please check your network connection status and ";
-      if (err.config.url.indexOf("localhost") >= 0) {
-        funcErrorMsg += `make sure to start Azure Function locally (Run "npm run start" command inside api folder from terminal) first before running this App`;
+  } catch (err: unknown) {
+    if (axios.default.isAxiosError(err)) {
```
henzhang-ms, line 21: NIT: prefer to use early return to reduce indent size

eriolchan (approved):

templates/tab/ts/default/src/components/sample/AzureFunctions.tsx
hund030: I also considered early return, but I think by convention, we put the unhandled case at last of catch block:
```
catch (err: unknown) {
  if (err instanceof SomeError) {
    ...
  } else if (err instanceof AnotherError) {
    ...
  } else {
    throw err;
  }
}
```

PR #2208, opened 2021-09-06
fix(migratev1): hidden welcome treeview after migration

XiaofuHuang:
hidden welcome treeview after migration

### DIFF
--- a/packages/vscode-extension/src/handlers.ts
+++ b/packages/vscode-extension/src/handlers.ts
@@ -226,6 +226,7 @@ export async function migrateV1ProjectHandler(args?: any[]): Promise<Result<null
   );
   const result = await runCommand(Stage.migrateV1);
   await openMarkdownHandler();
+  await vscode.commands.executeCommand("setContext", "fx-extension.sidebarWelcome", false);
   return result;
 }
 


packages/vscode-extension/src/handlers.ts
```diff
@@ -226,6 +226,7 @@ export async function migrateV1ProjectHandler(args?: any[]): Promise<Result<null
   );
   const result = await runCommand(Stage.migrateV1);
   await openMarkdownHandler();
+  await vscode.commands.executeCommand("setContext", "fx-extension.sidebarWelcome", false);
```
HuihuiWu-Microsoft, line 229: Even you set"fx-extension.sidebarWelcome" context to false, the tree view still will not be updated. If the user is not in sidebar welcome treatment group, he still can't see tree items like 'provision', 'deploy',...

1openwindow:
why do we need to hide welcome treeview in v1 case? Most of treeitem items are only available to v2 project, such as "provision", "deploy". So we should only show these operations items in v2 project.

packages/vscode-extension/src/handlers.ts
XiaofuHuang: Because the migrate from v1 project cannot provision, deploy, ......, may be we can keep this logic and discuss a better way to fix it in the next sprint.

HuihuiWu-Microsoft (approved):
LGTM

XiaofuHuang:
> why do we need to hide welcome treeview in v1 case? Most of treeitem items are only available to v2 project, such as "provision", "deploy". So we should only show these operations items in v2 project.

v1 project neednto use m365 login to localdebug. welcome page hide the account view.

HuihuiWu-Microsoft:
> why do we need to hide welcome treeview in v1 case? Most of treeitem items are only available to v2 project, such as "provision", "deploy". So we should only show these operations items in v2 project.

Because welcome tree view doesn't have account info.

1openwindow (approved):

PR #2209, opened 2021-09-07
chore: support vscode native tree item properties like description an…

HuihuiWu-Microsoft:
…d command
image

### DIFF
--- a/packages/vscode-extension/src/controls/tree.tsx
+++ b/packages/vscode-extension/src/controls/tree.tsx
@@ -31,107 +31,63 @@ class App extends React.Component {
     return (
       <div>
         <TreeItem
-          name="Provision in the cloud"
+          label="Provision in the cloud"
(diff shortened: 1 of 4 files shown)


packages/vscode-extension/src/controls/tree.tsx
```diff
@@ -31,107 +31,63 @@ class App extends React.Component {
...
       <div>
         <TreeItem
-          name="Provision in the cloud"
+          label="Provision in the cloud"
           tooltip={StringResources.vsc.commandsTreeViewProvider.provisionDescription}
           icon="codicon codicon-type-hierarchy"
           customized={false}
           disable={false}
-          onclick={this.onProvision}
+          command="fx-extension.provision"
```
1openwindow, line 39: like this "command" property.

1openwindow (approved):

PR #2213, opened 2021-09-07
feat(spfx): add spfx local debug on teams workbench

jiayi-ruan:
1. add spfx local debug on teams workbench
2. add UT

SPFx_Local_debug_Speedup


packages/fx-core/src/plugins/resource/localdebug/tasks.ts
```diff
@@ -122,6 +122,16 @@ export function generateSpfxTasks(): Record<string, unknown>[] {
       },
       dependsOn: "gulp trust-dev-cert",
```
swatDong, line 123: Since gulp has its own dev cert, does SPFx require local-debug plugin to setup teamsfx certificate as well?

Currently, command `fx-extension.pre-debug-check` generates teamsfx certificate no matter spfx or not.

swatDong: My suggestion is to try this new spfx local debug experience on a clean machine (e.g., windows sandbox) to see if any functionality is redundant (cert, deps checker, ...).

swatDong (approved):
The local debug part looks good to me.

nliu-ms (approved):

jayzhang (approved):

PR #2220, opened 2021-09-07
fix(multi-env): update the resource path for SPFx project

qinezh:
1. skip to generate bicep template for SPFx project scaffolding.
2. add new manifest template for SPFx, the difference compared to original manifest template:
   * the path of color is updated to `resources/color.png`
   * the path of outline is updated to `resources/outline.png`


packages/fx-core/src/plugins/resource/appstudio/plugin.ts
```diff
@@ -499,7 +504,7 @@ export class AppStudioPluginImpl {
       );
     }
 
-    const outlineFile = `${appDirectory}/${manifest.icons.outline}`;
+    const outlineFile = `${resourcesDir}/${manifest.icons.outline}`;
```
a1exwang, line 507: the outline file path is already the relative path to `resourceDir`, so you cannot prepend `resourceDir` here.
Screenshot from 2021-09-07 16-27-35

qinezh: Nice catch! it affect the azure project indeed. Will fix it soon.

a1exwang (approved):

packages/fx-core/src/core/index.ts
```diff
@@ -966,5 +969,10 @@ export class FxCore implements Core {
   }
 }
 
+function isAzureProject(solutionSettings: SolutionSettings): boolean {
```
dooriya, line 972: This API is also useful in `solution.ts`,  to reduce duplicate code between core & solution, can we move it to some common utility?

qinezh: That would bring a lots of change in `solution.ts`, and make the target of this PR not clear . How about refactoring it in another PR?

dooriya: Sure~ Another PR will be more reasonable.

dooriya (approved):
:shipit:

nliu-ms (approved):

jayzhang (approved):

PR #2229, opened 2021-09-08
perf(collaboration): add grant permission for VSCode

SLdragon:


packages/vscode-extension/package.json
```diff
@@ -405,6 +414,12 @@
         "title": "Activate environment",
         "icon": "$(check)",
         "enablement": "fx-extension.isMultiEnvEnabled"
+      },
+      {
+        "command": "fx-extension.grantPermission",
+        "title": "Activate environment",
```
KennethBWSong, line 420: Here should be updated

SLdragon: Thanks, will update

KennethBWSong (approved):

PR #2231, opened 2021-09-08
fix(multi-env): cli config file path for multi-env

a1exwang:
- support multi-env paths

### DIFF
--- a/packages/cli/src/utils.ts
+++ b/packages/cli/src/utils.ts
@@ -24,11 +24,17 @@ import {
   Inputs,
   Platform,
   Colors,
+  PublishProfilesFolderName,
+  EnvNamePlaceholder,
+  ProjectSettingsFileName,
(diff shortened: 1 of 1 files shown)


packages/cli/src/utils.ts
```diff
@@ -114,19 +120,54 @@ export async function sleep(ms: number): Promise<void> {
 }
 
 export function getActiveEnv(): string {
-  return "default";
+  if (isMultiEnvEnabled()) {
```
qinezh, line 123: Can we use `environmentManager.getDefaultEnvName()` here?

Alive-Fish (approved):
lgtm

packages/cli/src/utils.ts
a1exwang: Yes. Fixed

qinezh (approved):

PR #2235, opened 2021-09-08
chore(multi-env): support create environment copy

dooriya:
## User experience for adding environment:
1. On scaffold, a default environment "dev" will be created for user;
2. When project is scaffolded, user can user "create new environment copy" command to add a new environment.
    -  User need to provide an existing environment name to copy (by default, the new env is copied from the active environment)
    - The environment input file (`config.{env}.json` and `azure.parameters.{env}.json`) are copied from the specified env.
    
## Steps to add new env copy:
Select "+" button:
image

Provide new environment name:
image

Select an existing env to create copy (default is the active env):
image

### Follow-ups:
1. @jayzhang will help to consolidate the duplicate code and sync the behavior between v1 and v2 APIs.
2. @blackchoey will help tp modify the resource base name in bicep parameter file when copy env.
3. @blackchoey will help to remove parameter.template.json.

### DIFF
--- a/packages/fx-core/src/core/error.ts
+++ b/packages/fx-core/src/core/error.ts
@@ -59,6 +59,13 @@ export function ReadFileError(e: Error): SystemError {
   return error;
 }
 
+export function CopyFileError(e: Error): SystemError {
+  const error = assembleError(e);
+  error.name = "CopyFileError";
+  error.source = CoreSource;
+  return error;
+}
+
 export function NoneFxError(e: any): FxError {
   const err = assembleError(e);
   err.name = "NoneFxError";
--- a/packages/fx-core/src/core/index.ts
+++ b/packages/fx-core/src/core/index.ts
@@ -41,7 +41,6 @@ import * as fs from "fs-extra";
 import * as jsonschema from "jsonschema";
 import * as path from "path";
 import * as uuid from "uuid";
-import { environmentManager } from "..";
 import { globalStateUpdate } from "../common/globalState";
 import {
   Component,
@@ -58,9 +57,8 @@ import {
   isMultiEnvEnabled,
   saveFilesRecursively,
 } from "../common/tools";
-import { getParameterJson } from "../plugins/solution/fx-solution/arm";
-import { LocalCrypto } from "./crypto";
 import {
+  CopyFileError,
   FetchSampleError,
   FunctionRouterError,
   InvalidInputError,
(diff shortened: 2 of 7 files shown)


qinezh:
I'd like to double confirm the user behavoir to create an environment copy: does user need to select the existing environment to be copied? Or the active environment will be selected directly to keep consistent with the behavoir of provision, deploy, publish?

dooriya:
> I'd like to double confirm the user behavoir to create an environment copy: does user need to select the existing environment to be copied? Or the active environment will be selected directly to keep consistent with the behavoir of provision, deploy, publish?


@yanjiang222 What's your opinion here? For me, I would prefer explicitly show the env list to copy from, since we set the active env is the default option to copy env, user just simply press `ENTER` to use the active env.

yanjiang222:
> > I'd like to double confirm the user behavoir to create an environment copy: does user need to select the existing environment to be copied? Or the active environment will be selected directly to keep consistent with the behavoir of provision, deploy, publish?
> 
> @yanjiang222 @jaeyonglee8687 What's your opinion here? For me, I would prefer explicitly show the env list to copy from, since we set the active env is the default option to copy env, user just simply press `ENTER` to use the active env.

Personally, the design make sense to me. User both have a way to copy from default active environment but also select from an inactive environment. This cover both scenarios! thanks!

jayzhang (approved):

qinezh (approved):

jaeyonglee8687:
Do we want to achieve 'creating a (completely) new env' as well? If then we may need to separate the actions, duplicate or create new one.

If not, we should say 'duplicate', or 'copy from' with different icon instead of 'create new' with + icon.

dooriya:
> Do we want to achieve 'creating a (completely) new env' as well? If then we may need to separate the actions, duplicate or create new one.
> 
> If not, we should say 'duplicate', or 'copy from' with different icon instead of 'create new' with + icon.

Each time user add an environment to his project, the toolkit will **copy the env related configuration** (e.g. bicep parameters, manifest configs, etc) from an existing environment. In this sense, we don't want to provide "creating a (completely) new env" as well, which means **a new env can always be created by copying from existing one**. 

For the command naming as well as the icon, I just reference Visual Studio for the Cloud Service environment manager. It uses "**create copy**" to add new env for the cloud service. And I'm open to any suggestion on this for better understanding from user perspective :)

image

PR #2236, opened 2021-09-08
perf: add icon for grant permission

KennethBWSong:


packages/vscode-extension/src/handlers.ts
```diff
@@ -856,6 +866,31 @@ export async function listCollaborator(env: string): Promise<TreeItem[]> {
   return result;
 }
 
+export async function checkPermission(env: string): Promise<boolean> {
+  let result = false;
+  try {
+    const checkCoreRes = checkCoreNotEmpty();
+    if (checkCoreRes.isErr()) {
+      throw checkCoreRes.error;
```
SLdragon, line 874: If it throw error, what will happen on the tree view

KennethBWSong: result will be assigned false, and thus it will return false. The icon should not be displayed

SLdragon (approved):

PR #2249, opened 2021-09-09
chore(sql): move checking sql existing into pre-provision

xzf0587:
1. move checking sql existing into pre-provision step.
2. update  getQuestions for only CLI_HELP scenario


packages/fx-core/src/plugins/resource/sql/plugin.ts
```diff
@@ -83,70 +84,26 @@ export class SqlPluginImpl {
...
-    }
-
-    if (!this.config.prepareQuestions) {
-      this.config.existSql = true;
-    }
+    await this.loadSkipAddingUser(ctx);
+    await this.checkSqlExisting(ctx);
 
     if (!this.config.existSql) {
+      await this.askInputs(ctx);
```
adashen, line 106: so we still only ask password when admin not exists? So in re-provision with rg deleted, question is not asked?

xzf0587: In no-arm support scenario,  the flag config.existSql depends on checking sql endpoint. It will ask password if check endpoint failed. It works well in re-provision regardless rg deleted or not.
In arm support scenario, the flag config.existSql depends on the admin existing. The admin will put in config once provision succeeded.
Currently it will fail for creating new sql instance if admin is existing and rg is deleted, since password is required but we store admin but not password.

blackchoey (approved):

PR #2252, opened 2021-09-09
fix(function): clean up function config condition

dilin-MS:


packages/fx-core/src/plugins/resource/function/plugin.ts
```diff
@@ -400,14 +400,8 @@ export class FunctionPluginImpl {
...
-    const resourceGroupName = this.checkAndGet(
-      this.config.resourceGroupName,
-      FunctionConfigKey.resourceGroupName
-    );
-    const subscriptionId = this.checkAndGet(
-      this.config.subscriptionId,
-      FunctionConfigKey.subscriptionId
-    );
+    const resourceGroupName = this.getFunctionAppResourceGroupName(ctx);
+    const subscriptionId = this.getFunctionAppSubscriptionId(ctx);
```
hund030, line 404: In my understanding, if arm enabled, this provision method will never be invoked. Do we need to change this?

henzhang-ms (commented):
Overall looks good to me. Let @hund030 give the final approve.

packages/fx-core/src/plugins/resource/function/plugin.ts
dilin-MS: The method `getFunctionAppResourceGroupName()` is used to get resource group name. The method will handle different situation, arm-enabled or non-arm situation. The caller just call the method to get resource group name without handling how to get it.

hund030 (approved):

PR #2254, opened 2021-09-09
chore: send telemetry when fail to read deployment info

hund030:
By comparing `count EventName=deployment-info-not-found` and `distinct count MachineId+ProjectId`, we can find out how many users remove the .deployment folder and decide whether to remove the skip deployment feature.


henzhang-ms (approved):
good idea, very happy to see we get used to make decision with BI data.

PR #2257, opened 2021-09-09
fix(multi-env): add local settings and userdata to git ignore

qinezh:
Update the .gitignore content while scaffolding new project when multi-env is enabled.

image

### DIFF
--- a/packages/fx-core/src/core/index.ts
+++ b/packages/fx-core/src/core/index.ts
@@ -721,7 +721,7 @@ export class FxCore implements Core {
             description: "",
             author: "",
             scripts: {
-              test: "echo \"Error: no test specified\" && exit 1",
+              test: 'echo "Error: no test specified" && exit 1',
             },
             devDependencies: {
               "@microsoft/teamsfx-cli": "0.*",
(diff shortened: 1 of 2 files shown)


dooriya (approved):

packages/fx-core/src/core/index.ts
```diff
@@ -731,7 +734,9 @@ export class FxCore implements Core {
       );
       await fs.writeFile(
         path.join(inputs.projectPath!, `.gitignore`),
-        `node_modules\n/.${ConfigFolderName}/*.env\n/.${ConfigFolderName}/*.userdata\n.DS_Store\n${ArchiveFolderName}\n${ArchiveLogFileName}`
+        isMultiEnvEnabled()
```
a1exwang, line 737: Is something like this easier to read and maintain?

```typescript
const ignoreFilePatterns = [
  "node_modules",
  `.${ConfigFolderName}/${InputConfigsFolderName}/${localSettingsFileName}`,
  `.${ConfigFolderName}/${PublishProfilesFolderName}/*.userdata`,
  ".DS_Store",
  ...
];
const content = ignoreFilePatterns.join("\n");
```

qinezh: Looks better. Updated~

a1exwang (approved):

jayzhang (approved):

PR #2262, opened 2021-09-09
refactor(multi-env): move getActiveEnv to core

a1exwang:
- move `getActiveEnv` to core so extension and CLI can share the function
- add error handling for `getActiveEnv`
- add error handling for callers of `getActiveEnv`

### DIFF
--- a/packages/cli/src/cmds/config.ts
+++ b/packages/cli/src/cmds/config.ts
@@ -302,7 +302,7 @@ export class ConfigSet extends YargsCommand {
     }
     const writeFileResult = writeSecretToFile(secretData, rootFolder);
     if (writeFileResult.isErr()) {
-      return err(writeFileResult.error);
+      return writeFileResult;
     }
     CLILogProvider.necessaryLog(
       LogLevel.Info,
--- a/packages/cli/src/cmds/config.ts
+++ b/packages/cli/src/cmds/config.ts
@@ -300,7 +300,10 @@ export class ConfigSet extends YargsCommand {
       }
       secretData[option] = encrypted.value;
     }
-    writeSecretToFile(secretData, rootFolder);
+    const writeFileResult = writeSecretToFile(secretData, rootFolder);
+    if (writeFileResult.isErr()) {
+      return err(writeFileResult.error);
+    }
     CLILogProvider.necessaryLog(
       LogLevel.Info,
       `Successfully configured project setting ${option}.`
--- a/packages/cli/src/utils.ts
+++ b/packages/cli/src/utils.ts
@@ -34,7 +34,12 @@ import {
 import { ConfigNotFoundError, InvalidEnvFile, ReadFileError } from "./error";
 import AzureAccountManager from "./commonlib/azureLogin";
 import { FeatureFlags } from "./constants";
-import { isMultiEnvEnabled, getActiveEnv, environmentManager } from "@microsoft/teamsfx-core";
+import {
+  isMultiEnvEnabled,
+  getActiveEnv,
+  environmentManager,
+  WriteFileError,
+} from "@microsoft/teamsfx-core";
 
 type Json = { [_: string]: any };
 
@@ -242,14 +247,27 @@ export async function readProjectSecrets(
   }
 }
 
-export function writeSecretToFile(secrets: dotenv.DotenvParseOutput, rootFolder: string): void {
-  const secretFile = `${rootFolder}/.${ConfigFolderName}/${getActiveEnv(rootFolder)}.userdata`;
+export function writeSecretToFile(
+  secrets: dotenv.DotenvParseOutput,
+  rootFolder: string
+): Result<undefined, FxError> {
+  const envResult = getActiveEnv(rootFolder);
+  if (envResult.isErr()) {
+    return err(envResult.error);
+  }
+
+  const secretFile = `${rootFolder}/.${ConfigFolderName}/${envResult.value}.userdata`;
   const array: string[] = [];
   for (const secretKey of Object.keys(secrets)) {
     const secretValue = secrets[secretKey];
     array.push(`${secretKey}=${secretValue}`);
   }
-  fs.writeFileSync(secretFile, array.join("\n"));
+  try {
+    fs.writeFileSync(secretFile, array.join("\n"));
+  } catch (e) {
+    return err(WriteFileError(e));
+  }
+  return ok(undefined);
 }
 
 export async function getSolutionPropertyFromEnvFile(
--- a/packages/cli/src/utils.ts
+++ b/packages/cli/src/utils.ts
@@ -125,20 +125,22 @@ export function getConfigPath(projectFolder: string, filePath: string): string {
 }
 
 // TODO: move config read/write utils to core
-export function getEnvFilePath(projectFolder: string): string {
+export function getEnvFilePath(projectFolder: string): Result<string, FxError> {
   if (isMultiEnvEnabled()) {
-    // NOTE: The fallback logic is to prevent this function from returning undefined, which will cause a lot of useless refactoring work.
-    // Because sooner or later we will move the config file related util functions to core, it is meaningless to refactor here.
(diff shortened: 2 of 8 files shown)


kuojianlu (approved):

jayzhang (approved):

packages/fx-core/src/core/tools.ts
```diff
@@ -118,6 +119,24 @@ export function isValidProject(workspacePath?: string): boolean {
...
+  if (!isMultiEnvEnabled()) {
+    return "default";
+  }
+  try {
+    if (isValidProject(projectRoot)) {
+      const settingsJsonPath = path.join(
+        projectRoot,
+        `.${ConfigFolderName}/${InputConfigsFolderName}/${ProjectSettingsFileName}`
+      );
+      const settingsJson = JSON.parse(fs.readFileSync(settingsJsonPath, "utf8"));
```
qinezh, line 132: async method would be always preferred: `fs.readFileSync`

packages/cli/src/utils.ts
```diff
@@ -222,7 +222,7 @@ export async function readProjectSecrets(
 }
 
 export function writeSecretToFile(secrets: dotenv.DotenvParseOutput, rootFolder: string): void {
-  const secretFile = `${rootFolder}/.${ConfigFolderName}/${getActiveEnv()}.userdata`;
+  const secretFile = `${rootFolder}/.${ConfigFolderName}/${getActiveEnv(rootFolder)}.userdata`;
```
qinezh, line 225: do we need to check if `getActiveEnv(rootFolder)` return undefined?

packages/fx-core/src/core/tools.ts
```diff
@@ -118,6 +119,24 @@ export function isValidProject(workspacePath?: string): boolean {
...
+      return settingsJson.activeEnvironment;
+    }
+  } catch (e) {
+    return undefined;
```
qinezh, line 136: should we handle the error like `InvalidJson` or `activeEnvironment` is not set in projectSettings.json?

a1exwang: `InvalidJson` and other IO errors are handled and returns undefined. If activeEnvironment is not set, it will return undefined. But I think it makes sense to check whether `activeEnvironment` is a string

packages/fx-core/src/core/tools.ts
a1exwang: Currently, the config related util functions are all sync functions. We need to refactor other util functions that use this function. It makes sense to change them to async but the refactor effort kind of big, so I won't cover in this pr.

packages/fx-core/src/core/tools.ts
qinezh: We need to report accurate user error, sometimes it would be `InvalidJson` and sometimes it would be `activeEnvironment is missing`. My point the return value of `undefined` cannot help us distinguish above cases.

packages/fx-core/src/core/tools.ts
```diff
@@ -118,6 +119,24 @@ export function isValidProject(workspacePath?: string): boolean {
   }
 }
 
+export function getActiveEnv(projectRoot: string): string | undefined {
```
dooriya, line 122: Since you return `undefined`, please make sure you will not reference an `undefined` value before using the active env returned by this function.

packages/fx-core/src/core/tools.ts
a1exwang: It makes sense. I will let the function to return `Promise<Result<string, FxError>>`

packages/fx-core/src/core/tools.ts
qinezh: Sure thing.

HuihuiWu-Microsoft (approved):
LGTM for extension part

packages/cli/src/utils.ts
a1exwang: Added error handling

packages/fx-core/src/core/tools.ts
a1exwang: Yes, it makes sense. Added error handling

packages/fx-core/src/core/tools.ts
```diff
@@ -98,6 +102,15 @@ export function validateSettings(projectSettings?: ProjectSettings): string | un
         return `${PluginNames.APIM} setting is missing in settings.json`;
     }
   }
+
+  if (isMultiEnvEnabled()) {
+    if (
+      !projectSettings.activeEnvironment ||
+      typeof projectSettings.activeEnvironment !== "string"
+    ) {
+      return `activeEnvironment is missing or not a string in settings.json`;
```
qinezh, line 111: settings.json => projectSettings.json?

a1exwang: fixed

packages/cli/src/utils.ts
```diff
@@ -314,6 +357,11 @@ export function getLocalTeamsAppId(rootfolder: string | undefined): any {
   }
 
   if (isWorkspaceSupported(rootfolder)) {
+    // TODO: read local teams app ID from localSettings.json instead of env file
+    if (isMultiEnvEnabled()) {
+      return undefined;
+    }
```
dooriya, line 363: Will you handle reading local Teams app ID from `localSettings.json`? otherwise it will break local debug if multi-env is enabled.

Alive-Fish (approved):

packages/cli/src/utils.ts
a1exwang: It won't break local debug in CLI because this function is only used for telemetry. The sideloading teams app id is retrieve by `executeUserTask`

dooriya: I see. There's another `getLocalTeamsAppId` in `preview.ts`, which is used to preview a local Teams app. Can we consolidate this two APIs into one, since this is only use for local preview?

a1exwang: In my understanding,
- `getLocalTeamsAppId()` is only for local debug. And this is the localdebug teams app id.
- `getTeamsAppId()` is for all telemetry and this is the remote appId. Other operations like provision, deploy, etc. can be linked together with this appid in telemetry. In multi-env scenarios, we need to use the current active env teamsapp id.

They are not the same thing. So I don't think we should merge them into one function, because I think one function should do one thing.

qinezh (approved):

packages/cli/src/utils.ts
dooriya: Both these two APIs are `getLocalTeamsAppId()`, that's why I'm considering merge these tow APIs into one. 

image

image

`

packages/cli/src/cmds/config.ts
```diff
@@ -300,7 +300,10 @@ export class ConfigSet extends YargsCommand {
       }
       secretData[option] = encrypted.value;
     }
-    writeSecretToFile(secretData, rootFolder);
+    const writeFileResult = writeSecretToFile(secretData, rootFolder);
+    if (writeFileResult.isErr()) {
+      return err(writeFileResult.error);
```
henzhang-ms, line 305: why not just return writeFileResult

packages/cli/src/utils.ts
a1exwang: Sorry, I forgot to expand the code on github. You are correct. I will refactor later.

packages/cli/src/cmds/config.ts
a1exwang: Yes, we can do that.

packages/cli/src/utils.ts
dooriya: We can refine this in a another PR~BTW, you can use the Microsoft CodeFlow tool to do code review, which will be more clearer than GitHub :).

dooriya (approved):

packages/cli/src/utils.ts
a1exwang: Thanks for the advice~

PR #2264, opened 2021-09-09
chore(multi-env): mark active env on env selection for creating env copy

dooriya:
Append `(active)` on the active env when doing env selection for create env copy:
image

### DIFF
--- a/packages/fx-core/src/core/middleware/envInfoLoader.ts
+++ b/packages/fx-core/src/core/middleware/envInfoLoader.ts
@@ -43,7 +43,7 @@ import { newEnvInfo } from "../tools";
 import { CoreHookContextV2 } from "../v2";
 
 const newTargetEnvNameOption = "+ new environment";
-const lastUsedMark = " (active)";
+const activeMark = " (active)";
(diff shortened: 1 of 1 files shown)


packages/fx-core/src/core/middleware/envInfoLoader.ts
```diff
@@ -43,7 +43,7 @@ import { newEnvInfo } from "../tools";
 import { CoreHookContextV2 } from "../v2";
 
 const newTargetEnvNameOption = "+ new environment";
-const lastUsedMark = " (activate)";
+const lastUsedMark = " (active)";
```
qinezh, line 46: nit: the variable name can be updated for its new meaning, maybe `activeMark` or something better.

qinezh (approved):

packages/fx-core/src/core/middleware/envInfoLoader.ts
dooriya: Sounds good. Updated.

jayzhang (approved):

PR #2268, opened 2021-09-09
feat(dotnet-sdk): make configuration to options pattern and utilize DI

tecton:


jiayi-ruan (commented):
reminder: update readme doc

jiayi-ruan (approved):

PR #2269, opened 2021-09-09
perf(arm support): support frontend hosting resource provisoned to customized resource group

dilin-MS:
Support frontend hosting resource provisoned to customized resource group.

Task Item:
- [[frontend hosting] bicep output storage resource id instead of name ](https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10793636)
- [[frontend hosting] use storage id instead of name to deploy app code ](https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10793638)


packages/fx-core/src/plugins/resource/frontend/configs.ts
```diff
@@ -109,6 +80,67 @@ export class FrontendConfig {
...
+    }
+    return result;
+  }
+
+  static getStorageResourceId(ctx: PluginContext, getFromArmOutput = false): string {
+    const result = getFromArmOutput
+      ? getArmOutput(ctx, ArmOutput.FrontendStorageResourceId)
+      : ctx.config.getString(FrontendConfigInfo.StorageResourceId);
+    if (!result) {
+      throw new Error("Failed to get storaget accounts resource id");
```
hund030, line 111: This error eventually be a `Unhandled Error`. better to throw `InvalidConfigError("resourceId")`. 
Since arm output is not config. Maybe you can define a new error type.

packages/fx-core/src/plugins/resource/frontend/configs.ts
```diff
@@ -109,6 +80,67 @@ export class FrontendConfig {
...
+
+  static getSubscriptionId(ctx: PluginContext, getFromArmOutput = false): string {
+    const result = isArmSupportEnabled()
+      ? getSubscriptionIdFromResourceId(FrontendConfig.getStorageResourceId(ctx, getFromArmOutput))
+      : FrontendConfig.getConfig<string>(
+          DependentPluginInfo.SubscriptionId,
+          ctx.envInfo.profile.get(DependentPluginInfo.SolutionPluginName)
+        );
+    if (!result) {
+      throw new Error("Failed to get subscription id");
```
hund030, line 124: Also here.

packages/fx-core/src/plugins/resource/frontend/configs.ts
```diff
@@ -109,6 +80,67 @@ export class FrontendConfig {
       });
   }
 
+  static getStorageName(ctx: PluginContext, getFromArmOutput = false): string {
+    let result = isArmSupportEnabled()
+      ? getStorageAccountNameFromResourceId(
```
hund030, line 85: Since this method throw error, we need to handle it.

packages/fx-core/src/plugins/resource/frontend/configs.ts
dilin-MS: Fixed.

packages/fx-core/src/plugins/resource/frontend/configs.ts
dilin-MS: The error will be thrown and handled by caller.

packages/fx-core/src/plugins/resource/frontend/configs.ts
dilin-MS: Fixed.

packages/fx-core/src/plugins/resource/frontend/configs.ts
dilin-MS: Fixed.

hund030 (approved):

PR #2276, opened 2021-09-10
perf(collaboration): add telemetry, log, update warning message

SLdragon:


packages/vscode-extension/src/handlers.ts
```diff
@@ -847,18 +847,33 @@ export async function listCollaborator(env: string): Promise<TreeItem[]> {
...
           commandId: `fx-extension.listcollaborator.${env}`,
-          label: "No permission to list all collaborators.",
+          label: "No permission to list collaborators",
           icon: "warning",
           isCustom: true,
           parent: "fx-extension.environment." + env,
         },
       ];
     }
+    ExtTelemetry.sendTelemetryEvent(Stage.listCollaborator, {
```
KennethBWSong, line 857: Also send start telemetry at the beginning？

SLdragon: OK, thanks for the tip

packages/vscode-extension/src/handlers.ts
```diff
@@ -892,7 +907,11 @@ export async function checkPermission(env: string): Promise<boolean> {
     result =
       (teamsAppPermission.roles?.includes("Administrator") ?? false) &&
       (aadPermission.roles?.includes("Owner") ?? false);
+    ExtTelemetry.sendTelemetryEvent(Stage.checkPermission, {
```
KennethBWSong, line 910: same as above. Also another question, why telemetry is not sent in grant permission?

KennethBWSong: By the way, seems log is not added?

SLdragon: processResult function at end will send the telemetry and show error dialog, so not send here

packages/vscode-extension/src/handlers.ts
SLdragon: Updated

KennethBWSong (approved):

PR #2278, opened 2021-09-10
refactor(multi-env): move getActiveEnv to environmentManager so we can mock in UT

a1exwang:
Global functions in ESM exported by reimporting from another file cannot be mocked by sinon. Take `setActiveEnv` as an example, in core's index.ts, there are code like this:
  
```
export { setActiveEnv } from "./core/middleware/envInfoLoader";
```

This will be compiled to 
```
Object.defineProperty(exports, "setActiveEnv", { enumerable: true, get: function () { return envInfoLoader_1.setActiveEnv; } });
```

From [mozzila docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty#description), we can know that this is wrapped as an readonly function (`defineProperty` without `writable: true`). This will cause sinon to fail to mock it.

This is a known issue of sinon and they said these types of issues related to module system is [out of scope](https://github.com/sinonjs/sinon/issues/1711).

One possible solution is to convert the whole core to a common js module which is too much efforts and may affect too many things.

Another solution I have tried is to manually write dependency injection code which will also affect too much because these are util functions used everywhere. And it is not consistent with other parts of CLI.

So I take the third solution to move the function to a modifiable object
- To make is possible to mock with sinon 
- and it also makes sense to make `getActiveEnv` belong to `environmentManager`.

Thanks @tecton to help me debug.

### DIFF
--- a/packages/fx-core/src/core/environment.ts
+++ b/packages/fx-core/src/core/environment.ts
@@ -397,23 +397,18 @@ class EnvironmentManager {
     if (!isMultiEnvEnabled()) {
       return ok("default");
     }
-    if (!isValidProject(projectRoot)) {
-      return err(InvalidProjectError());
-    }
+
     const settingsJsonPath = path.join(
       projectRoot,
       `.${ConfigFolderName}/${InputConfigsFolderName}/${ProjectSettingsFileName}`
     );
-    let settingsContent;
-    try {
-      settingsContent = fs.readFileSync(settingsJsonPath, "utf8");
-    } catch (error) {
-      return err(ReadFileError(error));
+    if (!fs.existsSync(settingsJsonPath)) {
+      return err(PathNotExistError(settingsJsonPath));
     }
 
     let settingsJson;
     try {
-      settingsJson = JSON.parse(settingsContent);
+      settingsJson = fs.readJSONSync(settingsJsonPath, { encoding: "utf8" });
     } catch (error) {
       return err(
         InvalidProjectSettingsFileError(
--- a/packages/fx-core/src/core/environment.ts
+++ b/packages/fx-core/src/core/environment.ts
@@ -18,6 +18,7 @@ import {
   EnvNamePlaceholder,
   EnvInfo,
   Json,
+  ProjectSettingsFileName,
 } from "@microsoft/teamsfx-api";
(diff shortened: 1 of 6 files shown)


packages/fx-core/src/core/environment.ts
```diff
@@ -385,6 +392,50 @@ class EnvironmentManager {
       return this.defaultEnvName;
     }
   }
+
+  public getActiveEnv(projectRoot: string): Result<string, FxError> {
+    if (!isMultiEnvEnabled()) {
+      return ok("default");
+    }
+    if (!isValidProject(projectRoot)) {
```
dooriya, line 400: nit: keep return consistent error in envManager if project root path doesn't exist.
Other APIs return `PathNotExistError` while `getActiveEnv` return `InvalidProjectError`:
```ts
if (!(await fs.pathExists(projectPath))) {
  return err(PathNotExistError(projectPath));
}
```

jayzhang (approved):

packages/fx-core/src/core/environment.ts
```diff
@@ -385,6 +392,50 @@ class EnvironmentManager {
...
+    let settingsJson;
+    try {
+      settingsJson = JSON.parse(settingsContent);
+    } catch (error) {
+      return err(
+        InvalidProjectSettingsFileError(
+          `Project settings file is not a valid JSON, error: '${error}'`
+        )
+      );
+    }
```
qinezh, line 423: nit: Here we can use `fs.readJSON` for simplicity.

qinezh (approved):

dooriya (approved):

tecton (approved):

packages/fx-core/src/core/environment.ts
a1exwang: It makes sense. Fixed

packages/fx-core/src/core/environment.ts
a1exwang: Changed `PathNotExistsError` to be consistent with other APIs.

nliu-ms (approved):

chagong (approved):

PR #2280, opened 2021-09-10
feat(apim): add scaffold to user task 'addResource' for V2 API

XiaofuHuang:


jayzhang (approved):

packages/fx-core/src/plugins/resource/apim/constants.ts
```diff
@@ -399,3 +399,7 @@ export enum OpenApiSchemaVersion {
   V2 = "v2",
   V3 = "v3",
 }
+
+export enum UserTask {
```
dooriya, line 403: Do we really need an `enum` type here since we only have one element?

XiaofuHuang: Changed it to string.

dooriya (approved):

PR #2282, opened 2021-09-10
refactor: add development view container and support AB testing

HuihuiWu-Microsoft:
1. add development view container
2. add A/B testing treatment variable logic
3. refactor code

native tree view: 
native_render

webview rendered tree view:
webview_render


jayzhang (approved):

packages/vscode-extension/src/commandsTreeViewProvider.ts
```diff
@@ -4,13 +4,14 @@
 import * as vscode from "vscode";
 import * as path from "path";
 import { ext } from "./extensionVariables";
```
1openwindow, line 6: how about put this ts file also under tree folder?

1openwindow (approved):

packages/vscode-extension/src/commandsTreeViewProvider.ts
HuihuiWu-Microsoft: Will do in next pr.

PR #2284, opened 2021-09-10
refactor: core

jayzhang:
1. move azure parameter create code from core to solution
2. Fix API signatures

### DIFF
--- a/packages/api/src/v2/types.ts
+++ b/packages/api/src/v2/types.ts
@@ -17,9 +17,9 @@ export interface Context {
   userInteraction: UserInteraction;
   logProvider: LogProvider;
   telemetryReporter: TelemetryReporter;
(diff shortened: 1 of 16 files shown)


packages/api/src/v2/types.ts
```diff
@@ -17,9 +17,9 @@ export interface Context {
   userInteraction: UserInteraction;
   logProvider: LogProvider;
   telemetryReporter: TelemetryReporter;
-  cryptoProvider: CryptoProvider;
+  cryptoProvider?: CryptoProvider;
```
1yefuwang1, line 20: Is there any strong reason to make cryptoProvider and permissionRequestProvideroptional?

1yefuwang1 (approved):

HuihuiWu-Microsoft (approved):

PR #2300, opened 2021-09-14
fix(local collaboration): add file path check

KennethBWSong:


packages/fx-core/src/core/middleware/projectUpgrader.ts
```diff
@@ -107,7 +111,16 @@ export async function upgradeContext(ctx: CoreHookContext): Promise<Result<undef
         EnvProfileFileNameTemplate.replace("@envName", defaultEnvName)
       )
     : path.resolve(confFolderPath, `env.${defaultEnvName}.json`);
+  if (!(await fs.pathExists(contextPath))) {
```
adashen, line 114: I saw there is also some check in https://github.com/OfficeDev/TeamsFx/blob/860c1f21f25ef8f8f0f699c414e2bc694453a6f6/packages/fx-core/src/core/middleware/projectUpgrader.ts#L195 . Can we refact a little bit to have the check in the "readXXX" method and just call the method in upgrade?

KennethBWSong: Sure. I will refact this part.

jayzhang (approved):

adashen (approved):

PR #2301, opened 2021-09-14
feat: core will support register callback on predefined events

LongOddCode:


1openwindow (approved):

packages/fx-core/src/core/callback.ts
```diff
@@ -0,0 +1,34 @@
...
+    return this.registry.has(event);
+  }
+
+  public static set(event: CoreCallbackEvent, func: CoreCallbackFunc): void {
+    if (!this.registry.has(event)) {
+      this.registry.set(event, []);
+    }
+    const funcs = this.registry.get(event) as CoreCallbackFunc[];
+    funcs.push(func);
+    this.registry.set(event, funcs);
```
HuihuiWu-Microsoft, line 19: In corner case, if the event is registered twice, the map will include duplicated callbacks and the callback will be called twice when the event fired.

LongOddCode: By design. I already test this kind of case in UT

HuihuiWu-Microsoft (approved):

1yefuwang1 (approved):

PR #2306, opened 2021-09-14
ci: webpack out of memory

Alive-Fish:


.github/workflows/cd.yml
```diff
@@ -20,6 +20,8 @@ on:
 
 jobs:
   cd:
+    env:
+      NODE_OPTIONS: "--max-old-space-size=5120"
```
wenytang-ms, line 24: if this large enough?

PR #2310, opened 2021-09-14
fix: out of memory when webpacking

Alive-Fish:


wenytang-ms (approved):

packages/cli/package.json
```diff
@@ -15,7 +15,7 @@
     "copy-files": "copyfiles -u 1 src/**/*.html src/**/*.css src/**/*.sh src/**/*.ps1 lib/",
     "build": "rimraf lib && tsc -p ./ && npm run copy-files",
     "watch": "rimraf lib && npm run copy-files && tsc -w -p ./",
-    "package": "rimraf lib && npx webpack --mode production --config ./webpack.config.js",
+    "package": "rimraf lib && node --max-old-space-size=3072 node_modules/webpack/bin/webpack.js --mode production --config ./webpack.config.js",
```
wenytang-ms, line 18: this will require VM ram size?

Alive-Fish: I'm not sure. If the process uses more memory, the process may be crashed.

chagong (approved):

PR #2311, opened 2021-09-14
feat(ab test): disable related tree items when operation in progress …

HuihuiWu-Microsoft:
…to avoid concurrency error
disable_concurrency


swatDong (approved):
LGTM for `azureLogin`. (Only import and format changes)

HuihuiWu-Microsoft:
@jaeyonglee8687 Can you help confirm if current UX is OK to you？ We'll deliver in this release if everything is fine.

jaeyonglee8687:
I am slightly lacking the contextual info, but it seems like some sidebar actions are disabled while waiting for a response from a command? Is it only applies to Provision? or does it for others as well? What's the main motivation for it?

HuihuiWu-Microsoft:
> I am slightly lacking the contextual info, but it seems like some sidebar actions are disabled while waiting for a response from a command? Is it only applies to Provision? or does it for others as well? What's the main motivation for it?

Sorry for not giving you enough info. The main purpose of this feature is to avoid concurrency error. For example, user is not able to deploy when he already clicked deploy button and in the process of deploying. Also, that applies to multiple commands like provision, validate manifest, publish....If any of these commands are in progress, user is not supposed to execute other commands since they might modify the same file and cause error. 
Our former behavior is to show concurrency error notification in this situation which is not elegant. In this feature, we'll disable those commands to indicate they're not available currently.

jaeyonglee8687:
Thanks for the explanation. I like that idea, showing the dependencies and preventing possible errors. One thing we could consider is visualizing what's in progress. See the example of Azure. If we see the duration of time takes quite a while, we can clearly visualize what's going on in which action, and show the dependencies together. 

image

This is my rough sketch for it, 

- Show what's going on (which action). 
- Explain implicitly why some actions are disabled
- Visualize the dependencies.

image


> > I am slightly lacking the contextual info, but it seems like some sidebar actions are disabled while waiting for a response from a command? Is it only applies to Provision? or does it for others as well? What's the main motivation for it?
> 
> Sorry for not giving you enough info. The main purpose of this feature is to avoid concurrency error. For example, user is not able to deploy when he already clicked deploy button and in the process of deploying. Also, that applies to multiple commands like provision, validate manifest, publish....If any of these commands are in progress, user is not supposed to execute other commands since they might modify the same file and cause error.
> Our former behavior is to show concurrency error notification in this situation which is not elegant. In this feature, we'll disable those commands to indicate they're not available currently.

HuihuiWu-Microsoft:
> Thanks for the explanation. I like that idea, showing the dependencies and preventing possible errors. One thing we could consider is visualizing what's in progress. See the example of Azure. If we see the duration of time takes quite a while, we can clearly visualize what's going on in which action, and show the dependencies together.
> 
> image
> 
> This is my rough sketch for it,
> 
> * Show what's going on (which action).
> * Explain implicitly why some actions are disabled
> * Visualize the dependencies.
> 
> image
> 
> > > I am slightly lacking the contextual info, but it seems like some sidebar actions are disabled while waiting for a response from a command? Is it only applies to Provision? or does it for others as well? What's the main motivation for it?
> > 
> > 
> > Sorry for not giving you enough info. The main purpose of this feature is to avoid concurrency error. For example, user is not able to deploy when he already clicked deploy button and in the process of deploying. Also, that applies to multiple commands like provision, validate manifest, publish....If any of these commands are in progress, user is not supposed to execute other commands since they might modify the same file and cause error.
> > Our former behavior is to show concurrency error notification in this situation which is not elegant. In this feature, we'll disable those commands to indicate they're not available currently.

That sounds good. Actually we also have the progress bar at the bottom right of VS Code to indicate which is running. But this design makes it more clear to users what's going on. Maybe we can do this as an enhancement in next release since the pre-release will be delivered today.

1openwindow (approved):

PR #2313, opened 2021-09-14
chore(vsc): add upgrade user propery in telemetry

1openwindow:
ADO: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10892498

### DIFF
--- a/packages/vscode-extension/src/utils/commonUtils.ts
+++ b/packages/vscode-extension/src/utils/commonUtils.ts
@@ -208,6 +208,6 @@ export function getAllFeatureFlags(): string[] | undefined {
   return result;
 }
 
-export function getIsUpgradeUser(): string | undefined {
-  return ext.context.globalState.get<string>(UserState.IsUpgrade);
+export function getIsExistingUser(): string | undefined {
+  return ext.context.globalState.get<string>(UserState.IsExisting);
 }
--- a/packages/vscode-extension/src/utils/commonUtils.ts
+++ b/packages/vscode-extension/src/utils/commonUtils.ts
@@ -14,7 +14,7 @@ import {
 import { environmentManager, isMultiEnvEnabled, isValidProject } from "@microsoft/teamsfx-core";
 import { workspace, WorkspaceConfiguration } from "vscode";
 import * as commonUtils from "../debug/commonUtils";
-import { ConfigurationKey, CONFIGURATION_PREFIX } from "../constants";
+import { ConfigurationKey, CONFIGURATION_PREFIX, UserState } from "../constants";
 
 export function getPackageVersion(versionStr: string): string {
   if (versionStr.includes("alpha")) {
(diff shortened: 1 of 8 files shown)


packages/vscode-extension/src/utils/commonUtils.ts
```diff
@@ -207,3 +207,7 @@ export function getAllFeatureFlags(): string[] | undefined {
 
   return result;
 }
+
+export function getIsUpgradeUser(): string | undefined {
+  return ext.context.globalState.get<string>(UserState.IsUpgrade);
```
HuihuiWu-Microsoft, line 212: Why not check this.context.globalState.get<string>(SyncedState.Version) directly?

packages/vscode-extension/src/utils/upgrade.ts
```diff
@@ -24,6 +24,11 @@ export class ExtensionUpgrade {
       syncedVersion === undefined ||
       versionUtil.compare(teamsToolkitVersion, syncedVersion) === 1
     ) {
+      // if syncedVersion is undefined, then it is not upgrade user
```
HuihuiWu-Microsoft, line 27: If user didn't upgrade, then the global state of UserState.IsUpgrade is undefined. In this case, the telemetry property will be "". Is that expected?

packages/vscode-extension/src/utils/commonUtils.ts
1openwindow: both new users and upgraded users, their synced version will be update to latest version finally. so there is not way to distinguish is new users or upgrade users.

packages/vscode-extension/src/utils/upgrade.ts
1openwindow: this will not happen. in if condition, we check **versionUtil.compare(teamsToolkitVersion, syncedVersion) === 1**, so only upgraded user will go into this if condition.

packages/vscode-extension/src/utils/commonUtils.ts
HuihuiWu-Microsoft: 1. Actually we can know that when we check the version change like you do now. But we can save it in local memory or as global state and you choose the latter one, right?
2. Another thing is if current user is upgrade user and the global state is set to true. Then next time when he start VSC and no version change, the global state is still true which will cause wrong telemtry. Let me know if I'm wrong

1openwindow: to point 2: it is expected. if user already upgrade before, then it will always true. Maybe I need to change the **upgrade user** to **existing users**, which may remove the confusion.

packages/vscode-extension/src/utils/upgrade.ts
HuihuiWu-Microsoft: Yes, as you said, only upgraded user will go into this if condition and set global state of IsUpgrade. For user who didn't upgrade, this global state is not set and when you send telemetry, the value is defined and property will be "". Actually for this condition, these events should also be treated as existed user. Maybe you can also use query statement to group them to existing user group.

packages/vscode-extension/src/utils/commonUtils.ts
HuihuiWu-Microsoft: I got your point.

HuihuiWu-Microsoft (approved):

PR #2314, opened 2021-09-14
chore(arm): improve error handling

xzf0587:
1. get deployment error message and detailed error
2. add help link and doc


xzf0587:
the error message in output pasted in
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10877773

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -734,3 +714,76 @@ async function areFoldersEmpty(folderPaths: string[]): Promise<boolean> {
...
+async function getDeploymentError(
+  deployCtx: DeployContext,
+  resourceGroupName: string,
+  deploymentName: string
+): Promise<any> {
+  let deployment;
+  try {
+    deployment = await deployCtx.client.deployments.get(resourceGroupName, deploymentName);
+  } catch (error) {
+    return undefined;
```
blackchoey, line 727: return or throw error that mentions we failed to get deployment error?

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -734,3 +714,76 @@ async function areFoldersEmpty(folderPaths: string[]): Promise<boolean> {
...
+        }
+      }
+    }
+  }
+  return deploymentError;
+}
+
+function getResourceGroupFromDeploymentId(deploymentId: string | undefined) {
+  if (deploymentId) {
+    const resultArr = deploymentId.match(/\/resourceGroups\/[a-zA-Z_0-9-]*/);
```
blackchoey, line 773: add `i` flag to do case insensitive regex match, in case resource**G**roups may be resource**g**roups?

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -734,3 +714,76 @@ async function areFoldersEmpty(folderPaths: string[]): Promise<boolean> {
...
+    if (resultArr && resultArr.length > 0) {
+      return resultArr[0].substring(16);
```
blackchoey, line 775: why we need do `substring` here?

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -734,3 +714,76 @@ async function areFoldersEmpty(folderPaths: string[]): Promise<boolean> {
...
+    }
+  }
+  return "";
+}
+
+function buildDeploymentErrorMessage(failedDeployments: string[]): Result<void, FxError> {
+  const format = failedDeployments.map((deployment) => deployment + " module");
```
blackchoey, line 782: better naming? maybe formattedDeploymentName or something else

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -734,3 +714,76 @@ async function areFoldersEmpty(folderPaths: string[]): Promise<boolean> {
...
+      if (
+        operation.properties.targetResource?.resourceType ===
+          ConstantString.DeploymentResourceType &&
+        operation.properties.targetResource?.resourceName &&
+        operation.properties.targetResource?.id
+      ) {
+        const resourceGroupName: string = getResourceGroupFromDeploymentId(
+          operation.properties.targetResource.id
+        );
+        const subError = await getDeploymentError(
```
blackchoey, line 757: the resource group name from `getResourceGroupFromDeploymentId` may be empty string, maybe we need to handle this case? by the way, when will `operation.properties.targetResource.id` be null?

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -114,16 +114,14 @@ async function pollDeploymentStatus(deployCtx: DeployContext) {
...
+    const operations = await deployCtx.client.deploymentOperations.list(
+      deployCtx.resourceGroupName,
+      deployCtx.deploymentName
     );
-    deployments.forEach(async (deployment) => {
-      if (
-        deployment.properties?.timestamp &&
-        deployment.properties.timestamp.getTime() > deployCtx.deploymentStartTime
-      ) {
+    operations.forEach(async (operation) => {
```
adashen, line 121: we didn't await is it as designed here?

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
It will not be an error if the sub-deployment is not found. 
I will throw out error except this scenario.

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
xzf0587: It is fixed. Use the common function to replace it.

packages/fx-core/src/plugins/solution/fx-solution/arm.ts

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
xzf0587: It is fixed. Thanks.

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
xzf0587: It is fixed. Use the common function to replace it. If operation.properties.targetResource.id is null, throw an error.

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
xzf0587: It is fixed. Thanks.

PR #2315, opened 2021-09-14
feat(multi-env): add "teamsfx env add" command in CLI

a1exwang:
- add `teamsfx env add <name> [--env sourceEnv]` command. If --env is not specified, it will use current active env.
- add test cases (the test cases might be somewhat too long, so I split them into 3 commits for easier reading)

Screenshot from 2021-09-14 16-57-03

The --help output
Screenshot from 2021-09-15 11-39-53

https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10921113

### DIFF
--- a/packages/cli/src/cmds/env.ts
+++ b/packages/cli/src/cmds/env.ts
@@ -112,7 +112,7 @@ class EnvAdd extends YargsCommand {
       sourceEnv = activeEnvResult.value;
     }
 
-    const validNewTargetEnvResult = await this.validNewTargetEnvName(projectDir, targetEnv);
+    const validNewTargetEnvResult = await this.validateNewTargetEnvName(projectDir, targetEnv);
     if (validNewTargetEnvResult.isErr()) {
       return err(validNewTargetEnvResult.error);
     }
--- a/packages/cli/src/cmds/env.ts
+++ b/packages/cli/src/cmds/env.ts
@@ -129,7 +129,7 @@ class EnvAdd extends YargsCommand {
     return ok(null);
   }
 
-  private async validNewTargetEnvName(
+  private async validateNewTargetEnvName(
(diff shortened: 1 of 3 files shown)


packages/cli/src/cmds/env.ts
```diff
@@ -60,6 +66,87 @@ export default class Env extends YargsCommand {
...
+      require: true,
+    });
+    return yargs.version(false).options(this.params);
+  }
+
+  public async runCommand(args: { [argName: string]: string }): Promise<Result<null, FxError>> {
+    const projectDir = args.folder || process.cwd();
+    // args.name always exists because or `require: true` in builder
+    const targetEnv = args.name as string;
+    const argsEnv = args.env as string | undefined;
```
qinezh, line 91: nit: would be more straightforward:
```suggestion
    const argsEnv = args.env as string;
```

a1exwang: The `--env` in `teamsfx add` command is optional, so it is not always a string.

qinezh (approved):

packages/cli/src/cmds/env.ts
```diff
@@ -60,6 +66,87 @@ export default class Env extends YargsCommand {
...
+
+    const result = await fxCore.createEnv(inputs);
+    if (result.isErr()) {
+      return err(result.error);
+    }
+
+    return ok(null);
+  }
+
+  private async validNewTargetEnvName(
```
dooriya, line 132: using verb instead?
```suggestion
  private async validateNewTargetEnvName(
```

a1exwang: It makes sense. Changed

Alive-Fish (approved):
LGTM

PR #2327, opened 2021-09-15
chore(multi-env): show sub and resource group in env tree

dooriya:
If the env has already been provisioned, then the resource group and sub info will be showed in the tree view:
image

### DIFF
--- a/packages/vscode-extension/src/envTree.ts
+++ b/packages/vscode-extension/src/envTree.ts
@@ -121,7 +121,7 @@ async function localSettingsExists(projectRoot: string): Promise<boolean> {
 }
 
 export async function addSubscriptionAndResourceGroupNode(env: string) {
-  if (!environmentTreeProvider) {
+  if (!environmentTreeProvider || env === LocalEnvironment) {
     return;
   }
 
--- a/packages/vscode-extension/src/envTree.ts
+++ b/packages/vscode-extension/src/envTree.ts
@@ -17,6 +17,7 @@ import { checkPermission, listCollaborator } from "./handlers";
 import { signedIn } from "./commonlib/common/constant";
 import { AppStudioLogin } from "./commonlib/appStudioLogin";
 import * as fs from "fs-extra";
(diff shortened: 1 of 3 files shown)


packages/vscode-extension/src/utils/commonUtils.ts
```diff
@@ -211,3 +217,80 @@ export function getAllFeatureFlags(): string[] | undefined {
 export function getIsExistingUser(): string | undefined {
   return ext.context.globalState.get<string>(UserState.IsExisting);
 }
+
+export function getDefaultEnv(): string {
+  return isMultiEnvEnabled() ? "dev" : "default";
+}
```
qinezh, line 223: As we already have `environmentManager.getDefaultEnvName()`, there's no need to create a duplicated one?

dooriya: It makes sense. Let me fix it

packages/vscode-extension/src/envTree.ts
```diff
@@ -117,3 +119,43 @@ async function localSettingsExists(projectRoot: string): Promise<boolean> {
   const provider = new LocalSettingsProvider(projectRoot);
   return await fs.pathExists(provider.localSettingsFilePath);
 }
+
+export async function addSubscriptionAndResourceGroupNode(env: string) {
+  if (!environmentTreeProvider) {
+    return;
+  }
+
+  const parentCommand = environmentTreeProvider.findCommand("fx-extension.environment." + env);
```
qinezh, line 128: Shall we return directly if the env value equal to local?

dooriya: Sounds good as we already include 'local' env node in the tree view now.

qinezh (approved):

a1exwang (approved):

xiaolang124 (approved):

HuihuiWu-Microsoft (approved):
LGTM

PR #2330, opened 2021-09-15
feat(multi-env): add --resource-group argument for provision command in CLI

a1exwang:
- add resource group argument for provision command
- change get resource group logic in core

It will use the specified rg without asking user
Screenshot from 2021-09-15 14-33-47

It will ask user if rg is not specified
Screenshot from 2021-09-15 14-02-53


jayzhang (approved):

Alive-Fish (approved):
Approve for now and when refactor to API@v2, the hardcode parameters may be removed.

PR #2334, opened 2021-09-15
feat(sdk): add certificate support for JS SDK

formulahendry:
- add certificate support for M365TenantCredential and onBehalfOfUserCredential
  - Read cert path from process.env.M365_CERTIFICATE_PATH
  - If both process.env.M365_CLIENT_SECRET and process.env.M365_CERTIFICATE_PATH are set, then respect process.env.M365_CERTIFICATE_PATH 
- update and add unit tests


packages/sdk/src/credential/m365TenantCredential.ts
```diff
@@ -48,16 +49,24 @@ export class M365TenantCredential implements TokenCredential {
...
-      authorityHost: config.authorityHost,
-    };
-
-    this.clientSecretCredential = new ClientSecretCredential(
-      config.tenantId!,
-      config.clientId!,
-      config.clientSecret!,
-      tokenCredentialOptions
-    );
+    if (config.certificatePath) {
```
SLdragon, line 52: If user both provide certificatePath and clientSecret, do we need to give some tips/or throw error that we will use certificate by default?

packages/sdk/src/credential/onBehalfOfUserCredential.ts
```diff
@@ -214,4 +229,37 @@ export class OnBehalfOfUserCredential implements TokenCredential {
...
   }
+
+  private parseCertificate(certificatePath: string | undefined): ClientCertificate | undefined {
+    if (!certificatePath) {
+      return undefined;
+    }
+
+    const certificateContent = fs.readFileSync(certificatePath, "utf8");
+    const certificatePattern =
+      /(-+BEGIN CERTIFICATE-+)(\n\r?|\r\n?)([A-Za-z0-9+/\n\r]+=*)(\n\r?|\r\n?)(-+END CERTIFICATE-+)/;
```
SLdragon, line 240: Add comments about this regex

formulahendry:
Close this PR, as we will use cert content instead of cert path in SDK.

PR #2341, opened 2021-09-16
fix: bot questions are removed

jayzhang:
fix:
image

### DIFF
--- a/packages/fx-core/src/plugins/resource/bot/v2/index.ts
+++ b/packages/fx-core/src/plugins/resource/bot/v2/index.ts
@@ -48,12 +48,7 @@ export class BotPluginV2 implements ResourcePlugin {
   activate(solutionSettings: AzureSolutionSettings): boolean {
     return this.plugin.activate(solutionSettings);
   }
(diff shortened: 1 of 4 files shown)


packages/fx-core/src/plugins/resource/bot/v2/index.ts
```diff
@@ -48,12 +48,7 @@ export class BotPluginV2 implements ResourcePlugin {
   activate(solutionSettings: AzureSolutionSettings): boolean {
     return this.plugin.activate(solutionSettings);
   }
-  async getQuestionsForScaffolding(
```
1yefuwang1, line 51: Why is this method removed?

YitongFeng-git (approved):
lgtm

packages/fx-core/src/plugins/resource/bot/v2/index.ts
jayzhang: Bot remove getQuestions method.

1yefuwang1 (approved):

PR #2356, opened 2021-09-16
perf(collaboration): do not send user error telemetry

SLdragon:
Do not send telemetry of user error for collaboration, these errors are related user not login or not provisioned yet


KennethBWSong (approved):

jayzhang (approved):

SLdragon:
Close this due to we have another better design to fix this

PR #2357, opened 2021-09-16
chore(multi-env): check azure account status and sub info in env tree view

dooriya:
**No azure account signed in:**
image

**No permission to the target subscription (used to provision resource in an environment):**
image


packages/vscode-extension/src/envTree.ts
```diff
@@ -121,42 +149,85 @@ async function localSettingsExists(projectRoot: string): Promise<boolean> {
...
-    envSubItems.push(resourceGroupTreeItem);
+async function checkSubscriptionPermission(
+  env: string,
+  subscriptionId: string
+): Promise<TreeItem | undefined> {
+  if (tools.tokenProvider.azureAccountProvider.getAccountInfo()) {
+    const subscriptions: SubscriptionInfo[] | undefined =
+      await tools.tokenProvider.azureAccountProvider.listSubscriptions();
+
+    const targetSub = subscriptions.find((sub) => sub.subscriptionId === subscriptionId);
```
qinezh, line 218: Need to check if subscriptions is undefined here?

dooriya: Good catch. Will fix.

PR #2358, opened 2021-09-16
fix(bot): fix duplicate detail errors

YitongFeng-git:
Due to recent api/ 's change bot plugin has duplicated error shown to user, for api now will join the inner error msg, so remove it from plugin now.

### DIFF
--- a/packages/fx-core/src/plugins/resource/bot/index.ts
+++ b/packages/fx-core/src/plugins/resource/bot/index.ts
@@ -176,13 +176,14 @@ export class TeamsBot implements Plugin {
     sendTelemetry: boolean,
     name: string
   ): FxResult {
+    let errorMsg = e.message;
     if (e.innerError) {
-      e.message += ` Detailed error: ${e.innerError.message}.`;
+      errorMsg += ` Detailed error: ${e.innerError.message}.`;
       if (e.innerError.response?.data?.errorMessage) {
-        e.message += ` Reason: ${e.innerError.response?.data?.errorMessage}`;
+        errorMsg += ` Reason: ${e.innerError.response?.data?.errorMessage}`;
       }
     }
-    Logger.error(e.message);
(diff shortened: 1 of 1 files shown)


packages/fx-core/src/plugins/resource/bot/index.ts
```diff
@@ -203,9 +204,13 @@ export class TeamsBot implements Plugin {
...
           name,
-          ResultFactory.SystemError(UnhandledErrorCode, `Got an unhandled error: ${e.message}`)
+          ResultFactory.SystemError(
+            UnhandledErrorCode,
+            `Got an unhandled error: ${e.message}`,
+            e.innerError
+          )
         );
-      return ResultFactory.SystemError(UnhandledErrorCode, e.message, e);
+      return ResultFactory.SystemError(UnhandledErrorCode, e.message, e.innerError);
```
IvanJobs, line 213: I'm wondering if the main purpose is to remove message of inner error from e.message?

packages/fx-core/src/plugins/resource/bot/index.ts
```diff
@@ -176,13 +176,14 @@ export class TeamsBot implements Plugin {
     sendTelemetry: boolean,
     name: string
   ): FxResult {
+    let errorMsg = e.message;
     if (e.innerError) {
-      e.message += ` Detailed error: ${e.innerError.message}.`;
+      errorMsg += ` Detailed error: ${e.innerError.message}.`;
```
IvanJobs, line 181: Nit: Using join instead of + would be better.

packages/fx-core/src/plugins/resource/bot/index.ts
YitongFeng-git: yes, for in api its recent change will joint the inner error msg

IvanJobs (approved):

PR #2364, opened 2021-09-17
perf(collaboration): update collaboration return type

SLdragon:


packages/fx-core/src/common/permissionInterface.ts
```diff
@@ -8,6 +8,30 @@ export interface ResourcePermission {
   roles: string[] | undefined;
 }
 
+export enum CollaborationState {
+  OK,
+  NotProvisioned,
+  M365AccountNotMatch,
```
KennethBWSong, line 14: Not match seems confusing. What do you mean by not match here?

SLdragon: User login a different m365 account that is not belong to current project

KennethBWSong: You mean whether the account has the permission? Then sth. like NoPermission may be better?

SLdragon: it is not just no permission, it is M365 account in different tenant

KennethBWSong: That's OK, and maybe we can consider about a better naming here.

adashen: Can we re-name it to show the tenant not match?

packages/fx-core/src/plugins/solution/fx-solution/solution.ts
```diff
@@ -1512,21 +1543,24 @@ export class TeamsAppSolution implements Solution {
           : "undefined",
       });
 
-      return ok(permissions);
+      return ok({
+        state: CollaborationState.OK,
+        permissions,
+      });
     } finally {
       ctx.envInfo.profile.get(GLOBAL_CONFIG)?.delete(USER_INFO);
```
1yefuwang1, line 1551: Seems like you are using the config interface as an message passing mechanism... which is not config interface was intended to be used.

SLdragon: Thanks, as discussed offline, we can add parameters to plugin to not use config. I will add a task for this

1yefuwang1 (approved):

packages/fx-core/src/common/permissionInterface.ts
SLdragon: Sure, will update

jayzhang (approved):

adashen (approved):

PR #2367, opened 2021-09-17
fix(multi-env): refresh env tree if env config added/deleted

qinezh:
Use vscode API to monitor the file changes and refresh env tree view when env config file is added or deleted.

### DIFF
--- a/packages/vscode-extension/src/handlers.ts
+++ b/packages/vscode-extension/src/handlers.ts
@@ -187,11 +187,11 @@ export async function activate(): Promise<Result<Void, FxError>> {
     if (workspacePath) {
       // refresh env tree when env config files added or deleted.
       workspace.onDidCreateFiles(async (event) => {
-        await refreshEnvTree(workspacePath, event.files);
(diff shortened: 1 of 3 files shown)


packages/vscode-extension/src/handlers.ts
```diff
@@ -196,6 +208,21 @@ export async function activate(): Promise<Result<Void, FxError>> {
   return result;
 }
 
+async function refreshEnvTree(workspacePath: string, files: readonly Uri[]) {
```
dooriya, line 211: nit: how about change `refreshEnvTree` to `refreshEnvTreeOnFileChanged`?

dooriya (approved):

packages/vscode-extension/src/handlers.ts
qinezh: sure, updated.

jayzhang (approved):

HuihuiWu-Microsoft (approved):
LGTM for extension part

PR #2369, opened 2021-09-17
chore(arm): add resource id and resource group for sql and identity

xzf0587:
1. add resource id and resource group output 
2. update config load
3. update empty admin case in sql bicep


packages/fx-core/src/plugins/resource/identity/index.ts
```diff
@@ -253,9 +239,44 @@ export class IdentityPlugin implements Plugin {
...
 
   private syncArmOutput(ctx: PluginContext) {
-    ctx.config.set(Constants.identityName, getArmOutput(ctx, IdentityArmOutput.identityName));
+    const resourceId = getArmOutput(ctx, IdentityArmOutput.identityName);
+    const resourceGroup = getResourceGroupNameFromResourceId(resourceId!);
+
+    ctx.config.set(Constants.identityName, resourceId);
     ctx.config.set(Constants.identityId, getArmOutput(ctx, IdentityArmOutput.identityId));
     ctx.config.set(Constants.identity, getArmOutput(ctx, IdentityArmOutput.identity));
+    ctx.config.set(Constants.resourceGroupName, resourceGroup);
```
adashen, line 248: so we set both resource id and resourcegroup?

xzf0587: If developers use customized resource group, it could be clear to show it in output.
As the resource id contains the resource group,  we can remove the resource group as well.

wenytang-ms:
lint-pr check failed, please rebase dev and update this pr

PR #2370, opened 2021-09-17
fix(bot): re-create bot on Appstudio failed due to non idempotent of AppStudio API

YitongFeng-git:
fix bug https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10941861

### DIFF
--- a/packages/fx-core/src/plugins/resource/bot/resources/messages.ts
+++ b/packages/fx-core/src/plugins/resource/bot/resources/messages.ts
@@ -46,6 +46,8 @@ export class Messages {
     "Falling back to use local template zip.";
   public static readonly ResourceProviderExist = (rp: string): string =>
     `Resource Provider ${rp} already exists, do not need to register.`;
+  public static readonly BotResourceExist = (where: string): string =>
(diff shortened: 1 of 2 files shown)


packages/fx-core/src/plugins/resource/bot/resources/messages.ts
```diff
@@ -46,6 +46,8 @@ export class Messages {
     "Falling back to use local template zip.";
   public static readonly ResourceProviderExist = (rp: string): string =>
     `Resource Provider ${rp} already exists, do not need to register.`;
+  public static readonly BotResourceExist = (where: string): string =>
+    `Bot resource already existed on ${where}, skip creating Bot resource.`;
```
IvanJobs, line 50: Nit: The bot resource is already existing.

IvanJobs (approved):
lgtm

packages/fx-core/src/plugins/resource/bot/resources/messages.ts
eriolchan: I think Yitong's wording is OK, no need to change

eriolchan (approved):

PR #2380, opened 2021-09-17
fix(cli): fix rootFolder undefined when publishing

a1exwang:
we should use `inputs.projectRoot` rather than `inputs.folder` because `argsToInputs` has already converted inputs.folder to inputs.projectPath.

This bug should also exist in old versions, but will only cause `AzureAccountManager.setSubscription()`. not called before calling core's `publishApplication`.
But in multienv case, it will cause active environment not set.

Screenshot from 2021-09-17 17-59-32

https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10950624/

### DIFF
--- a/packages/cli/src/cmds/publish.ts
+++ b/packages/cli/src/cmds/publish.ts
@@ -41,8 +41,7 @@ export default class Publish extends YargsCommand {
       CliTelemetry.sendTelemetryEvent(TelemetryEvent.PublishStart);
       result = await activate();
     } else {
-      const rootFolder = answers["folder"] as string;
-      delete answers.folder;
+      const rootFolder = answers.projectPath || ".";
       CliTelemetry.withRootFolder(rootFolder).sendTelemetryEvent(TelemetryEvent.PublishStart);
       result = await activate(rootFolder);
     }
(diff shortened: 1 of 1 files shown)


packages/cli/src/cmds/publish.ts
```diff
@@ -41,8 +41,7 @@ export default class Publish extends YargsCommand {
       CliTelemetry.sendTelemetryEvent(TelemetryEvent.PublishStart);
       result = await activate();
     } else {
-      const rootFolder = answers["folder"] as string;
-      delete answers.folder;
+      const rootFolder = answers.projectPath || ".";
```
qinezh, line 44: `|| "."` 

as you mentioned, the logic has already been handled by `argsToInputs` and it already has similar fallback, do we still need it here?

a1exwang: Not needed in theory. It is because the type of `projectPath` is `string | undefined`. Changed to `projectPath!` to prevent confusion

Alive-Fish (approved):

qinezh (approved):

PR #2388, opened 2021-09-18
chore(arm): add resource id for sql and identity

xzf0587:
1.add resource id and resource group output
2.update config load
3.update empty admin case in sql bicep


wenytang-ms:
sorry, seems bicep files not filter by prettier, I will commit PR to by-pass format check for bicep

PR #2391, opened 2021-09-18
chore(arm): add resource id for sql and identity

xzf0587:
1.add resource id and resource group output
2.update config load
3.update empty admin case in sql bicep

### DIFF
--- a/packages/fx-core/src/plugins/resource/sql/plugin.ts
+++ b/packages/fx-core/src/plugins/resource/sql/plugin.ts
@@ -42,18 +42,18 @@ import { Bicep, ConstantString } from "../../../common/constants";
 import { ScaffoldArmTemplateResult } from "../../../common/armInterface";
 import * as fs from "fs-extra";
 import { getArmOutput } from "../utils4v2";
-import { getResourceGroupNameFromResourceId, isArmSupportEnabled } from "../../../common";
+import {
+  getResourceGroupNameFromResourceId,
+  getSubscriptionIdFromResourceId,
+  isArmSupportEnabled,
+} from "../../../common";
 import { IdentityArmOutput } from "../identity/constants";
 
 export class SqlPluginImpl {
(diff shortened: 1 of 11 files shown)


packages/fx-core/src/plugins/resource/sql/plugin.ts
```diff
@@ -42,42 +42,36 @@ import { Bicep, ConstantString } from "../../../common/constants";
...
 import { IdentityArmOutput } from "../identity/constants";
 
 export class SqlPluginImpl {
   config: SqlConfig = new SqlConfig();
 
-  async init(ctx: PluginContext) {
-    ContextUtils.init(ctx);
-    this.config.azureSubscriptionId = ContextUtils.getConfigString(
+  async loadConfig(ctx: PluginContext) {
+    this.config.azureSubscriptionId = ContextUtils.getConfig<string>(
```
blackchoey, line 52: It's possible that user provision the resource to another subscription, we also need to read subscriptionid from resource id

packages/fx-core/src/plugins/resource/sql/plugin.ts
```diff
@@ -434,4 +432,47 @@ export class SqlPluginImpl {
       );
     }
   }
+
+  private loadConfigResourceGroup(ctx: PluginContext) {
+    if (isArmSupportEnabled()) {
+      this.config.sqlResourceId = ctx.config.get(Constants.sqlResourceId) as string;
+      if (this.config.sqlResourceId) {
```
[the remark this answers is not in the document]
xzf0587: this function could be called before provision, so the resource id may be empty and it is not an error.

packages/fx-core/src/plugins/resource/sql/plugin.ts
xzf0587: It is fixed. Thanks.

blackchoey (approved):

PR #2394, opened 2021-09-18
fix: update azure time out logic

xiaolang124:
1. Throw timeout error when azure login does not success.

### DIFF
--- a/packages/vscode-extension/src/commonlib/azureLogin.ts
+++ b/packages/vscode-extension/src/commonlib/azureLogin.ts
@@ -161,6 +161,14 @@ export class AzureAccountManager extends login implements AzureAccountProvider {
       [TelemetryProperty.AccountType]: AccountType.Azure,
     });
     await vscode.commands.executeCommand("azure-account.login");
+    const azureAccount = this.getAzureAccount();
+    if (azureAccount.status != loggedIn) {
+      throw new UserError(
+        StringResources.vsc.codeFlowLogin.loginTimeoutTitle,
+        StringResources.vsc.codeFlowLogin.loginTimeoutDescription,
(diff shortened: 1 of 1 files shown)


wenytang-ms:
Lint-PR check fail, just rebase or merge latest dev code

swatDong (commented):
Is this only to change system error to user error? Any UX requirement?

packages/vscode-extension/src/commonlib/azureLogin.ts
```diff
@@ -161,6 +161,14 @@ export class AzureAccountManager extends login implements AzureAccountProvider {
       [TelemetryProperty.AccountType]: AccountType.Azure,
     });
     await vscode.commands.executeCommand("azure-account.login");
+    const azureAccount = this.getAzureAccount();
+    if (azureAccount.status != loggedIn) {
```
a1exwang, line 165: Does CLI also have this bug?

xiaolang124:
> Is this only to change system error to user error? Any UX requirement?

This is only fix for system error. No UX requirement.
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10957287

packages/vscode-extension/src/commonlib/azureLogin.ts
xiaolang124: CLI Azure login is handled by codeFlow and will throw error if timeout. VsCode azure-account-extension will catch this error and cause this bug.

swatDong (approved):

xiaolang124:
> Lint-PR check fail, just rebase or merge latest dev code

Thanks for the advice.

PR #2398, opened 2021-09-18
chore(dotnet-sdk): implement TeamsUserCredentialV2

tecton:
Rewrite the logic using C# and make TeamsUserCredential only depends on teams JS SDK.
Will replace TeamsUserCredential in another PR.


packages/dotnet-sdk/src/TeamsFx/TeamsUserCredential2.cs
```diff
@@ -0,0 +1,531 @@
...
+                var instance = await _teamsSdkTask.Value.ConfigureAwait(false);
+                await instance.DisposeAsync().ConfigureAwait(false);
+            }
+        }
+
+        private async ValueTask<IJSObjectReference> ImportTeamsSdk(IJSRuntime jsRuntime)
+        {
+            try
+            {
+                await jsRuntime.InvokeVoidAsync("import", "https://statics.teams.cdn.office.net/sdk/v1.7.0/js/MicrosoftTeams.min.js").ConfigureAwait(false);
```
jiayi-ruan, line 279: VS scaffolded project use v1.10.0 Teams library, maybe we can use the latest version?

jiayi-ruan (approved):

packages/dotnet-sdk/src/TeamsFx/TeamsUserCredential2.cs
tecton: Fixed.

PR #2407, opened 2021-09-22
fix(multi-env): process env is not ready when hooks init

qinezh:
Similar as the previous fix: https://github.com/OfficeDev/TeamsFx/pull/2387


a1exwang (approved):
Tested the vscode settings.

LongOddCode (approved):

PR #2427, opened 2021-09-23
fix: spfx mustache template app name

HuihuiWu-Microsoft:
Use the config value of current environment to replace manifest template app name for spfx project

Tested scaffold, build package, validate manifest, publish, provision operations for spfx under multi-env(I think current code change shouldn't have effect on other scenarios).


HuihuiWu-Microsoft:
Close this one for lint issue.

PR #2428, opened 2021-09-23
fix: spfx mustache template app name

HuihuiWu-Microsoft:
Use the config value of current environment to replace manifest template app name for spfx project

Tested scaffold, build package, validate manifest, publish, provision operations for spfx under multi-env(I think current code change shouldn't have effect on other scenarios).


packages/fx-core/src/plugins/resource/appstudio/plugin.ts
```diff
@@ -373,6 +373,11 @@ export class AppStudioPluginImpl {
     let appDefinition: IAppDefinition;
     if (this.isSPFxProject(ctx)) {
       appDefinition = this.convertToAppDefinition(manifest, false);
+
+      if (isMultiEnvEnabled()) {
+        appDefinition.appName = ctx.envInfo.config.manifest.values.appName.short;
```
nliu-ms, line 378: what if users customized other values in manifest and config file?

I'd suggest to render the mustache template in manifest string, and then convert it to manifest/app definition.

HuihuiWu-Microsoft: Yes, you're right! I was also wondering why in current logic non-spfx project used mustache to render and spfx set the values directly. I think that's the right way.

nliu-ms (approved):

PR #2429, opened 2021-09-23
fix(collaboration): fix output to show env name in output

KennethBWSong:
image

### DIFF
--- a/packages/fx-core/src/plugins/solution/fx-solution/solution.ts
+++ b/packages/fx-core/src/plugins/solution/fx-solution/solution.ts
@@ -1356,6 +1356,7 @@ export class TeamsAppSolution implements Solution {
 
       if (ctx.answers?.platform === Platform.CLI) {
         const aadAppTenantId = ctx.envInfo.profile?.get(PluginNames.AAD)?.get(REMOTE_TENANT_ID);
+        const envName = ctx.envInfo.envName ?? "dev";
 
         const message = [
           { content: `Account to grant permission: `, color: Colors.BRIGHT_WHITE },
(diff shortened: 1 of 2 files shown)


packages/fx-core/src/plugins/solution/fx-solution/solution.ts
```diff
@@ -1365,7 +1366,7 @@ export class TeamsAppSolution implements Solution {
             color: Colors.BRIGHT_WHITE,
           },
           // Todo, when multi-environment is ready, we will update to current environment
```
SLdragon, line 1368: now we can remove this todo

KennethBWSong: Removed. Thanks!

SLdragon (approved):

packages/fx-core/src/plugins/solution/fx-solution/solution.ts
```diff
@@ -1635,15 +1635,16 @@ export class TeamsAppSolution implements Solution {
 
       if (ctx.answers?.platform === Platform.CLI) {
         const aadAppTenantId = ctx.envInfo.profile?.get(PluginNames.AAD)?.get(REMOTE_TENANT_ID);
+        const envName = ctx.envInfo.envName ?? "dev";
```
1yefuwang1, line 1638: If envInfo.envName is undefined, the Toolkit is in an invalid state. I think it would be better to just fail.

KennethBWSong: Will this happen when calling into solution? Or shall we check at the beginning of every stage?

1yefuwang1: In theory, envName will not be undefined. But IMHO, checking the null-ness of envName is good defensive programming practice.

KennethBWSong: Fixed. Thanks!

1yefuwang1 (approved):

PR #2432, opened 2021-09-23
perf(collaboration): do not use config to pass parameter to plugin  

SLdragon:


packages/api/src/plugin.ts
```diff
@@ -81,9 +81,9 @@ export interface Plugin {
   /**
    * For grant and check permission in remote collaboration
    */
-  grantPermission?: (ctx: PluginContext) => Promise<Result<any, FxError>>;
+  grantPermission?: (ctx: PluginContext, userInfo: any) => Promise<Result<any, FxError>>;
```
1yefuwang1, line 84: Is it possible to change this any type to a more concrete type? If you don't want to expose too much in the interface, Json type can be used instead of any.

SLdragon: OK, thanks, will update

1yefuwang1 (approved):

KennethBWSong (approved):

nliu-ms (approved):

PR #2438, opened 2021-09-23
feat(bot): apply scaffold component api

YitongFeng-git:


packages/fx-core/src/plugins/resource/bot/resources/messages.ts
```diff
@@ -128,8 +128,8 @@ export class Messages {
     "Please check log in output channel and try to fix this issue.";
   public static readonly RegisterRequiredRP = (resourceProviders: string[]): string =>
     `Register ${resourceProviders.join(",")} resource provider for your subscription manually.`;
-  public static readonly ReopenWorkingDir =
-    "Please check whether the working directory path is valid, if no, please move the project to a valid working directory path.";
+  public static readonly ReopenWorkingDir = (path?: string): string =>
```
henzhang-ms, line 131: It would be better to remove the `?` or giving a default value. It would print 'Please check whether the path undefined exists and you have write access to it', which looks like we didn't give a  good error handling.

henzhang-ms (approved):
Overall it looks good to me.

packages/fx-core/src/plugins/resource/bot/resources/messages.ts
YitongFeng-git: OK, will add a default value

IvanJobs (approved):
lgtm

PR #2446, opened 2021-09-24
test(multi-env): add EnvInfoLoader UT for skip logic and fix a bug

a1exwang:
- when skip is set, EnvInfo should not be loaded
- add UT for skip logic
- add a test case for NonExistEnvError to prevent the coverage drop because after the bugfix, this exception is not thrown in the happy path case.

### DIFF
--- a/packages/fx-core/tests/core/hooks.test.ts
+++ b/packages/fx-core/tests/core/hooks.test.ts
@@ -1535,7 +1535,6 @@ describe("Middleware", () => {
 
   describe("LocalSettingsLoaderMW, ContextInjectorMW", () => {
     it("NoProjectOpenedError", async () => {
-      console.log("no project opened");
       const original = process.env[FeatureFlagName.MultiEnv];
       process.env[FeatureFlagName.MultiEnv] = "true";
 
--- a/packages/fx-core/tests/core/hooks.test.ts
+++ b/packages/fx-core/tests/core/hooks.test.ts
@@ -10,6 +10,8 @@ import {
   ConfigFolderName,
   ConfigMap,
   CoreCallbackEvent,
+  EnvConfig,
+  EnvInfo,
   err,
   Func,
   FunctionRouter,
@@ -23,10 +25,12 @@ import {
   ProjectSettings,
   QTreeNode,
   Result,
+  returnSystemError,
   Solution,
   SolutionContext,
   Stage,
   SystemError,
+  Tools,
   UserCancelError,
   UserError,
   UserInteraction,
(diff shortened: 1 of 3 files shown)


packages/fx-core/tests/core/hooks.test.ts
```diff
@@ -1668,4 +1675,110 @@ describe("Middleware", () => {
       }
     });
   });
+  describe("EnvInfoLoaderMW", () => {
```
qinezh, line 1678: how about update the UT description to distinguish the status that multi-env is enabled or not?

a1exwang: Good point. I forgot that we also need to test cases with multenv disabled.

packages/fx-core/tests/core/hooks.test.ts
```diff
@@ -1529,6 +1535,7 @@ describe("Middleware", () => {
 
   describe("LocalSettingsLoaderMW, ContextInjectorMW", () => {
     it("NoProjectOpenedError", async () => {
+      console.log("no project opened");
```
qinezh, line 1538: do we need to keep the log here?

qinezh (approved):

packages/fx-core/tests/core/hooks.test.ts
a1exwang: removed debug code

packages/fx-core/tests/core/hooks.test.ts
```diff
@@ -1668,4 +1674,110 @@ describe("Middleware", () => {
...
+
+      hooks(MyClass, {
+        myMethod: [MockProjectSettingsLoaderMW(), EnvInfoLoaderMW(true), SolutionContextSpyMW],
+      });
+      const my = new MyClass();
+      const res = await my.myMethod(inputs);
+      assert.isTrue(res.isOk());
+      assert(solutionContext);
+      // envInfo should be set to a default value when envInfo loading is skipped.
+      assert.isTrue(solutionContext?.envInfo.envName === environmentManager.getDefaultEnvName());
```
qinezh, line 1756: nit: assert.equal would be preferred here, as it provided more info such like expected string and the actual string when the test case failed.

packages/fx-core/tests/core/hooks.test.ts
```diff
@@ -1668,4 +1674,110 @@ describe("Middleware", () => {
...
+    });
+
+    it("skip dynamically with inputs.ignoreEnvInfo", async () => {
```
qinezh, line 1759: another case we might need to added is `input.env` is provided.

a1exwang: Yes. More case will be added in later PRs. As this PR main contains cases related to the bug fix

packages/fx-core/tests/core/hooks.test.ts
a1exwang: It makes sense. Fixed

LongOddCode (approved):

PR #2450, opened 2021-09-24
perf(arm support): bot support deploy to customized resource group

dilin-MS:
bot bicep files returns bot web app resource id. During deploy project step, bot plugin get subscription id and resource group id from bot web app resource id. So we can deploy bot azure resources to customized resource group.

Task Item
- [[bot] bicep output resource id](https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10793666)
- [[bot] deploy use bot resource id instead of resource group + sitename](https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10793667)


packages/fx-core/src/plugins/resource/bot/plugin.ts
```diff
@@ -540,28 +575,21 @@ export class TeamsBotImpl {
...
     await LanguageStrategy.localBuild(
       this.config.scaffold.programmingLanguage!,
-      packDir,
+      workingDir,
       this.config.deploy.unPackFlag === "true" ? true : false
     );
 
     await handler?.next(ProgressBarConstants.DEPLOY_STEP_ZIP_FOLDER);
-    const zipBuffer = utils.zipAFolder(packDir, DeployConfigs.UN_PACK_DIRS, [
+    const zipBuffer = utils.zipAFolder(workingDir, DeployConfigs.UN_PACK_DIRS, [
```
blackchoey, line 586: it seems there're some unrelated changes, is it caused by git merge?

packages/fx-core/src/plugins/resource/bot/plugin.ts
```diff
@@ -517,17 +543,26 @@ export class TeamsBotImpl {
...
     this.ctx = context;
     await this.config.restoreConfigFromContext(context);
+
+    this.config.provision.subscriptionId = getSubscriptionIdFromResourceId(
+      this.config.provision.botWebAppResourceId!
+    );
+    this.config.provision.resourceGroup = getResourceGroupNameFromResourceId(
+      this.config.provision.botWebAppResourceId!
+    );
+
```
blackchoey, line 553: we also need to extract the web app name from resource id?

packages/fx-core/src/plugins/resource/bot/plugin.ts
dilin-MS: No. It is just some tiny code refactoring and I change it in passing.

dilin-MS: The original `packDir` is just the same variable as the `workingDir`. They are just the same value and change the name. So I remove one of them.

packages/fx-core/src/plugins/resource/bot/plugin.ts
dilin-MS: Yes. Fixed.

blackchoey (approved):

YitongFeng-git (approved):

PR #2452, opened 2021-09-24
chore(arm): improve error handling

xzf0587:
1. update poll status to remove duplicated status print.
2. make deployment error message clean.


xzf0587:
the error message
image

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -115,26 +115,45 @@ type DeployContext = {
 export async function pollDeploymentStatus(deployCtx: DeployContext) {
   const failedCount = 4;
   let tryCount = 0;
+  let preStatus: { [key: string]: string } = {};
```
blackchoey, line 118: better naming? maybe previousStatus or something else

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -822,3 +841,22 @@ function formattedDeploymentName(failedDeployments: string[]): Result<void, FxEr
...
+
+function formattedDeploymentError(deploymentError: any): any {
+  if (deploymentError.subErrors) {
+    const result: any = {};
+    for (const key in deploymentError.subErrors) {
+      const subError = deploymentError.subErrors[key];
+      if (subError.inner) {
+        result[key] = formattedDeploymentError(subError.inner);
+      } else {
+        if (!subError.error?.message?.includes("Template output evaluation skipped")) {
```
blackchoey, line 853: is there an error code for this? if yes, can we use the error code to filter the error? The major concern is that ARM may return different error message for a similar issue, though I'm not sure whether it will do this...

xzf0587: The error code is a general code "DeploymentOperationFailed", which contains many detailed errors. It will filter else errors if we use this error code. I will add the error code as filter with error message as well.

jayzhang (approved):

PR #2466, opened 2021-09-27
perf(collaboration): add list all collaborators

SLdragon:


packages/vscode-extension/src/handlers.ts
```diff
@@ -901,61 +901,96 @@ export async function grantPermission(env: string): Promise<Result<Void, FxError
...
 
-    const inputs: Inputs = getSystemInputs();
-    inputs.env = env;
-    const userList = await core.listCollaborator(inputs);
-    if (userList.isErr()) {
-      throw userList.error;
-    }
+  const inputs: Inputs = getSystemInputs();
+  // Todo, remove this when we have better solution not to select env
+  inputs.env = envs[0];
```
KennethBWSong, line 916: Seems we don't need this line?

SLdragon: Thanks, will update it according to the discussion

packages/vscode-extension/src/handlers.ts
```diff
@@ -901,61 +901,94 @@ export async function grantPermission(env: string): Promise<Result<Void, FxError
...
 
-  try {
-    const checkCoreRes = checkCoreNotEmpty();
-    if (checkCoreRes.isErr()) {
-      throw checkCoreRes.error;
-    }
+  let fatalError;
+  const checkCoreRes = checkCoreNotEmpty();
+  if (checkCoreRes.isErr()) {
+    fatalError = checkCoreRes.error;
```
qinezh, line 911: Need to throw the error here? Otherwise, below line 915 will fail as core is undefined/null.

SLdragon: Oh, you're right, thanks and will update

SLdragon: Updated

packages/vscode-extension/src/handlers.ts
SLdragon: Updated

KennethBWSong (approved):

PR #2474, opened 2021-09-28
perf: resource base name be appname + envname

wenytang-ms:


packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -335,9 +333,14 @@ export async function copyParameterJson(
   );
   const targetParameterFilePath = path.join(parameterFolderPath, targetParameterFileName);
   const sourceParameterFilePath = path.join(parameterFolderPath, sourceParameterFileName);
+  const targetParameterContent = await fs.readJson(sourceParameterFilePath);
+  const appName = ctx.projectSettings!.appName;
+  targetParameterContent["parameters"][resourceBaseName] = {
```
blackchoey, line 338: let's do nothing if resourceBaseName property does not exist (which means user manually deletes it)

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -676,14 +682,18 @@ function normalizeToEnvName(input: string): string {
   return input.toUpperCase().replace(/-/g, "_").replace(/\./g, "__"); // replace "-" to "_" and "." to "__"
 }
 
-function generateResourceName(ctx: SolutionContext): void {
+function generateResourceBaseName(appName: string, envName: string): string {
   const maxAppNameLength = 10;
-  const appName = ctx.projectSettings!.appName;
-  const suffix = ctx.envInfo.profile.get(GLOBAL_CONFIG)?.getString("resourceNameSuffix");
   const normalizedAppName = appName.replace(/[^a-zA-Z0-9]/g, "").toLowerCase();
+  return normalizedAppName.substr(0, maxAppNameLength) + envName.substr(0, maxAppNameLength);
```
blackchoey, line 688: for `envName.substr(0, maxAppNameLength);`, use another variable for the max length? so we can tweak the 2 values separately.

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -676,14 +682,18 @@ function normalizeToEnvName(input: string): string {
...
+}
+
+function generateResourceName(ctx: SolutionContext): void {
+  const appName = ctx.projectSettings!.appName;
+  const suffix = ctx.envInfo!.envName;
   ctx.envInfo.profile
     .get(GLOBAL_CONFIG)
-    ?.set("resource_base_name", normalizedAppName.substr(0, maxAppNameLength) + suffix);
+    ?.set("resource_base_name", generateResourceBaseName(appName, suffix));
```
blackchoey, line 696: since the resource base name is now an actual value instead of placeholder in parameter file, maybe we do not need to add this config to profile anymore?

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
wenytang-ms: ok, but we can also add the resource base name on new JSON file.

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
wenytang-ms: sure

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
wenytang-ms: sure

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -335,10 +333,12 @@ export async function copyParameterJson(
   const targetParameterFilePath = path.join(parameterFolderPath, targetParameterFileName);
   const sourceParameterFilePath = path.join(parameterFolderPath, sourceParameterFileName);
   const targetParameterContent = await fs.readJson(sourceParameterFilePath);
-  const appName = ctx.projectSettings!.appName;
-  targetParameterContent["parameters"][resourceBaseName] = {
-    value: generateResourceBaseName(appName, targetEnvName),
-  };
+  if (targetParameterContent["parameters"][resourceBaseName]) {
```
blackchoey, line 336: should we also make `"parameters"` as a const? the coding style is not aligned

blackchoey (approved):

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
wenytang-ms: sure

1yefuwang1 (approved):

PR #2479, opened 2021-09-28
fix(arm error handling): fix error handling issue

dilin-MS:
fix arm related error handling:
1. `Solution.ArmDeploymentFailed` should be `Teams Toolkit.FailedToDeployArmTemplatesToAzure`
2. `generateArmTemplate()` error should be `Teams Toolkit.FailedToGenerateArmTemplates` instead of `Teams Toolkit.FailedToDeployArmTemplatesToAzure`.
3. Path in bicep command should be wrapped in quotes in case there are blanks in the path and cause the  `The input file path cannot be specified multiple times` error.


xzf0587 (dismissed):

packages/vscode-extension/.vscode/launch.json
```diff
@@ -7,7 +7,7 @@
   "configurations": [
     {
       "name": "Run Extension",
-      "type": "pwa-extensionHost",
```
wenytang-ms, line 10: what's the difference between pwa-ext vs. ext?

jayzhang (dismissed):

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -496,7 +496,7 @@ async function compileBicepToJson(
   bicepCommand: string,
   bicepOrchestrationFilePath: string
 ): Promise<JSON> {
-  const command = `${bicepCommand} build ${bicepOrchestrationFilePath} --stdout`;
+  const command = `${bicepCommand} build "${bicepOrchestrationFilePath}" --stdout`;
```
blackchoey, line 499: is there any library to do the escape? I'm not sure whether `"` works for all cases

packages/vscode-extension/.vscode/launch.json
blackchoey: +1, why modify this?

dilin-MS: latest vscode does not support "pwa-extensionHost"

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
a1exwang: +1

From my previous experience when implementing dependency checker, it is better to use `child_process.spawn` with `shell: false` and pass in an array to prevent shell escape issues. 
somethiong like: 
```
child_process.spawn([bicepCommand, "build", bicepOrchestrationFilePath, "--stdout"], {shell: false})
```

[Node.js docs](https://nodejs.org/api/child_process.html#child_process_child_process_spawn_command_args_options
)

dilin-MS: From [Node.js docs](https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback):
```typescript
const { exec } = require('child_process');

exec('"/path/to/test file/test.sh" arg1 arg2');
// Double quotes are used so that the space in the path is not interpreted as
// a delimiter of multiple arguments.
```

blackchoey: what if the path contains `"`? the comment only said `"` solves the space issue in path. Since we need to escape the path, it's better to consider other corner cases if possible.
Of course if we confirm `"` works for other cases, we can use it.

dilin-MS: What do you mean by the path has `"`? `"` is not allowed to appeared in file or directory name.

a1exwang: 1. File name can contain `"` and [other special characters](https://ss64.com/osx/syntax-filenames.html) like `'` and `$` on macOS and Linux. 
2. If you use `child_process.exec`, Node.js will spawn a default shell (mostly `cmd.exe` on Windows and `/bin/bash` on macOS and Linux, but users can customize this which is not managed by us) and use the shell the execute the command. `"` can handle some cases. But some other examples of shell escape issues:
    - If there is a `$` in the path, bash will treat `$xxx` as an environment variable and replace them with actual values, which leads to incorrect result. 
    - There are even more dangerous cases, if the filename contains a `!something`, bash will interpret it as "executing the last command that starts with `something` in history".  If the last command is a `rm -rf xxx` or `shutdown`, this can cause issues. 
    - The users can set some other shell as their default shell. In this case we don't even know what the escape rules for these shells are.

    These are just examples of some shell escape issues and there are more.

blackchoey (commented):
I only reviewed `arm.ts`. Please let me know if other files require review

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
dilin-MS: I will try to replace `child_process.exec()` with `child_process.spawn()` but will leave it to next PR to avoid making this PR to big.

[Task Item](https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/11043412)

packages/fx-core/src/core/middleware/projectMigrator.ts
```diff
@@ -458,7 +458,7 @@ async function generateArmTempaltesFiles(ctx: CoreHookContext) {
     throw err(
       returnSystemError(
         new Error(`Failed to generate ${parameterEnvFileName} on migration`),
-        "Solution",
+        CoreSource,
```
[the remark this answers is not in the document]
dilin-MS: This is to replace `Teams Toolkit.FailedToGenerateArmTemplates` to `Solution.FailedToGenerateArmTemplates`. The changes in  these files is just to use a constant string `SolutionSource` to replace the string `"Solution"`. And here in *packages/fx-core/src/core/middleware/projectMigrator.ts* the changes is that the error is thrown by the core component instead of the solution component. It might be an issue and fixed in this pr. @wenytang-ms could you please review it? Thx!

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -444,7 +441,7 @@ async function doGenerateArmTemplate(ctx: SolutionContext): Promise<Result<any,
 
     // Output .gitignore file
     const gitignoreContent = await fs.readFile(
-      path.join(getTemplatesFolder(), "plugins", "solution", "armGitignore"),
+      path.join(getTemplatesFolder(), "plugins", SolutionSource, "armGitignore"),
```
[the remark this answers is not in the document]
dilin-MS: Nice catch. This is a mischanged code. Fixed.

jayzhang (approved):

jayzhang (approved):

KennethBWSong (approved):

PR #2483, opened 2021-09-28
fix(bot): recreate secret to adapt to collaboration scenario

YitongFeng-git:
fix bug https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10957268

### DIFF
--- a/packages/fx-core/src/plugins/resource/bot/aadRegistration.ts
+++ b/packages/fx-core/src/plugins/resource/bot/aadRegistration.ts
@@ -6,7 +6,7 @@ import { AxiosInstance, default as axios } from "axios";
 import { AADRegistrationConstants } from "./constants";
 import { IAADDefinition } from "./appStudio/interfaces/IAADDefinition";
 import { AppStudio } from "./appStudio/appStudio";
-import { ProvisionError } from "./errors";
+import { CheckThrowSomethingMissing, ProvisionError } from "./errors";
 import { CommonStrings } from "./resources/strings";
 import { BotAuthCredential } from "./botAuthCredential";
 
@@ -72,7 +72,9 @@ export class AADRegistration {
 
   public static async registerAADAppAndGetSecretByAppStudio(
     appStudioToken: string,
-    displayName: string
+    displayName: string,
+    objectId?: string,
+    msAppId?: string
   ): Promise<BotAuthCredential> {
     const result = new BotAuthCredential();
 
@@ -80,10 +82,18 @@ export class AADRegistration {
       displayName: displayName,
     };
 
-    const app = await AppStudio.createAADAppV2(appStudioToken, appConfig);
-    result.clientId = app.appId;
-    result.objectId = app.id;
+    if (!objectId && !msAppId) {
+      const app = await AppStudio.createAADAppV2(appStudioToken, appConfig);
+      result.clientId = app.appId;
+      result.objectId = app.id;
+    } else {
+      CheckThrowSomethingMissing("objectId", objectId);
+      CheckThrowSomethingMissing("msAppId", msAppId);
+      result.objectId = objectId;
+      result.clientId = msAppId;
+    }
 
+    // if has objectId, just create password for this AAD
     const password = await AppStudio.createAADAppPassword(appStudioToken, result.objectId);
 
     if (!password || !password.value) {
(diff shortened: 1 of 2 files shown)


packages/fx-core/src/plugins/resource/bot/plugin.ts
```diff
@@ -742,14 +742,15 @@ export class TeamsBotImpl {
       Logger.info(Messages.ProvisioningBotRegistration);
       botAuthCreds = await AADRegistration.registerAADAppAndGetSecretByAppStudio(
         appStudioToken!,
-        aadDisplayName
+        aadDisplayName,
+        this.config.localDebug.localObjectId
```
IvanJobs, line 746: The overall logic is changed significantly. I recommend to refactor the code to describe the new logic instead of patching on the legacy version.

packages/fx-core/src/plugins/resource/bot/plugin.ts
```diff
@@ -789,18 +790,15 @@ export class TeamsBotImpl {
...
-      if (!this.config.scaffold.botId) {
-        this.config.scaffold.botId = botAuthCreds.clientId;
-      }
-      if (!this.config.scaffold.botPassword) {
-        this.config.scaffold.botPassword = botAuthCreds.clientSecret;
-      }
-      if (!this.config.scaffold.objectId) {
-        this.config.scaffold.objectId = botAuthCreds.objectId;
-      }
+      this.config.scaffold.botId = botAuthCreds.clientId ?? this.config.scaffold.botId;
```
IvanJobs, line 797: Why change it like this? The logic is different.

eriolchan: Can you explain more about the general logic here first so that we can see if this change is reasonable?

packages/fx-core/src/plugins/resource/bot/aadRegistration.ts
```diff
@@ -72,18 +72,23 @@ export class AADRegistration {
...
-    result.clientId = app.appId;
-    result.objectId = app.id;
+    if (!objectId) {
+      const app = await AppStudio.createAADAppV2(appStudioToken, appConfig);
+      result.clientId = app.appId;
+      result.objectId = app.id;
+      objectId = app.id;
+    }
 
+    // if has objectId, just create password for this AAD
```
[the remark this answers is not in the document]
YitongFeng-git: If no appid and no objectid, create a new AAD; otherwise, the appid and objectid should both existed, and in this case, if the password is missing, create for the AAD.

packages/fx-core/src/plugins/resource/bot/plugin.ts
YitongFeng-git: For this PR, I think no need.

packages/fx-core/src/plugins/resource/bot/plugin.ts
YitongFeng-git: @IvanJobs for the above has the logic `if (!this.config.scaffold.botAADCreated()) {` , and no need to judge again, and the bothAuthCreds's value should override the value in config.scaffold

YitongFeng-git: @eriolchan explained in below comments

packages/fx-core/src/plugins/resource/bot/plugin.ts
```diff
@@ -789,18 +791,15 @@ export class TeamsBotImpl {
       );
       botAuthCreds = await AADRegistration.registerAADAppAndGetSecretByAppStudio(
         appStudioToken!,
-        aadDisplayName
+        aadDisplayName,
+        this.config.scaffold.objectId,
+        this.config.scaffold.botId
       );
 
-      if (!this.config.scaffold.botId) {
```
eriolchan, line 795: Good to refactor it as single of truth. If we refactor this code of L795, do we need to also refactor for that in local debug of L764?

packages/fx-core/src/plugins/resource/bot/aadRegistration.ts
```diff
@@ -72,18 +72,28 @@ export class AADRegistration {
...
+      result.clientId = app.appId;
+      result.objectId = app.id;
+    } else {
+      CheckThrowSomethingMissing("objectId", objectId);
+      CheckThrowSomethingMissing("msAppId", msAppId);
+      result.objectId = objectId;
+      result.clientId = msAppId;
+    }
 
+    // if has objectId, just create password for this AAD
```
IvanJobs, line 96: The comment needs to be adapted. When it comes to this line, result.objectId is definitely non-null, so remove `if has objectId` in the comment or just remove the line of comment totally.

PR #2489, opened 2021-09-28
fix(tab): Show how to use AzFn after adding it to a project incrementally

hund030:


eriolchan (approved):

templates/tab/js/default/src/components/Tab.jsx
```diff
@@ -0,0 +1,12 @@
+import React from "react";
+import { Welcome } from "./sample/Welcome";
```
eriolchan, line 2: Is there any code change in plugin to set this env variable?

hund030: No. We have already set the variable.

PR #2495, opened 2021-09-29
chore: support telemetry test

chagong:
chore: support telemetry test

To support telemetry UITEST we need to send telemetry into a log file instead of sending to AI. The test need to set a env variable  (TEAMSFX_TELEMETRY_TEST) to enable it.

image

https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10676875

### DIFF
--- a/packages/vscode-extension/src/commonlib/telemetry.ts
+++ b/packages/vscode-extension/src/commonlib/telemetry.ts
@@ -3,9 +3,21 @@
 import * as vscode from "vscode";
 import Reporter from "vscode-extension-telemetry";
 import { TelemetryReporter } from "@microsoft/teamsfx-api";
-import { getAllFeatureFlags, getPackageVersion, getProjectId } from "../utils/commonUtils";
+import {
+  getAllFeatureFlags,
+  getPackageVersion,
+  getProjectId,
+  isFeatureFlagEnabled,
+  FeatureFlags,
+} from "../utils/commonUtils";
 import { TelemetryProperty } from "../telemetry/extTelemetryEvents";
 import { Correlator } from "@microsoft/teamsfx-core";
+import { ConfigFolderName } from "@microsoft/teamsfx-api";
+import { configure, getLogger, Logger } from "log4js";
+import * as os from "os";
+import * as path from "path";
+
+const TelemetryTestLoggerFile = "telemetryTest.log";
 
 /**
  *  VSCode telemetry reporter used by fx-core.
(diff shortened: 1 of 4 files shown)


packages/vscode-extension/src/commonlib/telemetry.ts
```diff
@@ -19,19 +31,53 @@ import { Correlator } from "@microsoft/teamsfx-core";
...
   constructor(key: string, extensionVersion: string, extensionId: string) {
     super(async () => await this.reporter.dispose());
     this.reporter = new Reporter(extensionId, extensionVersion, key, true);
     this.extVersion = getPackageVersion(extensionVersion);
+    const logFile = path.join(os.homedir(), `.${ConfigFolderName}`, TelemetryTestLoggerFile);
+    configure({
+      appenders: { everything: { type: "file", filename: logFile } },
+      categories: { default: { appenders: ["everything"], level: "debug" } },
+    });
+    this.logger = getLogger("TelemTest");
```
swatDong, line 48: Is the log file created here or on the first log generated?

XiaofuHuang (approved):

packages/vscode-extension/src/commonlib/telemetry.ts
chagong: created at 'getLogger' call.

chagong: oh, was wrong, it's at 'configure' call.

HuihuiWu-Microsoft:
Is it a good practice to include testing code in production? What about stub methods of this.reporter before test cases?

packages/vscode-extension/src/commonlib/telemetry.ts
swatDong: So how about move `configure` and `getLogger` in if condition? For normal user we don't want to generate the test log file.

```
if (TelemetryTest) {
   configure...
   getLogger...
}
```

chagong:
> Is it a good practice to include testing code in production? What about stub methods of this.reporter before test cases?

It's for UITEST. It's not like UT that we can mock some objects or functions. it has to be in production code.

swatDong (approved):

HuihuiWu-Microsoft (approved):

HuihuiWu-Microsoft (approved):

swatDong (approved):

PR #2499, opened 2021-09-29
Fix/update local debug secret

xiaolang124:
1. Add new local debug secret encrypt and decrypt. (For auth.clientSecret, auth.SimpleAuthEnvironmentVariableParams, bot.botPassword)

image

### DIFF
--- a/packages/fx-core/src/common/localSettingsProvider.ts
+++ b/packages/fx-core/src/common/localSettingsProvider.ts
@@ -6,6 +6,7 @@ import * as fs from "fs-extra";
 import {
   ConfigFolderName,
   ConfigMap,
+  CryptoProvider,
   InputConfigsFolderName,
   Json,
   LocalSettings,
@@ -20,6 +21,11 @@ import {
 import { isMultiEnvEnabled } from "./tools";
 
 export const localSettingsFileName = "localSettings.json";
+
+const crypto = "crypto";
+const clientSecret = "clientSecret";
+const SimpleAuthEnvironmentVariableParams = "SimpleAuthEnvironmentVariableParams";
+const botPassword = "botPassword";
 
 export class LocalSettingsProvider {
   public readonly localSettingsFilePath: string;
@@ -106,7 +112,7 @@ export class LocalSettingsProvider {
     return localSettings;
   }
 
-  public async load(): Promise<LocalSettings | undefined> {
+  public async load(cryptoProvider?: CryptoProvider): Promise<LocalSettings | undefined> {
     if (await fs.pathExists(this.localSettingsFilePath)) {
       const localSettingsJson = await fs.readJSON(this.localSettingsFilePath);
       const localSettings: LocalSettings = {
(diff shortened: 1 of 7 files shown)


jayzhang (approved):

packages/fx-core/src/common/localSettingsProvider.ts
```diff
@@ -21,6 +22,11 @@ import { isMultiEnvEnabled } from "./tools";
 
 export const localSettingsFileName = "localSettings.json";
 
+const crypto = "crypto";
```
dooriya, line 25: Can we move these secret key names to `localSettingsConstants.ts`?

packages/fx-core/src/common/localSettingsProvider.ts
```diff
@@ -21,6 +22,11 @@ import { isMultiEnvEnabled } from "./tools";
...
+const clientSecret = "clientSecret";
+const SimpleAuthEnvironmentVariableParams = "SimpleAuthEnvironmentVariableParams";
+const botPassword = "botPassword";
```
dooriya, line 28: Can we move these secret key names to `localSettingsConstants.ts`?

packages/fx-core/src/common/localSettingsProvider.ts
```diff
@@ -117,21 +123,95 @@ export class LocalSettingsProvider {
...
+        this.decryptLocalSettings(localSettingsJson, cryptoProvider);
+      }
       return localSettingsJson;
     } else {
       return undefined;
     }
   }
-  public async save(localSettings: LocalSettings | Json): Promise<void> {
+
+  public decryptLocalSettings(
```
dooriya, line 147: How about make it `private` as the method only used in this class?

packages/vscode-extension/src/extension.ts
```diff
@@ -266,9 +270,19 @@ export async function activate(context: vscode.ExtensionContext) {
       ? `**/.${ConfigFolderName}/${PublishProfilesFolderName}/*.userdata`
       : `**/.${ConfigFolderName}/*.userdata`,
   };
+  const localDebugDataSelector = {
+    language: "json",
+    scheme: "file",
+    pattern: isMultiEnvEnabled()
+      ? `**/.${ConfigFolderName}/${InputConfigsFolderName}/localSettings.json`
```
dooriya, line 277: Use `localSettingsJsonName` constant instead.

packages/vscode-extension/src/codeLensProvider.ts
```diff
@@ -8,17 +8,22 @@ import * as vscode from "vscode";
...
+    this.localDebugRegex =
+      /( *|\t*)"((clientSecret)|(SimpleAuthEnvironmentVariableParams)|(botPassword))": "(.*)"/g;
   }
 
   public provideCodeLenses(
     document: vscode.TextDocument
   ): vscode.CodeLens[] | Thenable<vscode.CodeLens[]> {
     if (document.fileName.endsWith("userdata")) {
       return this.computeCodeLenses(document, this.userDataRegex);
+    } else if (document.fileName.endsWith("localSettings.json")) {
```
dooriya, line 25: Same here. Use `localSettingsJsonName` constant instead.

packages/vscode-extension/src/codeLensProvider.ts
```diff
@@ -51,4 +56,32 @@ export class CryptoCodeLensProvider implements vscode.CodeLensProvider {
     }
     return codeLenses;
   }
+
+  private computeCodeLensesForLocal(
```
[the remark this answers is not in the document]
xiaolang124: Merge into one method.

packages/vscode-extension/src/codeLensProvider.ts
xiaolang124: Fixed.

packages/fx-core/src/common/localSettingsConstants.ts
```diff
@@ -48,3 +48,10 @@ export const LocalSettingsBotKeys = Object.freeze({
   BotDomain: "botDomain",
   BotEndpoint: "botEndpoint",
 });
+
+export const LocalSettingsEncryptKeys = Object.freeze({
+  Crypto: "crypto",
```
dooriya, line 53: `Crypto` is not the config key name, just a constant used as secret prefix, so moving it out from `LocalSettingsEncryptKeys` would be better.

xiaolang124: Fixed.

dooriya (approved):

packages/fx-core/src/common/localSettingsConstants.ts
xiaolang124: Fixed.

PR #2507, opened 2021-09-30
perf(collaboration): use cache after create new environment

SLdragon:
1. CreateEnv in fx-core now return a string represent the newly added environment name
2. Use cache after create new environment
3. Simplify some code logic


KennethBWSong (approved):

1yefuwang1 (approved):
Changes to solution LGTM.

jayzhang (approved):

PR #2508, opened 2021-09-30
fix: add check m365 azure account logic

xiaolang124:
1. Add azure subscription check logic before provision and deploy.
2. Add m365 tenant check logic before provision and deploy.

image
image


jayzhang (approved):

packages/fx-core/src/plugins/solution/fx-solution/commonQuestions.ts
```diff
@@ -314,6 +351,10 @@ async function askCommonQuestions(
     );
   }
 
+  const m365TenantResult = await checkM365Tenant(ctx.envInfo, appstudioTokenJson);
```
a1exwang, line 354: Should we use `ctx.solutionContext.envInfo` rather than `ctx.envInfo` currently? Because ctx.envInfo is for v2, and v2 is not ready yet.

packages/fx-core/src/plugins/solution/fx-solution/commonQuestions.ts
```diff
@@ -76,23 +86,24 @@ export async function checkSubscription(
...
-  // TODO: verify valid subscription (permission)
+  // verify valid subscription (permission)
   const subscriptions = await azureAccountProvider.listSubscriptions();
   const targetSubInfo = subscriptions.find((item) => item.subscriptionId === subscriptionId);
   if (!targetSubInfo) {
     return err(
       new UserError(
         SolutionError.SubscriptionNotFound,
-        `The subscription '${subscriptionId}' is not found in the current account, please check the '${EnvConfigFileNameTemplate.replace(
+        `The subscription '${subscriptionId}'(${subscriptionName}) is not found in the current account, please check the '${EnvConfigFileNameTemplate.replace(
```
a1exwang, line 106: Only put `subscriptionName` here when subscriptionName is not undefined?

packages/fx-core/src/plugins/solution/fx-solution/commonQuestions.ts
a1exwang: OK, I was wrong.

a1exwang (dismissed):

LongOddCode (approved):

PR #2513, opened 2021-09-30
chore: add insider preview flag

blackchoey:


packages/vscode-extension/package.json
```diff
@@ -573,9 +573,9 @@
       {
         "title": "Teams Toolkit",
         "properties": {
-          "fx-extension.armSupportEnabled": {
+          "fx-extension.insiderPreview": {
             "type": "boolean",
-            "description": "Enable provision Azure resources with ARM templates. (Requires reload of VS Code)",
+            "description": "Enable insider preview features. (requires reload of VS Code)",
             "default": false
           },
```
dooriya, line 580: As we combined the feature flags, we need to remove the multi-env user setting.

blackchoey: thanks, removed it in package.json

packages/fx-core/tests/plugins/solution/solution.provision.test.ts
```diff
@@ -500,10 +500,10 @@ describe("provision() happy path for SPFx projects", () => {
   it("should succeed if app studio returns successfully", () =>
     provisionSpfxProjectShouldSucceed(false));
 
-  it("should succeed if arm support feature flag enabled", () =>
+  it.skip("should succeed if insider feature flag enabled", () =>
```
blackchoey, line 503: @nliu-ms  As discussed offline, this test case is skipped temporary. Please help enable this test case again after you helps fix this.

1openwindow (approved):

KennethBWSong (dismissed):

packages/vscode-extension/package.json
```diff
@@ -601,7 +596,7 @@
           "fx-extension.validateBicep": {
             "type": "boolean",
```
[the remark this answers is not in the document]
blackchoey: all dependency checkers has their own settings. So I didn't merge it, but turned it on by default.

KennethBWSong (approved):

jayzhang (approved):

PR #2515, opened 2021-09-30
fix(local-debug): pop-up error if user local debug in another tenant

dooriya:
If user re-run local debug in another tenant, we should pop-up error message to block the local debug:

image

User can:
- Switch back to the original M365 account to re-run local debug.
- Or, delete `localSettings.json` to re-run local debug using current tenant.


packages/fx-core/src/plugins/solution/fx-solution/solution.ts
```diff
@@ -1182,6 +1184,39 @@ export class TeamsAppSolution implements Solution {
...
+
+        const m365UserAccount = parseUserName(await ctx.appStudioToken?.getJsonObject());
+        if (m365UserAccount.isErr()) {
+          throw err(m365UserAccount.error);
+        }
+
+        if (m365TenantId.value !== localDebugTenantId) {
+          const errorMessage: string = util.format(
+            getStrings().solution.LocalDebugTenantConfirmNotice,
+            localDebugTenantId,
```
qinezh, line 1206: how about adding the tenant name to the error message for user friendly?

packages/fx-core/resource/strings.json
```diff
@@ -22,6 +22,7 @@
     "NotProvisionedNotice": "'%s' is not provisioned in the cloud yet. Please provision first.",
     "GenerateArmTemplateFailNotice": "'%s' Failed to create ARM template for required Azure services.",
     "RegenerateArmTemplateConfirmNotice": "We will update your Azure resource template based on your resource and capability selection. The existing template file will be moved to backup folder under Azure resource template folder for your reference. We recommend to initialize your project workspace with git init to track all changes.",
+    "LocalDebugTenantConfirmNotice": "You already run local debug for the Teams app in another tenant '%s'. You can signed out %s and sign in with the matched M365 account to re-use the previous local debug resource. If you really want to switch tenant for local debug, you can delete the localSettings.json file and re-run local debug.",
```
qinezh, line 25: typo: signed out => sign out?

dooriya: Nice catch! Will fix later.

qinezh (approved):

PR #2520, opened 2021-10-08
chore: add resource Id after migration

wenytang-ms:
after migration ARM and multi-env, will add sqlResourceId functionAppId storageResourceId

### DIFF
--- a/packages/fx-core/src/core/middleware/projectMigrator.ts
+++ b/packages/fx-core/src/core/middleware/projectMigrator.ts
@@ -54,11 +54,14 @@ class EnvConfigName {
   static readonly StorageName = "storageName";
   static readonly IdentityName = "identity";
   static readonly SqlEndpoint = "sqlEndpoint";
+  static readonly SqlResourceId = "sqlResourceId";
   static readonly SqlDataBase = "databaseName";
   static readonly SkuName = "skuName";
   static readonly AppServicePlanName = "appServicePlanName";
   static readonly StorageAccountName = "storageAccountName";
+  static readonly StorageResourceId = "storageResourceId";
   static readonly FuncAppName = "functionAppName";
+  static readonly FunctionId = "functionAppId";
 }
 
 class ArmParameters {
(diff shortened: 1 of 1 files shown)


packages/fx-core/src/core/middleware/projectMigrator.ts
```diff
@@ -409,6 +412,40 @@ async function removeBotConfig(ctx: CoreHookContext) {
     delete envConfig[ResourcePlugins.Bot];
     envConfig[ResourcePlugins.Bot] = { wayToRegisterBot: "create-new" };
   }
+  let needUpdate = false;
```
xzf0587, line 415: need to update function name removeBotConfig

packages/fx-core/src/core/middleware/projectMigrator.ts
```diff
@@ -409,6 +412,40 @@ async function removeBotConfig(ctx: CoreHookContext) {
...
+      EnvConfigName.StorageResourceId
+    ] = `${configPrefix}/providers/Microsoft.Storage/storageAccounts/${
+      envConfig[ResourcePlugins.FrontendHosting][EnvConfigName.StorageName]
+    }`;
+  }
+  if (needUpdate && envConfig[ResourcePlugins.AzureSQL]) {
+    envConfig[ResourcePlugins.AzureSQL][
+      EnvConfigName.SqlResourceId
+    ] = `${configPrefix}/providers/Microsoft.Sql/servers/${
+      envConfig[ResourcePlugins.AzureSQL][EnvConfigName.SqlEndpoint].split(
```
xzf0587, line 434: The sql endpoint will exist after provision. Will the migration happen in scaffolding step?

wenytang-ms: no, this middle ware not trigger on scaffolding step

packages/fx-core/src/core/middleware/projectMigrator.ts
```diff
@@ -409,6 +412,40 @@ async function removeBotConfig(ctx: CoreHookContext) {
...
+    }
+  }
+  if (needUpdate && envConfig[ResourcePlugins.FrontendHosting]) {
+    envConfig[ResourcePlugins.FrontendHosting][
+      EnvConfigName.StorageResourceId
+    ] = `${configPrefix}/providers/Microsoft.Storage/storageAccounts/${
+      envConfig[ResourcePlugins.FrontendHosting][EnvConfigName.StorageName]
+    }`;
+  }
+  if (needUpdate && envConfig[ResourcePlugins.AzureSQL]) {
```
blackchoey, line 430: we may need to handle cases that provision partially success: for example, create resource group succeeded but create sql failed because of the quota issue. 
By the way, since 'fx-resource-sql` config (as well as other config) exists in `env.default.json` before we run provision, `envConfig[ResourcePlugins.AzureSQL]` seems will always be true?

wenytang-ms: sure, thanks!

packages/fx-core/src/core/middleware/projectMigrator.ts
```diff
@@ -401,14 +404,54 @@ export async function migrateArm(ctx: CoreHookContext) {
...
   const inputs = ctx.arguments[ctx.arguments.length - 1] as Inputs;
   const fx = path.join(inputs.projectPath as string, `.${ConfigFolderName}`);
   const envConfig = await fs.readJson(path.join(fx, "env.default.json"));
   if (envConfig[ResourcePlugins.Bot]) {
     delete envConfig[ResourcePlugins.Bot];
     envConfig[ResourcePlugins.Bot] = { wayToRegisterBot: "create-new" };
   }
+  let needUpdate = false;
+  let configPrefix = "";
+  if (envConfig["solution"]["subscriptionId"] && envConfig["solution"]["resourceGroupName"]) {
```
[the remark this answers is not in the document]
wenytang-ms: sure

blackchoey (approved):

jayzhang (approved):

PR #2523, opened 2021-10-08
feat(sdk): add certificate support for JS SDK

formulahendry:
- Add certificate support for M365TenantCredential and onBehalfOfUserCredential
  - Read cert content from `certificateContent` property of `AuthenticationConfiguration`
  - If both `clientSecret` and `certificateContent` are set in `AuthenticationConfiguration`, then respect `certificateContent`
- Update and add unit tests

### DIFF
--- a/packages/sdk/src/credential/m365TenantCredential.ts
+++ b/packages/sdk/src/credential/m365TenantCredential.ts
@@ -135,8 +135,7 @@ export class M365TenantCredential implements TokenCredential {
     }
 
     if (!config.clientSecret && !config.certificateContent) {
-      missingValues.push("clientSecret");
-      missingValues.push("certificateContent");
+      missingValues.push("clientSecret or certificateContent");
     }
 
     if (!config.tenantId) {
--- a/packages/sdk/src/credential/onBehalfOfUserCredential.ts
+++ b/packages/sdk/src/credential/onBehalfOfUserCredential.ts
@@ -61,8 +61,7 @@ export class OnBehalfOfUserCredential implements TokenCredential {
     }
 
     if (!config?.authentication?.clientSecret && !config?.authentication?.certificateContent) {
-      missingConfigurations.push("clientSecret");
-      missingConfigurations.push("certificateContent");
+      missingConfigurations.push("clientSecret or certificateContent");
     }
 
     if (!config?.authentication?.tenantId) {
--- a/packages/sdk/test/unit/node/core/onBehalfOfUserCredential.spec.ts
+++ b/packages/sdk/test/unit/node/core/onBehalfOfUserCredential.spec.ts
@@ -205,7 +205,7 @@ fakeCert
     })
       .to.throw(
         ErrorWithCode,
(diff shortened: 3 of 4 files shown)


packages/sdk/src/credential/m365TenantCredential.ts
```diff
@@ -48,16 +56,28 @@ export class M365TenantCredential implements TokenCredential {
...
-      config.tenantId!,
-      config.clientId!,
-      config.clientSecret!,
-      tokenCredentialOptions
-    );
+    if (config.certificateContent) {
+      const auth: NodeAuthOptions = {
+        clientId: config.clientId!,
+        authority: getAuthority(config.authorityHost!, config.tenantId!),
+        clientCertificate: parseCertificate(config.certificateContent),
```
adashen, line 63: It will throw error if certificate is invalid right?

packages/sdk/test/unit/node/msGraphAuthProvider.spec.ts
```diff
@@ -49,13 +49,16 @@ describe("MsGraphAuthProvider Tests - Node", () => {
...
-    });
+    mockedEnvRestore = mockedEnv(
+      {
+        INITIATE_LOGIN_ENDPOINT: initiateLoginEndpoint,
+        M365_CLIENT_ID: clientId,
+        M365_CLIENT_SECRET: clientSecret,
+        M365_TENANT_ID: tenantId,
+        M365_AUTHORITY_HOST: authorityHost,
+      },
+      { clear: true }
```
blackchoey, line 60: can you share why we need `clear: true`? the concern is this may prevents us from setting a global env.

packages/sdk/src/models/configuration.ts
```diff
@@ -54,6 +54,13 @@ export interface AuthenticationConfiguration {
    */
   readonly clientSecret?: string;
 
+  /**
+   * The content of a PEM-encoded public/private key certificate. Default value comes from M365_CERTIFICATE_CONTENT environment variable
```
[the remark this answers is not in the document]
formulahendry: Good catch!

packages/sdk/src/credential/m365TenantCredential.ts
```diff
@@ -48,16 +56,28 @@ export class M365TenantCredential implements TokenCredential {
...
+
+      this.msalClient = new ConfidentialClientApplication({
+        auth,
+      });
+    } else {
+      const tokenCredentialOptions: TokenCredentialOptions = {
+        authorityHost: config.authorityHost,
+      };
+
+      this.clientSecretCredential = new ClientSecretCredential(
```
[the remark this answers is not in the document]
formulahendry: All are changed to `msal` lib now

packages/sdk/test/unit/node/msGraphAuthProvider.spec.ts
formulahendry: It is no need now. Changed back. Thx!

blackchoey (approved):

PR #2528, opened 2021-10-08
fix(aad): fix error message

KennethBWSong:
bug: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/11012874?src=WorkItemMention&src-action=artifact_link

### DIFF
--- a/packages/fx-core/src/plugins/resource/aad/constants.ts
+++ b/packages/fx-core/src/plugins/resource/aad/constants.ts
@@ -273,5 +273,5 @@ export class ConfigFilePath {
   static readonly Default = "env.default.json";
   static readonly LocalSettings = "localSettings.json";
   static readonly Profile = (env: string) => `profile.${env}.json`;
-  static readonly Input = (env: string) => `azure.parameters.${env}.json`;
+  static readonly Input = (env: string) => `config.${env}.json`;
 }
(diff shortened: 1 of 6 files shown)


packages/fx-core/src/plugins/resource/aad/constants.ts
```diff
@@ -268,3 +268,10 @@ export class TemplatePathInfo {
     "bicep"
   );
 }
+
+export class ConfigFilePath {
+  static readonly Default = "env.default.json";
+  static readonly LocalSettings = "localSettings.json";
+  static readonly Profile = (env: string) => `profile.${env}.json`;
+  static readonly Input = (env: string) => `azure.parameters.${env}.json`;
```
dooriya, line 276: The existing AAD configuration should be set in `config.${env}.json` rather than `azure.parameters.${env}.json`.
No matter user choose to use an existing AAD or create a new one, the AAD info will be persistent in the output file - `profile.${env}.json`. And the value of the  AAD client id and secret in `azure.parameters.${env}.json` will be resolved from `profile.${env}.json`

KennethBWSong: Will update, thx!

KennethBWSong: Fixed, thx!

PR #2536, opened 2021-10-09
fix: logging status notify

xiaolang124:
1. Fix logging notify logic, when signing in show singing status in tree view. (Cli does not need notify logic)

image


packages/vscode-extension/src/commonlib/codeFlowLogin.ts
```diff
@@ -61,7 +61,7 @@ export class CodeFlowLogin {
         this.account = dataCache;
         this.status = loggedIn;
       }
-    } else {
+    } else if (this.status != loggingIn) {
```
swatDong, line 64: !==

xiaolang124: Fixed.

PR #2538, opened 2021-10-09
refactor: manifest for local debug

nliu-ms:
1) For local debug, add manifest.local.json in {Root}/templates/appPackage for Azure project
2) Create manifest.local.json in project migration
3) Update param from "localdebug"/"remote" strings to boolean.
4) Some format


packages/fx-core/src/plugins/resource/appstudio/index.ts
```diff
@@ -389,9 +389,15 @@ export class AppStudioPlugin implements Plugin {
       return await this.buildTeamsPackage(ctx);
     } else if (func.method === "getAppDefinitionAndUpdate") {
       if (func.params && func.params.type && func.params.manifest) {
+        let isLocalDebug;
```
qinezh, line 392: nit: seems the logic can be simplified like `const isLocalDebug = (func.params.type as string) === "localDebug"`

packages/fx-core/src/core/middleware/projectMigrator.ts
```diff
@@ -202,6 +215,17 @@ async function migrateMultiEnv(projectPath: string): Promise<void> {
...
   await moveIconsToResourceFolder(templateAppPackage);
 
+  if (!isSPFx) {
+    const localManifest: TeamsAppManifest = createLocalManifest(
+      appName,
+      hasFrontend,
+      hasBotCapability,
+      hasMessageExtensionCapability
+    );
+    const localManifestFile = path.join(templateAppPackage, "manifest.local.json");
```
qinezh, line 225: why not use the constant `MANIFEST_LOCAL` here?

packages/fx-core/src/plugins/resource/appstudio/utils/utils.ts
```diff
@@ -0,0 +1,9 @@
+export function replaceConfigValue(config: string, id: string, value: string): string {
```
qinezh, line 1: missing copyright?

qinezh (approved):

swatDong (approved):

PR #2542, opened 2021-10-09
chore: prepare resolveTask for custom local debug

swatDong:
Resolve `npm run dev` and `npm run watch` tasks to support custom local debug.
UT added.

Workitem: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/10878741

### DIFF
--- a/packages/vscode-extension/src/debug/teamsfxTaskProvider.ts
+++ b/packages/vscode-extension/src/debug/teamsfxTaskProvider.ts
@@ -78,8 +78,69 @@ export class TeamsfxTaskProvider implements vscode.TaskProvider {
     task: vscode.Task,
     token?: vscode.CancellationToken | undefined
   ): Promise<vscode.Task | undefined> {
-    // Return undefined since all tasks are provided and fully resolved
-    return undefined;
+    // Resolve "npm run dev" and "npm run watch" tasks
+    if (vscode.workspace.workspaceFolders) {
+      const workspaceFolder: vscode.WorkspaceFolder = vscode.workspace.workspaceFolders[0];
+      const workspacePath: string = workspaceFolder.uri.fsPath;
+      if (!(await commonUtils.isFxProject(workspacePath))) {
+        vscodeLogger.error(
+          `No ${TeamsfxTaskProvider.type} project. Cannot resolve ${TeamsfxTaskProvider.type} task.`
+        );
+        return undefined;
+      }
+
+      const command: string | undefined = task.definition.command;
+      const component: string | undefined = task.definition.component;
+      if (
+        command &&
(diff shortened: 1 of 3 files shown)


packages/vscode-extension/src/debug/teamsfxTaskProvider.ts
```diff
@@ -78,8 +78,69 @@ export class TeamsfxTaskProvider implements vscode.TaskProvider {
...
+          return undefined;
+        }
+
+        const resolvedTask = new vscode.Task(
+          task.definition,
+          task.scope,
+          command,
+          TeamsfxTaskProvider.type,
+          new vscode.ShellExecution(command, { cwd: cwd, env: env }),
+          problemMatcher
```
kuojianlu, line 135: What will be the behavior if user has customized shell options (like env) and problem matcher in tasks.json?

packages/vscode-extension/src/debug/teamsfxTaskProvider.ts
```diff
@@ -78,8 +78,69 @@ export class TeamsfxTaskProvider implements vscode.TaskProvider {
...
+        task.scope !== undefined &&
+        task.scope !== vscode.TaskScope.Global &&
+        task.scope !== vscode.TaskScope.Workspace
+      ) {
+        let env: { [key: string]: string } | undefined;
+        let cwd: string | undefined;
+        let problemMatcher: string;
+        const isWatchTask = /^npm run watch/i.test(command + "");
+        if (component?.toLowerCase() === "frontend") {
+          env = isWatchTask ? undefined : await commonUtils.getFrontendLocalEnv();
```
kuojianlu, line 105: So according to our definition, `npm run watch` is just a task to watch files (like `tsc -w`) but not a task to start service directly with files watched?

swatDong: Right, and we already have `watch` in ts function package.json (https://github.com/OfficeDev/TeamsFx/blob/dev/templates/function-base/ts/default/package.json#L6)

Actually we only need `watch` for backend `func`.

packages/vscode-extension/src/debug/teamsfxTaskProvider.ts
swatDong: Based on my test:
- `options.env` is appended
- `problemMatcher` is over written
- `presentationOptions` is over written

So vscode does gracefully handle customized cases.

kuojianlu (dismissed):

kuojianlu (approved):

1openwindow (approved):

PR #2544, opened 2021-10-09
chore(arm): update identity output

xzf0587:
1. update identity output: identity resource id and identity client id
2. update function, bot, SQL reading the output
3. update unit test

### DIFF
--- a/packages/fx-core/src/plugins/resource/function/ops/provision.ts
+++ b/packages/fx-core/src/plugins/resource/function/ops/provision.ts
@@ -170,7 +170,7 @@ export class FunctionProvision {
     identityId: string,
     databaseName: string,
     sqlEndpoint: string,
-    identityName: string
+    identityResourceId: string
   ): void {
     this.pushAppSettings(site, FunctionAppSettingKeys.identityId, identityId);
     this.pushAppSettings(site, FunctionAppSettingKeys.databaseName, databaseName);
@@ -179,7 +179,7 @@ export class FunctionProvision {
     site.identity = {
       type: DefaultProvisionConfigs.siteIdentityTypeUserAssigned,
       userAssignedIdentities: {
-        [identityName]: {},
+        [identityResourceId]: {},
       },
     };
   }
--- a/packages/fx-core/src/plugins/resource/identity/config.ts
+++ b/packages/fx-core/src/plugins/resource/identity/config.ts
@@ -7,6 +7,6 @@ export class IdentityConfig {
   location = "";
 
   identity = "";
-  identityId = "";
-  identityName = "";
+  clientId = "";
+  identityResourceId = "";
 }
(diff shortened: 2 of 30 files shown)


packages/fx-core/src/plugins/resource/identity/config.ts
```diff
@@ -7,6 +7,6 @@ export class IdentityConfig {
   location = "";
 
   identity = "";
```
dilin-MS, line 9: What's this `identity` means?

packages/fx-core/src/plugins/resource/identity/config.ts
```diff
@@ -7,6 +7,6 @@ export class IdentityConfig {
...
-  identityId = "";
-  identityName = "";
+  clientId = "";
```
dilin-MS, line 10: what about `identityClientId`?

packages/fx-core/src/plugins/resource/identity/index.ts
```diff
@@ -111,9 +111,9 @@ export class IdentityPlugin implements Plugin {
       return err(error);
     }
 
-    ctx.config.set(Constants.identityName, this.config.identityName);
-    ctx.config.set(Constants.identityId, this.config.identityId);
-    ctx.config.set(Constants.identity, this.config.identity);
+    ctx.config.set(Constants.identityName, this.config.identity);
+    ctx.config.set(Constants.identityClientId, this.config.clientId);
```
blackchoey, line 115: Do we need to update the migration logic to change the name of related properties in `env.default.json`?

packages/fx-core/src/plugins/resource/identity/index.ts
```diff
@@ -242,8 +242,14 @@ export class IdentityPlugin implements Plugin {
     const resourceId = getArmOutput(ctx, IdentityArmOutput.identityName);
 
     ctx.config.set(Constants.identityName, resourceId);
```
blackchoey, line 244: do we also need to update this line? the identity name is no longer resource id

packages/fx-core/src/plugins/resource/function/ops/provision.ts
```diff
@@ -179,7 +179,7 @@ export class FunctionProvision {
     site.identity = {
       type: DefaultProvisionConfigs.siteIdentityTypeUserAssigned,
       userAssignedIdentities: {
-        [identityName]: {},
+        [identityResourceId]: {},
```
hund030, line 182: We simply replace identity name with resource id, does this also work?

YitongFeng-git (approved):

packages/fx-core/src/plugins/resource/function/ops/provision.ts
xzf0587: We change the key name, but the value is always the resource id.

blackchoey (approved):

YitongFeng-git (approved):

PR #2545, opened 2021-10-09
chore(multi-env): add env config telemetry

a1exwang:
- send envConfig file with unknown properties and all values redacted when provisioning

https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/11012613

### DIFF
--- a/packages/fx-core/src/common/tools.ts
+++ b/packages/fx-core/src/common/tools.ts
@@ -591,7 +591,7 @@ function isBasicJsonSchema(jsonSchema: unknown): jsonSchema is BasicJsonSchema {
 //  }
 //  Output:
 //  {"name": null}
-export function redactObject(
+export function _redactObject(
   obj: unknown,
   jsonSchema: unknown,
   maxRecursionDepth = 8,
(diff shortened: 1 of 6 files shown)


jayzhang (approved):

packages/fx-core/src/common/tools.ts
```diff
@@ -561,3 +561,79 @@ export function isSPFxProject(projectSettings?: ProjectSettings): boolean {
...
+//  obj = {"name": "some name", "user defined property": {"key1": "value1"}}
+//  jsonSchema = {
+//    "type": "object",
+//    "properties": {
+//      "name": { "type": "string" }
+//    }
+//  }
+//  Output:
+//  {"name": null}
+export function _redactObject(
```
qinezh, line 594: This function seems to be internal, do we need to expose it?

a1exwang: No. I forgot to delete the export.

packages/fx-core/src/common/tools.ts
```diff
@@ -561,3 +561,79 @@ export function isSPFxProject(projectSettings?: ProjectSettings): boolean {
...
+  };
+}
+function isBasicJsonSchema(jsonSchema: unknown): jsonSchema is BasicJsonSchema {
+  if (!jsonSchema || typeof jsonSchema !== "object") {
+    return false;
+  }
+  return typeof (jsonSchema as { type: unknown })["type"] === "string";
+}
+
+// Redact user content of "obj";
```
qinezh, line 578: move the comment to the exported function?

PR #2547, opened 2021-10-09
docs: Update outdated content in VSC extention readme

zyxiaoyuer:
And remove samples to the end of the page because we don't want to distract users but want them start with create new projects.


jayzhang (approved):

packages/vscode-extension/README.md
```diff
@@ -22,9 +22,11 @@ Teams apps are a combination of [capabilities](https://docs.microsoft.com/en-us/
 
 <a href=https://docs.microsoft.com/en-us/microsoftteams/platform/messaging-extensions/what-are-messaging-extensions>Messaging extensions</a> allow users to interact with your web service through buttons and forms in the Microsoft Teams client.
 
-## Build a Teams app in less than 5 minutes
+## Getting started
+
+After installing the Teams toolkit, follow the [Get Started](https://aka.ms/teamsfx-build-first-app) instructions in our documentation to smoothly start with.
 
-Build a Teams app from the scratch or explore our [samples](https://github.com/OfficeDev/TeamsFx-Samples) to help you quickly get started with the basic Teams app concepts and code structures.
+In the Teams Toolkit for Visual Sutido Code, you can easily discover all applicable commands in the sidebar and Command Palette with the keyword ‘Teams’. It also supports [Command Line Interface (CLI)](https://www.npmjs.com/package/@microsoft/teamsfx-cli) to increase efficiency.
```
eriolchan, line 29: typo -> Studio

eriolchan: Another is why we use single quote here, instead of double quote"" for "the keyword 'Teams'"? Besides the single quote doesn't look like English style

zyxiaoyuer: Thanks, I have correct the typo and single quote.  For single quote, it's nothing special but just single quote is more predominate in writing class at my university. so it become part of my writing habits. I have google it and both are acceptable. Maybe double quotes is more American style and single is more English style. so I changed to double quotes.

PR #2570, opened 2021-10-12
fix(multi-env): remove the active env remains

a1exwang:
- remove all code related to "activeEnv". active env is no more.
- fix resource and config command arguments

tested `teamsfx resource show`, `teamsfx resource list` and `teamsfx config get/set`

E2E test case passed: https://github.com/OfficeDev/TeamsFx/actions/runs/1335886350

https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/11033065


jayzhang (dismissed):

packages/cli/src/error.ts
```diff
@@ -125,3 +126,13 @@ export function InvalidEnvFile(msg: string, path: string): UserError {
     "InvalidEnvFile"
   );
 }
+
+export class EnvUndefined extends SystemError {
+  constructor() {
+    super(
+      new.target.name,
+      `env is undefined, isMultiEnvEnabled = ${isMultiEnvEnabled()}`,
```
qinezh, line 134: no need to expose the feature flag isMultiEnvEnabled in error message?

PR #2575, opened 2021-10-12
chore: fix e2e test for dotnet envchecker

meifans:


packages/vscode-extension/test/integration/envChecker/utils/dotnet.ts
```diff
@@ -130,7 +130,7 @@ export async function createTmpBackendProjectDir(
 
   const csprojPath = path.resolve(
     __dirname,
-    "../../../../../../../templates/function-base/ts/default/extensions.csproj"
```
a1exwang, line 133: also change the github action yaml files so the PR can trigger the test.

PR #2577, opened 2021-10-12
fix(multi-env): validate the length of environment name

qinezh:
Bug: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/11645755

### DIFF
--- a/packages/fx-core/src/core/question.ts
+++ b/packages/fx-core/src/core/question.ts
@@ -13,6 +13,7 @@ import {
 import * as jsonschema from "jsonschema";
 import * as path from "path";
 import * as fs from "fs-extra";
+import * as os from "os";
 import { environmentManager } from "./environment";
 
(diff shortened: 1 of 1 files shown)


packages/fx-core/src/core/question.ts
```diff
@@ -124,6 +126,11 @@ export function getQuestionNewTargetEnvironmentName(projectPath: string): TextIn
           return "Environment name can only contain letters, digits, _ and -.";
         }
 
+        const envFilePath = environmentManager.getEnvConfigPath(targetEnvName, projectPath);
+        if (os.type() === "Windows_NT" && envFilePath.length >= WINDOWS_MAX_PATH_LENGTH) {
```
a1exwang, line 130: I think we should check windows max length even on macOS or Linux. Because they can share the project with others.

jayzhang (approved):

PR #2578, opened 2021-10-12
fix(frontend): check plugin activated or not by solution settings

hund030:
Plugin config instance is always assigned in v2 api, so stop checking it. Use solution settings instead.

### DIFF
--- a/packages/fx-core/src/plugins/resource/frontend/plugin.ts
+++ b/packages/fx-core/src/plugins/resource/frontend/plugin.ts
@@ -7,6 +7,7 @@ import {
   ArchiveFolderName,
   ArchiveLogFileName,
   AppPackageFolderName,
+  AzureSolutionSettings,
 } from "@microsoft/teamsfx-api";
 import path from "path";
 
@@ -59,6 +60,7 @@ import * as fs from "fs-extra";
 import { Bicep, ConstantString } from "../../../common/constants";
 import { EnvironmentUtils } from "./utils/environment-utils";
 import { copyFiles } from "../../../common";
+import { AzureResourceFunction } from "../../solution/fx-solution/question";
 
 export class FrontendPluginImpl {
   public async scaffold(ctx: PluginContext): Promise<TeamsFxResult> {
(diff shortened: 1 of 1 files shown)


packages/fx-core/src/plugins/resource/frontend/plugin.ts
```diff
@@ -245,24 +247,26 @@ export class FrontendPluginImpl {
...
       envs[EnvironmentVariables.FuncEndpoint] = functionPlugin.get(
         DependentPluginInfo.FunctionEndpoint
       ) as string;
     }
 
-    const authPlugin = ctx.envInfo.profile.get(DependentPluginInfo.RuntimePluginName);
-    if (authPlugin) {
+    if (solutionSettings.activeResourcePlugins.includes(DependentPluginInfo.RuntimePluginName)) {
+      const authPlugin = ctx.envInfo.profile.get(DependentPluginInfo.RuntimePluginName);
       envs[EnvironmentVariables.RuntimeEndpoint] = authPlugin.get(
```
henzhang-ms, line 262: so `authPlugin` cannot be undefined now?

henzhang-ms: or It would return an empty map?

hund030: It would return an empty map.

hund030: Updated.

henzhang-ms (dismissed):

packages/fx-core/src/plugins/resource/frontend/plugin.ts
```diff
@@ -244,26 +246,41 @@ export class FrontendPluginImpl {
 
   private async updateDotenv(ctx: PluginContext): Promise<void> {
     const envs: { [key: string]: string } = {};
+    const addToEnvs = (key: string, value: string | undefined) => {
+      if (value) {
```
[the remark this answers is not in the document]
hund030: After more test, adding `""` to .env is different from skip adding.
If we add `""`, in .env file there will be `REACT_APP_FUNC_NAME=`, and `process.env.REACT_APP_FUNC_NAME=""`, but if we skip adding the variable, `process.env.REACT_APP_FUNC_NAME`.
So add a fix to only skip adding when value is null or undefined.

packages/fx-core/src/plugins/resource/frontend/plugin.ts
```diff
@@ -244,26 +246,42 @@ export class FrontendPluginImpl {
 
   private async updateDotenv(ctx: PluginContext): Promise<void> {
     const envs: { [key: string]: string } = {};
+    const addToEnvs = (key: string, value: string | undefined) => {
+      // Check for both null and undefined, add to envs when value is "", 0 or false.
+      if (value != null) {
```
henzhang-ms, line 251: Nit: this may cause warning, please check the compilation log

PR #2581, opened 2021-10-12
perf: adjust parameter.json placeholder as object, align with manifest

wenytang-ms:


packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -642,37 +644,45 @@ function expandParameterPlaceholders(
...
+    const solutionVariables: Record<string, string> = {};
     for (const configItem of solutionConfig) {
       if (typeof configItem[1] === "string") {
         // Currently we only config with string type
-        const variableName = `SOLUTION__${normalizeToEnvName(configItem[0])}`;
-        availableVariables[variableName] = configItem[1];
+        solutionVariables[configItem[0]] = configItem[1];
       }
     }
+    availableVariables[solutionName] = solutionVariables;
```
blackchoey, line 675: should be availableVariables[profileName], because "solution" is also part of profile.dev.json

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -642,37 +644,45 @@ function expandParameterPlaceholders(
...
   }
   // Add environment variable to available variables
-  Object.assign(availableVariables, process.env); // The environment variable has higher priority
 
-  if (expandSecrets === false) {
-    escapeSecretPlaceholders(availableVariables);
-  }
+  const processVariables: Record<string, string> = Object.keys(process.env)
+    .filter((key) => ![solutionName, profileName].includes(key))
```
blackchoey, line 680: can we print a warning if there's an environment variable named `profile` or `solution`? We can do this in later PR

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -642,37 +644,45 @@ function expandParameterPlaceholders(
...
-        const variableName = `${normalizeToEnvName(plugin.name)}__${normalizeToEnvName(
-          configItem[0]
-        )}`;
-        availableVariables[variableName] = configItem[1];
+        pluginVariables[configItem[0]] = configItem[1];
       }
     }
+    profileVariables[plugin.name] = pluginVariables;
   }
+  if (expandSecrets === false) {
```
blackchoey, line 661: is there still cases that expandSecret = false?

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
wenytang-ms: sure

packages/fx-core/tests/plugins/resource/aad/unit/generateArmTemplates.test.ts
```diff
@@ -74,11 +74,7 @@ describe("AadGenerateArmTemplates", () => {
         "parameters.json"
       );
       chai.assert.strictEqual(
-        JSON.stringify(
-          expectedResult.Orchestration.ParameterTemplate!.ParameterJson,
-          undefined,
-          2
-        ) + "\n", // workaround test issue caused by prettier
+        JSON.stringify(expectedResult.Orchestration.ParameterTemplate!.ParameterJson, undefined, 2), // workaround test issue caused by prettier
```
blackchoey, line 77: why we no longer need to append `\n` anymore?

blackchoey (approved):

packages/fx-core/tests/plugins/resource/aad/unit/generateArmTemplates.test.ts
wenytang-ms: since the test file parameters.json files remove the \n at the end of file

YitongFeng-git (approved):

PR #2598, opened 2021-10-13
perf(apim): apim arm support deploy

dilin-MS:


packages/fx-core/src/plugins/resource/apim/utils/commonUtils.ts
```diff
@@ -73,3 +74,30 @@ export class Lazy<T> {
...
   }
 }
+
+export function getApimServiceNameFromResourceId(resourceId: string): string {
+  const result = parseFromResourceId(
+    /providers\/Microsoft.ApiManagement\/service\/([^\/]*)/i,
+    resourceId
+  );
+  if (!result) {
+    throw new Error("Failed to get API Management service name from resource id: " + resourceId);
```
XiaofuHuang, line 84: Can we use BuildError function to create an FxError?

XiaofuHuang (approved):

PR #2606, opened 2021-10-14
fix(core): make sure all APIs of core are lock-protected

jayzhang:


LongOddCode (approved):

packages/fx-core/src/core/index.ts
```diff
@@ -1026,7 +1038,13 @@ export class FxCore implements Core {
     }
   }
 
-  @hooks([ErrorHandlerMW, ProjectSettingsLoaderMW, EnvInfoLoaderMW(true), ContextInjectorMW])
+  @hooks([
+    ErrorHandlerMW,
+    ConcurrentLockerMW,
```
qinezh, line 1043: Just curious to know why not allow the operation to decrypt secrets and other operations like deploy happen at the same time?

PR #2609, opened 2021-10-14
fix(simple auth): improve error handling

KennethBWSong:
task: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_sprints/taskboard/AuthAndData/Microsoft%20Teams%20Extensibility/CY21-10.1?workitem=12010015
Bug:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_sprints/taskboard/AuthAndData/Microsoft%20Teams%20Extensibility/CY21-10.2?workitem=12381795

### DIFF
--- a/packages/fx-core/src/plugins/resource/simpleauth/utils/common.ts
+++ b/packages/fx-core/src/plugins/resource/simpleauth/utils/common.ts
@@ -1,13 +1,12 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
-import { FxError, LogProvider, PluginContext } from "@microsoft/teamsfx-api";
+import { LogProvider, PluginContext } from "@microsoft/teamsfx-api";
 import * as path from "path";
 import * as fs from "fs-extra";
 import { Constants, Message } from "../constants";
 import {
   EndpointInvalidError,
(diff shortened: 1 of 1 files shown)


formulahendry (dismissed):

packages/fx-core/src/plugins/resource/simpleauth/utils/common.ts
```diff
@@ -145,6 +146,17 @@ export class Utils {
     return configValue;
   }
 
+  public static handleError(error: any, errorDetail: PluginError, ...args: string[]): FxError {
+    if (
+      error?.response?.status >= Constants.statusCodeUserError &&
```
adashen, line 151: Do we make sure all the error content structure has response.status? Do we need have type to restrict this?

KennethBWSong: Here handleError method only works on ARM response for creating webapp, update webapp config, etc. It should be better if we have a type for the error. I will update this pr, thx!

formulahendry (approved):
Just one small comment.

adashen (approved):

PR #2612, opened 2021-10-14
perf: add simple auth server farm and webapp

wenytang-ms:


packages/fx-core/src/core/middleware/projectMigrator.ts
```diff
@@ -660,6 +663,17 @@ async function generateArmParameterJson(ctx: CoreHookContext) {
       value: envConfig[ResourcePlugins.SimpleAuth][EnvConfigName.SkuName],
     };
   }
+
+  if (envConfig[ResourcePlugins.SimpleAuth]?.[EnvConfigName.Endpoint]) {
```
blackchoey, line 667: So the previous `env.default.json` does not have a dedicated property for the web app name?

PR #2621, opened 2021-10-14
chore: generate .env.teamsfx.local for multienv local debug

swatDong:
This is a preparation for customized local debug, which is to generate `.env.teamsfx.local` for each component, then user can partial launch local service(s) or add customized environment variables.

- Only enabled for multi-env
- Add `.env.teamsfx.local` to git ignore
- Add unit tests

### DIFF
--- a/packages/fx-core/src/plugins/resource/localdebug/index.ts
+++ b/packages/fx-core/src/plugins/resource/localdebug/index.ts
@@ -36,6 +36,13 @@ import * as Settings from "./settings";
 import * as Tasks from "./tasks";
 import { LocalEnvProvider } from "./localEnv";
 import {
+  EnvKeysFrontend,
+  EnvKeysBackend,
+  EnvKeysBot,
+  EnvKeysBotV1,
+  LocalEnvMultiProvider,
+} from "./localEnvMulti";
+import {
   LocalBotEndpointNotConfigured,
   MissingStep,
   NgrokTunnelNotConnected,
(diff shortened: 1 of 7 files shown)


dooriya:
As we support env variable for local debug, is there any plan for the env variable support in remote env? For example, the react tab app can consume some env variables at build time (targeting on a remote env).

packages/fx-core/src/plugins/resource/localdebug/index.ts
```diff
@@ -326,25 +343,144 @@ export class LocalDebugPlugin implements Plugin {
...
+          backendEnvs!.teamsfxLocalEnvs[EnvKeysBackend.AuthorityHost] =
+            "https://login.microsoftonline.com";
+          backendEnvs!.teamsfxLocalEnvs[EnvKeysBackend.TenantId] = teamsAppTenantId;
+          backendEnvs!.teamsfxLocalEnvs[EnvKeysBackend.ApiEndpoint] = localFuncEndpoint;
+          backendEnvs!.teamsfxLocalEnvs[EnvKeysBackend.ApplicationIdUri] = applicationIdUri;
+          backendEnvs!.teamsfxLocalEnvs[EnvKeysBackend.AllowedAppIds] = [
+            teamsMobileDesktopAppId,
+            teamsWebAppId,
+          ].join(";");
+        }
```
dooriya, line 428: The local env generation is quite similar to `getLocalDebugEnvs`.
As we will support env for local debug, can we consolidate this part? And we can leverage the new `.env.teamsfx.local` to pass the local debug env to VS Code extension.

dooriya (approved):

PR #2629, opened 2021-10-15
fix(local-debug): remove duplicate getLocalDebugEnvs

kuojianlu:
To fix https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/11033697.

Avoid calling getLocalDebugEnvs every time when getting frontend/backend/auth/bot envs.

### DIFF
--- a/packages/cli/src/cmds/preview/commonUtils.ts
+++ b/packages/cli/src/cmds/preview/commonUtils.ts
@@ -187,10 +187,9 @@ export function createTaskStopCb(
   };
 }
 
-async function getLocalEnv(
+export async function getLocalEnv(
   core: FxCore,
-  workspaceFolder: string,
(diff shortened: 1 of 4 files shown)


a1exwang (approved):
Curious how much performance this will improve

swatDong (approved):

kuojianlu:
> Curious how much performance this will improve

For a project with tab+func+bot, one `provideTasks` call saves 4 times `getLocalDebugEnvs` calls. One `getLocalDebugEnvs` call consumes about 180 ms. There are totally 8 times `provideTasks` calls during one local debug. So this saves about 8 * 4 * 180 = 5760 ms at most.
BTW, it also reduces the duplicate logs.

PR #2630, opened 2021-10-15
fix(collaborate): fix f5 login not refresh tree view bug

KennethBWSong:


swatDong (changes requested):
Use `AppStudioLogin.setStatusChangeMap` to setup login callback, instead of changing login component code.

SLdragon (approved):

PR #2631, opened 2021-10-15
feat(bot): register AAD by Graph

YitongFeng-git:


packages/fx-core/src/plugins/resource/bot/aadRegistration.ts
```diff
@@ -25,29 +27,34 @@ export class AADRegistration {
...
+      } catch (e) {
+        throw new ProvisionError(CommonStrings.AAD_APP, e);
+      }
+
+      if (!regResponse || !utils.isHttpCodeOkOrCreated(regResponse.status)) {
+        throw new ProvisionError(CommonStrings.AAD_APP);
+      }
+      result.clientId = regResponse.data.appId;
+      result.objectId = regResponse.data.id;
+    } else {
```
hund030, line 50: NIT: I prefer put the smaller condition block ahead.
```
if (condition) {
    CheckThrowSomethingMissing("objectId", objectId);
    CheckThrowSomethingMissing("msAppId", msAppId);
    result.objectId = objectId;
    result.clientId = msAppId;
    return;
}
```

packages/fx-core/src/plugins/resource/bot/aadRegistration.ts
```diff
@@ -25,29 +27,34 @@ export class AADRegistration {
...
-          signInAudience: AADRegistrationConstants.AZURE_AD_MULTIPLE_ORGS,
-        }
-      );
-    } catch (e) {
-      throw new ProvisionError(CommonStrings.AAD_APP, e);
-    }
-
-    if (!regResponse || !utils.isHttpCodeOkOrCreated(regResponse.status)) {
-      throw new ProvisionError(CommonStrings.AAD_APP);
+    if (!objectId && !msAppId) {
```
hund030, line 30: What if only objectId assigned or only msAppId assigned?
```
objectId = "xxx-xxxx";
msAppId = undefined;
```

packages/fx-core/src/plugins/resource/bot/aadRegistration.ts
YitongFeng-git: But I think put the unimportant logic in "else" is better

hund030: Sure, depends on you.

packages/fx-core/src/plugins/resource/bot/aadRegistration.ts
hund030: Exactly. I mean we don't register new aad when one of the ids is undefined. Does the bot works in that case?

packages/fx-core/src/plugins/resource/bot/aadRegistration.ts
eriolchan: It depends on you to make the final decision. What Zhijie suggest is one best practice that return early for non-core logic, so that people can focus on the remaining code, which is all about core business logic. Another benefit of this tip is that it can reduce the number of if/else condition to make the code be more clear

eriolchan, line 50: It depends on you to make the final decision. What Zhijie suggest is one best practice that return early for non-core logic, so that people can focus on the remaining code, which is all about core business logic. Another benefit of this tip is that it can reduce the number of if/else condition to make the code be more clear

IvanJobs (approved):
looks good to me in terms of :
  1. align with AAD Plugin
  2. align with registerByAppStudio
  3. regen secret anyway when calling this method

PR #2633, opened 2021-10-15
chore: fix dialog popped up twice when open teams project

meifans:
FIx [bug](https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/12193819)

### DIFF
--- a/packages/fx-core/src/core/middleware/projectMigrator.ts
+++ b/packages/fx-core/src/core/middleware/projectMigrator.ts
@@ -107,7 +107,7 @@ class ArmParameters {
 }
 
 export const ProjectMigratorMW: Middleware = async (ctx: CoreHookContext, next: NextFunction) => {
-  if ((await needMigrateToArmAndMultiEnv(ctx)) && (await checkMethod(ctx))) {
+  if ((await needMigrateToArmAndMultiEnv(ctx)) && checkMethod(ctx)) {
     const core = ctx.self as FxCore;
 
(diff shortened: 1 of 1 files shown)


packages/fx-core/src/core/middleware/projectMigrator.ts
```diff
@@ -142,6 +144,13 @@ export const ProjectMigratorMW: Middleware = async (ctx: CoreHookContext, next:
   await next();
 };
 
+async function checkMethod(ctx: CoreHookContext) {
+  const getProjectConfigMethod = "getProjectConfig";
+  if (ctx.method === getProjectConfigMethod && fromReloadFlag) return false;
```
a1exwang, line 149: Why not just delete `ProjectMigratorMW` for `getProjectConfig` method in `fx-core/src/core/index.ts`

packages/fx-core/src/core/middleware/projectMigrator.ts
```diff
@@ -142,6 +144,13 @@ export const ProjectMigratorMW: Middleware = async (ctx: CoreHookContext, next:
...
+  fromReloadFlag = ctx.method === getProjectConfigMethod;
+  return true;
+}
```
wenytang-ms, line 152: why this method is async?

packages/fx-core/src/core/middleware/projectMigrator.ts
meifans: If delete, dialog will not be popped up when user open project.

PR #2643, opened 2021-10-18
fix: update deploy check account logic

xiaolang124:
1. Fix deploy check account logic.
2. Update notify message content.

### DIFF
--- a/packages/fx-core/src/plugins/solution/fx-solution/commonQuestions.ts
+++ b/packages/fx-core/src/plugins/solution/fx-solution/commonQuestions.ts
@@ -113,7 +113,7 @@ export async function checkSubscription(
     return err(
       new UserError(
         SolutionError.SubscriptionNotFound,
-        `The subscription '${subscriptionId}'${subscriptionName} is not found in the current account, please check the '${EnvConfigFileNameTemplate.replace(
+        `The subscription '${subscriptionId}'${subscriptionName} is not found in the current account, please use the right Azure account or check the '${EnvConfigFileNameTemplate.replace(
           EnvNamePlaceholder,
           envInfo.envName
         )}' file.`,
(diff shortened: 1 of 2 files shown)


packages/fx-core/src/plugins/solution/fx-solution/commonQuestions.ts
```diff
@@ -142,7 +142,10 @@ export async function checkM365Tenant(
     return err(
       new UserError(
         SolutionError.TeamsAppTenantIdNotRight,
-        `The m365 tenant id '${m365TenantId}' is not found in the current account.`,
+        `The M365 tenant id '${m365TenantId}'(which is used before) does not match the current account, please use the right M365 account or check the '${EnvConfigFileNameTemplate.replace(
```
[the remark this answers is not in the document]
xiaolang124: Good point. Fixed.

packages/fx-core/src/plugins/solution/fx-solution/commonQuestions.ts
```diff
@@ -142,7 +142,7 @@ export async function checkM365Tenant(
     return err(
       new UserError(
         SolutionError.TeamsAppTenantIdNotRight,
-        `The m365 tenant id '${m365TenantId}' is not found in the current account.`,
+        `The M365 tenant id '${m365TenantId}'(which is used before) does not match the current account, please use the right M365 account.`,
```
qinezh, line 145: How about including env name in this error message? So user can know the M365 tenant id in which env doesn't matched.

xiaolang124: Already add.

qinezh (approved):

nliu-ms (approved):

PR #2645, opened 2021-10-18
chore: add .backup folder

meifans:
Add a folder name .backup under the project root directory and it have two folders. (named .fx and appPackage)


packages/fx-core/src/core/middleware/projectMigrator.ts
```diff
@@ -456,8 +456,11 @@ async function getMultiEnvFolders(projectPath: string): Promise<any> {
 
 async function backup(projectPath: string): Promise<void> {
   const fx = path.join(projectPath, `.${ConfigFolderName}`);
-  const backup = path.join(fx, "migrationbackup");
-  await fs.ensureDir(backup);
+  const backup = path.join(projectPath, ".backup");
```
a1exwang, line 459: What if there is an existing `.backup` folder under the project root, as the directory name is quite common?

wenytang-ms:
what about migrationbackup folder?

PR #2673, opened 2021-10-19
fix(apim): remove publisher info from profile.xxx.json

dilin-MS:


packages/fx-core/src/plugins/resource/apim/managers/apimManager.ts
```diff
@@ -110,8 +116,9 @@ export class ApimManager {
       apimConfig.serviceResourceId = getArmOutput(ctx, ApimArmOutput.ServiceResourceId);
       apimConfig.productResourceId = getArmOutput(ctx, ApimArmOutput.ProductResourceId);
       apimConfig.authServerResourceId = getArmOutput(ctx, ApimArmOutput.AuthServerResourceId);
-      apimConfig.publisherEmail = getArmOutput(ctx, ApimArmOutput.PublisherEmail);
-      apimConfig.publisherName = getArmOutput(ctx, ApimArmOutput.PublisherName);
+      // Remove publisherEmail and publisherName from config to avoid sensitive info being committed into git history
+      apimConfig.publisherEmail = undefined;
```
blackchoey, line 120: can we just remove these 2 lines?

blackchoey (dismissed):

jayzhang (dismissed):

packages/fx-core/src/plugins/resource/apim/managers/apimManager.ts
dilin-MS: No. Without setting these two properties to be undefined, they will be record into profile.xxx.json.

packages/fx-core/src/plugins/resource/apim/constants.ts
```diff
@@ -431,6 +431,4 @@ export class ApimArmOutput {
   static readonly ServiceResourceId = "apimServiceResourceId";
   static readonly ProductResourceId = "apimProductResourceId";
   static readonly AuthServerResourceId = "apimAuthServiceResourceId";
-  static readonly PublisherEmail = "apimPublisherEmail";
-  static readonly PublisherName = "apimPublisherName";
```
dooriya, line 435: Just to want to know what's the motivation to remove these two config values from the provision output?  Because it contains user sensitive info?

dooriya: Sounds good to me. Thanks for the explaination.

dooriya (approved):

jayzhang (approved):

PR #2677, opened 2021-10-19
chore: prepare func checker for custom local debug

swatDong:
- Expose bin folder so that `func` can be used directly from cmd
- Delete `func.ps1` for portable installation as well

@a1exwang @meifans - please help review

### DIFF
--- a/packages/vscode-extension/src/debug/depsChecker/funcToolChecker.ts
+++ b/packages/vscode-extension/src/debug/depsChecker/funcToolChecker.ts
@@ -86,15 +86,8 @@ export class FuncToolChecker implements IDepsChecker {
       // to avoid "func -v" and "func new" work well, but "func start" fail.
       hasSentinel = await fs.pathExists(FuncToolChecker.getSentinelPath());
 
-      if (isWindows()) {
-        // delete func.ps1 when checking portable function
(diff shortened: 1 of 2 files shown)


packages/vscode-extension/src/debug/depsChecker/funcToolChecker.ts
```diff
@@ -85,6 +85,15 @@ export class FuncToolChecker implements IDepsChecker {
         portableFuncVersion !== null && supportedVersions.includes(portableFuncVersion);
       // to avoid "func -v" and "func new" work well, but "func start" fail.
       hasSentinel = await fs.pathExists(FuncToolChecker.getSentinelPath());
+
+      // delete func.ps1 when checking portable function
+      for (const funcFolder in this.getPortableFuncBinFolders()) {
```
a1exwang, line 90: 1. Extract this to a function?
2. Delete files only when `isVersionSupported && hasSentinel` (i.e. isPortableFuncInstalled() == true) to only delete things in a known correct state?

swatDong: good point, fixed.

a1exwang (approved):

PR #2683, opened 2021-10-19
fix(aad): fix aad to support multi env input

KennethBWSong:
task: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_sprints/taskboard/AuthAndData/Microsoft%20Teams%20Extensibility/CY21-10.2?workitem=12382197

### DIFF
--- a/packages/fx-core/src/plugins/resource/aad/plugin.ts
+++ b/packages/fx-core/src/plugins/resource/aad/plugin.ts
@@ -192,9 +192,7 @@ export class AadAppForTeamsImpl {
       isLocalDebug
     );
 
-    const skip: boolean = isMultiEnvEnabled()
-      ? (ctx.envInfo.config.auth?.[ConfigKeys.skip] as boolean)
-      : (ctx.config.get(ConfigKeys.skip) as boolean);
+    const skip = Utils.skipAADProvision(ctx);
     if (skip) {
       ctx.logProvider?.info(Messages.SkipProvision);
       Utils.addLogAndTelemetryWithLocalDebug(
--- a/packages/fx-core/tests/plugins/resource/aad/helper.ts
+++ b/packages/fx-core/tests/plugins/resource/aad/helper.ts
@@ -35,6 +35,7 @@ import {
   LocalSettingsBotKeys,
   LocalSettingsFrontendKeys,
 } from "../../../../src/common/localSettingsConstants";
+import { EnvConfig } from "@microsoft/teamsfx-api";
 
 const permissions = '[{"resource": "Microsoft Graph","delegated": ["User.Read"],"application":[]}]';
 const permissionsWrong =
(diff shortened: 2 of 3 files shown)


packages/fx-core/src/plugins/resource/aad/utils/common.ts
```diff
@@ -63,4 +64,30 @@ export class Utils {
...
+      return skip;
+    }
+
+    const objectId = ctx.envInfo.config.auth?.objectId;
+    const clientId = ctx.envInfo.config.auth?.clientId;
+    const oauth2PermissionScopeId = ctx.envInfo.config.auth?.accessAsUserScopeId;
+    const clientSecret = ctx.envInfo.config.auth?.clientSecret;
+    const skip = ctx.envInfo.config.auth?.[ConfigKeys.skip];
+
+    if (objectId && clientId && oauth2PermissionScopeId && clientSecret) {
```
SLdragon, line 80: Do we need to throw errors when something is undefined?

KennethBWSong: Now check will be performed separately when provisioning or skipping provision. It's a good idea to throw error here. I will update the code. Thx!

adashen: Failure one information not complete?

packages/fx-core/src/plugins/resource/aad/utils/common.ts
```diff
@@ -63,4 +64,30 @@ export class Utils {
...
+    if (!isMultiEnvEnabled()) {
+      const skip = ctx.config.get(ConfigKeys.skip) as boolean;
+      return skip;
+    }
+
+    const objectId = ctx.envInfo.config.auth?.objectId;
+    const clientId = ctx.envInfo.config.auth?.clientId;
+    const oauth2PermissionScopeId = ctx.envInfo.config.auth?.accessAsUserScopeId;
+    const clientSecret = ctx.envInfo.config.auth?.clientSecret;
+    const skip = ctx.envInfo.config.auth?.[ConfigKeys.skip];
```
dooriya, line 78: According to the env config schema: https://aka.ms/teamsfx-env-config-schema, we don't have the `skip` config key in the new env config.

KennethBWSong: I will update the schema. Thx!

packages/fx-core/src/plugins/resource/aad/utils/common.ts
```diff
@@ -63,4 +64,30 @@ export class Utils {
     const tenantId: string = (tokenObject as any).tid;
     return tenantId;
   }
+
+  public static getAndMoveInput(ctx: PluginContext): boolean {
```
dooriya, line 68: I cannot guess what this function is used for without go through the code? Can we provide a better naming for it?

packages/fx-core/src/plugins/resource/aad/utils/common.ts
KennethBWSong: Will fix in another pr, thx!

dooriya: OK. So can you remove this line which is reading `skip` property from env config? You can consolidate the skip aad provision/configuration logic in another PR.

PR #2691, opened 2021-10-20
test: make frontend plugin ut support insider preview

a1exwang:
The unit test will run twice, one with TEAMSFX_INSIDER_PREVIEW=true and one without it. This PR make frontend plugin to support both cases.

https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/12217265


hund030 (approved):
Do we have plan to remove `isArmSupportEnabled` after the go production story?

PR #2700, opened 2021-10-21
chore: generate npm install scripts to project package.json

swatDong:
Add npm install scripts to root package.json. This is for automated npm install feature, and also enables developer to install all sub project with one command.
Put code logic in local debug plugin scaffold to work for both "create project" and "add capability".

### DIFF
--- a/packages/fx-core/tests/plugins/resource/localdebug/unit/index.test.ts
+++ b/packages/fx-core/tests/plugins/resource/localdebug/unit/index.test.ts
@@ -5,13 +5,12 @@ import * as dotenv from "dotenv";
 import * as fs from "fs-extra";
 import { ConfigFolderName, Platform, PluginContext } from "@microsoft/teamsfx-api";
 import * as path from "path";
-import * as sinon from "sinon";
 
 import { LocalDebugPluginInfo } from "../../../../../src/plugins/resource/localdebug/constants";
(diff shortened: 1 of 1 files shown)


packages/fx-core/src/plugins/resource/localdebug/index.ts
```diff
@@ -201,6 +204,38 @@ export class LocalDebugPlugin implements Plugin {
...
+            packageJson = await PackageJson.load(packageJsonPath);
+          } catch (error) {
+            ctx.logProvider?.error(`Cannot load package.json from ${ctx.root}. ${error}`);
+          }
+
+          if (packageJson !== undefined) {
+            const scripts = packageJson.content.scripts ?? {};
+            const installAll: string[] = [];
+
+            if (includeBackend) {
```
kuojianlu, line 221: Maybe it is also necessary to check if tabs/api/bot folder exist?

swatDong: This is to change root package.json so nothing to do with tabs/api/bot folder.
BTW, plugin scaffolding may happen concurrently so tabs/api/bot folder may not be created yet. The only flag to use is `includeXxx`.
That's also the same logic as generating `launch.json` and `tasks.json`.

kuojianlu: My mistake.

PR #2706, opened 2021-10-21
fix(cputil): use cp.spwan() to build bicep files

dilin-MS:
Fix bug: [[Multi-Env][BugBash] Bicep deployment failed if the project path contains special characters](https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/11032976)

Use `cp.spawn()` to replace `cp.exec()`.

### DIFF
--- a/packages/fx-core/src/common/cpUtils.ts
+++ b/packages/fx-core/src/common/cpUtils.ts
@@ -31,7 +31,7 @@ export async function executeCommand(
   return result.stdout;
 }
 
-async function tryExecuteCommand(
+export async function tryExecuteCommand(
   command: string,
   args: string[],
   logger?: LogProvider,


packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -494,12 +495,20 @@ async function getResourceManagementClientForArmDeployment(
...
 ): Promise<JSON> {
-  const command = `${bicepCommand} build "${bicepOrchestrationFilePath}" --stdout`;
   try {
-    const result = await Executor.execCommandAsync(command);
-    return JSON.parse(result.stdout as string);
+    const result = await executeCommand(
+      undefined,
+      logger,
+      { shell: false },
+      bicepCommand,
```
a1exwang, line 506: When `shell: false`, you may also need to change bicep checker to remove the `"` in 
```
return `"${this.getBicepExecPath()}"`;
```
in `bicepChecker.ts`

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -494,12 +495,20 @@ async function getResourceManagementClientForArmDeployment(
...
+      "build",
+      bicepOrchestrationFilePath,
+      "--stdout"
+    );
+    return JSON.parse(result as string);
```
a1exwang, line 511: is `as string` needed here?

a1exwang (dismissed):

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
dilin-MS: Removed.

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -494,12 +495,20 @@ async function getResourceManagementClientForArmDeployment(
...
+  logger: LogProvider | undefined
 ): Promise<JSON> {
-  const command = `${bicepCommand} build "${bicepOrchestrationFilePath}" --stdout`;
   try {
-    const result = await Executor.execCommandAsync(command);
-    return JSON.parse(result.stdout as string);
+    const result = await executeCommand(
+      undefined,
+      logger,
+      { shell: false },
```
meifans, line 505: Shell is used by default in exec(), if set shell false in spawn(), need to test it for some corner case. For example, bicep path contains special characters like this "/home/bicep/bin test1/bicep"

a1exwang: `shell: false` is to deal with cases like these, because shell escape is not needed anymore.

packages/fx-core/src/common/cpUtils.ts
```diff
@@ -0,0 +1,113 @@
...
+  logger: LogProvider | undefined,
+  additionalOptions: cp.SpawnOptions | undefined,
+  command: string,
+  ...args: string[]
+): Promise<ICommandResult> {
+  return await new Promise(
+    (resolve: (res: ICommandResult) => void, reject: (e: Error) => void): void => {
+      let cmdOutput = "";
+      let cmdOutputIncludingStderr = "";
+      const formattedArgs: string = args.join(" ");
```
blackchoey, line 44: JSON.stringify the arrary? Since args may contain space, args.join(" ") will impact troubleshooting since we don't know it's 1 arg with space, or 2 args

packages/fx-core/src/common/cpUtils.ts
```diff
@@ -0,0 +1,113 @@
...
+      );
+
+      childProc.stdout?.on("data", (data: string | Buffer) => {
+        data = data.toString();
+        cmdOutput = cmdOutput.concat(data);
+      });
+
+      childProc.stderr?.on("data", (data: string | Buffer) => {
+        data = data.toString();
+        cmdOutputIncludingStderr = cmdOutputIncludingStderr.concat(data);
```
blackchoey, line 80: based on the variable name, my understanding is the output here include both stdout and stderr? Is it possible to only provide stderr to consumers? Since there's already a cmdOutput that stores stdout.

packages/fx-core/src/common/cpUtils.ts
```diff
@@ -0,0 +1,113 @@
...
+  if (result.code !== 0) {
+    const errorMessage = `Failed to run command: "${command} ${result.formattedArgs}", code: "${result.code}",
+                            output: "${result.cmdOutput}", error: "${result.cmdOutputIncludingStderr}"`;
+    await logger?.debug(errorMessage);
+    throw new Error(errorMessage);
+  } else {
+    await logger?.debug(`Finished running command: "${command} ${result.formattedArgs}".`);
+  }
+
+  return result.cmdOutput;
```
blackchoey, line 30: as a common utility, should we return exit code, stdout, stderr to consumers together?

packages/fx-core/src/common/cpUtils.ts
```diff
@@ -0,0 +1,113 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+import { LogProvider } from "@microsoft/teamsfx-api";
+import * as cp from "child_process";
+import * as os from "os";
+
+export async function executeCommand(
+  workingDirectory: string | undefined,
```
blackchoey, line 8: can we adjust the sequence of parameters and give default value for some parameters? I saw you set some undefined value in arm implementation. maybe we can try to reduce the need of setting undefined value as parameter (not total remove the requirement).
For example, workingDirectory may put to latter position and set undefined as default value because it's not always required.

PR #2800, opened 2021-11-01
feat(blazor): add new resource plugin for blazor project

hund030:


hund030:
E2E TEST: https://github.com/OfficeDev/TeamsFx/blob/bb110600e8feccf400fb050598f7ff7b4e62d32f/packages/cli/tests/e2e/blazor/TestDeployBlazor.tests.ts#L6

packages/fx-core/src/plugins/resource/blazor/constants.ts
```diff
@@ -0,0 +1,100 @@
...
+      name: "B1",
+    },
+  });
+}
+
+export class BlazorPluginInfo {
+  static readonly pluginName = "fx-resource-blazor";
+  static readonly displayName = "Blazor";
+  static readonly alias = "bz";
+  static readonly issueLink = "https://github.com/OfficeDev/TeamsFx/issues/new";
```
eriolchan, line 46: Since the issue link is the same across all plugin, I am not sure if it is filled by solution/telemetry helper level. Can you help check if we need to manually set this value in our plugin?

packages/fx-core/src/plugins/resource/blazor/enum.ts
```diff
@@ -0,0 +1,34 @@
...
+}
+
+export enum AppSettingsKey {
+  clientSecret = "CLIENT_SECRET",
+  clientId = "CLIENT_ID",
+  oauthHost = "OAUTH_AUTHORITY",
+  tabAppEndpoint = "TAB_APP_ENDPOINT",
+  aadMetadataAddress = "AAD_METADATA_ADDRESS",
+  botId = "BOT_ID",
+  botPassword = "BOT_PASSWORD",
```
eriolchan, line 27: I see bot config here, which bot blazor plugin support here? Bot from Bot plugin by TS/JS or Bot in the Blazor plugin of .NET project?

hund030: Bot of .NET project. The bot config works If we provision Azure Bot Service with bot plugin and host both bot code and blazor frontend code in the same Azure Web App.
Because we don't suppose to support bot in step 0, this is for future use.

packages/fx-core/src/plugins/resource/blazor/constants.ts
hund030: Seems it's filled by extension. Let me remove it.

henzhang-ms (approved):
lgtm

packages/fx-core/src/plugins/resource/blazor/plugin.ts
```diff
@@ -0,0 +1,309 @@
...
+      ConfigKey.resourceGroupName
+    );
+    const subscriptionId = this.checkAndGet(this.config.subscriptionId, ConfigKey.subscriptionId);
+    const credential = this.checkAndGet(
+      await ctx.azureAccountProvider?.getAccountCredentialAsync(),
+      ConfigKey.credential
+    );
+
+    // ? Do we support user customize build? If yes, how?
+    // * If we support user customize runtime, we need to validate its value because we use it to concat build command.
```
henzhang-ms, line 295: I think the runtime is bounded to the azure resource's env. So I think we don't need to support custom now. Or we can retrieve it from the azure resource's config.

PR #2806, opened 2021-11-01
perf(arm): solution help apim and function fill arm output

dilin-MS:
* [[apim] solution help fill in arm output](https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/12461160)
* [[function] solution help fill in arm output](https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/12461172)

Design docs: [Auto fill resource plugin context after ARM deployment](https://microsoftapc-my.sharepoint.com/:w:/g/personal/chyuan_microsoft_com/Eb5sguPTQwlOiJXo0se6feIBMOatMLn9w_0_MhrTdkcarg?e=KviQnO&isSPOFile=1)

### DIFF
--- a/packages/fx-core/src/plugins/resource/apim/constants.ts
+++ b/packages/fx-core/src/plugins/resource/apim/constants.ts
@@ -8,7 +8,7 @@ export class ProjectConstants {
   public static readonly pluginShortName: string = "APIM";
   public static readonly pluginName: string = "fx-resource-apim";
   public static readonly pluginDisplayName: string = "API Management";
-  public static readonly configFilePath: string = "state.xxx.json";
+  public static readonly configFilePath: string = "env.default.json";
   public static readonly workingDir: string = "openapi";
   public static readonly openApiDocumentFileName: string = "openapi.json";
   public static readonly readMeFileName: string = "README.md";
--- a/packages/fx-core/src/plugins/resource/apim/constants.ts
+++ b/packages/fx-core/src/plugins/resource/apim/constants.ts
@@ -8,7 +8,7 @@ export class ProjectConstants {
   public static readonly pluginShortName: string = "APIM";
   public static readonly pluginName: string = "fx-resource-apim";
   public static readonly pluginDisplayName: string = "API Management";
-  public static readonly configFilePath: string = "env.default.json";
+  public static readonly configFilePath: string = "state.xxx.json";
   public static readonly workingDir: string = "openapi";
   public static readonly openApiDocumentFileName: string = "openapi.json";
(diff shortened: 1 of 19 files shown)


packages/fx-core/src/plugins/resource/apim/constants.ts
```diff
@@ -8,7 +8,7 @@ export class ProjectConstants {
   public static readonly pluginShortName: string = "APIM";
   public static readonly pluginName: string = "fx-resource-apim";
   public static readonly pluginDisplayName: string = "API Management";
-  public static readonly configFilePath: string = "env.default.json";
+  public static readonly configFilePath: string = "state.xxx.json";
```
blackchoey, line 11: what's the purpose of this variable? the value seems strange... should we add env name to it dynamically?

packages/fx-core/src/plugins/resource/function/constants.ts
```diff
@@ -25,7 +25,7 @@ export class FunctionPluginInfo {
     FunctionConfigKey.storageAccountName,
     FunctionConfigKey.appServicePlanName,
     FunctionConfigKey.functionEndpoint,
-    FunctionConfigKey.functionAppId,
+    FunctionConfigKey.functionAppResourceId,
```
blackchoey, line 28: seems the key is changed? do we need update migration logic?

XiaofuHuang (approved):

packages/fx-core/src/plugins/resource/function/constants.ts
dilin-MS: Fixed.

packages/fx-core/src/plugins/resource/apim/constants.ts
dilin-MS: This config file path variable is used in error message of `EmptyConfigValue` error type. 

> Project configuration '${name}' of '${component}' is missing in **'${ProjectConstants.configFilePath}'**. Retry ${retryOperation} or set the value manually.


Since the original `env.default.json` does not exist, it is srtange to say somehing is missing in env.default.json. So update it to `state.xxx.json`. Any better idea?

blackchoey (approved):

packages/fx-core/src/plugins/resource/apim/constants.ts
dilin-MS: Revert the bug-fix and open a [bug](https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/12493238) for @XiaofuHuang .

PR #2808, opened 2021-11-01
chore(samples): extract the sample info in common lib

qinezh:
Extract the samples info into common lib in fx-core, so that there will be single source of truth for the samples info.

### DIFF
--- a/packages/vscode-extension/src/controls/SampleGallery.tsx
+++ b/packages/vscode-extension/src/controls/SampleGallery.tsx
@@ -12,7 +12,8 @@ import { EventMessages } from "./messages";
 interface SampleInfo {
   id: string;
   title: string;
-  description: string;
+  shortDescription: string;
+  fullDescription: string;
   tags: string[];
   link: string;
 }
@@ -110,7 +111,7 @@ class SampleAppCardList extends React.Component<any, any> {
             image={imageMapping[sample.id]}
             tags={sample.tags}
             title={sample.title}
-            description={sample.description}
+            description={sample.fullDescription}
             sampleAppFolder={sample.id}
             sampleAppUrl={sample.link}
           />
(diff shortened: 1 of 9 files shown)


Alive-Fish (dismissed):

packages/vscode-extension/src/controls/SampleGallery.tsx
```diff
@@ -39,61 +72,54 @@ export default class SampleGallery extends React.Component<any, any> {
           styles={{ root: { overflow: "visible" } }}
           tokens={{ childrenGap: 20 }}
         >
-          <SampleAppCard
```
HuihuiWu-Microsoft, line 42: The contents on sample gallery page for vscode extension is not totally the same with CLI/core as you can see the description has more details.

qinezh: is it by design to have two kinds (scaffold from sample & view sample) of description even for the same sample app?

HuihuiWu-Microsoft: Yes, it's by design because we don't have much space to show more detailed info when scaffolding from sample. But for sample gallery, there's enough space and we want the description as detailed as possible. You can also discuss with PM&UX whether we can align the descriptions and just have one version.

qinezh: Thanks for the clarification. I'm not going to change the behavoir if it's by design as it's out of scope of this PR. I will add `shortDescription` and `fullDescription` to `samples-config.json` to mitigate the issue. Does it sound good to you?

HuihuiWu-Microsoft (approved):
LGTM

Alive-Fish (approved):

jayzhang (approved):

Issue #2299, opened 2021-09-13
bug report: AppStudioPlugin.GetRemoteConfigFailed


bpasham:
This happens every time - forces to create azure resources everytime manifest validation is run 
Missing configuration data for manifest. Run 'provision' first. Data required: webApplicationInfoResource.

stack:
GetRemoteConfigFailed: Missing configuration data for manifest. Run 'provision' first. Data required: webApplicationInfoResource.
	at Function.SystemError (<REDACTED: user-file-path>:2:5229528)
	at AppStudioPluginImpl.getConfigForCreatingManifest (<REDACTED: user-file-path>:2:5215987)
	at AppStudioPluginImpl.getAppDefinitionAndManifest (<REDACTED: user-file-path>:2:5226269)
	at AppStudioPluginImpl.validateManifest (<REDACTED: user-file-path>:2:5194099)
	at async AppStudioPlugin.validateManifest (<REDACTED: user-file-path>:2:5177219)
	at async AppStudioPlugin.executeUserTask (<REDACTED: user-file-path>:2:5185720)
	at async TeamsAppSolution.executeUserTask (<REDACTED: user-file-path>:2:5860412)
	at async TeamsAppSolution.exports.ErrorHandlerMW (<REDACTED: user-file-path>:2:4720301)
	at async FxCore.executeUserTask (<REDACTED: user-file-path>:2:4689876)
	at async FxCore.exports.LocalSettingsWriterMW (<REDACTED: user-file-path>:2:4723899)
	at async FxCore.<anonymous> (<REDACTED: user-file-path>:2:4718896)
	at async FxCore.exports.ProjectSettingsWriterMW (<REDACTED: user-file-path>:2:4729186)
	at async FxCore.exports.ContextInjecterMW (<REDACTED: user-file-path>:2:4711050)
	at async FxCore.exports.QuestionModelMW (<REDACTED: user-file-path>:2:4739676)
	at async FxCore.<anonymous> (<REDACTED: user-file-path>:2:4740135)
	at async FxCore.exports.LocalSettingsLoaderMW (<REDACTED: user-file-path>:2:4723531)
	at async FxCore.<anonymous> (<REDACTED: user-file-path>:2:4715054)
	at async FxCore.exports.ProjectSettingsLoaderMW (<REDACTED: user-file-path>:2:4728021)
	at async <REDACTED: user-file-path>:2:4710420
[...]

PR #2821, opened 2021-11-02
fix: do not throw error if user manually update manifest

nliu-ms:
Previous:
Throw error if some configurations are missing from env.default.json

Current:
If user manually updated manifest template, replace placeholders with actual values. Then do not throw error for missing configurations.

image

Resolve:
https://github.com/OfficeDev/TeamsFx/issues/2299


nliu-ms:
E2T TEST:
https://github.com/OfficeDev/TeamsFx/blob/2040f63d9aa23fc666ede164b3f61926362284a1/packages/cli/tests/e2e/smoke/AzureAppHappyPath.tests.ts#L125

jayzhang (approved):

PR #2823, opened 2021-11-02
chore: remove resource name suffix when arm enabled

blackchoey:
Remove the resourceNameSuffix in state.{env}.json when insider preview enabled per request, since this value is not used any more.
Further code cleanup is required when we make insider preview feature as stable release.


dooriya (approved):

blackchoey:
Some other resources depends on the suffix. Closing this PR temporary. Will reopen this PR when we confirmed all resources does not rely on this context value.

PR #2835, opened 2021-11-03
fix: click m365 account will trigger sign in logic

HuihuiWu-Microsoft:
ADO: 
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/12401147
Issue:
the signInAzure(M365)Callback will be triggered when user click related tree item even if user already sign in. This will introduce performance issue on collaborator feature. 
E2E TEST: Only affect UI


HuihuiWu-Microsoft:
E2E TEST: Only affect UI

1openwindow (approved):

PR #2838, opened 2021-11-03
perf(ARM): resource plugins support updateArmTemplates 

wenytang-ms:


packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -416,18 +422,41 @@ async function doGenerateArmTemplate(ctx: SolutionContext): Promise<Result<any,
...
+      } else {
+        const msg = format(
+          getStrings().solution.GenerateArmTemplateFailNotice,
+          ctx.projectSettings?.appName
+        );
+        ctx.logProvider?.error(msg);
+        return result;
+      }
+    } else if (pluginWithArm.updateArmTemplates) {
+      const pluginContext = getPluginContext(ctx, pluginWithArm.name);
```
blackchoey, line 449: no need to run update arm templates for selected plugin?

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -443,7 +472,7 @@ async function doGenerateArmTemplate(ctx: SolutionContext): Promise<Result<any,
 
   // Write bicep content to project folder
   if (bicepOrchestrationTemplate.needsGenerateTemplate()) {
-    await backupExistingFilesIfNecessary(ctx);
+    // await backupExistingFilesIfNecessary(ctx);
```
blackchoey, line 475: remove this comment?

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -468,18 +497,20 @@ async function doGenerateArmTemplate(ctx: SolutionContext): Promise<Result<any,
...
     }
 
-    await fs.writeFile(
-      path.join(templateFolderPath, bicepOrchestrationProvisionFileName),
-      bicepOrchestrationProvisionContent
-    );
-    const res = bicepOrchestrationTemplate.applyReference(bicepOrchestrationConfigContent);
+    let res = bicepOrchestrationTemplate.applyReference(bicepOrchestrationProvisionContent);
+    await fs.appendFile(path.join(templateFolderPath, bicepOrchestrationProvisionFileName), res);
+    res = bicepOrchestrationTemplate.applyReference(bicepOrchestrationConfigContent);
```
blackchoey, line 502: better naming?

packages/fx-core/src/plugins/resource/bot/plugin.ts
```diff
@@ -189,6 +189,22 @@ export class TeamsBotImpl {
...
+    const result: ArmTemplateResult = {
+      Configuration: {
+        Modules: {
+          botConfiguration: await fs.readFile(
+            path.join(bicepTemplateDir, PathInfo.ConfigurationModuleTemplateV2FileName),
+            ConstantString.UTF8Encoding
+          ),
+        },
+      },
+    };
```
xzf0587, line 204: The return of updateArmTemplates is different from generateArmTemplates. If we add orchestration part in this return, what will happen? 
Should we replace ArmTemplateResult with a specified class?

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
wenytang-ms: sure

packages/fx-core/src/plugins/resource/bot/plugin.ts
wenytang-ms: updateArmTemplates means will update moduleConfiguration.bicep files if need to regenerate.
if orchestration is needed in updateArmTemplates? if so, the orchestraction will do append files to provision.bicep or config.bicep files. this operation can support on solution as well

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
wenytang-ms: sure

PR #2841, opened 2021-11-03
chore: support resolving secret value from environment variables

dooriya:
In current multi-env design, user may need to config some secret value in the `config.<env>.json` file, such as AAD client secret, bot app secret, as well as some other user customized secret data.

To prevent check-in the credential into the project's repo, we are now supporting resolving the credential in env config from environment variables, you can define your config like this:
```json
{
   "clientId": "{{$env.my_env_name}}"
}
```

### DIFF
--- a/packages/fx-core/src/core/environment.ts
+++ b/packages/fx-core/src/core/environment.ts
@@ -38,18 +38,19 @@ import {
   ModifiedSecretError,
 } from "..";
 import { GLOBAL_CONFIG } from "../plugins/solution/fx-solution/constants";
-import { readJson } from "../common/fileUtils";
 import { Component, sendTelemetryErrorEvent, TelemetryEvent } from "../common/telemetry";
-import { isMultiEnvEnabled } from "../common";
+import { compileHandlebarsTemplateString, isMultiEnvEnabled } from "../common";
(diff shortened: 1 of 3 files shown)


dooriya:

qinezh:
According to our previous discussion, the syntax to reference environment variables should be like:

```json
{
   "clientId": "${{my_env_name}}"
}
```
And the syntax is consistent with the existing Azure pipeline and GitHub action syntax. Is the design changed?

dooriya:
> According to our previous discussion, the syntax to reference environment variables should be like:
> 
> ```json
> {
>    "clientId": "${{my_env_name}}"
> }
> ```
> 
> And the syntax is consistent with the existing Azure pipeline and GitHub action syntax. Is the design changed?

Yes, after discussion, the final design will be `{{$env: my_env_name}}` to align with other placeholders in our teams toolkit, and we can use `$env:` prefix to differentiate the environment variables. I'll update the design document in One Note.

jayzhang (dismissed):

qinezh (dismissed):

a1exwang (approved):

jayzhang (dismissed):

jayzhang (approved):

PR #2845, opened 2021-11-04
chore: export create manifest function to handle migrate multienv cor…

HuihuiWu-Microsoft:
…ner case

Tested locally the scenario that manifest file is successfully created for local if user deleted the file.

### DIFF
--- a/packages/fx-core/src/plugins/resource/appstudio/plugin.ts
+++ b/packages/fx-core/src/plugins/resource/appstudio/plugin.ts
@@ -253,58 +253,6 @@ export class AppStudioPluginImpl {
   }
 
   /**
-   * ask app common questions to generate app manifest
-   * @param settings
-   * @returns
-   */
-  private async createManifest(settings: ProjectSettings): Promise<TeamsAppManifest | undefined> {
-    const solutionSettings: AzureSolutionSettings =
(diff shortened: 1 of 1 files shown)


packages/fx-core/src/plugins/resource/appstudio/plugin.ts
```diff
@@ -594,13 +542,20 @@ export class AppStudioPluginImpl {
         await fs.writeFile(`${appDir}/${MANIFEST_LOCAL}`, JSON.stringify(localManifest, null, 4));
       }
     } else {
-      manifest = await this.createManifest(ctx.projectSettings!);
+      const solutionSettings: AzureSolutionSettings = ctx.projectSettings
+        ?.solutionSettings as AzureSolutionSettings;
+      const hasFrontend = solutionSettings.capabilities.includes(TabOptionItem.id);
+      const hasBot = solutionSettings.capabilities.includes(BotOptionItem.id);
+      const hasMessageExtension = solutionSettings.capabilities.includes(MessageExtensionItem.id);
+      manifest = await createManifest(
```
HuihuiWu-Microsoft, line 550: I think it's better to align function signature

nliu-ms (approved):

PR #2848, opened 2021-11-04
feat(vsc): remove hosting type from ui

1openwindow:
ADO: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/12400825

VSC experience:
RemoveHostingTypeVcs

CLI experience:
RemoveHostingTypeCli

CLI experience (interactive false): 
image

### DIFF
--- a/packages/fx-core/src/plugins/solution/fx-solution/solution.ts
+++ b/packages/fx-core/src/plugins/solution/fx-solution/solution.ts
@@ -122,6 +122,7 @@ import {
   ProgrammingLanguageQuestion,
   TabOptionItem,
   GetUserEmailQuestion,
+  TabSPFxItem,
 } from "./question";
 import {
   getActivatedResourcePlugins,
@@ -965,9 +966,9 @@ export class TeamsAppSolution implements Solution {
       node.addChild(capNode);
 
       // 1.1 hostType
-      const hostTypeNode = new QTreeNode(FrontendHostTypeQuestion);
-      hostTypeNode.condition = { contains: TabOptionItem.id };
-      capNode.addChild(hostTypeNode);
+      //const hostTypeNode = new QTreeNode(FrontendHostTypeQuestion);
+      //hostTypeNode.condition = { contains: TabOptionItem.id };
+      //capNode.addChild(hostTypeNode);
 
       // 1.1.1 SPFX Tab
       const spfxPlugin: Plugin = new SpfxPlugin();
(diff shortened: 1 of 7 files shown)


1openwindow:
E2E TEST: https://github.com/OfficeDev/TeamsFx/blob/f6f87441aa4176fabddb2feb3a60bd5983d971dc/packages/fx-core/tests/plugins/solution/solution.create.test.ts#L40

packages/fx-core/src/plugins/solution/fx-solution/v2/utils.ts
```diff
@@ -196,8 +198,13 @@ export function fillInSolutionSettings(
...
   }
   let hostType = answers[AzureSolutionQuestionNames.HostType] as string;
-  if (capabilities.includes(BotOptionItem.id) || capabilities.includes(MessageExtensionItem.id))
+  if (
+    capabilities.includes(BotOptionItem.id) ||
+    capabilities.includes(MessageExtensionItem.id) ||
+    capabilities.includes(TabOptionItem.id)
+  )
     hostType = HostTypeOptionAzure.id;
+  if (capabilities.includes(TabSPFxItem.id)) hostType = HostTypeOptionSPFx.id;
```
HuihuiWu-Microsoft, line 207: I think there's much more than here that you need to update since currently there're many places use capabilities.includes(TabOptionItem.id) to indicate whether there's tab cap.

HuihuiWu-Microsoft:
Have you confirmed with PM that this change also applies for CLI?

chagong:
Looks like you cut one param off from QM, that will need E2E test change because E2E is using non-interactive command line.

packages/fx-core/src/plugins/solution/fx-solution/solution.ts
```diff
@@ -965,9 +966,9 @@ export class TeamsAppSolution implements Solution {
       node.addChild(capNode);
 
       // 1.1 hostType
-      const hostTypeNode = new QTreeNode(FrontendHostTypeQuestion);
-      hostTypeNode.condition = { contains: TabOptionItem.id };
-      capNode.addChild(hostTypeNode);
+      //const hostTypeNode = new QTreeNode(FrontendHostTypeQuestion);
+      //hostTypeNode.condition = { contains: TabOptionItem.id };
+      //capNode.addChild(hostTypeNode);
```
Alive-Fish, line 971: Removing the host type parameter needs to change these e2e test cases which uses it:
image

packages/fx-core/src/plugins/solution/fx-solution/v2/utils.ts
1openwindow: fixed.

packages/fx-core/src/plugins/solution/fx-solution/solution.ts
1openwindow: fixed

jayzhang (approved):

Alive-Fish (approved):

a1exwang (approved):

PR #2852, opened 2021-11-04
refactor(api): add new platform enum item for onboarding blazor plugin

1yefuwang1:
closes https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_sprints/taskboard/Teams%20Extensibility%20E2E%20Team/Microsoft%20Teams%20Extensibility/CY21-11.1?workitem=12517463

In VS extension, identiy plugin should not be activated. So the activate() method needs to know in which platform it is operating.


1yefuwang1:
E2E test: modifying API only thus no tests provided

PR #2859, opened 2021-11-05
refactor: move package to build

nliu-ms:
build/appPackage/appPackage.{env}.zip
build/appPackage/manifest.{env}.zip

Trigger from:
- Remote: zip package / provision / publish
- Local: local debug


nliu-ms:
E2E TEST: https://github.com/OfficeDev/TeamsFx/blob/149cdd67807a59f3325ba5e6e167d0e9897df727/packages/cli/tests/e2e/multienv/TestRemoteHappyPath.tests.ts#L132

packages/fx-core/src/plugins/resource/appstudio/plugin.ts
```diff
@@ -590,26 +591,35 @@ export class AppStudioPluginImpl {
...
     if (!ctx.envInfo?.envName) {
       throw new Error("Failed to get target environment name from plugin context.");
     }
 
     const appDirectory = await getAppDirectory(ctx.root);
-    const zipFileName: string = isMultiEnvEnabled()
-      ? `${ctx.root}/${AppPackageFolderName}/appPackage.${ctx.envInfo.envName}.zip`
-      : `${ctx.root}/${AppPackageFolderName}/appPackage.zip`;
+    let zipFileName: string;
+    if (isMultiEnvEnabled()) {
```
HuihuiWu-Microsoft, line 603: So this partially applies to multi-env, app package will be generated when publish/zip/provision, but the manifest file will not be generated?

nliu-ms: It applies to multi-env, the manifest file will also be generated from line 684

HuihuiWu-Microsoft: Sorry, typo. I mean non multi env. If it's not multi-env,  the app package will still be generated when publish/zip/provison in current logic, right?

nliu-ms: no, for non multi-env, only zip/publish will trigger build. Provision/local debug won't.

HuihuiWu-Microsoft (dismissed):

nliu-ms:
E2E passed: https://github.com/OfficeDev/TeamsFx/actions/runs/1433926695

jayzhang (approved):

a1exwang (approved):

PR #2860, opened 2021-11-05
fix(multi-env): fix teamsfx env list and teamsfx env add

a1exwang:
`teamsfx env list` previously required `action` option be set, but apparently there is no `action` option in the `list` command.

Use `<>` instead of `[]` to make the `action` option in `teamsfx env` command a required option. This aligns with other commands like `teamsfx config` and `teamsfx permission` 

E2E TEST https://github.com/OfficeDev/TeamsFx/blob/d3ab1f1be30620255343538fb1839bec273613f1/packages/cli/tests/e2e/multienv/TestRemoteHappyPath.tests.ts#L38

e2e test result: https://github.com/OfficeDev/TeamsFx/actions/runs/1424222811

### DIFF
--- a/packages/cli/src/cmds/env.ts
+++ b/packages/cli/src/cmds/env.ts
@@ -30,18 +30,12 @@ import { EnvNodeNoCreate } from "../constants";
 
 export default class Env extends YargsCommand {
   public readonly commandHead = `env`;
-  public readonly command = `${this.commandHead} [action]`;
+  public readonly command = `${this.commandHead} <action>`;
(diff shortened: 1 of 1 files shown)


a1exwang:
E2E TEST: https://github.com/OfficeDev/TeamsFx/blob/d3ab1f1be30620255343538fb1839bec273613f1/packages/cli/tests/e2e/multienv/TestRemoteHappyPath.tests.ts#L38

Alive-Fish (approved):

PR #2873, opened 2021-11-08
perf(apim): update apim generate v2 arm template

dilin-MS:


blackchoey (approved):
Can we revert the changes to packages/cli/package-lock.json and packages/sdk/pcakge-lock.json?

xzf0587 (approved):

dilin-MS:
> Can we revert the changes to packages/cli/package-lock.json and packages/sdk/pcakge-lock.json?

@blackchoey The changes exists when execute `npm run setup` under fx-core. Do you think we should not include the changes?

PR #2876, opened 2021-11-08
test(migrate-v1): enable migrate v1 test cases for multi env

XiaofuHuang:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_boards/board/t/Infrastructure%20Team/Stories/?workitem=12023943

### DIFF
--- a/packages/fx-core/tests/core/core.test.ts
+++ b/packages/fx-core/tests/core/core.test.ts
@@ -210,7 +210,7 @@ describe("Core basic APIs", () => {
       {
         description: "skip ask app name",
         appName: appName,
-        projectPath: path.resolve(os.tmpdir(), "v1projectpath", `${appName}`),
+        projectPath: path.resolve(os.tmpdir(), "v1projectpath", appName),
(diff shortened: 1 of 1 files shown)


packages/fx-core/tests/core/core.test.ts
```diff
@@ -283,17 +283,19 @@ describe("Core basic APIs", () => {
           const validSettingsResult = validateSettings(projectSettings);
           assert.isTrue(validSettingsResult === undefined);
 
-          const envInfoResult = await loadSolutionContext(tools, inputs, projectSettings);
-          if (envInfoResult.isErr()) {
-            assert.fail("failed to load env info");
-          }
+          if (!commonTools.isMultiEnvEnabled()) {
```
a1exwang, line 286: why don't you check `solutionContext` anymore?

XiaofuHuang: solutionContext should be in an env, migrate v1 do not have any env.

a1exwang (approved):

PR #2880, opened 2021-11-08
chore(blazor): integrate blazor plugin with solution

hund030:


henzhang-ms (approved):
Is it a short term workaround?

KennethBWSong (approved):

hund030:
> Is it a short term workaround?
The feature flag is a workaround. In the future, we just rely on project setting and question model to decide which plugin should be activated, rather than a feature flag.

PR #2899, opened 2021-11-10
chore: remove local debug ab test code and merge notifications

HuihuiWu-Microsoft:


packages/vscode-extension/src/handlers.ts
```diff
@@ -819,14 +819,13 @@ async function openMarkdownHandler() {
     const workspaceFolder = workspace.workspaceFolders[0];
     const workspacePath: string = workspaceFolder.uri.fsPath;
     let targetFolder: string | undefined;
-    showLocalDebugMessage();
     if (await isMigrateFromV1Project(workspacePath)) {
       targetFolder = workspacePath;
     } else if (await isSPFxProject(workspacePath)) {
-      showChangeLocationMessage();
```
HuihuiWu-Microsoft, line 826: One thing to note is that the local debug notification will not pop up for migrate v1 project. I think  that's acceptable. Add @XiaofuHuang for awareness.

1openwindow (approved):

PR #2900, opened 2021-11-10
fix: download sample will display wrong location

HuihuiWu-Microsoft:
ADO:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/12493620
E2E TEST: only affect extension UI, will add UI test later


HuihuiWu-Microsoft:
Will automatically fixed after download sample logic centralized.

PR #2901, opened 2021-11-10
fix(multi-env): improve resource group permission error message

a1exwang:
When selecting resource group before provision, and if the user does not have permission to acess current selected subscription, it will display a more user friendly error message.
image

https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/12545873

### DIFF
--- a/packages/fx-core/src/plugins/solution/fx-solution/commonQuestions.ts
+++ b/packages/fx-core/src/plugins/solution/fx-solution/commonQuestions.ts
@@ -34,9 +34,9 @@ import {
   RESOURCE_GROUP_NAME,
   SolutionError,
   SolutionSource,
-  FailedToCheckResourceGroupExistence,
+  FailedToCheckResourceGroupExistenceError,
   SUBSCRIPTION_ID,
-  UnauthorizedToCheckResourceGroup,
+  UnauthorizedToCheckResourceGroupError,
 } from "./constants";
 import { v4 as uuidv4 } from "uuid";
 import { ResourceManagementClient } from "@azure/arm-resources";
@@ -702,11 +702,11 @@ function handleRestError<T>(
   // ARM API will return 403 with empty body when users does not have permission to access the resource group
   if (restError.statusCode === 403) {
     return err(
-      new UnauthorizedToCheckResourceGroup(resourceGroupName, subscriptionId, subscriptionName)
+      new UnauthorizedToCheckResourceGroupError(resourceGroupName, subscriptionId, subscriptionName)
     );
   } else {
     return err(
-      new FailedToCheckResourceGroupExistence(
+      new FailedToCheckResourceGroupExistenceError(
         restError,
         resourceGroupName,
         subscriptionId,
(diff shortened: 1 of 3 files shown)


a1exwang:

packages/fx-core/src/plugins/solution/fx-solution/commonQuestions.ts
```diff
@@ -671,24 +691,68 @@ async function createNewResourceGroup(
...
+      message: restError.message,
+    });
+    return err(
+      returnUserError(
+        new Error(
+          `Failed to check the existence of resource group '${resourceGroupName}' in subscription '${subscriptionInfoString}', raw error: '${rawErrorString}'`
+        ),
+        SolutionSource,
+        SolutionError.FailedToCheckResourceGroupExistence
+      )
```
dooriya, line 728: We have several `UserError` related to the resource group check, is that possible to wrap/abstract these kinds of `UserError` for better error handling/monitoring, I remember Huajie has talked about how to define the Error in Teams toolkit before.

packages/fx-core/src/plugins/solution/fx-solution/commonQuestions.ts
```diff
@@ -671,24 +691,68 @@ async function createNewResourceGroup(
...
-    // Wrap the error to prevent showing empty error message to the users.
     let error;
-    const errorMessage = `Failed to check the existence of the resource group '${resourceGroupName}', error: '${e}'`;
-    if (e instanceof Error) {
+    if (e instanceof RestError) {
+      return handleRestError(e, resourceGroupName, subscriptionId, subscriptionName);
+    } else if (e instanceof Error) {
       // reuse the original error object to prevent losing the stack info
-      e.message = errorMessage;
+      e.message = `Failed to check the existence of resource group '${resourceGroupName}', error: '${e}'`;
```
qinezh, line 748: > `error: '${e}'`
do we need to stringify the object as well here?

a1exwang: The reason here is for some other types of non-RestError (like network error), the user can know:
1. The error is related to checking resource group
2. The true reason of the error so he can take action or report issue.

So I wrap these two pieces of information into this error message.

packages/fx-core/src/plugins/solution/fx-solution/commonQuestions.ts
a1exwang: It makes sense. I will refactor it.

packages/fx-core/src/plugins/solution/fx-solution/constants.ts
```diff
@@ -234,3 +234,51 @@ export enum SolutionTelemetrySuccess {
...
+      subscriptionId + (subscriptionName.length > 0 ? `(${subscriptionName})` : "");
+    super(
+      new.target.name,
+      `Unauthorized to check the existence of resource group '${resourceGroupName}' in subscription '${subscriptionInfoString}'. Please check your Azure subscription.`,
+      SolutionSource
+    );
+  }
+}
+
+export class FailedToCheckResourceGroupExistence extends UserError {
```
dooriya, line 250: nit: can we name it with `Error` suffix?

qinezh (approved):

dooriya (approved):

jayzhang (approved):

PR #2913, opened 2021-11-11
fix: make getLocalDebugEnvs readonly operation

swatDong:
Make `getLocalDebugEnvs` a read-only operation so it could be called before `localDebug`.
(This fixes ADO bug 12514524)

### DIFF
--- a/packages/fx-core/tests/plugins/resource/localdebug/unit/index.test.ts
+++ b/packages/fx-core/tests/plugins/resource/localdebug/unit/index.test.ts
@@ -64,14 +64,14 @@ describe(LocalDebugPluginInfo.pluginName, () => {
         numConfigurations: 5,
         numCompounds: 2,
         numTasks: 9,
-        numLocalEnvs: isMultiEnvEnabled() ? 29 : 30,
+        numLocalEnvs: isMultiEnvEnabled() ? 30 : 30,
       },
       {
         programmingLanguage: "typescript",
         numConfigurations: 5,
         numCompounds: 2,
         numTasks: 9,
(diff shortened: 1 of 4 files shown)


swatDong:
E2E TEST: https://github.com/OfficeDev/TeamsFx-UI-Test/blob/85e2ed14b33b839e603e7b6f9e8abb82c77a9fcc/src/ui-test/localdebug/localdebug-tab.test.ts#L8

packages/fx-core/tests/plugins/resource/localdebug/unit/index.test.ts
```diff
@@ -64,14 +64,14 @@ describe(LocalDebugPluginInfo.pluginName, () => {
         numConfigurations: 5,
         numCompounds: 2,
         numTasks: 9,
-        numLocalEnvs: isMultiEnvEnabled() ? 29 : 30,
+        numLocalEnvs: isMultiEnvEnabled() ? 30 : 30,
```
dooriya, line 67: Since the env count is consistent whether the `MultiEnv` feature flag is enabled or not, we can just set to a fix number.

swatDong: Not all (just for tab+auth it's the same), and it's possible to adjust when adding new features to multi-env.

dooriya: OK, That would make sense.

dooriya (approved):

kuojianlu (approved):

PR #2923, opened 2021-11-11
fix: download  azure sample will download two samples

HuihuiWu-Microsoft:
ADO:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/12565786

E2E TEST: ignore for quick hotfix and will add E2E test later

### DIFF
--- a/packages/fx-core/src/common/tools.ts
+++ b/packages/fx-core/src/common/tools.ts
@@ -284,7 +284,12 @@ export async function saveFilesRecursively(
   await Promise.all(
     zip
       .getEntries()
-      .filter((entry) => !entry.isDirectory && entry.entryName.includes(appFolder))
+      .filter(
+        (entry) =>
+          !entry.isDirectory &&
+          entry.entryName.includes(appFolder) &&
+          entry.entryName.split("/").includes(appFolder)
+      )
       .map(async (entry) => {
         const entryPath = entry.entryName.substring(entry.entryName.indexOf("/") + 1);
         const filePath = path.join(dstPath, entryPath);


packages/fx-core/src/common/tools.ts
```diff
@@ -284,7 +284,12 @@ export async function saveFilesRecursively(
   await Promise.all(
     zip
       .getEntries()
-      .filter((entry) => !entry.isDirectory && entry.entryName.includes(appFolder))
+      .filter(
+        (entry) =>
+          !entry.isDirectory &&
+          entry.entryName.includes(appFolder) &&
+          entry.entryName.split("/").includes(appFolder)
+      )
```
kuojianlu, line 292: Suppose the following scenario:
1. There is one sample named `sample-1`
2. Ther is another sample named `sample-2`.
3. In `sample-2`, there is a file accidentally named `sample-1`, like `sample-2/some/path/sample-1`

If user downloads `sample-1`, will `sample-2/some/path/sample-1` be also downloaded?

HuihuiWu-Microsoft: Good catch. That's correct. But there's not an ideal way to check which one belongs to current app folder. 

One way I can think of is to check the second string after splitting using '/' but actually we can't make sure the folder structure is fixed either. So since the scenario you mentioned should have little possibility. I think we can fix it in this way for now.

HuihuiWu-Microsoft:
E2E TEST: ignore for quick hotfix and will add E2E test later

packages/fx-core/src/common/tools.ts
kuojianlu: Understood. Agree to just quick fix the bug introduced by `todo-list-with-Azure-backend-M365` sample.

kuojianlu (approved):

PR #2924, opened 2021-11-11
chore: migrate SQL skipAddingUser config

blackchoey:


packages/fx-core/src/core/middleware/projectMigrator.ts
```diff
@@ -436,16 +437,15 @@ async function migrateMultiEnv(projectPath: string): Promise<void> {
...
+  await ensureProjectSettings(projectSettings, envDefaultFilePath);
 
   const appName = await getAppName(projectSettings);
   if (!migrateFromV1) {
     //config.dev.json
-    await fs.writeFile(
-      path.join(fxConfig, "config.dev.json"),
-      JSON.stringify(getConfigDevJson(appName), null, 4)
-    );
+    await fs.writeFile(configDevJsonFilePath, JSON.stringify(getConfigDevJson(appName), null, 4));
```
meifans, line 448: We can move  the logic of "migrate skipAddingSqlUser" to here. 
Cause they are all about the code of generating config.dev.json file.

blackchoey: thanks, updated

meifans (approved):

PR #2927, opened 2021-11-11
feat(extension,core,cli,api): add project-id in download-sample telemetry event

jayzhang:
1. vs code extension and cli call core.createProject() API to download sample
2. core add project-id in download-sample telemetry
3. collect sample id instead of sample object for sample select question


jayzhang:

LongOddCode (approved):

packages/vscode-extension/src/controls/webviewPanel.ts
```diff
@@ -154,76 +149,25 @@ export class WebviewPanel {
...
-        }
-      }
+    };
+    ExtTelemetry.sendTelemetryEvent(TelemetryEvent.DownloadSampleStart, props);
+    const inputs: Inputs = getSystemInputs();
+    inputs.stage = Stage.create;
+    inputs["scratch"] = "no";
+    inputs["samples"] = msg.data.appFolder;
+    const res = await runCommand(Stage.create, inputs);
+    if (inputs.projectId) {
```
HuihuiWu-Microsoft, line 162: After runCommand(Stage.create, inputs) is executed, there's chance that the project is opened and VS Code is reloaded, so the codes after this line might get lost.

LongOddCode (dismissed):

packages/vscode-extension/src/handlers.ts
```diff
@@ -482,8 +483,10 @@ export async function runCommand(
           const uri = Uri.file(tmpResult.value);
           // 15% events are lost due to open other vs code instance, so collect the data before open folder.
           await processResult(eventName, result, inputs);
-          commands.executeCommand("vscode.openFolder", uri);
-          result = ok(null);
+          if (reloadForCreate) {
```
HuihuiWu-Microsoft, line 497: Do you need to set true as default value for other entries?

HuihuiWu-Microsoft (approved):

LongOddCode (approved):

packages/vscode-extension/src/handlers.ts
```diff
@@ -360,12 +360,16 @@ export function getSystemInputs(): Inputs {
   return answers;
 }
 
-export async function createNewProjectHandler(args?: any[]): Promise<Result<null, FxError>> {
+export async function createNewProjectHandler(args?: any[]): Promise<Result<any, FxError>> {
   ExtTelemetry.sendTelemetryEvent(TelemetryEvent.CreateProjectStart, getTriggerFromProperty(args));
-  return await runCommand(Stage.create);
+  const result = await runCommand(Stage.create);
+  if (result.isOk()) {
+    commands.executeCommand("vscode.openFolder", result.value);
```
chagong, line 367: So this will make sure all the telemetries been sending out before we open folder, right?

chagong (approved):

packages/vscode-extension/src/handlers.ts
jayzhang: Yes

tecton (approved):

PR #2929, opened 2021-11-12
chore(multi-env): remove local env node from env tree view

dooriya:
Per discussed with Zhen, we will remove the `local` env node from tree view to avoid confusion.


HuihuiWu-Microsoft (commented):
LGTM

HuihuiWu-Microsoft (approved):
LGTM

qinezh (approved):

PR #2934, opened 2021-11-15
fix: including activities in app definition

### LINKED ISSUE #2909: activitites in manifest.source.json
We are working to implement SendActivityNotifications from our teams app, and we need to add activities in the manifest, however every time we add them in the manifest.source.json it will not update the app manifest in the developer portal for the app. If we add them in the developer portal it will be removed on the first debug session even though we have them in the manifest.source.json.
To reproduce:
1. add the following to the manifest,source.json:
 "activities": {
        "activityTypes": [
          {
            "type": "mentionTask",
            "description": "Task message for x",
            "templateText": "Hallo there is a new action for {userPrincipalName}"
          }
        ]
      }
2. save and run debug 
3. Check developer portal for the app settings and activity feed notification is not enabled.

**Expected behavior**
That the manifest is updated correctly, it works for setting names and etc.

nliu-ms:
resolves #2909


nliu-ms:

jayzhang (approved):

PR #2940, opened 2021-11-15
chore(upgrade): add backup folder to .gitignore

qinezh:
Add below files/folder to .gitignore to avoid unexpected check-in:

- .backup/
- .fx/configs/localSettings.json
- **/.env.teamsfx.local

work item: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/12554473/

### DIFF
--- a/packages/fx-core/src/core/middleware/projectMigrator.ts
+++ b/packages/fx-core/src/core/middleware/projectMigrator.ts
@@ -30,6 +30,7 @@ import { LocalSettingsProvider } from "../../common/localSettingsProvider";
 import { Middleware, NextFunction } from "@feathersjs/hooks/lib";
 import fs from "fs-extra";
 import path from "path";
+import os from "os";
 import { readJson } from "../../common/fileUtils";
 import { PluginNames } from "../../plugins/solution/fx-solution/constants";
 import { CoreSource, FxCore } from "..";
@@ -69,6 +70,7 @@ import {
 import * as dotenv from "dotenv";
 import { PlaceHolders } from "../../plugins/resource/spfx/utils/constants";
 import { Utils as SPFxUtils } from "../../plugins/resource/spfx/utils/utils";
+import { LocalEnvMultiProvider } from "../../plugins/resource/localdebug/localEnvMulti";
 
 const programmingLanguage = "programmingLanguage";
 const defaultFunctionName = "defaultFunctionName";
@@ -82,6 +84,7 @@ const parameterFileNameTemplate = "azure.parameters.@envName.json";
 const learnMoreLink = "https://aka.ms/teamsfx-migration-guide";
 const manualUpgradeLink = `${learnMoreLink}#upgrade-your-project-manually`;
 const upgradeReportName = `upgrade-change-logs.md`;
+const gitignoreFileName = ".gitignore";
 let updateNotificationFlag = false;
 let fromReloadFlag = false;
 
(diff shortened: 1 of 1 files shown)


packages/fx-core/src/core/middleware/projectMigrator.ts
```diff
@@ -678,6 +689,24 @@ async function backup(projectPath: string): Promise<void> {
   }
 }
 
+// append folder path to .gitignore under the project root.
+async function addPathToGitignore(projectPath: string, ignoredPath: string) {
+  const relativePath = path.relative(projectPath, ignoredPath);
```
a1exwang, line 694: `path.relative` uses platform dependent path separator.

on Windows:
image

But [gitignore](https://git-scm.com/docs/gitignore#_pattern_format) only uses slash (not backslash) for path separator. So this may not work on Windows.

qinezh: Good point, I have only tested in WSL. Let me have a check.

a1exwang (approved):

packages/fx-core/src/core/middleware/projectMigrator.ts
```diff
@@ -444,6 +447,12 @@ async function migrateMultiEnv(projectPath: string): Promise<void> {
   await localSettingsProvider.save(
     localSettingsProvider.init(hasFrontend, hasBackend, hasBot, migrateFromV1)
   );
+  await addPathToGitignore(projectPath, localSettingsProvider.localSettingsFilePath);
```
meifans, line 450: Move this logic to `postMigration` method or recover it in `cleanup` method. 
Cause if upgrade failed, user shouldn't see it in .gitignore file.

packages/fx-core/src/core/middleware/projectMigrator.ts
```diff
@@ -678,6 +689,24 @@ async function backup(projectPath: string): Promise<void> {
...
 
+// append folder path to .gitignore under the project root.
+async function addPathToGitignore(projectPath: string, ignoredPath: string) {
+  const relativePath = path.relative(projectPath, ignoredPath).replace(/\\/g, "/");
+  await addItemToGitignore(projectPath, relativePath);
+}
+
+// append item to .gitignore under the project root.
+async function addItemToGitignore(projectPath: string, item: string) {
+  const gitignorePath = path.join(projectPath, gitignoreFileName);
```
meifans, line 700: It is better to add try catch for losing write permission. 

```suggestion
try{
     const gitignorePath = path.join(projectPath, gitignoreFileName);
     await fs.ensureFile(gitignorePath);
  
    const gitignoreContent = await fs.readFile(gitignorePath, "UTF-8");
    if (gitignoreContent.indexOf(item) === -1) {
      const appendedContent = os.EOL + item;
      await fs.appendFile(gitignorePath, appendedContent);
    }
  }catch(err) {
   // do nothing
  }
```

packages/fx-core/src/core/middleware/projectMigrator.ts
qinezh: Yeah, it makes sense to me.

packages/fx-core/src/core/middleware/projectMigrator.ts
qinezh: Shall we report this error instead of hiding it, otherwise .backup might be checked-in by mistake?

qinezh: Offline discussed, we're going to catch the error and add user warning to let user do it manually without blocking the migration.

meifans (approved):

PR #2945, opened 2021-11-15
chore: add ignore script on lerna bootstrap

wenytang-ms:
add npm run "install-all" to just install dependency without build project.


zhenyasav:
I think we had a few objectives behind this work:
- [ ] remove an unnecessary manual post-installation step such as `npm run setup` and handle any such logic using [`postinstall`] (https://docs.npmjs.com/cli/v8/using-npm/scripts)
- [ ] do not perform any builds by default during initial npm install
- [ ] ensure there is an easy to use build-all command runnable from the root
- [ ] consider if augmenting lerna with npm workspaces helps avoid doing a separate eslint install after main install, and if the main npm install gets faster with npm workspaces. [See here on workspaces](https://docs.npmjs.com/cli/v7/using-npm/workspaces) and this [ts template](https://github.com/Quramy/npm-ts-workspaces-example)

zhenyasav (changes requested):
see comment above

wenytang-ms:
@zhenyasav thanks for your comment
According to your comment, I understand that there are two implementation stages, one is install, the other is what post-install does.
install help developers handle dependencies, and post-install do prepare for building or debugging work.
My idea is to keep setup as most developers are used to this command. setup just do install work.
for post-install, create a build script for users to build the project.

You mentioned using workspace, which is a very nice feature. However, considering that many of our developers are using the node 14 version, the functions of the existing version can continue to be used, and the processing of eslint will continue to be optimized in the future.

Please help review this PR:
https://github.com/OfficeDev/TeamsFx/pull/3904

thanks!

PR #2958, opened 2021-11-16
chore(upgrade): no need do next step if need upgrade project or update Toolkit

qinezh:
fix bug: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/12606051


qinezh:
Wait to check whether the support of auto-reload after upgrade can resolve the bug mentioned in PR description. Closed for now.

PR #2962, opened 2021-11-16
feat(bot): improve error handling for graph aad api

YitongFeng-git:


YitongFeng-git:
e2e: https://github.com/OfficeDev/TeamsFx/blob/b8a7fdcddd116020ec6b158b8baba326703d8297/packages/cli/tests/e2e/bot/ProvisionResourceProjectWithNewBot.tests.ts#L21

packages/fx-core/src/plugins/resource/bot/aadRegistration.ts
```diff
@@ -66,7 +69,7 @@ export class AADRegistration {
         }
       );
     } catch (e) {
-      throw new ProvisionError(CommonStrings.AAD_CLIENT_SECRET, e);
+      throw this.handleError(e, CreateSecretError);
```
IvanJobs, line 72: Since handleError is a static method, better to call it using `AADRegistration.handleError()`

YitongFeng-git: I use the handleError in this class, it is different from AADRegistration.handleError()

IvanJobs (approved):
lgtm

PR #2968, opened 2021-11-17
chore(upgrade): no need next step if need upgrade project or update Toolkit

qinezh:
fix bug: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/12606051


meifans (approved):
LGTM

PR #2977, opened 2021-11-17
style(vsc): show notification in the middle

1openwindow:
ADO: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_sprints/taskboard/Teams%20Extensibility%20E2E%20Team/Microsoft%20Teams%20Extensibility/CY21-11.2?workitem=12544264
AskProvision1


jaeyonglee8687:
Not so sure we should have two dialogs asking to provision. Customers can make a decision whether or not to provision the resources right after the deployment command, and just execute the provisioning since we can merge the messages together. After that, I would expect to deploy automatically if possible. Is there any further user input required after the provision is finished?

image

HuihuiWu-Microsoft (dismissed):

1openwindow:
> Not so sure we should have two dialogs asking to provision. Customers can make a decision whether or not to provision the resources right after the deployment command, and just execute the provisioning since we can merge the messages together. After that, I would expect to deploy automatically if possible. Is there any further user input required after the provision is finished?
> 
> image

Yes, user may need to input some further answers after provisioning, but it is case by case, depend on what resource the users try to provision. From this point of view, combine deploy and provision process do not quick make sense.

PR #2980, opened 2021-11-17
chore: remove error log for upgrade cancelled

meifans:
- Remove error log for upgrade cancelled
- Pre-check does not return error
- Send reload telemetry when platform is VSCode


packages/fx-core/src/core/error.ts
```diff
@@ -265,7 +265,8 @@ export function ProjectSettingError(): UserError {
 
 export function UpgradeCanceledError(): UserError {
   return new UserError(
-    "UpgradeCanceledError",
+    // @see tools.isUserCancelError()
+    "User Cancel",
```
a1exwang, line 269: Use "UserCancel" instead of "User Cancel"? Because other error codes use CamelCase style and `tools.isUserCancelError()` also supports it.

packages/fx-core/src/core/middleware/projectMigrator.ts
```diff
@@ -269,7 +261,9 @@ async function migrateToArmAndMultiEnv(ctx: CoreHookContext): Promise<void> {
     telemetryProperties
   );
 
-  await preCheckKeyFiles(projectPath, ctx);
+  if (!(await preCheckKeyFiles(projectPath, ctx))) {
```
a1exwang, line 264: Is it better to put this check in `needMigrateToArmAndMultiEnv()`? If the key project files are broken, do not even pop up migrate dialog?

nliu-ms (approved):

packages/fx-core/src/core/error.ts
meifans: ok

packages/fx-core/src/core/middleware/projectMigrator.ts
meifans: preCheckKeyFiles is part of `migrateToArmAndMultiEnv`, I think we shouldn't check key files if user doesn't want to upgrade.

a1exwang: Then add a telemetry event if preCheckKeyFiles fails. Otherwise, we will lose this info (previously, errorHandlerMW can catch this exception and send telemetry).

LongOddCode (approved):

a1exwang (approved):

PR #2984, opened 2021-11-17
chore: add npm install handler

swatDong:
This is to prepare for customized local debug - to detect whether `npm install` is done and to run `npm install` for each component.
Tests added.


swatDong:
@1openwindow @HuihuiWu-Microsoft @nliu-ms  You are in reviewer list since there're test changes...
I've adjusted code owner as well but this PR still need your approval.

HuihuiWu-Microsoft (dismissed):

HuihuiWu-Microsoft:
> @1openwindow @HuihuiWu-Microsoft @nliu-ms You are in reviewer list since there're test changes... I've adjusted code owner as well but this PR still need your approval.

Done

a1exwang (dismissed):

kuojianlu:
One question that might be releated to this PR as well as the feature design: since we give back the work of installing packages to user, what if user uses `yarn` instead of `npm`? In this case, there will be no `package-lock.json` but `yarn.lock`.

packages/vscode-extension/src/debug/npmInstallHandler.ts
```diff
@@ -0,0 +1,54 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import * as fs from "fs-extra";
+import * as path from "path";
+import * as vscode from "vscode";
+
+import { loadPackageJson } from "./commonUtils";
+
+export async function hasNpmInstalled(folder: string): Promise<boolean> {
```
kuojianlu, line 10: `isNpmInstalled` may be more accurate.

swatDong: Changed to `checkDependencies` to not confuse with *npm is installed*

swatDong:
> One question that might be releated to this PR as well as the feature design: since we give back the work of installing packages to user, what if user uses `yarn` instead of `npm`? In this case, there will be no `package-lock.json` but `yarn.lock`.

Good catch. Updated to use [@npmcli/arborist](https://www.npmjs.com/package/@npmcli/arborist) to load installed dependencies from `node_modules`, which is the same package used by npm cli.

kuojianlu (approved):

HuihuiWu-Microsoft (approved):

PR #2985, opened 2021-11-17
feat: allow user to manually update manifest file

HuihuiWu-Microsoft:
User can update manifest file manually from editor context menu/preview file codelens/command palette.
For the former two scenarios, update current manifest. For  command palette, ask user to select env to update
update manifest

### DIFF
--- a/packages/fx-core/src/plugins/resource/appstudio/plugin.ts
+++ b/packages/fx-core/src/plugins/resource/appstudio/plugin.ts
@@ -449,7 +449,7 @@ export class AppStudioPluginImpl {
     isLocalDebug: boolean
   ): Promise<Result<string, FxError>> {
     const teamsAppId = await this.getTeamsAppId(ctx, isLocalDebug);
-    let manifest: TeamsAppManifest;
+    let manifest: any;
     let manifestString: string;
     const appDirectory = await getAppDirectory(ctx.root);
     const manifestPath = await this.getManifestTemplatePath(ctx.root, isLocalDebug);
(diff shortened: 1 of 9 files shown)


HuihuiWu-Microsoft:
E2E TEST: only for extension UI

packages/vscode-extension/src/codeLensProvider.ts
```diff
@@ -149,4 +157,15 @@ export class ManifestTemplateCodeLensProvider implements vscode.CodeLensProvider
 
     return codeLenses;
   }
+
+  private computePreviewCodeLenses(document: vscode.TextDocument) {
+    const codeLenses: vscode.CodeLens[] = [];
+    const command = {
+      title: "🔄Update",
```
HuihuiWu-Microsoft, line 164: Will update the icon after UX confirmed

packages/fx-core/src/plugins/resource/appstudio/plugin.ts
```diff
@@ -443,6 +444,124 @@ export class AppStudioPluginImpl {
...
+      }
+    }
+
+    const appStudioToken = await ctx?.appStudioToken?.getAccessToken();
+    const result = await this.updateApp(
+      ctx,
+      appDefinition,
+      appStudioToken!,
+      isLocalDebug,
+      !isLocalDebug,
```
nliu-ms, line 547: This param is indicating if toolkit will create a new teams app.

If here is false, and the app doesn't exist(e.g. user manually delete it), will cause some 404 error.

HuihuiWu-Microsoft: Yes, I'm thinking whether I should create the app for local debug if user haven't done local debug before. Maybe I can confirm with PM.

HuihuiWu-Microsoft: updated

nliu-ms (approved):

HuihuiWu-Microsoft:
Verified the spfx case, it can be updated successfully if user manually changed it to another provisioned id.

nliu-ms (approved):

PR #2986, opened 2021-11-17
chore(arm): add principalId for identity

xzf0587:
1. add principalId in output for identity


formulahendry (dismissed):

packages/fx-core/templates/plugins/resource/identity/bicep/provision.template.bicep
```diff
@@ -11,4 +11,5 @@ output identityOutput object = {
   identityName: userAssignedIdentityProvision.outputs.identityName
   identityResourceId: userAssignedIdentityProvision.outputs.identityResourceId
   identityClientId: userAssignedIdentityProvision.outputs.identityClientId
+  identityPrincipalId: userAssignedIdentityProvision.outputs.identityPrincipalId
```
blackchoey, line 14: do we need to persist it to state.dev.json? It seems just used inside ARM template (bicep).

xzf0587: It is fixed.

PR #2988, opened 2021-11-17
chore: change solution version to 2.0.0 for upgrade

meifans:
Fix [bug](https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/12554334)

### DIFF
--- a/packages/fx-core/src/core/middleware/projectMigrator.ts
+++ b/packages/fx-core/src/core/middleware/projectMigrator.ts
@@ -772,7 +772,7 @@ async function ensureProjectSettings(
       settings.defaultFunctionName || envDefault[PluginNames.FUNC]?.[defaultFunctionName];
   }
   if (!settings.version) {
-    settings.version = "1.0.0";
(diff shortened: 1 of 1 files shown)


a1exwang (approved):

packages/fx-core/src/core/middleware/projectMigrator.ts
```diff
@@ -772,7 +772,7 @@ async function ensureProjectSettings(
       settings.defaultFunctionName || envDefault[PluginNames.FUNC]?.[defaultFunctionName];
   }
   if (!settings.version) {
```
qinezh, line 774: For the case that settings.version is already set to 1.0.0, should we upgrade it as well?

qinezh: Here's an example: https://github.com/OfficeDev/TeamsFx-Samples/blob/main/NPM-search-connector-M365/.fx/settings.json#L18

qinezh (approved):

PR #2992, opened 2021-11-17
chore(vsc): add local debug and remote preview button in treeview

XiaofuHuang:
image

Localdebug: 
treeview-localdebug
Preview:
treeview-preview


packages/vscode-extension/package.json
```diff
@@ -232,11 +232,29 @@
...
         },
         {
           "command": "fx-extension.viewEnvironmentWithIcon",
-          "when": "view == teamsfx-environment && viewItem =~ /environment/",
+          "when": "view == teamsfx-environment && viewItem =~ /environment.*/",
+          "group": "inline@3"
+        },
+        {
+          "command": "fx-extension.preview",
+          "when": "view == teamsfx-environment && viewItem =~ /environment-provisioned/"
```
swatDong, line 244: Will tree view automatically refresh after provisioned?

dooriya: after deployment? otherwise user will got 404 when preview

XiaofuHuang: > Will tree view automatically refresh after provisioned?

Yes, the tree view will auto refresh

XiaofuHuang: > after deployment? otherwise user will got 404 when preview. Not sure if is there any way we can detect whether an app is deployed before?

image
image

swatDong: I have an idea but may change frontend provision behavior - create an default page at first provision.
(Similar to create new app service, after creating new app service without any code deployed, the homepage shows a welcome page instead of 404)

swatDong (approved):

dooriya (approved):

XiaofuHuang:
TODO: 
Confirm with PM about the preview icon
Remove view environment logic

yanjiang222:
> TODO: Confirm with PM about the preview icon Remove view environment logic

have one quick question: what's the difference between the play with a bug and the one without?

also, customers had feedback before that the arrow icon beside local environment is confusing since there is nothing down below, why need an arrow icon?

dooriya (approved):

XiaofuHuang:
> > TODO: Confirm with PM about the preview icon Remove view environment logic
> 
> have one quick question: what's the difference between the play icon with a bug and the one without?
> 
> also, customers had feedback before that the arrow icon beside local environment is confusing since there is nothing down below, why need an arrow icon?

If the play button with a bug, user can add break point in the code. (Local debug will start)
If the play button without a bug, the target env Teams App will be opened in the browser.

image
The latest UI design has removed the arrow icon.

packages/vscode-extension/src/telemetry/extTelemetryEvents.ts
```diff
@@ -139,6 +139,11 @@ export enum TelemetryEvent {
   MigrateTeamsTabAppCode = "migrate-teams-tab-app-code",
   MigrateTeamsManifestStart = "migrate-teams-manifest-start",
   MigrateTeamsManifest = "migrate-teams-manifest",
+
+  TreeViewLocalDebug = "treeview-localdebug",
```
HuihuiWu-Microsoft, line 143: Can we use existing RunIconDebug event with triggerFrom property to distinguish?

XiaofuHuang: They are different logic. One can choose to use local / remote debug, another only can choose local debug.

packages/vscode-extension/src/envTree.ts
```diff
@@ -66,16 +66,21 @@ export async function registerEnvTreeHandler(
...
         const provisionSucceeded = await getProvisionSucceedFromEnv(item);
+        const isLocal = item === LocalEnvironmentName;
         environmentTreeProvider.add([
           {
             commandId: "fx-extension.environment." + item,
             label: item,
             description: provisionSucceeded ? "(Provisioned)" : "",
             parent: TreeCategory.Environment,
-            contextValue: item === LocalEnvironmentName ? "local" : "environment",
+            contextValue: isLocal
```
HuihuiWu-Microsoft, line 76: Will this get updated after user provision?

XiaofuHuang: Yes, the treeview will update after provision.

HuihuiWu-Microsoft (approved):

dooriya (approved):

PR #2994, opened 2021-11-17
fix: spfx deploy cause 401 error

HuihuiWu-Microsoft:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/12587039?src=WorkItemMention&src-action=artifact_link

Root cause:
The scope is not updated if sharepoint.getAccessToken() called before user switch account.

E2E TEST: the token provider implementation for e2e is not the same with product code

### DIFF
--- a/packages/vscode-extension/src/commonlib/sharepointLogin.ts
+++ b/packages/vscode-extension/src/commonlib/sharepointLogin.ts
@@ -90,16 +90,14 @@ export class SharepointLogin extends login implements SharepointTokenProvider {
     }
 
     await SharepointLogin.codeFlowInstance.reloadCache();
-    if (!SharepointLogin.codeFlowInstance.account) {
-      try {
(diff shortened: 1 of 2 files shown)


packages/vscode-extension/src/commonlib/sharepointLogin.ts
```diff
@@ -90,16 +90,14 @@ export class SharepointLogin extends login implements SharepointTokenProvider {
...
-      try {
-        const scopes = await this.getScopes(showDialog);
-        if (!scopes) {
-          return undefined;
-        }
-        SharepointLogin.codeFlowInstance.scopes = scopes;
-      } catch (error) {
-        throw error;
+    try {
+      const scopes = await this.getScopes(showDialog);
```
swatDong, line 94: Will `getScopes` be called twice since line 77 has the same code?

swatDong (approved):

PR #3003, opened 2021-11-18
fix: fix bicepChecker 401 error

a1exwang:
- Do not set axios.default because this will affect all axios instances globally
- SPFx sets a global `Authorization` headers, so when bicepChecker calls GitHub API (this is public API, and it does not need auth), it includes an Authorization header. GitHub will treat this as a invalid PAT and return a 401 error
- This will only happen in release package because SPFx login is in `vscode-extension` package and bicepChecker is in `core` package. After packaging, webpack will merge `axios` package in `vscode-extension` and `core` into one


a1exwang:
E2E TEST: https://github.com/OfficeDev/TeamsFx-UI-Test/blob/main/src/ui-test/remotedebug/remotedebug-spfxreact.test.ts

nliu-ms (approved):

qinezh (approved):

PR #3014, opened 2021-11-18
chore: remove backup folder if upgrade failed

meifans:


packages/fx-core/src/core/middleware/projectMigrator.ts
```diff
@@ -789,6 +793,9 @@ async function cleanup(projectPath: string): Promise<void> {
   if (await fs.pathExists(path.join(fxConfig, "..", "new.env.default.json"))) {
     await fs.remove(path.join(fxConfig, "..", "new.env.default.json"));
   }
+  if (backupFolder) {
```
qinezh, line 796: We need to check if the path of `backupFolder` existed, otherwise fs.remove the file will fail and throw exception?

meifans: don't need, remove will check. refer: https://github.com/jprichardson/node-fs-extra/blob/master/docs/remove.md

meifans: 793 line is from wenyu, I just not modify it.

qinezh: Curious why we need the check for other files like `new.env.default.json`?

qinezh: ok, got it.

qinezh (approved):

PR #3016, opened 2021-11-18
fix: bicepChecker will stuck on write error

a1exwang:
The bug is: when it fails to write to `.fx/bin/bicep/bicep` (e.g. no permission or no space), the provision will be stuck forever.

Remove this defensive `writer.end()` because it will stuck on error

This is a possible node.js bug: https://github.com/nodejs/node/issues/40851


meifans (approved):

a1exwang:
https://github.com/OfficeDev/TeamsFx/blob/dev/packages/cli/tests/e2e/multienv/TestRemoteHappyPath.tests.ts

PR #3020, opened 2021-11-18
fix: let user select local env for zip app package

nliu-ms:
1) Select env when clicking `Edit manifest file` in Treeview
2) Select env when clicking `Zip app package` in Treeview
image
3) Fix manifest format


nliu-ms:
E2E TEST: ux change

HuihuiWu-Microsoft (approved):

jayzhang (approved):

PR #3022, opened 2021-11-18
fix(env-checker): fallback to use hardcoded bicep version if Github API fails

a1exwang:
- Unauthenticated GitHub API call has a limit of 60 requests per hour per IP. It is easily hit. If the API fails, fallback to use a hard-coded version of bicep

Tested by changing hosts file


a1exwang:

meifans (dismissed):
LGTM

meifans (approved):

PR #3023, opened 2021-11-18
feat(blazor): support arm

hund030:
For blazor scenario, we do scaffolding in VS extension, so we don't need to implement `generateArmTemplate` in blazor plugin, just update the arm bicep files in scaffold templates.
E2E TEST: https://github.com/OfficeDev/TeamsFx/blob/bb110600e8feccf400fb050598f7ff7b4e62d32f/packages/cli/tests/e2e/blazor/TestDeployBlazor.tests.ts#L6

### DIFF
--- a/packages/fx-core/src/plugins/resource/frontend/index.ts
+++ b/packages/fx-core/src/plugins/resource/frontend/index.ts
@@ -16,6 +16,7 @@ import { ErrorFactory, TeamsFxResult } from "./error-factory";
 import {
   ErrorType,
   FrontendPluginError,
+  NotImplemented,
   UnhandledErrorCode,
   UnhandledErrorMessage,
 } from "./resources/errors";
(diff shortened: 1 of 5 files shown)


packages/fx-core/src/plugins/resource/frontend/index.ts
```diff
@@ -49,7 +50,7 @@ export class FrontendPlugin implements Plugin, ArmResourcePlugin {
 
   public async scaffold(ctx: PluginContext): Promise<TeamsFxResult> {
     if (isVsCallingCli()) {
-      return ok(undefined);
+      throw new NotImplemented();
```
eriolchan, line 53: Since the bicep file is pack in zip file in VS extension, what if there is any breaking/schema change in Bicep file after we release VS extension in December?

hund030: Then we need to update the zip, other wise VS toolkit breaks. I think bicep should consider backward compatibility for any change after go production. @blackchoey

eriolchan (approved):

eriolchan (approved):

PR #3083, opened 2021-11-22
feat(dotnetsdk): c# 10 syntactic sugar

vnbaaij:
Use file scoped namespaces and enable ImplicitUsings
Update version to 0.3.1-rc


vnbaaij:
@adashen, @wenytang-ms, @blackchoey I accidentally pushed a new commit/PR to this branch. Reset it but you still show as reviewer. Sorry.

packages/dotnet-sdk/src/TeamsFx.Test/TeamsFxTests.cs
```diff
@@ -1,38 +1,40 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using System.Threading;
+using System.Threading.Tasks;
```
tecton, line 5: Are these 2 `using`s are not necessary any more? Since we have enabled implicit using in csproj file.
You could use "Analyze and Code Cleanup" -> "Run Code Cleanup" to auto format codes, I've configured the rules for it.

vnbaaij: Yes, you are right. These can be removed. I'll run the code cleanup

tecton (approved):

PR #3086, opened 2021-11-23
perf(error handling): add get help button when validate manifest / zip pkg

dilin-MS:
[add get help button when validate manifest / zip pkg](https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/12663844)

### DIFF
--- a/packages/fx-core/src/plugins/resource/appstudio/results.ts
+++ b/packages/fx-core/src/plugins/resource/appstudio/results.ts
@@ -8,9 +8,9 @@ export class AppStudioResultFactory {
   public static UserError(
     name: string,
     message: string,
-    innerError?: any,
+    helpLink?: string,
     stack?: string,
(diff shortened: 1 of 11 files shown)


nliu-ms (approved):

packages/fx-core/src/plugins/resource/appstudio/results.ts
```diff
@@ -8,9 +8,9 @@ export class AppStudioResultFactory {
   public static UserError(
     name: string,
     message: string,
-    innerError?: any,
+    helpLink?: string,
```
blackchoey, line 11: why we changes the parameter sequence? it might cause problem if we didn't modify all function calls

blackchoey (approved):

xzf0587 (approved):

PR #3093, opened 2021-11-23
fix(env-tree): no need refresh env tree if provision cancelled

qinezh:
Skip to refresh env tree when provision cancelled.
work item: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/12664126


qinezh:
E2E Test: UI changes.

a1exwang (dismissed):

PR #3104, opened 2021-11-24
chore(sdk): add required dependencies

tecton:
The declaration file needs the required libraries. Otherwise it will show error when building by `tsc`.


wenytang-ms (dismissed):

packages/sdk/package.json
```diff
@@ -51,16 +51,18 @@
   "dependencies": {
     "@azure/identity": "^2.0.1",
     "@azure/msal-node": "~1.1.0",
+    "@microsoft/microsoft-graph-client": "^3.0.1",
     "axios": "^0.24.0",
-    "botbuilder": ">=4.15.0-rc0 <5.0.0",
-    "botbuilder-dialogs": ">=4.15.0-rc0 <5.0.0",
+    "botbuilder": ">=4.15.0 <5.0.0",
```
blackchoey, line 56: I remember @dilin-MS 's previous investigation suggests putting botbuilder to peerDependencies to avoid the bot sdk version conflict issue (user references a version that is different than the version referenced by our SDK). Can we move the bot sdks to peerDependencies?

packages/sdk/package.json
```diff
@@ -87,6 +88,7 @@
     "@typescript-eslint/eslint-plugin": "^5.0.0",
     "@typescript-eslint/parser": "^4.13.0",
     "adm-zip": "^0.5.9",
+    "botbuilder-core": ">=4.15.0 <5.0.0",
```
blackchoey, line 91: why we need to add it to devDependencies?

packages/sdk/package.json
tecton: I moved 'tedious' to peerDependencies in previous PR and met a problem when building Azure Function. (Please check [this bug](https://msazure.visualstudio.com/9660fff2-2363-48b0-9e15-64df2283e932/_workitems/edit/12664346)).
The reason is that type definition file referenced interfaces/classes in tedious. Thus I move it back to devDependencies because npm6 doesn't install peerDependencies by default.
Bot framework library has the same issue and I remember @dilin-MS chose to use this solution.

packages/sdk/package.json
tecton: It's only used in test files so I moved it to devDependendies.

blackchoey (approved):

PR #3115, opened 2021-11-24
fix: local debug configuration

nliu-ms:
double question mark only works for undefined and null values. 

local settings configuration entry return empty string instead. 

image


nliu-ms:
E2E TEST: local debug

jayzhang (approved):

PR #3118, opened 2021-11-24
chore(template): revert 'add graph client as dependency'

tecton:
1. Reverts #2993
2. Remove unnecessary dependencies in template.


templates/function-base/js/default/package.json
```diff
@@ -5,7 +5,6 @@
         "start": "npx func start"
     },
     "dependencies": {
-        "@microsoft/microsoft-graph-client": "^3.0.0",
```
hund030, line 8: What about ts function?

tecton: ts function is matched with SDK because SDK upgraded graph library to ^3.0.0.

hund030 (dismissed):

PR #3124, opened 2021-11-24
chore: remove dialogs of upgrade for localdebug 

meifans:
Fix [bug](https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/12664689)

### DIFF
--- a/packages/fx-core/src/core/middleware/projectMigrator.ts
+++ b/packages/fx-core/src/core/middleware/projectMigrator.ts
@@ -7,6 +7,7 @@ import {
   ConfigFolderName,
   EnvConfig,
   err,
+  ok,
   InputConfigsFolderName,
   Inputs,
   Platform,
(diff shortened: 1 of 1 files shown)


packages/fx-core/src/core/middleware/projectMigrator.ts
```diff
@@ -147,6 +148,11 @@ export class ArmParameters {
 
 export const ProjectMigratorMW: Middleware = async (ctx: CoreHookContext, next: NextFunction) => {
   if ((await needMigrateToArmAndMultiEnv(ctx)) && checkMethod(ctx)) {
+    if (!checkUserTasks(ctx)) {
```
swatDong, line 151: What if moving this check before line 150? Just to save extra check time.

kuojianlu (approved):

packages/fx-core/src/core/middleware/projectMigrator.ts
meifans: @swatDong  You can see line 205, to avoid executing next middleware that will cause unrelated error.

swatDong (approved):

qinezh (approved):

PR #3125, opened 2021-11-24
fix: add function web pattern and update azure SQL server name

wenytang-ms:
E2E TEST: https://github.com/OfficeDev/TeamsFx/actions/runs/1498494500

https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/12674299

### DIFF
--- a/packages/fx-core/src/core/middleware/projectMigrator.ts
+++ b/packages/fx-core/src/core/middleware/projectMigrator.ts
@@ -132,7 +132,7 @@ export class ArmParameters {
   static readonly FEStorageName = "frontendHostingStorageName";
   static readonly IdentityName = "userAssignedIdentityName";
   static readonly SQLServer = "sqlServerName";
-  static readonly SQLDatabase = "azureSqlDatabaseName";
+  static readonly SQLDatabase = "sqlDatabaseName";
   static readonly SimpleAuthSku = "simpleAuthSku";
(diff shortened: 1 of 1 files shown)


blackchoey (commented):
the bug also mentions an error about SQL provision, do we also need to fix it?

wenytang-ms:
> the bug also mentions an error about SQL provision, do we also need to fix it?

update the azure SQL server name to fix the second provision error

packages/fx-core/src/core/middleware/projectMigrator.ts
```diff
@@ -131,7 +131,7 @@ class EnvConfigName {
 export class ArmParameters {
   static readonly FEStorageName = "frontendHostingStorageName";
   static readonly IdentityName = "userAssignedIdentityName";
-  static readonly SQLServer = "azureSqlServerName";
+  static readonly SQLServer = "sqlServerName";
   static readonly SQLDatabase = "azureSqlDatabaseName";
```
xzf0587: azureSqlDatabaseName should be sqlDatabaseName

wenytang-ms: already updated.

xzf0587 (approved):

blackchoey (approved):

PR #3128, opened 2021-11-24
fix: popup and error log when backup fails

a1exwang:
When the upgrade process fails in the backup stage, there will be no popup message or error log at all. Fix https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/12674853/

The new behavior:
image


a1exwang:

qinezh (approved):

meifans (approved):
LGTM

PR #3129, opened 2021-11-24
chore: support version <=2.3.1 (upgrade)

meifans:


packages/fx-core/src/core/middleware/projectMigrator.ts
```diff
@@ -524,6 +534,32 @@ async function getManifest(sourceManifestFile: string): Promise<TeamsAppManifest
...
+    // version <= 2.4.1
+    await fs.copy(path.join(fx, AppPackageFolderName), target);
+  } else {
+    // version <= 2.3.1
+    await fs.copy(path.join(fx, REMOTE_MANIFEST), path.join(target, REMOTE_MANIFEST));
+    const manifest: TeamsAppManifest = await getManifest(path.join(target, REMOTE_MANIFEST));
+    const color = (await fs.pathExists(path.join(fx, "color.png")))
+      ? "color.png"
+      : manifest.icons.color;
+    const outline = (await fs.pathExists(path.join(fx, "outline.png")))
```
a1exwang, line 550: Should we check `manifest.icons.color` first?
```
    const color = (await fs.pathExists(manifest.icons.color))
      ? path.relative(fx, manifest.icons.color)
      : color.png)
```

packages/fx-core/src/core/middleware/projectMigrator.ts
```diff
@@ -799,6 +828,10 @@ async function removeOldProjectFiles(projectPath: string): Promise<void> {
   await fs.remove(path.join(fx, "new.env.default.json"));
   // version <= 2.4.1, remove .fx/appPackage.
   await fs.remove(path.join(fx, AppPackageFolderName));
+  // version <= 3.2.1
+  await fs.remove(path.join(fx, REMOTE_MANIFEST));
```
a1exwang, line 832: Should we first determine the toolkit version that created the manifest and then delete files only of specific versions?
For example, if we know the project are from 2.10.0, do not delete `.fx/manifest.source.json` or `.fx/color.png`, etc.

This is to prevent accidental deletion of `.fx/color.png`.

For example, if `.fx/color.png`, `appPackage/manifest.source.json` and `appPackage/color.png` exist,  `appPackage/color.png` is correctly backed up, but  `.fx/color.png` will be deleted without backup.

packages/fx-core/src/core/middleware/projectMigrator.ts
```diff
@@ -333,12 +333,22 @@ async function migrateToArmAndMultiEnv(ctx: CoreHookContext): Promise<void> {
   await postMigration(projectPath, ctx, inputs, backupFolder);
 }
 
+async function getManifestPath(fx: string, projectPath: string): Promise<string> {
```
XiaofuHuang, line 336: Maybe we can first check `path.join(projectPath, AppPackageFolderName, REMOTE_MANIFEST)` and last check `path.join(fx, REMOTE_MANIFEST)` . If the three file path does not exist, may be we can throw an error.

packages/fx-core/src/core/middleware/projectMigrator.ts
meifans: offline sync

packages/fx-core/src/core/middleware/projectMigrator.ts
```diff
@@ -524,6 +534,32 @@ async function getManifest(sourceManifestFile: string): Promise<TeamsAppManifest
...
+  if (await fs.pathExists(path.join(projectPath, AppPackageFolderName))) {
+    await fs.copy(path.join(projectPath, AppPackageFolderName), target);
+  } else if (await fs.pathExists(path.join(fx, AppPackageFolderName))) {
+    // version <= 2.4.1
+    await fs.copy(path.join(fx, AppPackageFolderName), target);
+  } else {
+    // version <= 2.3.1
+    await fs.copy(path.join(fx, REMOTE_MANIFEST), path.join(target, REMOTE_MANIFEST));
+    const manifest: TeamsAppManifest = await getManifest(path.join(target, REMOTE_MANIFEST));
+    const color = (await fs.pathExists(path.join(fx, "color.png")))
```
XiaofuHuang, line 547: The icon may can be start with "https://"

packages/fx-core/src/core/middleware/projectMigrator.ts
a1exwang: Offline discussed. It is hard to determine the version. And it is  very rare to happen.

packages/fx-core/src/core/middleware/projectMigrator.ts
meifans: The reason for checking color.png is if user write manifest.icons.color in a wrong value and color.png exist, upgrade can still works.

packages/fx-core/src/core/middleware/projectMigrator.ts
```diff
@@ -333,12 +333,22 @@ async function migrateToArmAndMultiEnv(ctx: CoreHookContext): Promise<void> {
...
 async function preCheckKeyFiles(projectPath: string, ctx: CoreHookContext): Promise<void> {
   const core = ctx.self as FxCore;
   const fx = path.join(projectPath, `.${ConfigFolderName}`);
-  const manifestPath = (await fs.pathExists(path.join(fx, AppPackageFolderName, REMOTE_MANIFEST)))
-    ? path.join(fx, AppPackageFolderName, REMOTE_MANIFEST)
-    : path.join(projectPath, AppPackageFolderName, REMOTE_MANIFEST);
+  const manifestPath = await getManifestPath(fx, projectPath);
   await preReadJsonFile(path.join(fx, "env.default.json"), core);
   await preReadJsonFile(path.join(fx, "settings.json"), core);
   await preReadJsonFile(manifestPath, core);
```
meifans, line 354: @XiaofuHuang if manifestPath is invalid, will be caught here. Not only manifest path not found, but also format error, permission error.

packages/fx-core/src/core/middleware/projectMigrator.ts
meifans: @XiaofuHuang  Why we need to first check `path.join(projectPath, AppPackageFolderName, REMOTE_MANIFEST)` ?
Because no matter what order, we always try to get a exist manifest file. 
Here is according to the version order. From old version to new version.

a1exwang (approved):

XiaofuHuang:
@meifans Yes, I think we can check from the new version to the old version. User may manually update the manifest filepath and do not delete the old file path. Our manifest path fallback logic is from new to old.
image

meifans:
> @meifans Yes, I think we can check from the new version to the old version. User may manually update the manifest filepath and do not delete the old file path. Our manifest path fallback logic is from new to old. image

meifans:
> > @meifans Yes, I think we can check from the new version to the old version. User may manually update the manifest filepath and do not delete the old file path. Our manifest path fallback logic is from new to old. image

Changed.

packages/fx-core/src/core/middleware/projectMigrator.ts
meifans: Added check.

a1exwang (approved):

qinezh (approved):

XiaofuHuang (approved):

PR #3136, opened 2021-11-25
fix: no need to check local teams app id

HuihuiWu-Microsoft:
Shouldn't check local settings teams app id since it doesn't exist for the first time and will be created in later logic

Already verified bot app local debug
image


HuihuiWu-Microsoft:
E2E TEST: hotfix

nliu-ms (dismissed):

nliu-ms (approved):

PR #3148, opened 2021-11-25
chore: update manual configuration wording

HuihuiWu-Microsoft:
image

### DIFF
--- a/packages/fx-core/src/common/samples-config.json
+++ b/packages/fx-core/src/common/samples-config.json
@@ -45,7 +45,7 @@
             "fullDescription": "Todo List provides an easy way to manage to-do items in Teams Client. This app helps enabling task collaboration and management for your team. The frontend is a React app and the backend is hosted on Azure. You will need an Azure subscription to run the app.",
             "tags": ["React", "Azure function", "Azure SQL", "JS", "CI/CD"],
             "time": "30min to run",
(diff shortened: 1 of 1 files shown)


MSFT-yiz (approved):

HuihuiWu-Microsoft:
Please hold this PR, we're under discussion if include sample screenshot updates in this one

packages/fx-core/src/common/samples-config.json
```diff
@@ -90,7 +90,7 @@
             "fullDescription": "Todo List app helps to manage personal to do items and also runs across Microsoft 365, including Teams, Outlook and Office. The frontend is a React app and the backend is an Azure Function. You can optionally deploy and host the app in Azure.",
             "tags": ["React", "Azure function", "JS", "Outlook", "Office"],
             "time": "5min to run",
-            "configuration": "F5 ready"
+            "configuration": "Ready for local debug"
```
kuojianlu, line 93: This sample is not local debug targeted. The goal of this sample is to run the todo list app in Outlook and Office, which needs provision & deploy. Currently, local debug only works in Teams, but not in Outlook and Office. So I think `Ready for local debug` may be not very suitable. @MuyangAmigo @zyxiaoyuer What do you think about?

HuihuiWu-Microsoft: Updated

kuojianlu (approved):

PR #3150, opened 2021-11-25
fix: create app catalog

nliu-ms:
If tenant app catalog doesn't exist, we will call ensure `Ensure App Catalog` API to create one.

But it requires some time to take effect.
image

https://docs.microsoft.com/en-us/sharepoint/dev/sp-add-ins/retrieve-tenant-app-catalog-url-rest

So we just disabled automatic app catalog creation for the user, give doc to let user manually create tenant app catalog.
image


nliu-ms:
E2E TEST: https://github.com/OfficeDev/TeamsFx/blob/dev/packages/cli/tests/e2e/spfx/CreateSPFxProject.tests.ts

jayzhang (approved):

PR #3154, opened 2021-11-26
fix(cli): UnknownError when upgrading

a1exwang:
fix: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/12684229/

### DIFF
--- a/packages/fx-core/src/core/middleware/projectMigrator.ts
+++ b/packages/fx-core/src/core/middleware/projectMigrator.ts
@@ -177,6 +177,8 @@ export const ProjectMigratorMW: Middleware = async (ctx: CoreHookContext, next: 
 
     try {
       await migrateToArmAndMultiEnv(ctx);
+      // return ok for the lifecycle functions to prevent breaking error handling logic.
+      ctx.result = ok({});
     } catch (error) {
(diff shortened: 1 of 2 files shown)


a1exwang:

packages/fx-core/src/core/middleware/projectMigrator.ts
```diff
@@ -177,6 +177,8 @@ export const ProjectMigratorMW: Middleware = async (ctx: CoreHookContext, next:
 
     try {
       await migrateToArmAndMultiEnv(ctx);
+      // return ok for the lifecycle functions to prevent breaking error handling logic.
+      ctx.result = ok({});
```
meifans, line 181: Maybe `undefined` is better.
```suggestion
      ctx.result = ok(undefined);
```

meifans (approved):
LGTM

packages/fx-core/src/core/middleware/projectMigrator.ts
a1exwang: The method signature is `Result<Void, FxError>`, Void is {}, so I use {}

qinezh (approved):

meifans (approved):

dooriya (approved):

PR #3155, opened 2021-11-26
refactor: collaboration api v2

1yefuwang1:
closes https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_sprints/taskboard/Teams%20Extensibility%20E2E%20Team/Microsoft%20Teams%20Extensibility/CY21-11.2?workitem=10951279


packages/fx-core/src/plugins/resource/utils4v2.ts
```diff
@@ -435,3 +441,37 @@ export function setLocalSettingsV1(pluginContext: PluginContext, localSettings:
...
+  }
+  const pluginContext: PluginContext = convert2PluginContext(plugin.name, ctx, inputs);
+  pluginContext.azureAccountProvider = tokenProvider.azureAccountProvider;
+  pluginContext.appStudioToken = tokenProvider.appStudioToken;
+  pluginContext.graphTokenProvider = tokenProvider.graphTokenProvider;
+  pluginContext.envInfo = newEnvInfo();
+  pluginContext.envInfo.state = flattenConfigMap(state);
+  pluginContext.envInfo.config = envInfo.config as EnvConfig;
+  pluginContext.config = pluginContext.envInfo.state.get(plugin.name) ?? new ConfigMap();
+  const result = await fn(pluginContext, userInfo);
```
SLdragon, line 475: Seems we can `return fn(pluginContext, userInfo)` directly without await

packages/fx-core/src/plugins/solution/fx-solution/v2/adaptor.ts
```diff
@@ -107,3 +107,36 @@ export class ProvisionContextAdapter extends BaseSolutionContextAdaptor {
...
 }
+
+export class CollaboratorContextAdapter extends BaseSolutionContextAdaptor {
+  constructor(params: Parameters<NonNullable<v2.SolutionPlugin["grantPermission"]>>) {
+    super();
+    const v2context: v2.Context = params[0];
+    const inputs: Inputs = params[1];
+    const envInfo = params[2];
+    const tokenProvider = params[3];
+    this.root = inputs.projectPath ?? "";
```
SLdragon, line 118: What will happen when root is empty?

SLdragon:
Some questions related to v1 and v2:
What will we do for v1 logic in the future, do we have time line to remove v1?
And for future development, do we need to implement both v1 and v2 logic?

packages/fx-core/src/plugins/solution/fx-solution/v2/adaptor.ts
1yefuwang1: root should never be empty here.  What I want here is just to cast away the null-ness of projectPath for simplicity. I'll change to throw error to enforce the fact the root will not be undefined.

1yefuwang1:
> Some questions related to v1 and v2:
> 
> * What will we do for v1 logic in the future, do we have time line to remove v1?
> * And for future development, do we need to implement both v1 and v2 logic?
I think I have refactored collaboration related code in a V1-V2 agnostic manner as much as possible. Most of the code only depends on what it needs, not just v1.PluginContext, except for code that calls plugin APIs which is different between V1 and V2.
Now,  both v1 and v2 implementations serve just as shims, which, in the end, call into (mostly) V1-V2 agnostic code.
So, I think maintaining both v1 and v2 will not be a problem.

SLdragon (approved):

jayzhang (dismissed):

jayzhang (approved):

LongOddCode (approved):

SLdragon (approved):

PR #3158, opened 2021-11-26
fix: publish failed with wrong icon

nliu-ms:
When zip app package, get right relative path from manifest.

Test case 1:
1) Disable insider preview
2) Scaffold & provision & publish
3) Enable insider preview
4) Publish

Test case 2:
1) Enable insider preview
2) Scaffold & provision & publish


nliu-ms:
E2E TEST: https://github.com/OfficeDev/TeamsFx/blob/20804f1bc75157907076495b8cff137f93e90bc5/packages/cli/tests/e2e/smoke/AzureAppHappyPath.tests.ts#L168

jayzhang (approved):

XiaofuHuang (approved):

packages/fx-core/src/plugins/resource/appstudio/plugin.ts
```diff
@@ -780,24 +780,39 @@ export class AppStudioPluginImpl {
...
-    const colorFile = `${appDirectory}/${manifest.icons.color}`;
 
-    let fileExists = await this.checkFileExist(colorFile);
-    if (!fileExists) {
-      throw AppStudioResultFactory.UserError(
-        AppStudioError.FileNotFoundError.name,
-        AppStudioError.FileNotFoundError.message(colorFile)
-      );
+    // color icon
+    if (manifest.icons.color && !manifest.icons.color.startsWith("https://")) {
```
qinezh, line 789: do we need to consider the case that the url of icon starts from `http://`?

nliu-ms: Need more investigation to decide if we need to support this case. Refer to VSO 12681229.

qinezh (approved):

PR #3163, opened 2021-11-29
fix: no permission to open on mac

nliu-ms:


nliu-ms:

packages/fx-core/src/plugins/resource/appstudio/plugin.ts
```diff
@@ -825,7 +825,7 @@ export class AppStudioPluginImpl {
       if (await fs.pathExists(manifestFileName)) {
         await fs.chmod(manifestFileName, 0o777);
       }
-      await fs.writeFile(manifestFileName, manifestString, { mode: 0o000 });
+      await fs.writeFile(manifestFileName, manifestString, { mode: 0o555 });
```
qinezh, line 828: Is there any consideration to use the mode `0o555` instead of default mode `0o666` for `fs.writeFile` here?

nliu-ms: Cause PM has a requirement that this file should be readonly.

HuihuiWu-Microsoft (approved):

packages/fx-core/src/plugins/resource/appstudio/plugin.ts
a1exwang: If the requirement is readonly then this code has the following issues:
- file mode only works on Unix-like OSes (macOS and Linux)
- 0o555 is read and executable, 0o444 is readonly  https://linuxcommand.org/lc3_lts0090.php

qinezh: Looks the file permission mode only works on *nix, how do we ensure the file to be readonly for Windows platform?

BTW, if only *nix are considered, is `0o444` sufficient as manifest doesn't need to be executed?

HuihuiWu-Microsoft: Yes, for Windows, "only the write permission can be changed, and the distinction among the permissions of group, owner or others is not implemented"(https://nodejs.org/api/fs.html#fschmodpath-mode-callback). That's also why we only change write permission and use 0o555. Any better ideas?

a1exwang: Then why should this file be executable on macOS?

PR #3164, opened 2021-11-29
perf(ARM): implement updateArmTemplate in solution

wenytang-ms:


packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -464,14 +488,41 @@ async function doGenerateArmTemplate(
   // Get bicep content from each resource plugin
   for (const plugin of plugins) {
     const pluginWithArm = plugin as Plugin & ArmResourcePlugin; // Temporary solution before adding it to teamsfx-api
-    if (pluginWithArm.generateArmTemplates) {
+    // plugin not selected need to be update.
+    if (
+      pluginWithArm.updateArmTemplates &&
+      !selectedPlugins.find((pluginItem) => pluginItem === pluginWithArm)
```
blackchoey, line 494: in the future, when we implemented `updateArmTemplates` interface in all resource plugins, we should do nothing if the plugin is not selected and `updateArmTemplates` is not implemented, instead of trying to invoke `generateArmTemplate` again. We can a todo item in our task board for this.

blackchoey (approved):

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
wenytang-ms: sure

jayzhang (approved):

PR #3166, opened 2021-11-29
refactor(api): fxcore v3 API contract

jayzhang:
fxcore v3 API contract:
1. define v3 API for new solution settings
2. define resource env state model
3. define ScaffoldPlugin and ResourcePlugin interfaces
4. Upgrade solution and core API to V3
E2E TEST: AzureAppHappyPath


LongOddCode (approved):

nliu-ms (approved):

packages/api/src/v3/resourceStates.ts
```diff
@@ -0,0 +1,49 @@
...
+}
+
+export interface ResourceStates {
+  solution: Json;
+  [key: string]: CloudResource | CloudResource[];
+}
+
+////////////Azure Solution/////////////////////
+export interface AzureResource extends CloudResource {
+  resourceGroupName?: string;
```
1yefuwang1, line 26: Is it possible that we reduce optional fields as much as possible?

jayzhang: Discussed with Chaoyi, actually, for Azure resourceId already encode the resource group and subscription information, so is deleted

1yefuwang1 (approved):

PR #3171, opened 2021-11-30
Adding support for MGT in the scaffolding in TS

sebastienlevert:
As discussed with @zhenyasav, we're opening this PR to start a conversation on how we can collaborate between MGT (https://aka.ms/mgt) and TeamsFx. 

This sample uses a new `TeamsFxProvider` built for MGT that abstracts all the authentication and automatically plugs it in to our web components that are using the token provided by the `TeamsUserCredential`.

Adding @nmetulev to this discussion also!


waldekmastykarz:
I think this is a great idea to offer developers familiar with MGT a consistent way to use it across all toolchains 👏

adashen:
@sebastienlevert @waldekmastykarz It's great that we could start this discussion. I have added @zhenyasav (PM) and @SLdragon (Engineer) to this PR to start the discussion. From our side, the is an update in the latest VSCode Teams Toolkit version 3.0.0. We have added a "teams app sample with graph toolkit" in to the sample gallery. So user could download the graph toolkit sample from our toolkit. And also the source code is in the sample repo: https://github.com/OfficeDev/TeamsFx-Samples/tree/v2/graph-toolkit-contact-exporter.

SLdragon:
Great, thank you @sebastienlevert 
Did latest `TeamsFxProvider` implementation come from here?: https://github.com/microsoftgraph/microsoft-graph-toolkit/tree/next/teamsfx ?

templates/tab/ts/default/package.json
```diff
@@ -5,6 +5,9 @@
     "dependencies": {
         "@fluentui/react-northstar": "^0.54.0",
         "@microsoft/microsoft-graph-client": "^3.0.0",
```
SLdragon, line 7: I tried the template, we need to remove `"@microsoft/microsoft-graph-client": "^3.0.0"` reference and `tabs\src\components\sample\lib\useGraph.ts` file, otherwise, there will be a conflict when compile

templates/tab/ts/default/src/components/sample/lib/useGraphToolkit.ts
```diff
@@ -0,0 +1,32 @@
...
+export function useGraphToolkit() {
+  const [providerState, setProviderState] = React.useState<ProviderState>(ProviderState.Loading);
+      
+  React.useEffect(() => {
+    if (process.env.NODE_ENV === "development") {
+      setLogLevel(LogLevel.Verbose);
+      setLogFunction((level: LogLevel, message: string) => { console.log(message); });
+    }
+
+    Providers.globalProvider = new TeamsFxProvider({
```
SLdragon, line 15: Seems current TeamsFxProvider trigger the login pop up window automatically, this will be blocked by the browser by default, we will have error below:
image

It is better to have a login function, that developer can call this function when button clicked

.vscode/settings.json
```diff
@@ -2,5 +2,11 @@
   "cSpell.words": [
     "kusto",
     "logmining"
-  ]
+  ],
+  "markdownlint.config": {
```
SLdragon, line 6: This is not necessarily for the template

templates/tab/ts/default/package.json
sebastienlevert: Right, MGT is still on 2.x and TeamsFx uses 3.x. We're working on making it happen from the MGT side.

templates/tab/ts/default/src/components/sample/lib/useGraphToolkit.ts
sebastienlevert: Isn't this the promise of Teams SSO though? I'm curious to have @waldekmastykarz thoughts on that.

.vscode/settings.json
sebastienlevert: Agree.

sebastienlevert:
> Great, thank you @sebastienlevert Sébastien Levert FTE
> Did latest `TeamsFxProvider` implementation come from here?
> [microsoftgraph/microsoft-graph-toolkit@`next`/teamsfx](https://github.com/microsoftgraph/microsoft-graph-toolkit/tree/next/teamsfx?rgh-link-date=2021-12-03T04%3A09%3A59Z)

Yes, this is where it's coming from. When we'll be confident about its state, this will get merge as 2.x.

templates/tab/ts/default/package.json
sebastienlevert: Do we need to remove `useGraph.ts` in this case? This could probably run with a previous version of the Graph SDK anyway.

templates/tab/ts/default/src/components/sample/lib/useGraphToolkit.ts
waldekmastykarz: Isn't the consent separate from SSO and are we seeing here the consent fail not the SSO itself?

SLdragon: The login function in teamsfx sdk is used to exchange Graph token, it will popup a login page to exchange graph access token using Teams SSO token.

From your TeamsFxProvider implementation, the login function will be called in the constructor, this will trigger the pop up window:
https://github.com/microsoftgraph/microsoft-graph-toolkit/blob/a19718867b0ff15a72b0491257720d3928bb878f/samples/teamsfx-app/tabs/src/components/TeamsFxProvider.js#L87

It is better not to call the login function in the constructor to avoid automatically popup the login page, maybe we can have a public login function in TeamsFxProvider, which can be called by developer manually in a button click callback function

templates/tab/ts/default/package.json
SLdragon: Yes, I think we can remove it because it is not used anymore

SLdragon:
Another problem is that after I login success, the Person component is not show up, I need to refresh the page to see it. Not sure what's wrong with it, maybe related to react state not refresh properly?

templates/tab/ts/default/src/components/sample/lib/useGraphToolkit.ts
sebastienlevert: So the pattern would be to click a "Login" button when it requires consent. Would this also break in the client or this would get forced on the rich client? Could we identify when we are running in web and offer the button there, and offer the direct pop-up when running in the client?

SLdragon: Yes, that's true, login button is required for consent when in browser.

I think it is developer's responsibility to determine when he want to popup login window. In our SDK, we do not need to identify whether it inside browser or inside client, SDK just need to provide a login function for developers.

SLdragon:
Hi, @sebastienlevert 

@zhenyasav and we have a discussion recently, here are some discussion notes:

1. For our tab template, we plan to update our Tab template to contains both "Render UI using bare graph API" and "Render UI using graph toolkit", all the code will be placed inside the tab template.


1. For teamsFxAuthProvider, we plan to write graph toolkit auth provider inside TeamsFx SDK
    Advantage to provide auth provider inside teamsfx SDK:
    - TeamsFx SDK will change recently, auth provider inside our SDK is better to avoid breaking
    - We can decouple TeamsFx SDK with Graph Toolkit SDK, user can use peer dependency when he want to use Graph Toolkit

Here is the draft usage of the SDK (Just draft and need more discussion about the design)
  ```js
  import { Providers, ProviderState } from '@microsoft/mgt-element';
  import { PeoplePicker } from '@microsoft/mgt-react';
  
  const credential = new TeamsUserCredential();
  const authProvider = credential.getTeamsFxAuthProvider();
  Providers.globalProvider = authProvider;
  
  async function loginBtnClick() {
    await credential.login(scope);
    Providers.globalProvider.setState(ProviderState.SignedIn);
  }


<PeoplePicker placeholder="Typing name to select people to view contact info"></PeoplePicker>
```

When our implementation is ready, we will create a PR and let you help review it, thank you!

waldekmastykarz:
@SLdragon what if there's a major release of MGT with backwards incompatible change to auth providers? Wouldn't that require a new major version of teamsfx to align them? Wouldn't it be better to decouple both libraries and allow MGT to instantiate an auth provider based on the context information provided by teamsfx?

SLdragon:
Hi, @sebastienlevert ,

I am OK with your design about the teamsfx auth provider to have a separate npm package, and we may need more discussion about where to host the auth provider:

We release teamsfx toolkit each sprint, and the project status is public preview, so there maybe breaking changes in our SDK and template, and we have a plan recently to change the auth flow inside teamsfx SDK and will introduce breaking changes.

From our discussion internally, we suggest to put the auth provider in our repo (inside teamsfx SDK, or have a separate npm package), because teamsfx auth provider is integration with vscode extension, tab template and our tool chain deeply, each release we may have some changes, in our repo we can take care of the release cycle and make sure the auth provider works with latest graph toolkit, and avoid asking your team to align with us to merge changes and release packages. 

I will send a meeting request for you next week, and we can have more discussion about this, you can still leave comments in this discussion, thank you!

PR #3175, opened 2021-11-30
feat(env-checker): support node16 for non-function/non-spfx projects

a1exwang:
- For non-function/non-spfx projects, support Node.js 16 LTS
- Update the documents accordingly

See the rendered document at [here](https://github.com/OfficeDev/TeamsFx/blob/aochengwang/feat/node16/docs/vscode-extension/envchecker-help.md#nodenotsupportedazure-functions)

env-checker test result: https://github.com/OfficeDev/TeamsFx/actions/runs/1525007470
M1 for https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/12196596/

### DIFF
--- a/docs/vscode-extension/envchecker-help.md
+++ b/docs/vscode-extension/envchecker-help.md
@@ -17,7 +17,9 @@ Please Note:
 
 Please refer to [nodejs.org](https://nodejs.org/) to download and install the supported versions:
 
-- For `Azure` hosting, please refer to the supported versions [here](#nodenotsupportedazure-hosting).
+- For `Azure` hosting projects:
+    - If your project does not contain Azure Functions, please refer to the supported versions [here](#nodenotsupportedazure-hosting).
+    - If your project contains Azure Functions, please refer to the supported versions [here](#nodenotsupportedazure-functions).
 - For `SPFx` hosting, please refer to the supported versions [here](#nodenotsupportedspfx-hosting).
 
 **NOTE**: Please restart all your Visual Studio Code instances after the installation is finished.
@@ -49,7 +51,12 @@ Please refer to [the official website](https://docs.microsoft.com/en-us/azure/az
 As the Teams Toolkit project is implemented by `Node.js`, it's required to install the npm pacakges and run the project in local. 
 
 ### Mitigation
-Please refer to [nodejs.org](https://nodejs.org/) to install the right version: currently only LTS versions (v10, v12 and v14) are supported by Teams Toolkit, and `Node v14` would be recommended to be installed.
+Please refer to [nodejs.org](https://nodejs.org/) to install the right version: currently only LTS versions (v10, v12, v14 and v16) are supported by Teams Toolkit that does not containing Azure Functions or SPFx, and `Node v14` would be recommended to be installed.
+
+**NOTE**: Please restart all your Visual Studio Code instances after the installation is finished.
+
+### Mitigation
+Please refer to [nodejs.org](https://nodejs.org/) to install the right version: currently only LTS versions (v10, v12 and v14) are supported by Teams Toolkit for projects containing Azure Functions and `Node v14` would be recommended to be installed.
 
 **NOTE**: Please restart all your Visual Studio Code instances after the installation is finished.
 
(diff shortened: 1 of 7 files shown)


swatDong:
For documentation, how about using a table to describe supported node versions?
(Markdown supports simple table, or just use HTML to create complex table)

For example,

| Project Component | Supported Node Version | Comment |
| - | - | - |
| Azure Hosting without Function | 12, **14 (recommended)**, 16 | All LTS Node versions |
| Azure Hosting with Function | 12, **14 (recommended)** | (refer to azure function doc) |
| SPFx Hosting | ... | (refer to spfx doc) |

a1exwang:
> For documentation, how about using a table to describe supported node versions? (Markdown supports simple table, or just use HTML to create complex table)
> 
> For example,
> 
> Project Component	Supported Node Version	Comment
> Azure Hosting without Function	12, **14 (recommended)**, 16	All LTS Node versions
> Azure Hosting with Function	12, **14 (recommended)**	(refer to azure function doc)
> SPFx Hosting	...	(refer to spfx doc)

Good idea. I will add it.

docs/vscode-extension/envchecker-help.md
```diff
@@ -49,15 +51,29 @@ Please refer to [the official website](https://docs.microsoft.com/en-us/azure/az
...
+**NOTE**: Please restart all your Visual Studio Code instances after the installation is finished.
+
+### Mitigation
+Please refer to [nodejs.org](https://nodejs.org/) to install the right version: currently only LTS versions (v10, v12 and v14) are supported by Teams Toolkit for projects containing Azure Functions and `Node v14` would be recommended to be installed.
 
 **NOTE**: Please restart all your Visual Studio Code instances after the installation is finished.
 
 ## NodeNotSupported(Azure hosting)
 ### Notification Message
 > Node.js (*node_version*) is not in the supported version list (v10, v12, v14).
```
qinezh, line 65: The notification message needs to be updated as well? => (v10, v12, v14, v16)

packages/vscode-extension/src/debug/depsChecker/spfxNodeChecker.ts
```diff
@@ -2,8 +2,14 @@ import { DepsCheckerEvent, nodeNotFoundHelpLink, nodeNotSupportedForSPFxHelpLink
 import { NodeChecker } from "./nodeChecker";
 
 export class SPFxNodeChecker extends NodeChecker {
-  protected readonly _supportedVersions: string[] = ["10", "12", "14"];
   protected readonly _nodeNotFoundHelpLink = nodeNotFoundHelpLink;
-  protected readonly _nodeNotSupportedHelpLink = nodeNotSupportedForSPFxHelpLink;
   protected readonly _nodeNotSupportedEvent = DepsCheckerEvent.nodeNotSupportedForSPFx;
+
+  protected async getNodeNotSupportedHelpLink(): Promise<string> {
```
dooriya, line 8: Does it need to be `async`?

packages/vscode-extension/src/debug/depsChecker/spfxNodeChecker.ts
```diff
@@ -2,8 +2,14 @@ import { DepsCheckerEvent, nodeNotFoundHelpLink, nodeNotSupportedForSPFxHelpLink
...
+    return nodeNotSupportedForSPFxHelpLink;
+  }
+
+  protected async getSupportedVersions(): Promise<string[]> {
```
dooriya, line 12: Same as above.

packages/vscode-extension/src/debug/depsChecker/spfxNodeChecker.ts
a1exwang: 1. In AzureNodeChecker's implementation, we need to read config files to determine whether it contains function. So it has to return a Promise. So the base class needs to return Promise
2. In SpfxNodeChecker, it does not have to async. We can also `return Promise.resolve()` but I think these two are not much of a difference

docs/vscode-extension/envchecker-help.md
```diff
@@ -49,15 +51,29 @@ Please refer to [the official website](https://docs.microsoft.com/en-us/azure/az
 As the Teams Toolkit project is implemented by `Node.js`, it's required to install the npm pacakges and run the project in local. 
 
 ### Mitigation
-Please refer to [nodejs.org](https://nodejs.org/) to install the right version: currently only LTS versions (v10, v12 and v14) are supported by Teams Toolkit, and `Node v14` would be recommended to be installed.
+Please refer to [nodejs.org](https://nodejs.org/) to install the right version: currently only LTS versions (v10, v12, v14 and v16) are supported by Teams Toolkit that does not containing Azure Functions or SPFx, and `Node v14` would be recommended to be installed.
```
kuojianlu, line 54: 1. v10 is not LTS currently.
2. Consider using https://nodejs.org/en/about/releases/ instead of https://nodejs.org/, see [this bug](https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/12684771).

docs/vscode-extension/envchecker-help.md
a1exwang: fixed

qinezh (approved):

dooriya (approved):

PR #3184, opened 2021-12-01
fix: move get status to immediate call logic

xiaolang124:
1. Fix open share point login page unexpected when open a project.

Fix bug: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/12674790/


xiaolang124:
E2E Test: only affect UI.

swatDong (approved):

PR #3191, opened 2021-12-02
chore(arm): update poll status

xzf0587:
1. update arm poll status for across resource group and subscription

2. update e2e test name

### DIFF
--- a/packages/cli/tests/e2e/sql/ProvisionReactAppWithAzureFunctionAndSQL.tests.ts
+++ b/packages/cli/tests/e2e/sql/ProvisionReactAppWithAzureFunctionAndSQL.tests.ts
@@ -28,7 +28,7 @@ describe("Provision to Azure with SQL", function () {
   const subscription = getSubscriptionId();
   const projectPath = path.resolve(testFolder, appName);
 
-  it(`Provision react app with Azure Function and SQL - Test Plan ID 9454227`, async function () {
+  it(`Provision react app with Azure Function and SQL`, async function () {
     // new a project ( tab + function + sql )
     await execAsync(
       `teamsfx new --interactive false --app-name ${appName} --capabilities tab --azure-resources function sql`,
--- a/packages/fx-core/src/plugins/solution/fx-solution/arm.ts
+++ b/packages/fx-core/src/plugins/solution/fx-solution/arm.ts
@@ -179,19 +179,23 @@ export async function pollDeploymentStatus(deployCtx: DeployContext) {
             currentStatus[operation.resourceName] = operation.status;
             if (!polledOperations.includes(operation.resourceName)) {
               polledOperations.push(operation.resourceName);
-
-              if (operation.subscriptionId === deployCtx.client.subscriptionId) {
-                const subOperations = await deployCtx.client.deploymentOperations.list(
-                  operation.resourceGroupName,
-                  operation.resourceName
-                );
-                subOperations.forEach((sub) => {
-                  const subOperation = getRequiredOperation(sub, deployCtx);
-                  if (subOperation) {
-                    currentStatus[subOperation.resourceName] = subOperation.status;
-                  }
-                });
+              let client = deployCtx.client;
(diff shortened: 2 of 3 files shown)


LongOddCode (dismissed):

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -159,16 +179,19 @@ export async function pollDeploymentStatus(deployCtx: DeployContext) {
...
-                operation.resourceName
-              );
-              subOperations.forEach((sub) => {
-                const subOperation = getRequiredOperation(sub, deployCtx);
-                if (subOperation) {
-                  currentStatus[subOperation.resourceName] = subOperation.status;
-                }
-              });
+
+              if (operation.subscriptionId === deployCtx.client.subscriptionId) {
```
blackchoey, line 183: we may need to handle cases that subscriptionId is different?

packages/fx-core/tests/plugins/solution/solution.arm.test.ts
```diff
@@ -788,6 +788,7 @@ describe("Arm Template Failed Test", () => {
         properties: {
           targetResource: {
             resourceName: "test resource",
+            id: "/subscriptions/1756abc0-3554-4341-8d6a-46674962ffff/resourceGroups/test52-dev-rg/providers/Microsoft.Resources/deployments/addTeamsFxConfigurations",
```
blackchoey, line 791: maybe change the subscription id and resource group name to some value that people thinks they fake data at first glance? e.g. all zero guid and fake rg group name

packages/cli/tests/e2e/sql/ProvisionReactAppWithAzureFunctionAndSQL.tests.ts
```diff
@@ -28,7 +28,7 @@ describe("Provision to Azure with SQL", function () {
   const subscription = getSubscriptionId();
   const projectPath = path.resolve(testFolder, appName);
 
-  it(`Provision react app with Azure Function and SQL - Test Plan ID 9454227`, async function () {
+  it(`Provision react app with Azure Function and SQL`, async function () {
```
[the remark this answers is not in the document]
xzf0587: It is suggested by @LongOddCode. He will use the name to match the test plan name. The test plan id is useless.

packages/fx-core/tests/plugins/solution/solution.arm.test.ts
xzf0587: It is fixed.

packages/fx-core/src/plugins/solution/fx-solution/arm.ts

LongOddCode (dismissed):

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -179,19 +179,23 @@ export async function pollDeploymentStatus(deployCtx: DeployContext) {
...
-                subOperations.forEach((sub) => {
-                  const subOperation = getRequiredOperation(sub, deployCtx);
-                  if (subOperation) {
-                    currentStatus[subOperation.resourceName] = subOperation.status;
-                  }
-                });
+              let client = deployCtx.client;
+              if (operation.subscriptionId !== deployCtx.client.subscriptionId) {
+                const azureToken =
+                  await deployCtx.ctx.azureAccountProvider?.getAccountCredentialAsync();
```
blackchoey, line 185: will there be any problem if we getAccountCredentialAsync before every pull status?

blackchoey (approved):

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
xzf0587: Actually the getAccountCredentialAsync() is called to get the client of arm deployment before pull status. 
It will fail when preparing that client.

blackchoey (approved):

jayzhang (approved):

LongOddCode (approved):

PR #3194, opened 2021-12-02
perf: function implement updateArmTemplates interface

wenytang-ms:


packages/fx-core/src/plugins/resource/function/enums.ts
```diff
@@ -82,6 +82,7 @@ export enum FunctionEvent {
   skipDeploy = "skip-deploy",
   DeploymentInfoNotFound = "deployment-info-not-found",
   generateArmTemplates = "generate-arm-templates",
+  updateArmTemplates = "update-arm-templates",
```
hund030, line 85: You add a new telemetry event here, do you help classify it in GDPR inventory?

hund030 (approved):

packages/fx-core/src/plugins/resource/function/enums.ts
wenytang-ms: will add this telemetry event after all resource added

PR #3195, opened 2021-12-02
perf(identity): implement updateArmTemplate interface

wenytang-ms:


packages/fx-core/src/plugins/resource/function/index.ts
```diff
@@ -143,6 +143,14 @@ export class FunctionPlugin implements Plugin {
     return res;
   }
 
+  public async updateArmTemplates(ctx: PluginContext): Promise<FxResult> {
```
hund030, line 146: Seems you include the same change in two PRs?

wenytang-ms: sorry, will close this PR

xzf0587 (approved):

PR #3197, opened 2021-12-02
perf(sql): implement updateArmTemplate interface

wenytang-ms:


packages/fx-core/tests/plugins/resource/sql/unit/generateArmTemplates.test.ts
```diff
@@ -104,7 +104,7 @@ describe("generateArmTemplates", () => {
       version: "1.0.0",
       activeResourcePlugins: activeResourcePlugins,
     } as AzureSolutionSettings;
-    const result = await sqlPlugin.generateArmTemplates(pluginContext);
+    const result = await sqlPlugin.updateArmtemplates(pluginContext);
 
     // Assert
     const testModuleFileName = "sqlProvision.result.bicep";
```
blackchoey, line 110: you may also need to remove L125 when updating the test case

wenytang-ms: sure, thanks

xzf0587 (approved):

PR #3204, opened 2021-12-02
feat(bot): improve bot error handling of judge error type

YitongFeng-git:
- Add the generic logic when throw Error:
  - If the Error has InnerError, if the innerError has statuscode
  - If 400 < statuscode < 500 User Error
  - else SystemError
  
- In retryHandler，when retryCount reduced to 0, throw the error rather than undefined


YitongFeng-git:
E2E Test: https://github.com/OfficeDev/TeamsFx/blob/d18908a872b16f0c737d211bd056ffe9e927018b/packages/cli/tests/e2e/bot/ProvisionResourceProjectWithNewBot.tests.ts#L28

packages/fx-core/src/plugins/resource/bot/errors.ts
```diff
@@ -34,12 +37,45 @@ export class PluginError extends Error {
...
 
   genMessage(): string {
     return `${this.message} Suggestions: ${this.suggestions.join(" ")}`;
   }
+
+  inferFromInnerError() {
+    if (!this.innerError) return;
+
+    const errorCode = this.innerError.response?.data?.error?.code;
+    const helpLink = GraphErrorCodes.get(errorCode);
```
IvanJobs, line 52: Looks not consistent for me.
1. This line should only applies for Graph Errors, but the class name is PluginError.
2. Get out a helpLink by GraphErrorCodes.get? Not good from readability.

YitongFeng-git: Yes it only has graph help link, for other's error, the helpLink =GraphErrorCodes.get(errorCode) will be undefined, and use the parameter's helpLink. The GraphError help link is special for it needs to show different helpLinks according to ErrorType.

IvanJobs (dismissed):
Overall logic is good, readability/design can be better.
Nice for the throw part of retryHandler.

IvanJobs (approved):

PR #3205, opened 2021-12-02
refactor(key-vault): support passing SKU name from provisionParameters

formulahendry:
1. support passing SKU name from provisionParameters
2. check whether KeyVault is already added


packages/fx-core/templates/plugins/resource/keyvault/bicep/keyVaultProvision.template.bicep
```diff
@@ -27,7 +28,7 @@ resource keyVault 'Microsoft.KeyVault/vaults@2019-09-01' = {
       }
     ]
     sku: {
-      name: 'standard'
+      name: keyVaultSkuName // You can follow https://aka.ms/teamsfx-bicep-add-param-tutorial to add "keyVaultSkuName" property to provisionParameters to override the default value "standard".
       family: 'A'
```
blackchoey, line 32: should we also expose the family?

formulahendry: family only has 'A' option. So no need to expose.

1yefuwang1 (approved):

PR #3208, opened 2021-12-03
fix: add function twice append provision bicep twice

wenytang-ms:
1. fix append configuration files.

E2E TEST: https://github.com/OfficeDev/TeamsFx/actions/runs/1533422769

### DIFF
--- a/packages/fx-core/src/plugins/solution/fx-solution/solution.ts
+++ b/packages/fx-core/src/plugins/solution/fx-solution/solution.ts
@@ -1641,7 +1641,7 @@ export class TeamsAppSolution implements Solution {
     const notifications: string[] = [];
     const pluginsToScaffold: LoadedPlugin[] = [this.LocalDebugPlugin];
     const azureResource = Array.from(settings.azureResources || []);
-    if (addFunc || ((addSQL || addApim || addKeyVault) && !alreadyHaveFunction)) {
+    if (addFunc || ((addSQL || addApim) && !alreadyHaveFunction)) {
       if (!azureResource.includes(AzureResourceFunction.id)) {
         pluginsToScaffold.push(functionPlugin);
         azureResource.push(AzureResourceFunction.id);
--- a/packages/fx-core/src/plugins/solution/fx-solution/solution.ts
+++ b/packages/fx-core/src/plugins/solution/fx-solution/solution.ts
@@ -1642,8 +1642,8 @@ export class TeamsAppSolution implements Solution {
     const pluginsToScaffold: LoadedPlugin[] = [this.LocalDebugPlugin];
     const azureResource = Array.from(settings.azureResources || []);
     if (addFunc || ((addSQL || addApim) && !alreadyHaveFunction)) {
+      pluginsToScaffold.push(functionPlugin);
       if (!azureResource.includes(AzureResourceFunction.id)) {
-        pluginsToScaffold.push(functionPlugin);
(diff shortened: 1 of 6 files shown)


packages/fx-core/src/plugins/solution/fx-solution/solution.ts
```diff
@@ -1641,13 +1641,11 @@ export class TeamsAppSolution implements Solution {
     const notifications: string[] = [];
     const pluginsToScaffold: LoadedPlugin[] = [this.LocalDebugPlugin];
     const azureResource = Array.from(settings.azureResources || []);
-    if (addFunc || ((addSQL || addApim) && !alreadyHaveFunction)) {
+    if (addFunc && !alreadyHaveFunction) {
```
blackchoey, line 1644: we may also need to handle cases on addFunc = true and alreadyHaveFunction = true, which only scaffolded an new api to Azure Functions project

wenytang-ms: sure, restore the logic

packages/fx-core/src/plugins/solution/fx-solution/solution.ts
```diff
@@ -1641,9 +1641,9 @@ export class TeamsAppSolution implements Solution {
     const notifications: string[] = [];
     const pluginsToScaffold: LoadedPlugin[] = [this.LocalDebugPlugin];
     const azureResource = Array.from(settings.azureResources || []);
-    if (addFunc || ((addSQL || addApim) && !alreadyHaveFunction)) {
-      pluginsToScaffold.push(functionPlugin);
+    if (addFunc || ((addSQL || addApim || addKeyVault) && !alreadyHaveFunction)) {
       if (!azureResource.includes(AzureResourceFunction.id)) {
```
adashen, line 1645: one question, it uses id to identify the scaffold. If the add is before provision, what will happen?

wenytang-ms: these id read from azureResouruces in projectSettings.json file, provision has no effect on this value.

xzf0587 (approved):

packages/fx-core/src/plugins/solution/fx-solution/solution.ts
adashen: Got it!

formulahendry (approved):

packages/fx-core/tests/plugins/solution/solution.arm.test.ts
```diff
@@ -368,6 +368,10 @@ output teamsFxConfigurationOutput object = contains(reference(resourceId('Micros
     expect(
       await fs.readFile(path.join(projectArmTemplateFolder, "../provision/bot.bicep"), fileEncoding)
     ).equals("Mocked bot Provision content. simple auth endpoint: Mocked simple auth endpoint");
+
+    expect(
+      await fs.readFile(path.join(projectArmTemplateFolder, "../teamsFx/bot.bicep"), fileEncoding)
```
dilin-MS, line 373: Can remove the above
```ts
expect(await fs.pathExists(path.join(projectArmTemplateFolder, "../teamsFx/bot.bicep"))).to.be.true;
```

wenytang-ms: sure

blackchoey (approved):

jayzhang (approved):

PR #3211, opened 2021-12-03
chore: add created-from telemetry property for bi data

a1exwang:
- add created-from property to telemetry events so BI report will not need to aggregate all projects and some hacky methods to determine at what version the project is created
- BI reports for old projects will still use the old method
- create-from property is core's package version, because we cannot get extension/cli version from core. toolkit/cli version can be determined because they have 1:1 mapping
- its implementation is similar to is-from-sample
- ut added

tested local debug happy path and packaged version:

CopyQ VDTdZl
CopyQ AgAnJO

https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/12690560/

### DIFF
--- a/packages/vscode-extension/src/handlers.ts
+++ b/packages/vscode-extension/src/handlers.ts
@@ -222,6 +222,7 @@ export async function activate(): Promise<Result<Void, FxError>> {
     await openSampleReadmeHandler();
     await postUpgrade();
     ExtTelemetry.isFromSample = await getIsFromSample();
+    ExtTelemetry.createdFrom = await getCreatedFrom();
 
     if (workspacePath) {
       // refresh env tree when env config files added or deleted.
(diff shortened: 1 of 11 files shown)


jayzhang (approved):

XiaofuHuang (approved):

packages/vscode-extension/src/handlers.ts
```diff
@@ -222,6 +222,7 @@ export async function activate(): Promise<Result<Void, FxError>> {
     await openSampleReadmeHandler();
     await postUpgrade();
     ExtTelemetry.isFromSample = await getIsFromSample();
+    ExtTelemetry.createdFrom = await getCreatedFrom();
```
HuihuiWu-Microsoft, line 225: This might be undefined sometimes and we have an extra field in core for isFromSample to avoid the issue.

a1exwang: thanks for the advice, updated.

HuihuiWu-Microsoft (approved):

XiaofuHuang (approved):

chagong (approved):

LongOddCode (approved):

PR #3216, opened 2021-12-03
chore(arm): update error output

xzf0587:
1. remove duplicate error output when arm deployment failed.
2. update error message in notification.
3. add module error property in telemetry.


xzf0587:
image

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -399,10 +401,15 @@ export async function deployArmTemplates(ctx: SolutionContext): Promise<Result<v
         [SolutionTelemetryProperty.Success]: SolutionTelemetrySuccess.Yes,
       });
     } else {
+      const properties: { [key: string]: string } = {};
+      if (result.error.innerError) {
+        properties[SolutionTelemetryProperty.ModuleErrorMessage] = result.error.innerError;
```
blackchoey, line 406: can we name the property to `arm-deployment-error` or something else to make it more straightforward?

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -399,10 +401,15 @@ export async function deployArmTemplates(ctx: SolutionContext): Promise<Result<v
         [SolutionTelemetryProperty.Success]: SolutionTelemetrySuccess.Yes,
       });
     } else {
+      const properties: { [key: string]: string } = {};
```
blackchoey, line 404: better naming for the variable? maybe `errorProperties` or something else

packages/vscode-extension/src/handlers.ts
```diff
@@ -1626,7 +1626,10 @@ export async function showError(e: UserError | SystemError) {
       },
     };
 
-    const button = await window.showErrorMessage(`[${errorCode}]: ${e.message}`, help);
+    const button = await window.showErrorMessage(
+      `[${errorCode}]: ${e.userData ?? e.message}`,
```
blackchoey, line 1630: Is there any other code using `userData`? Per the comment of userData, seems it's for other purpose

PR #3217, opened 2021-12-03
chore(telemetry): fill createdFrom when upgrading

a1exwang:
this is to prevent GDPR classification error making it indistinguishable to know whether it is from v2 or v3

Tested migrate happy path

### DIFF
--- a/packages/fx-core/src/core/middleware/projectMigrator.ts
+++ b/packages/fx-core/src/core/middleware/projectMigrator.ts
@@ -884,6 +884,11 @@ async function ensureProjectSettings(
     settings.defaultFunctionName =
       settings.defaultFunctionName || envDefault[PluginNames.FUNC]?.[defaultFunctionName];
   }
+  if (!settings.createdFrom) {
+    // We won't know the exact version when the project is created.
+    // But we know it is from core <1.0 (extension < 3.0).
+    settings.createdFrom = "<1.0";
+  }
   settings.version = "2.0.0";
   await fs.writeFile(projectSettingPath, JSON.stringify(settings, null, 4), {
     encoding: "UTF-8",


packages/fx-core/src/core/middleware/projectMigrator.ts
```diff
@@ -884,6 +884,11 @@ async function ensureProjectSettings(
     settings.defaultFunctionName =
       settings.defaultFunctionName || envDefault[PluginNames.FUNC]?.[defaultFunctionName];
   }
+  if (!settings.createdFrom) {
+    // We won't know the exact version when the project is created.
+    // But we know it is from core <1.0 (extension < 3.0).
+    settings.createdFrom = "<1.0";
+  }
```
dooriya, line 891: Why not  just `1.0`?
@jayzhang  Should we introduce the version of `fx-core` to app developers? This is just the internal concept of TeamsFx.

a1exwang: `createdFrom` is the core version when the project is created. It is only for BI report and not for app developers. 
1. core 1.0 means 3.0 toolkit
2. core <1.0 means 2.0 toolkit. We won't know the exact version in which the project is created, so just use <1.0.

a1exwang: `createdFrom` is the core version when the project is created. It is only for BI report and not for app developers. 
1. core 1.0 means 3.0 toolkit
2. core <1.0 means 2.0 toolkit. We won't know the exact version in which the project is created, so just use <1.0.

meifans (approved):

packages/fx-core/src/core/middleware/projectMigrator.ts
a1exwang: The BI challenge was that due to toolkit reload and gdpr classification issue, it is hard to calculate things like these:
- provision/local debug failure rate after migration
- how many projects are created in v3 and how many projects are migrated

(we currently solve it by aggregating all historic telemetry data by projectId, but we will hit the Kusto limit soon. It is not a long term solution)

However, this pr is not needed anymore. Because after fixing [this bug](https://github.com/OfficeDev/TeamsFx/pull/3242). We can just reuse `version` property in `projectSettings.json` in BI.

So I will close this.

PR #3223, opened 2021-12-06
chore: add azure sp login

xiaolang124:
1. Add service principal login for CLI Azure
2. Save encrypted service principal data in ~/home/account/azure.sp

image


packages/cli/src/cmds/account.ts
```diff
@@ -169,14 +181,37 @@ class AccountLogin extends YargsCommand {
         description: "Authenticate with a specific Azure Active Directory tenant.",
         type: "string",
         default: "",
+      })
+      .options("service-principal", {
```
swatDong, line 185: Is this to follow the same behavior as az-cli? Add error handlings for all unsupported cases:
- -u xxx@xxx.com -p SECRET (user login)
- --service-principal -u CLIENTID -p CERTIFICATE -t TENANTID (cert credential)
- --service-principal -u CLIENTID -p SECRET (no tenant)

Call out in documentation/example that the only supported case is **--service-principal -u CLIENTID -p SECRET -t TENANTID**

swatDong, line 185: Anyway to show these options only for **azure**?

packages/cli/src/cmds/account.ts
```diff
@@ -169,14 +181,37 @@ class AccountLogin extends YargsCommand {
...
+      })
+      .options("username", {
+        alias: "u",
+        description: "Client ID for service principal",
+        type: "string",
+        default: "",
+      })
+      .options("password", {
+        alias: "p",
+        description: "Client ID for service principal",
```
swatDong, line 198: Any plan to support certificate credential? Certificate is more commonly used than secret in production environments. And to meet security requirements as well.

packages/cli/src/cmds/account.ts
xiaolang124: Seperate login command to m365 and azure.

xiaolang124: Already add parameter check logic. And add examples to command help.
image

packages/cli/src/cmds/account.ts
xiaolang124: Add cert login logic to CLI.

packages/cli/src/commonlib/azureLoginCI.ts
```diff
@@ -98,53 +154,155 @@ export class AzureAccountManager extends login implements AzureAccountProvider {
...
     throw new Error("Method not implemented.");
   }
 
-  getJsonObject(showDialog?: boolean): Promise<Record<string, unknown> | undefined> {
-    throw new Error("Method not implemented.");
+  async listSubscriptions(): Promise<SubscriptionInfo[]> {
+    const credential = await this.getAccountCredentialAsync();
+    if (credential) {
+      const client = new SubscriptionClient(credential);
+      const tenants = await listAll(client.tenants, client.tenants.list());
```
swatDong, line 181: SP is bound in one tenant, why need to list tenants for SP?

packages/cli/src/commonlib/cacheAccess.ts
```diff
@@ -201,3 +202,54 @@ export async function loadAccountId(accountName: string) {
...
   return undefined;
 }
+
+export async function saveAzureSP(
+  clientId: string,
+  secret: string,
+  tenantId: string
+): Promise<void> {
+  await fs.ensureDir(cacheDir);
+  const azureSPCrypto = new AccountCrypto("azure-sp");
```
swatDong, line 212: Use `this.accountCrypto`, do not create another AccountCrypto.

packages/cli/src/commonlib/azureLoginCI.ts
xiaolang124: Good point, I have removed multi tenants part.

packages/cli/src/commonlib/cacheAccess.ts
xiaolang124: Thanks for the advice, I have changed ```AzureSpCypto``` method to static method.

packages/cli/src/commonlib/cacheAccess.ts
```diff
@@ -201,3 +202,60 @@ export async function loadAccountId(accountName: string) {
 
   return undefined;
 }
+
+export class AzureSpCrypto {
+  private static readonly accountCrypto: AccountCrypto = new AccountCrypto("sp");
```
swatDong, line 207: Just use "azure" as account name to keep consistency with user azure account. No need to add another crypto key for SP.

swatDong (dismissed):

swatDong (dismissed):

swatDong (approved):

Alive-Fish (approved):

PR #3227, opened 2021-12-06
feat(collaboration): collaboration support spfx project, cli message update

SLdragon:
E2E TEST: Will add E2E test in next PR

image

### DIFF
--- a/packages/cli/src/cmds/permission.ts
+++ b/packages/cli/src/cmds/permission.ts
@@ -4,7 +4,19 @@
 "use strict";
 
 import path from "path";
-import { FxError, err, ok, Result, Stage, LogLevel } from "@microsoft/teamsfx-api";
+import {
+  FxError,
+  err,
+  ok,
+  Result,
+  Stage,
+  LogLevel,
+  AzureSolutionSettings,
+  Inputs,
+  Platform,
+  ProjectConfig,
+} from "@microsoft/teamsfx-api";
+
 import { Argv, Options } from "yargs";
 import { YargsCommand } from "../yargsCommand";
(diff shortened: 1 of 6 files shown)


nliu-ms (dismissed):

packages/cli/src/cmds/permission.ts
```diff
@@ -17,13 +29,40 @@ import activate from "../activate";
...
 
+const spfxMessage =
+  "Notice: SPFX deployment permission needs to be handled manually by SharePoint site administrator.\n" +
+  "Manage site admins using SharePoint admin center: " +
+  "https://docs.microsoft.com/en-us/sharepoint/manage-site-collection-administrators";
+
+async function isSpfxProject(
+  rootFolder: string,
+  core: FxCore
+): Promise<Result<boolean | undefined, FxError>> {
```
Alive-Fish, line 48: Is this better to add into core/solution? I think CLI/VSCode should not manage this hardcode logic.

SLdragon: Thanks, updated and use an existing core function for this purpose

nliu-ms (approved):

Alive-Fish (approved):

packages/cli/src/cmds/permission.ts
SLdragon: Add two tasks in our board to optimize this in the future
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/12763877
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/12763884

PR #3233, opened 2021-12-06
fix(solution): spfx postScaffold

jayzhang:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/12762364
E2E TEST: AzureAppHappyPath


nliu-ms (approved):

packages/fx-core/src/plugins/solution/fx-solution/v2/scaffolding.ts
```diff
@@ -96,6 +103,14 @@ export async function scaffoldSourceCode(
         [SolutionTelemetryProperty.ProgrammingLanguage]:
           ctx.projectSetting?.programmingLanguage ?? "",
       });
+    } else {
+      //For SPFx plugin, execute it alone lastly
+      if (SPFxPlugin.scaffoldSourceCode) {
+        const spfxRes = await SPFxPlugin.scaffoldSourceCode(ctx, inputs);
```
HuihuiWu-Microsoft, line 109: It seems tricky that you do spfx scaffold after showing scaffold successful notification

PR #3236, opened 2021-12-06
fix(vsc): support treeview local debug button for spfx project

XiaofuHuang:
1. Fix the local debug button does not work  in the spfx project
2. If the user removes the default local debug configurations in launch.json, show all the choices in launch.json.

Detail:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_boards/board/t/Infrastructure%20Team/Stories/?workitem=12757449


packages/fx-core/src/plugins/resource/localdebug/launch.ts
```diff
@@ -524,3 +524,21 @@ export function generateSpfxCompounds(): Record<string, unknown>[] {
   );
   return launchCompounds;
 }
+
+export function getLaunchNamePrefix(launchConfig: any, isSPFxProject: boolean): string | undefined {
+  try {
+    const defaultCompoundNames = isSPFxProject
+      ? ["Teams workbench (Edge)", "Teams workbench (Chrome)"]
+      : ["Debug (Edge)", "Debug (Chrome)"];
+    const compounds = launchConfig.compounds as any[];
```
swatDong, line 533: The most accurate logic needs to check `presentation:hidden` as well...
BTW, I don't think we need to cover all corner cases here, how about just show all debug list?

XiaofuHuang: Got it, changed to show all the choices include "launch remote"

swatDong (approved):

HuihuiWu-Microsoft (approved):

kimizhu (approved):

dooriya (approved):

PR #3242, opened 2021-12-07
fix: fix project settings version

a1exwang:
- fix the bug that project settings is not changed in 3.0.0
- remove `createdFrom` (not needed and it is a duplicate) and use `version` instead in telemetry

### DIFF
--- a/packages/fx-core/src/core/index.ts
+++ b/packages/fx-core/src/core/index.ts
@@ -172,7 +172,7 @@ export let TOOLS: Tools;
 export class FxCore implements Core {
   tools: Tools;
   isFromSample?: boolean;
-  createdFrom?: string;
+  settingsVersion?: string;
 
   constructor(tools: Tools) {
     this.tools = tools;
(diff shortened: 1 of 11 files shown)


jayzhang (approved):

HuihuiWu-Microsoft (approved):

chagong (approved):

packages/fx-core/src/core/index.ts
```diff
@@ -1552,7 +1551,7 @@ export function undefinedName(objs: any[], names: string[]) {
 }
 
 export function getProjectSettingsVersion() {
-  if (isFeatureFlagEnabled(FeatureFlagName.InsiderPreview, false)) return "2.0.0";
+  if (isMultiEnvEnabled()) return "2.0.0";
```
dooriya, line 1554: So for the latest toolkit (v3.0.0), the project version in `projectSettings.json` is `2.0.0` rather than `3.0.0`?

a1exwang: I know the issue. This PR mainly focuses on telemetry. The semantics of `version` will be in the next PR

dooriya: Cool! Thanks for the explaination.

dooriya (approved):

PR #3245, opened 2021-12-07
fix: apiv2 spfx deploy

1yefuwang1:
fixes https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_queries/query/eded2ac0-3291-4335-9382-75f270bc8ffa/

### DIFF
--- a/packages/fx-core/src/plugins/solution/fx-solution/v2/provision.ts
+++ b/packages/fx-core/src/plugins/solution/fx-solution/v2/provision.ts
@@ -187,7 +187,7 @@ export async function provisionResource(
     }
   }
 
-  if (isV2() && !isAzureProject(azureSolutionSettings)) {
+  if (isV2() && isAzureProject(azureSolutionSettings)) {
     solutionInputs.remoteTeamsAppId =
       newEnvInfo.state[PluginNames.APPST]["output"][Constants.TEAMS_APP_ID];
   }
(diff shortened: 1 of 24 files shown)


jayzhang (approved):

hund030 (approved):

YitongFeng-git (approved):

packages/fx-core/src/plugins/solution/fx-solution/v2/provision.ts
```diff
@@ -187,7 +187,7 @@ export async function provisionResource(
     }
   }
 
-  if (isV2() && !isAzureProject(azureSolutionSettings)) {
+  if (isV2() && isAzureProject(azureSolutionSettings)) {
     solutionInputs.remoteTeamsAppId =
```
nliu-ms, line 191: how about spfx? no need for remote teams app id?

nliu-ms (approved):

packages/fx-core/src/plugins/solution/fx-solution/v2/provision.ts
1yefuwang1: Yeah. When multiple env is enabled, solution doesn't need to remember remote teams app id any more.

XiaofuHuang (approved):

PR #3247, opened 2021-12-07
chore: add more info for subscription item in env tree view

dooriya:
Fix bug: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/12607395
- Add subscription id info as the tree item description for each subscription node.
- Add tooltip for subscription node (user can mouse over to see more info about this node)

image

### DIFF
--- a/packages/vscode-extension/src/envTree.ts
+++ b/packages/vscode-extension/src/envTree.ts
@@ -255,6 +255,7 @@ async function appendSubscriptionAndResourceGroupNode(env: string): Promise<void
   ) {
     const envSubItems: TreeItem[] = [];
     const subscriptionInfo = await getSubscriptionInfoFromEnv(env);
+
     if (subscriptionInfo) {
       const subscriptionTreeItem: TreeItem = {
         commandId: `fx-extension.environment.subscription.${env}`,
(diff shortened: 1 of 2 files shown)


qinezh:
Looks it's hardly to display the full subscription id without resizing the tree view, how about putting the subscription id to the tooltip instead?

dooriya:
> Looks it's hardly to display the full subscription id without resizing the tree view, how about putting the subscription id to the tooltip instead?

Sounds good. In the latest commit, I also added subscription id info in the tooltip.

qinezh (dismissed):

packages/vscode-extension/src/envTree.ts
```diff
@@ -255,11 +255,23 @@ async function appendSubscriptionAndResourceGroupNode(env: string): Promise<void
...
+
     if (subscriptionInfo) {
       const subscriptionTreeItem: TreeItem = {
         commandId: `fx-extension.environment.subscription.${env}`,
         contextValue: "openSubscriptionInPortal",
         label: subscriptionInfo.subscriptionName ?? subscriptionInfo.subscriptionId,
+        description: subscriptionInfo.subscriptionId,
+        tooltip: {
+          isMarkdown: false,
+          value: `'${env}' environment is provisioned in Azure subscription '${
```
HuihuiWu-Microsoft, line 267: Can you extract the string to Strings.json for later localization?

a1exwang (dismissed):

HuihuiWu-Microsoft (approved):

packages/vscode-extension/src/envTree.ts
dooriya: Sure. Thanks for suggestion.

PR #3250, opened 2021-12-07
fix: add retry as a workaround for unstable App Studio API

nliu-ms:
Most of the cases retry 1 or 2 times can work.
image

image

But retry cannot resolve all the cases:
image

### DIFF
--- a/packages/fx-core/src/plugins/resource/appstudio/appStudio.ts
+++ b/packages/fx-core/src/plugins/resource/appstudio/appStudio.ts
@@ -8,7 +8,7 @@ import { AppStudioError } from "./errors";
 import { IPublishingAppDenition } from "./interfaces/IPublishingAppDefinition";
 import { AppStudioResultFactory } from "./results";
 import { getAppStudioEndpoint } from "../../..";
-import { Constants, ErrorMessages } from "./constants";
+import { Constants, ErrorMessages, RETRY_INTERVAL, RETRY_MAX_TIMES } from "./constants";
 import { sleep } from "../spfx/utils/utils";
 
 // eslint-disable-next-line @typescript-eslint/no-namespace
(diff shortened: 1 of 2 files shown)


packages/fx-core/src/plugins/resource/appstudio/constants.ts
```diff
@@ -45,6 +45,8 @@ export const DEFAULT_COLOR_PNG_FILENAME = "color.png";
 export const DEFAULT_OUTLINE_PNG_FILENAME = "outline.png";
 export const MANIFEST_RESOURCES = "resources";
 export const APP_PACKAGE_FOLDER_FOR_MULTI_ENV = "templates/appPackage";
+export const RETRY_INTERVAL = 30000;
```
IvanJobs, line 48: Nit: better to comment it with unit, or name the const with unit, for example, `RETRY_INTERVAL_MS`.

packages/fx-core/src/plugins/resource/appstudio/appStudio.ts
```diff
@@ -189,21 +188,29 @@ export namespace AppStudioClient {
...
-      );
-      if (response && response.data) {
-        const app = <IAppDefinition>response.data;
-
-        if (app && app.teamsAppId && app.teamsAppId === teamsAppId) {
-          return app;
-        } else {
-          await logProvider?.error(
-            `teamsAppId mismatch. Input: ${teamsAppId}. Got: ${app.teamsAppId}`
+      let retry = 0;
```
IvanJobs, line 191: Nit: To reduce duplicated code, better to leverage https://github.com/OfficeDev/TeamsFx/blob/dev/packages/fx-core/src/plugins/resource/bot/utils/retryHandler.ts

jayzhang (approved):

PR #3251, opened 2021-12-07
perf: migrate generate arm template with activePlugins

wenytang-ms:


blackchoey (approved):
UT failed, please take a look. By the way, if we don't have test case to cover this scenario, we may need to add one case.

wenytang-ms:
> 

sure, add task here 
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/12798624

blackchoey (approved):

PR #3270, opened 2021-12-08
fix: unexpected select-env window pop-up

dooriya:
Fix bug: https://dev.azure.com/msazure/Microsoft%20Teams%20Extensibility/_workitems/edit/12783952

### DIFF
--- a/packages/vscode-extension/src/handlers.ts
+++ b/packages/vscode-extension/src/handlers.ts
@@ -273,6 +273,8 @@ async function getIsFromSample() {
 async function getSettingsVersion(): Promise<string | undefined> {
   if (core) {
     const input = getSystemInputs();
+    input.ignoreEnvInfo = true;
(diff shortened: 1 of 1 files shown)


HuihuiWu-Microsoft (approved):

packages/vscode-extension/src/handlers.ts
```diff
@@ -273,6 +273,8 @@ async function getIsFromSample() {
 async function getSettingsVersion(): Promise<string | undefined> {
   if (core) {
     const input = getSystemInputs();
+    input.ignoreEnvInfo = true;
```
qinezh, line 276: would it be better to move this logic into `core.getProjectConfig` avoid similar issue occurs again?

a1exwang (approved):

qinezh (approved):

Issue #3207, opened 2021-12-02
bug report: EPERM: Operation not permitted, mkdir


q00u:
>EPERM: operation not permitted, mkdir 'C:\Program Files\Microsoft VS Code\%USERPROFILE%'

On Windows 10. Trying to set the Default Project Root Directory to be in the 'dev' folder, under my home directory.

e.g., `C:\Users\q00u\dev\TeamsApps`

The setting description is:

>Set the default root directory for creating new teams app project. By default, the root directory is $HOME_DIR/TeamsApps.

`$HOME_DIR/dev/TeamsApps` doesn't work. `%USERPROFILE%\dev\TeamsApps` doesn't work. Only a full path works, but I have different user folders on different computers.

1openwindow:
Thanks for report this issue @q00u . this is a common scenario, I will look at this. if you have any suggestion, feel free to comment.

1openwindow:
@q00u, in our next release, the extension will support "${homeDir}" global variable, please try it out once new release is published. The next release date should be early of next week. Thanks for your feedback, if any else, please kindly let us know.
![image](https://user-images.githubusercontent.com/26411726/146704431-96759cfe-4ef7-429a-b30d-0a9971e741c2.png)

PR #3273, opened 2021-12-08
fix: support homedir for project path setting

1openwindow:
fix this issue: #3207 

image

### DIFF
--- a/packages/fx-core/src/common/tools.ts
+++ b/packages/fx-core/src/common/tools.ts
@@ -453,7 +453,7 @@ export function getRootDirectory(): string {
   if (root === undefined || root === "") {
     return path.join(os.homedir(), ConstantString.rootFolder);
   } else {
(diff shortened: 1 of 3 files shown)


packages/fx-core/src/common/tools.ts
```diff
@@ -453,7 +453,7 @@ export function getRootDirectory(): string {
   if (root === undefined || root === "") {
     return path.join(os.homedir(), ConstantString.rootFolder);
   } else {
-    return root;
+    return root.replace("${homeDir}", os.homedir());
```
HuihuiWu-Microsoft, line 456: What if user fill in an absolute directory?

1openwindow: it works

HuihuiWu-Microsoft (approved):

PR #3276, opened 2021-12-08
refactor: integrate spfx solution with core for create

HuihuiWu-Microsoft:


packages/fx-core/src/core/SolutionPluginContainer.ts
```diff
@@ -40,6 +42,14 @@ export function getAllSolutionPlugins(): Solution[] {
   return plugins;
 }
 
+export function getSolutionPluginByCap(cap: string[]): v2.SolutionPlugin | undefined {
```
HuihuiWu-Microsoft, line 45: Since solution select question is removed, currently the map between capability and solution is hard coded here. Later we can extend the function of question model and set selected solution when a certain tree node is traversed.

packages/fx-core/src/core/index.ts
```diff
@@ -156,6 +159,10 @@ function featureFlagEnabled(flagName: string): boolean {
   }
 }
 
+export function isV3() {
```
HuihuiWu-Microsoft, line 162: @LongOddCode  You can update if you add feature flag in another place.

packages/fx-core/src/core/index.ts
```diff
@@ -1363,13 +1371,22 @@ export class FxCore implements Core {
...
-      if (res.isErr()) return res;
+      if (res.isErr()) return err(new SystemError(res.error, CoreSource, "QuestionModelFail"));
       if (res.value) {
-        const solutionNode = res.value as QTreeNode;
-        solutionNode.condition = { equals: solutionPlugin.name };
-        if (solutionNode.data) solutionSelectNode.addChild(solutionNode);
+        const solutionNode = Array.isArray(res.value)
+          ? (res.value as QTreeNode[])
+          : [res.value as QTreeNode];
+        for (const node of solutionNode) {
```
HuihuiWu-Microsoft, line 1379: Currently the solution node might be an array and each solution can return several nodes

packages/fx-core/src/plugins/solution/fx-solution/v2/getQuestions.ts
```diff
@@ -50,29 +49,27 @@ import {
...
-    ResourcePluginsV2.SpfxPlugin
-  );
-  if (spfxPlugin.getQuestionsForScaffolding) {
-    const res = await spfxPlugin.getQuestionsForScaffolding(ctx, inputs);
-    if (res.isErr()) return res;
-    if (res.value) {
-      const spfxNode = res.value as QTreeNode;
-      spfxNode.condition = { contains: TabSPFxItem.id };
-      if (spfxNode.data) capNode.addChild(spfxNode);
+  if (!isV3()) {
```
HuihuiWu-Microsoft, line 60: For V2: only TeamsAppSolution in the list and spfx is included
For V3: TeamsAppSolution + TeamsSPFxSolution in the list, and tab+bot+me in TeamsAppSolution, spfx in TeamsSPFxSolution

packages/fx-core/src/plugins/solution/spfx-solution/scaffolding.ts
```diff
@@ -0,0 +1,61 @@
...
+  Json,
+  Result,
+  v2,
+  AppStudioTokenProvider,
+  Void,
+  ok,
+} from "@microsoft/teamsfx-api";
+import Module from "module";
+
+export async function scaffoldSourceCode(
```
HuihuiWu-Microsoft, line 14: Will implement the methods in right position later

HuihuiWu-Microsoft:
Will open another PR

PR #3292, opened 2021-12-09
chore(localdebug): add local debug correlation id

a1exwang:
localdebug contains several parts in TaskProvider, DebugProvider and several command handlers.
- the correlation id life cycle is defined as follows:
    - starts from the start of env-checker
    - ends when all the teamsfx debug sessions end or another debug session starts

Implementation:
1. For env-checker/task providers/preDebugCheck/DebugProvider, use a global variable to store correlation id (assuming at most one debug session at anytime)
2. For onDidStart/TerminateDebugSessionHandler, use event.configuration.teamsfxCorrelationId. So concurrent issue will not happen in this stage

Stage 2 is longer than stage 1, so maybe stage 1 is less likely to encounter concurrent issue.

#### Tested tab local debug happy path (timestamp from latest to earliest):
image

#### Tab+Function happy path (multiple debug sessions are correctly handled):
image

#### Cancel local debug in env checker multiple times
image

#### Concurrent run in debug stage (note that correlation-id of the debug-stop event is still correct):
image

### DIFF
--- a/packages/vscode-extension/src/debug/commonUtils.ts
+++ b/packages/vscode-extension/src/debug/commonUtils.ts
@@ -3,6 +3,7 @@
 
 import * as fs from "fs-extra";
 import * as path from "path";
+import * as uuid from "uuid";
 import * as dotenv from "dotenv";
 import * as vscode from "vscode";
 import * as constants from "./constants";
@@ -395,4 +396,19 @@ export async function loadTeamsFxDevScript(componentRoot: string): Promise<strin
   } else {
     return undefined;
   }
+}
+
+// Helper functions for local debug correlation-id, only used for telemetry
+let localDebugCorrelationId: string | undefined = undefined;
+export function startLocalDebugSession(): string {
+  localDebugCorrelationId = uuid.v4();
+  return getLocalDebugSessionId();
+}
+
+export function endLocalDebugSession() {
+  localDebugCorrelationId = undefined;
+}
+
+export function getLocalDebugSessionId(): string {
+  return localDebugCorrelationId || "undefined";
 }
--- a/packages/vscode-extension/src/debug/commonUtils.ts
+++ b/packages/vscode-extension/src/debug/commonUtils.ts
@@ -410,5 +410,5 @@ export function endLocalDebugSession() {
 }
 
 export function getLocalDebugSessionId(): string {
-  return localDebugCorrelationId || "undefined";
+  return localDebugCorrelationId || "no-session-id";
 }
--- a/packages/vscode-extension/src/debug/teamsfxDebugProvider.ts
+++ b/packages/vscode-extension/src/debug/teamsfxDebugProvider.ts
@@ -39,6 +39,10 @@ export class TeamsfxDebugProvider implements vscode.DebugConfigurationProvider {
           return debugConfiguration;
         }
 
+        // Attach correlation-id to DebugConfiguration so concurrent debug sessions are correctly handled in this stage.
+        // For backend and bot debug sessions, debugConfiguration.url is undefined so we need to set correlation id early.
+        debugConfiguration.teamsfxCorrelationId = commonUtils.getLocalDebugSessionId();
+
         if (debugConfiguration.url === undefined) {
           return debugConfiguration;
         }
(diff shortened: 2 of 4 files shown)


packages/vscode-extension/src/debug/commonUtils.ts
```diff
@@ -396,3 +397,18 @@ export async function loadTeamsFxDevScript(componentRoot: string): Promise<strin
...
+  localDebugCorrelationId = uuid.v4();
+  return getLocalDebugSessionId();
+}
+
+export function endLocalDebugSession() {
+  localDebugCorrelationId = undefined;
+}
+
+export function getLocalDebugSessionId(): string {
+  return localDebugCorrelationId || "undefined";
```
swatDong, line 413: Use another string instead of "undefined", to distinguish with real undefined case. E.g., "local-debug-not-started".
(Just like do not set your display name to `null` :P )

packages/vscode-extension/src/debug/teamsfxDebugProvider.ts
```diff
@@ -9,20 +10,39 @@ import * as commonUtils from "./commonUtils";
...
+  teamsfxCorrelationId?: string;
 }
 
 export class TeamsfxDebugProvider implements vscode.DebugConfigurationProvider {
   public async resolveDebugConfiguration?(
     folder: vscode.WorkspaceFolder | undefined,
     debugConfiguration: TeamsfxDebugConfiguration,
     token?: vscode.CancellationToken
+  ): Promise<vscode.DebugConfiguration | undefined> {
+    return Correlator.runWithId(
```
swatDong, line 22: Will local debug correlator nest with command correlator?

packages/vscode-extension/src/debug/teamsfxTaskHandler.ts
```diff
@@ -376,17 +381,45 @@ function onDidTerminateDebugSessionHandler(event: vscode.DebugSession): void {
...
+        onDidStartDebugSessionHandler,
+        event
+      )
+    )
   );
   ext.context.subscriptions.push(
-    vscode.debug.onDidTerminateDebugSession(onDidTerminateDebugSessionHandler)
+    vscode.debug.onDidTerminateDebugSession((event: vscode.DebugSession) =>
+      Correlator.runWithId(
+        event.configuration.teamsfxCorrelationId || getLocalDebugSessionId(),
```
swatDong, line 419: Why not always `event.configuration.teamsfxCorrelationId` or `getLocalDebugSessionId()`. What's the mismatch case?

packages/vscode-extension/src/debug/commonUtils.ts
swatDong: Or, "no-session-{$uuid}", if we do want to provide correlation for **all** scenarios.

packages/vscode-extension/src/debug/teamsfxDebugProvider.ts
a1exwang: No. Because Correlation is related to a callstack. As long as the two function call are not in the same call stack, they will not overlap. 

- TeamsfxDebugProvider is called by vscode when user clicks Debug or hits F5
- Other `Correlator.run()` are called in command handlers and task providers, so they will not overlap

Even if they overlap AsyncLocalStorage can handle it correctly. With a quick test:
```
> cat asyncstore.js                                                                                                                                                                                                             16:44:55
const { AsyncLocalStorage } = require("async_hooks");
const l = new AsyncLocalStorage();

l.run("1", () => {
  console.log("before " + l.getStore());
  l.run("2", () => {
    console.log("inner " + l.getStore());
  });
  console.log("after " + l.getStore());
})
[16:44:58] alexwang@alex-p330 ~  
> node asyncstore.js                                                                                                                                                                                                            16:44:58
before 1
inner 2
after 1   <======= this is 1, not 2. so they handle re-entrance correctly
```

packages/vscode-extension/src/debug/teamsfxTaskHandler.ts
a1exwang: 1. Prefer using `event.configuration.teamsfxCorrelationId` because `event` is associate with one vscode debug session, so concurrent f5 will not cause incorrect correlation id after this.
2. fallback to `getLocalDebugSessionId()` because the backend debug will not call `TeamsfxDebugHandler.resolveDebugConfiguration()` so `event.configuration.teamsfxCorrelationId` is undefined.

packages/vscode-extension/src/debug/commonUtils.ts
a1exwang: Good point, thanks!

a1exwang: fixed

swatDong (approved):

HuihuiWu-Microsoft (approved):

PR #3299, opened 2021-12-09
feat(frontend): support customize environment variable in multi env

hund030:
Frontend plugin will generate .env.teamsfx.dev file for dev environment and consume it in building stage. So that customer can add their environment variable in .env.teamsfx.dev file.


packages/fx-core/src/plugins/resource/frontend/env.ts
```diff
@@ -0,0 +1,95 @@
...
+      customizedRemoteEnvs: { ...configs.customizedRemoteEnvs, ...envs.customizedRemoteEnvs },
+    };
+
+    if (JSON.stringify(newConfigs) === JSON.stringify(configs)) {
+      // Avoid updating dotenv file's modified time if nothing changes.
+      // We decide whether to skip deployment by comparing the mtime of all project files and last deployment time.
+      return;
+    }
+
+    const envPath = path.join(folder, envFileName(envName));
```
eriolchan, line 72: This line looks duplicated with L37 in packages/fx-core/src/plugins/resource/frontend/env.ts, maybe we can just use path as parameter instead of folder and envName?

packages/fx-core/src/plugins/resource/frontend/utils/environment-utils.ts
```diff
@@ -21,7 +22,7 @@ export class EnvironmentUtils {
 
     let envs = "";
     for (const key in newConfigs) {
-      envs += `${key}=${newConfigs[key]}\r\n`;
+      envs += `${key}=${newConfigs[key]}${os.EOL}`;
```
eriolchan, line 25: It looks the similar concat logic in packages/fx-core/src/plugins/resource/frontend/env.ts?

packages/fx-core/src/plugins/resource/frontend/utils.ts
```diff
@@ -76,18 +80,22 @@ export class Utils {
...
-          Logger.error(`Failed to run command: "${command}".`);
-          if (!ignoreError) {
+      exec(
+        command,
+        { cwd: workingDir, env: { ...process.env, ...env } },
+        (error, standardOutput, stderr) => {
+          Logger.debug(standardOutput);
+          if (error) {
+            Logger.error(`Failed to run command: "${command}".`);
+            if (stderr) {
```
eriolchan, line 90: Does this mean we will output error in both output & error pane?

packages/fx-core/src/plugins/resource/frontend/utils.ts
```diff
@@ -76,18 +80,22 @@ export class Utils {
...
       }
 
-      exec(command, { cwd: workingDir }, (error, standardOutput) => {
-        Logger.debug(standardOutput);
-        if (error) {
-          Logger.error(`Failed to run command: "${command}".`);
-          if (!ignoreError) {
+      exec(
+        command,
+        { cwd: workingDir, env: { ...process.env, ...env } },
```
eriolchan, line 85: Can you help me understand what's the difference between the value of  process.env and env?

packages/fx-core/src/plugins/resource/frontend/ops/deploy.ts
```diff
@@ -57,7 +58,10 @@ export class FrontendDeployment {
 
     await progressHandler?.next(DeploySteps.Build);
     await runWithErrorCatchAndThrow(new BuildError(), async () => {
-      await Utils.execute(Commands.BuildFrontend, componentPath);
+      await Utils.execute(Commands.BuildFrontend, componentPath, {
+        ...envs.customizedRemoteEnvs,
```
eriolchan, line 62: I find the second parameter of Utils.execute is workingDir, which confused me with this env value, packages/fx-core/src/plugins/resource/frontend/utils.ts

packages/fx-core/src/plugins/resource/frontend/plugin.ts
```diff
@@ -255,23 +260,35 @@ export class FrontendPluginImpl {
...
+        EnvKeys.ClientID,
         ctx.envInfo.state
           .get(DependentPluginInfo.AADPluginName)
           ?.get(DependentPluginInfo.ClientID) as string
       );
     }
+    return envs;
+  }
+
+  private async updateMultiEnv(ctx: PluginContext): Promise<void> {
```
eriolchan, line 282: I don't quite fully understand these two method of updateMultiEnv and updateDotenv, will check with you offline next week

packages/fx-core/src/plugins/resource/frontend/env.ts
```diff
@@ -0,0 +1,95 @@
...
+
+export async function saveEnvFile(
+  envName: string,
+  folder: string,
+  envs: RemoteEnvs
+): Promise<void> {
+  try {
+    const configs = await loadEnvFile(envName, folder);
+    const newConfigs: RemoteEnvs = {
+      teamsfxRemoteEnvs: { ...configs.teamsfxRemoteEnvs, ...envs.teamsfxRemoteEnvs },
```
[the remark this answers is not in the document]
hund030: envs overwrite configs.

packages/fx-core/src/plugins/resource/frontend/env.ts
```diff
@@ -0,0 +1,95 @@
...
+      customizedRemoteEnvs: { ...configs.customizedRemoteEnvs, ...envs.customizedRemoteEnvs },
+    };
+
+    if (JSON.stringify(newConfigs) === JSON.stringify(configs)) {
```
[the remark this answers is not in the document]
hund030: There's only one case that the order would be different: user change the order of those configs generated by toolkit. 
In that case, I think it's reasonable to update the env file because user modifies it. Even though the modification doesn't take effect.
In case the order changed, then we just update the env file and skip deployment feature may not work as expected. I think it's ok because everything work true.

packages/fx-core/src/plugins/resource/frontend/env.ts
```diff
@@ -0,0 +1,95 @@
...
+      // Avoid updating dotenv file's modified time if nothing changes.
+      // We decide whether to skip deployment by comparing the mtime of all project files and last deployment time.
+      return;
+    }
+
+    const envPath = path.join(folder, envFileName(envName));
+    await fs.ensureFile(envPath);
+
+    const envString =
+      `# Following variables are generated by TeamsFx${os.EOL}` +
```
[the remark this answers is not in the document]
hund030: It doesn't break parsing. This comment is from env.teamsfx.local. I like it so I follow the pattern.

packages/fx-core/src/plugins/resource/frontend/env.ts
```diff
@@ -0,0 +1,89 @@
...
+import { TelemetryHelper } from "./utils/telemetry-helper";
+import { TelemetryEvent } from "./constants";
+
+export interface RemoteEnvs {
+  teamsfxRemoteEnvs: { [key: string]: string };
+  customizedRemoteEnvs: { [name: string]: string };
+}
+
+export const envFileName = (envName: string): string => `env.teamsfx.${envName}`;
+export const envFilePath = (envName: string, folder: string): string =>
```
IvanJobs, line 18: Nit: Would it be better to use path type instead of string for the `folder` parameter?

packages/fx-core/src/plugins/resource/frontend/env.ts
```diff
@@ -0,0 +1,89 @@
...
+
+export const EnvKeys = Object.freeze({
+  FuncEndpoint: "REACT_APP_FUNC_ENDPOINT",
+  FuncName: "REACT_APP_FUNC_NAME",
+  RuntimeEndpoint: "REACT_APP_TEAMSFX_ENDPOINT",
+  StartLoginPage: "REACT_APP_START_LOGIN_PAGE_URL",
+  ClientID: "REACT_APP_CLIENT_ID",
+});
+
+export function initEnvs(): RemoteEnvs {
```
IvanJobs, line 29: Nit: naming refine, it looks to me that the initEnvs doesn't do any action, so it would be better not to use the verb `init`.
Since it just return a default config object, how about naming it, for example, defaultEnvsConfig, initialEnvsConfig, even emptyEnvsConfig?

packages/fx-core/src/plugins/resource/frontend/env.ts
```diff
@@ -0,0 +1,89 @@
...
+
+export async function saveEnvFile(envPath: string, envs: RemoteEnvs): Promise<void> {
+  try {
+    const configs = await loadEnvFile(envPath);
+    const newConfigs: RemoteEnvs = {
+      teamsfxRemoteEnvs: { ...configs.teamsfxRemoteEnvs, ...envs.teamsfxRemoteEnvs },
+      customizedRemoteEnvs: { ...configs.customizedRemoteEnvs, ...envs.customizedRemoteEnvs },
+    };
+
+    if (JSON.stringify(newConfigs) === JSON.stringify(configs)) {
```
IvanJobs, line 63: Same concern as above - it's better to implement a util method to check if two RemoteEnvs instances are value-wise equal.

packages/fx-core/src/plugins/resource/frontend/env.ts
hund030: Good catch! Will update.

IvanJobs (approved):
1. To reduce distraction, better to remove .env when by env `teamsfx.${env-name}.env` files are used.
2. JSON compare is non-deterministic, better to make it deterministic by implementing a compare method targeting interface RemoteEnvs.

Suppose smoke tests are passed, looks acceptable to me.

packages/fx-core/src/plugins/resource/frontend/utils.ts
```diff
@@ -76,18 +80,22 @@ export class Utils {
...
+        (error, standardOutput, stderr) => {
+          Logger.debug(standardOutput);
+          if (error) {
+            Logger.error(`Failed to run command: "${command}".`);
```
IvanJobs, line 89: Nit: combine L89 + L93?

IvanJobs (approved):
lgtm

PR #3317, opened 2021-12-10
fix(bot): query either bot exist should not throw err

YitongFeng-git:


YitongFeng-git:
E2E Test: [describe("Provision", function () {](https://github.com/OfficeDev/TeamsFx/blob/e93691f6885d5c2bcaf78630850b3118aa9cf9ba/packages/cli/tests/e2e/bot/ProvisionResourceProjectWithNewBot.tests.ts#L28)

packages/fx-core/src/plugins/resource/bot/utils/retryHandler.ts
```diff
@@ -13,7 +13,8 @@ export class RetryHandler {
         return response;
       } catch (e) {
         if (retries <= 0) {
-          throw e;
+          if (ignoreError) break;
```
eriolchan, line 16: I would like as below to let it retry even ignore error
```
if (!ignoreError) {
  throw e;
}
```

YitongFeng-git: only if(retries <= 0) then it will judge throw or break; anyway, i update the code to make it cleaner

eriolchan (approved):

PR #3360, opened 2021-12-15
fix: should encrypt client secret in localSettings.json

dooriya:


dooriya:
Will fix in main branch

PR #3377, opened 2021-12-16
fix(localdebug): enable portable ngrok in cli

kuojianlu:


packages/cli/src/cmds/preview/preview.ts
```diff
@@ -602,7 +606,7 @@ export default class Preview extends YargsCommand {
       undefined,
       {
         shell: true,
-        cwd: botRoot,
+        cwd: ngrokChecker.getNgrokBinFolder(),
```
kuojianlu, line 609: Please note that if we use
```typescript
{
  shell: true,
  cwd: botRoot,
  env: {
    PATH: `${ngrokChecker.getNgrokBinFolder()}${path.delimiter}${process.env.PATH ?? ""}`
  }
}
```
like that in extension, it does not work. `ngrok` will not be recognized as a shell command.

swatDong, line 609: Does cwd work for `npx ngrok`? For npx 6 / 7 / 8.

swatDong: Are we using `npx ngrok` or just `ngrok`? I suppose `Path` works for `npx`.

kuojianlu: We are using `npx ngrok`, and it does not work. I think it is because `PATH` will be injected into the ngrok process, but not used to resolve ngrok command.

kuojianlu: As we discussed offline, handle `PATH` when merging process env.

packages/cli/src/cmds/preview/commonUtils.ts
```diff
@@ -340,7 +340,16 @@ export function mergeProcessEnv(
   }
   const result = Object.assign({}, process.env);
   for (const key of Object.keys(env)) {
-    result[key] = env[key];
+    if (key === "PATH") {
+      for (const k of Object.keys(result)) {
+        if (k.toUpperCase() === "PATH") {
```
a1exwang, line 345: Linux and macOS environment variable name are case sensitive, so maybe it is better to only `toUpperCase()` on Windows?

kuojianlu: `toUpperCase()` is just for match, we respect the original name, see this line
```typescript
result[k] = `${env[key]}${path.delimiter}${result[k]}`;
```

kuojianlu: But I did not handle the case when there is no `PATH`, `Path`, etc. Let me add it.

a1exwang: Worse case is contaminating users environment variable named `path` or `Path` or something like that in ngrok process. As long as ngrok does not read environment variables like `Path` or `path`. It is fine.

packages/cli/src/cmds/preview/commonUtils.ts
```diff
@@ -340,7 +340,16 @@ export function mergeProcessEnv(
   }
   const result = Object.assign({}, process.env);
   for (const key of Object.keys(env)) {
-    result[key] = env[key];
+    if (key === "PATH") {
+      const pathKey = Object.keys(result).find((value) => value.toUpperCase() === "PATH");
+      if (pathKey !== undefined) {
+        result[pathKey] = `${env[key]}${path.delimiter}${result[pathKey]}`;
```
swatDong, line 346: It's wired to append here (this func `mergeProcessEnv` is to add or override envs)... just override, and let caller to input the full value.

Or, just apply the same logic as vscode.

kuojianlu: You are right, it is wired..
Updated to use the same logic as vscode: https://github.com/microsoft/vscode/blob/main/src/vs/workbench/contrib/terminal/common/terminalEnvironment.ts#L19.

swatDong (approved):

PR #3390, opened 2021-12-17
chore: ci scan password pattern

wenytang-ms:


.github/detect/regexes.json
```diff
@@ -0,0 +1,48 @@
+{
+    "Slack Token": "(xox[p|b|o|a]-[0-9]{12}-[0-9]{12}-[0-9]{12}-[a-z0-9]{32})",
+    "Generic Private Key POST Encapsulation Boundary": " PRIVATE KEY-----",
+    "Generic Private Key Block POST Encapsulation Boundary": " PRIVATE KEY BLOCK-----",
+    "Private key": "-----BEGIN PRIVATE KEY-----",
+    "RSA private key": "-----BEGIN RSA PRIVATE KEY-----",
```
swatDong, line 6: Just noticed this PR and would like to callout - we do have private key in localhost certificate unit test. Any action to do to exclude such unit test?

wenytang-ms: You can put these exception files in excludes.txt

LongOddCode: How about use annotation or comment around target code to evade inspection

wenytang-ms: It’s better to use documents for explanation than annotation/comments, especially when introducing anything about passwords.

LongOddCode (approved):

Issue #2594, opened 2021-10-13
VSCode Teams Toolkit extension won't load for a newly created project


arsdragonfly:
**Describe the bug**
VSCode Teams Toolkit extension won't load for a newly created project

**To Reproduce**
1. open a VSCode WSL window
2. create a new teams app with teams toolkit
3. launch Debug (Edge)
4. an error "command `fx-extension.validate-dependencies` not found" appears

**Expected behavior**
A debug window would launch.

**Screenshots**
![image](https://user-images.githubusercontent.com/4067473/137082029-4cc4c28f-9f72-4cae-b4c2-14d85450cb3f.png)

**VS Code Extension Information (please complete the following information):**
 - OS: WSL on Windows 11
 - Version v2.7.0 (preview)

kuojianlu:
@arsdragonfly Sometimes the toolkit may be not well activated. I suppose reloading VSCode window may solve this issue.
1. In this project, select and run `Developer: Reload Window` from command palette
2. Wait for a while util the toolkit is well activated
3. Launch Debug (Edge)

jayzhang:
@arsdragonfly: 
For short term, please follow what @kuojianlu said, don't trigger F5 before extension activated. 
For long term, @1openwindow , can you follow this issue by optimize the activation time for extension?

1openwindow:
@kuojianlu, @jayzhang , I think the root cause is not the activation time. The end users still can press F5 before teams toolkit activated. 
The right solution is monitor the onCommand trigger point. Please check this: https://github.com/OfficeDev/TeamsFx/blob/6835acba1cae9b1a83515019b309fbce148e218d/packages/vscode-extension/package.json#L62

PR #3393, opened 2021-12-17
fix: block debug before extension activated

kuojianlu:
To fix: (1) https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/12361074 and (2) #2594.

(1) can be stably reproduced, but (2) is hard to be reproduced.

From my test, task provider will not be called before activation finished. (1) can be definitely solved and we expect (2) can also be solved. See references below for more details.

References:
https://github.com/platformio/platformio-vscode-ide/issues/89#issuecomment-392447277

### DIFF
--- a/packages/vscode-extension/package.json
+++ b/packages/vscode-extension/package.json
@@ -58,8 +58,6 @@
     "onCommand:fx-extension.openAppManagement",
     "onCommand:fx-extension.openBotManagement",
     "onCommand:fx-extension.openReportIssues",
-    "onCommand:fx-extension.validate-dependencies",
-    "onCommand:fx-extension.pre-debug-check",
     "onCommand:fx-extension.migrateV1Project",
     "onCommand:workbench.action.tasks.runTask",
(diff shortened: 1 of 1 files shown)


swatDong:
Really coooooooooool solution.
Since `onCommand:workbench.action.tasks.runTask` works, can we remove `onCommand:fx-extension.validate-dependencies` and so on from the active event list? Just to keep the list cleaner to be better maintainable.

kuojianlu:
> Really coooooooooool solution. Since `onCommand:workbench.action.tasks.runTask` works, can we remove `onCommand:fx-extension.validate-dependencies` and so on from the active event list? Just to keep the list cleaner to be better maintainable.

Sure, it is safe to remove `onCommand:fx-extension.validate-dependencies` and `onCommand:fx-extension.pre-debug-check`.

swatDong (dismissed):

1openwindow (approved):

PR #3398, opened 2021-12-20
refactor: update getQuestionForScaffolding API and update question model

HuihuiWu-Microsoft:
1. revert API change from  
getQuestionsForScaffold?: (ctx: Context_2, inputs: InputsWithProjectPath) => Promise<Result<QTreeNode | QTreeNode[] | undefined, FxError>>
 to 
getQuestionsForScaffold?: (ctx: Context_2, inputs: InputsWithProjectPath) => Promise<Result<QTreeNode | undefined, FxError>>
2. Add another function node for solution in which set the selected solution and remove harded coded logic in core to get solution
3. Change plugin capability node condition according to input capabilities


packages/api/src/qm/visitor.ts
```diff
@@ -305,8 +305,10 @@ export async function traverse(
         continue; //ignore the following steps
       } else {
         //success or skip
-        question.value = inputResult.result;
-        inputs[question.name] = question.value;
+        if (question.type != "func") {
```
HuihuiWu-Microsoft, line 308: No need to set value for func node since there's no input for this kind of question

jayzhang (approved):

PR #3404, opened 2021-12-20
chore: need to be in TeamsFx project before asking for survey

chagong:
chore: need to be in TeamsFx project before asking for survey

https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/12816244


HuihuiWu-Microsoft (approved):
LGTM. Just one question, is the AB test still running? I see there's TreatmentVariable in the logics

Alive-Fish (approved):

PR #3405, opened 2021-12-20
refactor(localdebug): Prepare LocalEnvManager in fx-core

swatDong:
Prepare `LocalEnvManager` in fx-core:
- Add basic `getSettings` utilities with retry logic
- Define public APIs used by local debug
- UT added
- Code owner updated

Details: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/12864643/


packages/fx-core/src/common/local/localEnvManager.ts
```diff
@@ -0,0 +1,74 @@
...
+  public async getProgrammingLanguage() {}
+
+  public async getSkipNgrokConfig() {}
+
+  private async getLocalSettings(projectPath: string): Promise<Json | undefined> {
+    const projectSettings = await this.getProjectSettings(projectPath);
+    const localSettingsProvider = new LocalSettingsProvider(projectPath);
+    const crypto = new LocalCrypto(projectSettings.projectId);
+    return await this.retry(async () => {
+      return await localSettingsProvider.loadV2(crypto);
```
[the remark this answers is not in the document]
swatDong: Right, no V1, no `multiEnv:false` (Even no remote... but that will be removed after this feature)

kuojianlu (approved):

packages/fx-core/src/common/local/localEnvManager.ts
```diff
@@ -0,0 +1,88 @@
...
+    });
+  }
+
+  private async getProjectSettings(projectPath: string): Promise<ProjectSettings> {
```
XiaofuHuang, line 43: Do we need to use loadProjectSettings in packages\fx-core\src\core\middleware\projectSettingsLoader.ts?
There are some compatible logic there.

XiaofuHuang (approved):

packages/fx-core/src/common/local/localEnvManager.ts
swatDong: No, I'm trying to make `LocalEnvMananger` as independent as possible from `core`. `projectSettingsLoader` is tightly bound on core's lifecycle and context so hard to be used directly.
BTW, `LocalEnvManager` is read-only on project-level settings, so simple `readJson` is enough, no need to consider compatible logic.

XiaofuHuang: Got it, thx!

PR #3407, opened 2021-12-20
chore(sql): retry adding firewall

xzf0587:
1. retry adding firewall if the ip failed to access sql server.
2. update ut.


packages/fx-core/tests/plugins/resource/sql/unit/index.test.ts
```diff
@@ -151,26 +150,6 @@ describe("sqlPlugin", () => {
     chai.assert.isTrue(postProvisionResult.isOk());
   });
 
-  it("postProvision with skipAddingUser", async function () {
```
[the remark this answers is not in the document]
xzf0587: It is a duplicated test case. We have this case in the  "skipAddingUser.test.ts".

packages/fx-core/src/plugins/resource/sql/managementClient.ts
```diff
@@ -217,11 +217,12 @@ export class ManagementClient {
       startIpAddress: startIp,
       endIpAddress: endIp,
     };
+    const ruleName = Constants.firewall.localRule + index;
```
adashen, line 220: Can we update the rule instead of add multiple rules? Since the api is createOrUpdate. What is the reason to add multiple rules?

xzf0587: The old firewall rule is kept as the local IP may change in 2 address back and forth. It could successfully access SQL if the IP change back to the old one as well.

packages/fx-core/src/plugins/resource/sql/config.ts
```diff
@@ -20,4 +20,5 @@ export class SqlConfig {
   identity = "";
   existSql = false;
   skipAddingUser = false;
+  retryAddUser = 0;
```
adashen, line 23: This is not the retry time configuration right? If not, we should move it to the logic code part. Seems to be a temp variable.

packages/fx-core/src/plugins/resource/sql/plugin.ts
```diff
@@ -256,6 +248,29 @@ export class SqlPluginImpl {
...
+      try {
+        await sqlClient.addDatabaseUser();
+        return;
+      } catch (error) {
+        if (!SqlClient.isFireWallError(error?.innerError) || this.config.retryAddUser >= 3) {
+          throw error;
+        } else {
+          this.config.retryAddUser++;
+          ctx.logProvider?.warning(
+            `[${Constants.pluginName}] retry adding firewall rule to add database user [${this.config.retryAddUser}]`
```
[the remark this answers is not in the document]
xzf0587: it is fixed, thanks.

packages/fx-core/src/plugins/resource/sql/managementClient.ts
```diff
@@ -201,7 +201,7 @@ export class ManagementClient {
     }
   }
 
-  async addLocalFirewallRule(): Promise<void> {
+  async addLocalFirewallRule(index: number): Promise<void> {
```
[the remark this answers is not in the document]
xzf0587: it is fixed, thanks.

packages/fx-core/src/plugins/resource/sql/config.ts
xzf0587: It is fixed. Thanks.

packages/fx-core/src/plugins/resource/sql/plugin.ts
```diff
@@ -254,6 +245,30 @@ export class SqlPluginImpl {
...
+    sqlClient: SqlClient,
+    managementClient: ManagementClient
+  ): Promise<void> {
+    let retryAddUser = 0;
+    while (true) {
+      try {
+        await sqlClient.addDatabaseUser();
+        return;
+      } catch (error) {
+        if (!SqlClient.isFireWallError(error?.innerError) || retryAddUser >= 3) {
```
adashen, line 259: avoid magic number 3?

packages/fx-core/src/plugins/resource/sql/managementClient.ts
```diff
@@ -217,13 +218,15 @@ export class ManagementClient {
       startIpAddress: startIp,
       endIpAddress: endIp,
     };
+    const ruleName = Constants.firewall.localRule + this.nextIndex;
```
adashen, line 221: for the rule name can we have a extracted method to get rule name so that it could be shared between add and delete to avoid future regression?

packages/fx-core/src/plugins/resource/sql/managementClient.ts
```diff
@@ -11,6 +11,7 @@ export class ManagementClient {
   client: SqlManagementClient;
   config: SqlConfig;
   ctx: PluginContext;
+  nextIndex = 0;
```
blackchoey, line 14: I may suggest better variable naming here, e.g. totalFirewallRuleCount or something else

packages/fx-core/src/plugins/resource/sql/plugin.ts
```diff
@@ -251,19 +250,23 @@ export class SqlPluginImpl {
     sqlClient: SqlClient,
     managementClient: ManagementClient
   ): Promise<void> {
+    let retryAddUser = 0;
```
blackchoey, line 253: I may suggest better variable naming here, e.g. retryCount or something else

packages/fx-core/src/plugins/resource/sql/managementClient.ts
xzf0587: It is fixed. Thanks.

packages/fx-core/src/plugins/resource/sql/plugin.ts
xzf0587: It is fixed. Thanks.

blackchoey (approved):

adashen (approved):

packages/fx-core/src/plugins/resource/sql/managementClient.ts
xzf0587: OK.
Added a specific method getRuleName to get the rule name. Thanks.

PR #3418, opened 2021-12-21
refactor(localdebug): user tasks

swatDong:
Move several user tasks into `LocalEnvManager`. UT added.
(`getLocalDebugEnvs` is a bit complex so will be in another PR)


packages/fx-core/src/common/local/localEnvManager.ts
```diff
@@ -40,6 +51,14 @@ class LocalEnvManager {
     });
   }
 
+  // Load local settings without encryption
+  private async getRawLocalSettings(projectPath: string): Promise<Json | undefined> {
```
kuojianlu, line 55: Some methods may be called at the same moment, such as `getProgrammingLanguage`, `getProgrammingLanguage`. Is there a way to reduce the times of reading file?

swatDong: Hard to do so in `LocalEnvManager` since it's stateless.
Adjusted parameters to let caller to decide whether to read file once or multiple times.
Moved some user tasks logic to v2 api to apply v2 context.

XiaofuHuang (dismissed):

packages/fx-core/src/common/local/localEnvManager.ts
```diff
@@ -19,28 +19,43 @@ import { LocalCrypto } from "../../core/crypto";
...
+  }
 
-  public async getSkipNgrokConfig() {}
+  public getSkipNgrokConfig(rawLocalSettings: Json | undefined): boolean {
+    return (rawLocalSettings?.bot?.skipNgrok as boolean) === true;
+  }
 
-  private async getLocalSettings(projectPath: string): Promise<Json | undefined> {
-    const projectSettings = await this.getProjectSettings(projectPath);
+  public async getLocalSettings(projectPath: string, projectId: string): Promise<Json | undefined> {
```
kuojianlu, line 42: Why not use `crypto` instead of `projectId` as parameter, then `getLocalSettings` and `getRawLocalSettings` can be merged into one function?

swatDong: Good catch. Merged into one func but expose `cryptoOption` as parameter. (`CryptoProvider` is core internal so do not want to be exported to cli/vsc)

kuojianlu (approved):

PR #3423, opened 2021-12-21
refactor(localdebug): move provisionlocalresource from localdebug plugin to solution

XiaofuHuang:
Detail: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/12863187


swatDong (approved):

packages/fx-core/src/plugins/resource/localdebug/index.ts
```diff
@@ -75,122 +67,10 @@ export class LocalDebugPlugin implements Plugin {
...
-            ctx.localSettings?.bot?.set(LocalSettingsBotKeys.BotEndpoint, ngrokHttpUrl);
-            ctx.localSettings?.bot?.set(LocalSettingsBotKeys.BotDomain, ngrokHttpUrl.slice(8));
-          }
-        }
-      }
-    }
-
-    TelemetryUtils.sendSuccessEvent(TelemetryEventName.localDebug, telemetryProperties);
-    return ok(undefined);
+    return ok(Void);
```
swatDong, line 70: Curious - what's the difference between `ok(undefined)` and `ok(Void)`? Any behavior change?

XiaofuHuang: Here, they are no difference, I just want to follow the API v2 design.

swatDong (approved):

1yefuwang1 (approved):
LGTM

PR #3432, opened 2021-12-22
chore(solution): add notification message for error

xzf0587:
1. add optional property notificationMessage in error.

### DIFF
--- a/packages/vscode-extension/src/handlers.ts
+++ b/packages/vscode-extension/src/handlers.ts
@@ -1628,6 +1628,8 @@ export function cmdHdlDisposeTreeView() {
 }
 
 export async function showError(e: UserError | SystemError) {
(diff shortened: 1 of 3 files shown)


packages/vscode-extension/src/handlers.ts
```diff
@@ -1645,7 +1647,7 @@ export async function showError(e: UserError | SystemError) {
       },
     };
 
-    const button = await window.showErrorMessage(`[${errorCode}]: ${e.message}`, help);
```
jayzhang, line 1648: can we keep the current behavior by using 'message' as notification if notificationMessage is undefined?

jayzhang (approved):

jayzhang (dismissed):

jayzhang (approved):

HuihuiWu-Microsoft (approved):

PR #3435, opened 2021-12-22
feat: add a global setting 'interactive' for all commands

Alive-Fish:
E2E TEST: https://github.com/OfficeDev/TeamsFx/actions/runs/1610195003

CLI changes:
1. Add `teamsfx config set interactive true/false` to set it globally.
2. Add the `--interactive` argument for all commands which will overwrite the global `interactive` config.
3. When users open the `interactive` flag, CLI will ignore all the inputs that may be used by `fx-core` and output a log for user.
4. When users close the `interactive` flag, CLI will not let user answer any question when executing the command.
5. The `CI_ENABLED` flag is only used internally just for E2E tests. If it opens, it will set `interactive` to `false`.
image
image
image
image


chagong (approved):

MuyangAmigo (approved):
LGTM

packages/cli/src/index.ts
```diff
@@ -42,6 +42,10 @@ export async function start() {
       boolean: true,
       default: false,
     })
+    .options("interactive", {
+      description: "Run the command interactively.",
```
IvanJobs, line 46: Nit: An switch to toggle if the command is running interactively or not?

packages/cli/src/yargsCommand.ts
```diff
@@ -34,12 +35,26 @@ export abstract class YargsCommand {
...
+  public params: { [_: string]: Options } = {};
+
   /**
    * builds the command using supplied yargs handle.
    * @param yargs the yargs handle
    */
   abstract builder(yargs: Argv): Argv<any>;
 
+  /**
+   * before running command, some command may modify the arguments that users input.
```
IvanJobs, line 50: Nit: get the context.

IvanJobs (approved):
lgtm for the description.

PR #3442, opened 2021-12-23
chore: git commit check password pattern

wenytang-ms:


packages/eslint-config/index.js
```diff
@@ -34,6 +35,13 @@ module.exports = {
         ],
         "import/no-unresolved": [
             "warn"
-        ]
+        ],
+        "no-secrets/no-secrets": [
+            "error",
+            { "additionalRegexes": { 
+                "Basic Auth": "Authorization: Basic [A-Za-z0-9+/=]*",
+                "Common Pattern": "^(?=.*?[A-Z])(?=.*?[a-z])(?=.*?[0-9])(?=.*?[#?!@$%^&*-]).{8,}$"
```
adashen, line 43: the special char may also include ~ | and etc. We may check more.

wenytang-ms: sure

packages/eslint-config/index.js
```diff
@@ -34,6 +35,13 @@ module.exports = {
         ],
         "import/no-unresolved": [
             "warn"
-        ]
+        ],
+        "no-secrets/no-secrets": [
+            "warn",
+            { "additionalRegexes": { 
+                "Basic Auth": "Authorization: Basic [A-Za-z0-9+/=]*",
+                "Common Pattern": "^(?=.*?[A-Z])(?=.*?[a-z])(?=.*?[0-9])(?=.*?[#?!@$%^&*-~|]).{8,}$"
```
blackchoey, line 43: not blocking this PR, it may be better if we can have a document (maybe a readme in eslint-config folder?) talking about the common secret pattern (as well as other lint rules). It may be hard for others to understand the common pattern.

jayzhang (approved):

PR #3451, opened 2021-12-23
chore(arm): customize notification message for error

xzf0587:
customize notification message to update error handling.

### DIFF
--- a/packages/fx-core/src/plugins/solution/fx-solution/arm.ts
+++ b/packages/fx-core/src/plugins/solution/fx-solution/arm.ts
@@ -17,6 +17,7 @@ import {
   returnSystemError,
   returnUserError,
   SolutionContext,
+  UserError,
   v2,
   v3,
 } from "@microsoft/teamsfx-api";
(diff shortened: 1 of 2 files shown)


xzf0587:
The error looks like
image

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -1410,19 +1405,17 @@ async function getDeploymentError(
   return deploymentError;
 }
 
-function formattedDeploymentName(failedDeployments: string[]): FxError {
+function getNotificationMessage(deploymentError: any, deploymentName: string): string {
```
adashen, line 1408: do we need to handle the null/empty case of deploymentError and deploymentName. Per last modification if we provide notificationMessage, if it is empty will it also override the errormessage?

xzf0587: The code in line 505 handle the null/empty case of deploymentError. 
And the deploymentName is also not empty as we use it to fetch deploymentError.
Actually, only deploymentError exists, we will override the notificationMessage.

adashen (approved):

LongOddCode (approved):

PR #3453, opened 2021-12-24
perf: add BicepVersion and SchemaVersion to ArmTemplateResult

wenytang-ms:


packages/fx-core/tests/plugins/resource/aad/unit/generateArmTemplates.test.ts
```diff
@@ -54,6 +54,7 @@ describe("AadGenerateArmTemplates", () => {
       );
       chai.assert.isUndefined(armTemplateResult.Configuration);
       chai.assert.isUndefined(armTemplateResult.Provision);
+      chai.assert.strictEqual(armTemplateResult.SchemaVersion, "0.1.0");
```
KennethBWSong, line 57: reference constant instead of using "0.1.0"?

XiaofuHuang (approved):

packages/fx-core/tests/plugins/resource/aad/unit/generateArmTemplates.test.ts
wenytang-ms: test case use the value direct is better.

formulahendry:
Just curious. Why this pr is 'perf'? 

By checking this https://dev.to/i5han3/git-commit-message-convention-that-you-can-follow-1709, it says `perf: A code that improves performance`.

packages/fx-core/src/plugins/resource/bot/constants.ts
```diff
@@ -276,3 +276,8 @@ export class BotBicep {
   static readonly hostName: string = "provisionOutputs.botOutput.value.validDomain";
   static readonly webAppEndpoint: string = "provisionOutputs.botOutputs.value.botWebAppEndpoint";
 }
+
+export class ArmTemplateVersion {
+  static readonly BicepVersion = "0.4.1008";
```
formulahendry, line 281: 1. where is this number from?
2. could we put this in a central place, so that we would not change it in several files next time.

packages/fx-core/src/plugins/resource/bot/constants.ts
```diff
@@ -276,3 +276,8 @@ export class BotBicep {
...
+  static readonly SchemaVersion = "0.1.0";
```
formulahendry, line 282: Same for SchemaVersion, could we put this in a central place, so that we would not change it in several files next time.

PR #3455, opened 2021-12-24
chore: scaffold from local code for debug purpose

hund030:
How to use this action:
1. Set DEBUG_TEMPLATE=true to your environment variables.
1. Add your changes in templates source code.
2. cd to vscode-extension folder.
3. F5 to local debug and create new project.
4. The `FetchTemplateZipFromSourceCode` action will get template from the source code that you just changed.

* `FetchTemplatesUrlWithTag`, `FetchTemplatesZipFromUrl`, `FetchTemplateZipFromLocal`, these actions are skipped.
* Since bot's and message extension's templates are generated during CD pipeline, we can't debug those code with this action.

### DIFF
--- a/packages/fx-core/src/common/templatesActions.ts
+++ b/packages/fx-core/src/common/templatesActions.ts
@@ -49,11 +49,54 @@ const defaultTimeoutInMs = 10000;
 const missKeyErrorInfo = (key: string) => `Missing ${key} in template action.`;
 
 export enum ScaffoldActionName {
+  FetchTemplateZipFromSourceCode = "FetchTemplateZipFromSourceCode",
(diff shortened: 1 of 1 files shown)


packages/fx-core/src/common/templatesActions.ts
```diff
@@ -49,12 +49,55 @@ const defaultTimeoutInMs = 10000;
...
+
+    if (!isDebugMode()) {
+      return;
+    }
+
+    if (context.zip) {
+      return;
+    }
+
+    if (!context.group || !context.lang || !context.scenario) {
```
eriolchan, line 76: Can we add more comment on why it is skipped by this condition?

eriolchan (approved):

PR #3464, opened 2021-12-24
chore: add deps checker manager

meifans:


packages/fx-core/package.json
```diff
@@ -79,6 +79,7 @@
     "md5": "^2.3.0",
     "mime": "^2.5.2",
     "mustache": "^4.2.0",
+    "neverthrow": "^3.2.0",
```
swatDong, line 82: `import { Result } from "@microsoft/teamsfx-api"` so no need to change package.json.

meifans: Cool, will add it

meifans: Added

swatDong (approved):

PR #3474, opened 2021-12-27
fix: spfx local debug concurrency error

HuihuiWu-Microsoft:
ADO:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_sprints/taskboard/Teams%20Extensibility%20E2E%20Team/Microsoft%20Teams%20Extensibility/CY21-12.3?workitem=12873660
Root cause:
When F5, both pre debug check and validate dependency handlers will be executed. The two handlers will call APIs of core which will cause concurrency error. So I change the order to 'sequence' instead of 'parallel' to avoid the lock fail issue.
E2E TEST: only for extension


swatDong (approved):
Remember to update SPFx samples as well.

PR #3492, opened 2021-12-28
chore: ignore linesCovered property in ut report

LongOddCode:


IvanJobs (approved):
Assume the fix is aligned with dev.

PR #3499, opened 2021-12-29
refactor(localdebug): replace usertasks with LocalEnvManager

swatDong:
Replace user tasks with `LocalEnvManager` and remove unused common utils.
Two pending ones:
- `getSkipNgrokConfig` is still used by deps checker and will be covered in deps checker refactor.
- `getLocalDebugConfig` is still used by tree view and will be covered in another PR.

### DIFF
--- a/packages/vscode-extension/src/debug/teamsfxTaskProvider.ts
+++ b/packages/vscode-extension/src/debug/teamsfxTaskProvider.ts
@@ -6,11 +6,14 @@ import * as vscode from "vscode";
 
 import * as constants from "./constants";
 import * as commonUtils from "./commonUtils";
-import { ProductName, VsCodeEnv } from "@microsoft/teamsfx-api";
+import { Json, ProductName, ProjectSettings, VsCodeEnv } from "@microsoft/teamsfx-api";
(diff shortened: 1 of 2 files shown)


dooriya (dismissed):

packages/vscode-extension/src/debug/teamsfxTaskProvider.ts
```diff
@@ -28,8 +31,27 @@ export class TeamsfxTaskProvider implements vscode.TaskProvider {
...
+        localSettings = await localEnvManager.getLocalSettings(workspacePath, {
+          projectId: projectSettings.projectId,
+        });
+        localEnv = await localEnvManager.getLocalDebugEnvs(
+          workspacePath,
+          projectSettings,
+          localSettings
+        );
+      } catch (error: any) {
+        await showError(error);
```
kuojianlu, line 50: I remember `await showError` will block the thread since in `showError` it uses `await vscode.window.showErrorMessage`.

swatDong: Right... removed `await`.

kuojianlu (approved):

dooriya (approved):

PR #3519, opened 2022-01-04
chore(arm): fix poll status bug

xzf0587:
1. update poll status to get sub-operations when resource type is deployments
2. add error telemetry for poll status


packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -228,23 +231,26 @@ export async function pollDeploymentStatus(deployCtx: DeployContext) {
...
-              const subOperations = await client.deploymentOperations.list(
-                operation.resourceGroupName,
-                operation.resourceName
-              );
-              subOperations.forEach((sub) => {
-                const subOperation = getRequiredOperation(sub, deployCtx);
-                if (subOperation) {
-                  currentStatus[subOperation.resourceName] = subOperation.status;
+              // get sub operations when resource type is deployments.
+              if (operation.resourceType === ConstantString.DeploymentResourceType) {
```
adashen, line 235: if the resouceType is not deployment, we will not collect infomation?

xzf0587: When it is not a deployment, it has no sub operation. But we will show it in output as well.

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -261,11 +265,20 @@ export async function pollDeploymentStatus(deployCtx: DeployContext) {
...
+        );
       }
-      deployCtx.logProvider?.warning(
-        `[${PluginDisplayName.Solution}] ${deployCtx.deploymentName} -> waiting to get deplomyment status [${tryCount}]`
+      const pollError = returnSystemError(
+        error,
+        SolutionSource,
+        SolutionError.FailedToPollArmDeploymentStatus
+      );
+      sendErrorTelemetryThenReturnError(
```
blackchoey, line 278: we may just send the error telemetry when exceeds max retry count to reduce possible noises.

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -67,6 +67,8 @@ const bicepOrchestrationConfigFileName = "config.bicep";
 const templatesFolder = "./templates/azure";
 const configsFolder = `.${ConfigFolderName}/configs`;
 const parameterFileNameTemplate = `azure.parameters.${EnvNamePlaceholder}.json`;
+const pollWaitSeconds = 10;
+const MaxLogTimes = 4;
```
blackchoey, line 71: `m`axLogTimes to make the variable name aligned
by the way, can we have a better name for it? not sure whether `maxRetryTimes` suits for it.

packages/fx-core/src/plugins/solution/fx-solution/arm.ts

packages/fx-core/src/plugins/solution/fx-solution/arm.ts

blackchoey (approved):

PR #3520, opened 2022-01-04
chore: add cli deps checker

meifans:


packages/cli/src/cmds/preview/depsChecker/cliUtils.ts
```diff
@@ -0,0 +1,32 @@
+import UI from "../../../userInteraction";
```
kuojianlu, line 1: Add header.

meifans: Added

swatDong (approved):

PR #3528, opened 2022-01-04
chore(sdk): add IoC container

tecton:
Not exported now to keep it private until other components are ready.


packages/sdk/src/container/teamsfx.ts
```diff
@@ -0,0 +1,84 @@
...
+      if (method instanceof Function && !ComponentApiNames.includes(key) && !key.startsWith("_")) {
+        (this as Mapping)[key] = (...args: any) => {
+          const instance = this.resolve(component.name);
+          return method.call(instance, ...args);
+        };
+      }
+    });
+  }
+
+  public resolve(componentName: string, version?: string): unknown {
```
blackchoey, line 42: what's the purpose of version here?

packages/sdk/src/container/types.ts
```diff
@@ -0,0 +1,32 @@
...
+export type ComponentOption = Record<string, unknown>;
+
+export interface ComponentContainer {
+  resolve(componentName: string, version?: string): unknown;
+}
+
+export interface Component extends Mapping {
+  name: string;
+  version: string;
+  initialize(container: ComponentContainer, logger: Logger): void;
```
blackchoey, line 19: what is our solution if initialize requires additional parameters (inputs)?

blackchoey (commented):
I saw you created our own dependency injection implementation, is there any existing dependency injection library we can use?

packages/sdk/src/container/teamsfx.ts
```diff
@@ -0,0 +1,84 @@
...
+    const component = this.registry.get(componentName);
+    if (!component) {
+      throw new Error();
```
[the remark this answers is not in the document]
tecton: I'll add the error messages.

packages/sdk/src/container/teamsfx.ts
```diff
@@ -0,0 +1,84 @@
...
+  private addComponent(component: T): void {
+    this.registry.set(component.name, component);
+    const logger = new InternalLogger(component.name, this.logLevel);
+    this.loggers.set(component.name, logger);
+
+    const functionKeys = Object.getOwnPropertyNames(Object.getPrototypeOf(component));
+    functionKeys.forEach((key: string) => {
+      const method = component[key];
+      if (method instanceof Function && !ComponentApiNames.includes(key) && !key.startsWith("_")) {
+        (this as Mapping)[key] = (...args: any) => {
```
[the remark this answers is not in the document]
tecton: 2 same method names will be conflict and should be avoided.
The private methods will be cloned too but users cannot call them because they are not included in type definition.

packages/sdk/src/container/teamsfx.ts
tecton: The version is to avoid getting a component instance that has breaking changes.

packages/sdk/src/container/types.ts
tecton: I think we can add new function to ComponentContainer interface if that's a common scenario. The `logger` parameter is to ensure that the log message has the component name.

tecton:
> I saw you created our own dependency injection implementation, is there any existing dependency injection library we can use?

The implementation here is more like service locator pattern. I don't think existing DI libraries match our scenario because the component needs some customized options in constructor, so they cannot be easily created using parameter injection. And the singleton will introduce global states.

packages/sdk/src/container/teamsfx.ts
tecton: Thanks, removed.

blackchoey (approved):
Please help follow up on the function name conflict problem in later PRs

PR #3530, opened 2022-01-04
perf(local debug): update local debug for removing simple auth

KennethBWSong:


hund030 (approved):

packages/cli/src/cmds/preview/preview.ts
```diff
@@ -608,13 +611,14 @@ export default class Preview extends YargsCommand {
...
     const dotnetExecPath = await dotnetChecker.getDotnetExecPath();
-    const authStartTask = includeFrontend
-      ? this.prepareTask(
-          TaskDefinition.authStart(dotnetExecPath, commonUtils.getAuthServicePath(localEnv)),
-          constants.authStartStartMessage,
-          commonUtils.getAuthLocalEnv(localEnv)
-        )
-      : undefined;
+    const authStartTask =
+      includeFrontend && includeAuth
```
swatDong, line 615: `includeAuth` has two meanings:
1. Need AAD and auth functionality
2. Need SimpleAuth service.

I believe the goal is to remove 2) only, so better to directly use `includeSimpleAuth`.

swatDong:
Any `tasks.json` change? `includeAuth` is called by CLI/Core/VSC, but it seems this PR only updates CLI.

packages/cli/src/cmds/preview/preview.ts
KennethBWSong: Updated. Thx!

KennethBWSong:
> Any `tasks.json` change? `includeAuth` is called by CLI/Core/VSC, but it seems this PR only updates CLI.

@swatDong In extension there is already check on simple auth env here:
https://github.com/OfficeDev/TeamsFx/blob/21bca6c1fee7db60cb69c4cee62d45815bb0320e/packages/vscode-extension/src/debug/teamsfxTaskProvider.ts#L78
image
And for 'tasks.json', I think we will update later since we will keep simple auth in this release. We will update this in the release that simple auth plugin is removed from template.

swatDong (dismissed):

packages/fx-core/src/common/local/projectSettingsHelper.ts
```diff
@@ -26,15 +27,15 @@ export class ProjectSettingsHelper {
...
-    const includeSimpleAuth = selectedPlugins?.some(
-      (pluginName) => pluginName === ResourcePlugins.SimpleAuth
-    );
+    return selectedPlugins?.some((pluginName) => pluginName === ResourcePlugins.Aad);
+  }
 
-    return includeAad && (!this.includeFrontend(projectSettings) || includeSimpleAuth);
+  public static includeSimpleAuth(projectSettings: ProjectSettings | undefined): boolean {
+    const includeSimpleAuth = IsSimpleAuthEnabled(projectSettings);
+    return !this.includeFrontend(projectSettings) || includeSimpleAuth;
```
blackchoey, line 38: not blocking this PR: we can keep current logic, but we'd better have a follow up task on the condition. Is it you or qianhao to follow up on this?

hund030 (dismissed):

packages/fx-core/src/common/local/projectSettingsHelper.ts
swatDong: What will the final condition be?

packages/fx-core/src/plugins/solution/fx-solution/debug/provisionLocal.ts
```diff
@@ -54,7 +55,7 @@ export async function setupLocalDebugSettings(
     frontend: includeFrontend ? "true" : "false",
     function: includeBackend ? "true" : "false",
     bot: includeBot ? "true" : "false",
-    auth: includeAuth ? "true" : "false",
+    auth: includeAAD && includeSimpleAuth ? "true" : "false",
```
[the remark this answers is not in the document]
swatDong: Local debug cares how many local services are running, so here in telemetry it means "local auth service". So it's expected to be false if simple auth not enabled.
One optimization could be `includeAAD && includeSimpleAuth` -> `includeSimpleAuth`, but not a hard requirement. I believe `includeAAD: false; includeSimpleAuth: true` is an invalid case.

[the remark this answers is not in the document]
KennethBWSong: I expect to keep the same behavior here as current, and current includeAuth should include both includeAAD and includeSimpleAuth.

packages/fx-core/src/common/local/projectSettingsHelper.ts
KennethBWSong: I think finally we will remove the includeSimpleAuth here, but it will be kept for some time. I prefer to keep this until we abandon since there may be some migration V1 problem if we update the logic here.

packages/fx-core/src/plugins/solution/fx-solution/debug/provisionLocal.ts
```diff
@@ -197,12 +199,15 @@ export async function configLocalDebugSettings(
           frontendEnvs!.teamsfxLocalEnvs[EnvKeysFrontend.Port] = "53000";
         }
 
-        if (includeAuth) {
+        if (includeAAD) {
           frontendEnvs!.teamsfxLocalEnvs[EnvKeysFrontend.TeamsFxEndpoint] = localAuthEndpoint;
```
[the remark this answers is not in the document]
KennethBWSong: Fixed, thx!

packages/fx-core/src/common/local/localSettingsHelper.ts
```diff
@@ -92,12 +93,14 @@ export async function convertToLocalEnvs(
       localEnvs[LocalEnvFrontendKeys.Port] = "53000";
     }
 
-    if (includeAuth) {
+    if (includeAAD) {
       // frontend local envs
       localEnvs[LocalEnvFrontendKeys.TeamsFxEndpoint] = localAuthEndpoint;
```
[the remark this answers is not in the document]
KennethBWSong: Fixed, thx!

swatDong (approved):

blackchoey (approved):

hund030 (approved):

packages/fx-core/src/common/local/projectSettingsHelper.ts
blackchoey: thank you, as offline discussed, we can keep current logic and wait for removing the code totally to reduce some engineering effort.

PR #3537, opened 2022-01-05
chore: clear cache when logout

xiaolang124:
1. Delete cache on failure. (Set tokenCache storage cache to empty array in case msal overwrites cache)


packages/cli/src/commonlib/cacheAccess.ts
```diff
@@ -191,6 +191,15 @@ export async function saveAccountId(accountName: string, accountId?: string) {
   }
 }
 
+export async function clearCache(accountName: string) {
```
swatDong, line 194: Does `msalTokenCache.getAllAccounts()` works for such case?
Similar to https://github.com/OfficeDev/TeamsFx/blob/dev/packages/cli/src/commonlib/codeFlowLogin.ts#L327

xiaolang124: No, I have tried ```msalTokenCache.removeAccount(account)```, the older version can only remove its own token but not the newer version token.

xiaolang124: image

swatDong (approved):

swatDong (approved):

PR #3541, opened 2022-01-05
refactor(frontend): remove unused code & rename blazor plugin

hund030:
1. rename blazor to dotnet
2. remove dotnet's provision since we are going to implement arm.
3. clean up deprecated code in frontend plugin.
4. fix relative unit test.


hund030:
E2E test passed.
https://github.com/OfficeDev/TeamsFx/actions/runs/1661459681

eriolchan:
Add @blackchoey to review the part of clear ARM provision feature flag

blackchoey (dismissed):
the arm feature flag related part looks good to me

eriolchan (approved):

packages/fx-core/src/plugins/resource/frontend/index.ts
```diff
@@ -62,36 +62,18 @@ export class FrontendPlugin implements Plugin {
...
-      return ok(undefined);
-    }
-
-    FrontendPlugin.setContext(ctx);
-    return this.runWithErrorHandling(ctx, TelemetryEvent.Provision, () =>
-      isVsCallingCli()
-        ? this.blazorPluginImpl.provision(ctx)
-        : this.frontendPluginImpl.provision(ctx)
-    );
+    return ok(undefined);
```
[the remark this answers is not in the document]
hund030: I think there are still some plugins need to provision specific resources that arm can not cover. Bot plugin need to create aad in provision stage for example. So this API should be remained.

packages/fx-core/src/plugins/resource/frontend/configs.ts
```diff
@@ -105,25 +102,15 @@ export class FrontendConfig {
   }
 
   static getSubscriptionId(ctx: PluginContext): string {
-    const result = isArmSupportEnabled()
-      ? getSubscriptionIdFromResourceId(FrontendConfig.getStorageResourceId(ctx))
-      : FrontendConfig.getConfig<string>(
-          DependentPluginInfo.SubscriptionId,
-          ctx.envInfo.state.get(DependentPluginInfo.SolutionPluginName)
-        );
+    const result = getSubscriptionIdFromResourceId(FrontendConfig.getStorageResourceId(ctx));
```
IvanJobs, line 105: I have an assumption for getting subscriptionId, ie, the priority would be solution > others.
If the assumption is true, better to get subsId from solution first, if missed, then get it from other places for fallback.
What do you say?

packages/fx-core/src/plugins/resource/frontend/configs.ts
```diff
@@ -105,25 +102,15 @@ export class FrontendConfig {
...
   }
 
   static getResourceGroupName(ctx: PluginContext): string {
-    const result = isArmSupportEnabled()
-      ? getResourceGroupNameFromResourceId(FrontendConfig.getStorageResourceId(ctx))
-      : FrontendConfig.getConfig<string>(
-          DependentPluginInfo.ResourceGroupName,
-          ctx.envInfo.state.get(DependentPluginInfo.SolutionPluginName)
-        );
+    const result = getResourceGroupNameFromResourceId(FrontendConfig.getStorageResourceId(ctx));
```
IvanJobs, line 113: Same point as getting subsId. or maybe I missed some critical context?

packages/fx-core/src/plugins/resource/frontend/configs.ts
hund030: I have different opinion. I think the resource Id exactly describes what the resource is and where it is (subscription and resource group), rather than solution's config.
Anyway, this PR is just to refactor. We can discuss it later.

IvanJobs: Agree - not so critical, you can choose for your preference.

IvanJobs (approved):
lgtm - suppose smoke test for frontend and dotnet plugins are passed.

packages/fx-core/src/plugins/resource/frontend/dotnet/constants.ts
```diff
@@ -76,30 +76,9 @@ export class DependentPluginInfo {
...
   static readonly subscriptionId = "subscriptionId";
   static readonly resourceGroupName = "resourceGroupName";
-  static readonly resourceNameSuffix = "resourceNameSuffix";
-  static readonly location = "location";
-  static readonly remoteTeamsAppId = "remoteTeamsAppId";
-
-  static readonly functionPluginName = "fx-resource-function";
-  static readonly functionEndpoint = "functionEndpoint";
-
-  static readonly runtimePluginName = "fx-resource-simple-auth";
```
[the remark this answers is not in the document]
blackchoey: @eriolchan @hund030 FYI: @tecton will help drive the changes for VS, which removes the Simple Auth nuget package and use OBO flow to acquire MS Graph token from server side, since the blazor app now runs on server side.
While changing the .NET sdk and scaffolding template, Ning Tang will also help consider how our SDK should change when we support client side blazor app (via Web Assembly), to ensure a smooth change in the future.

PR #3551, opened 2022-01-06
feat: enable SharePoint app catalog creation

nliu-ms:
image


packages/fx-core/src/plugins/resource/spfx/plugin.ts
```diff
@@ -284,29 +289,38 @@ export class SPFxPluginImpl {
...
             appCatalogSite = await SPOClient.getAppCatalogSite(spoToken);
             while (appCatalogSite == null && retry < Constants.APP_CATALOG_MAX_TIMES) {
+              ctx.logProvider?.warning(`No tenant app catalog found, retry: ${retry}`);
               await sleep(Constants.APP_CATALOG_REFRESH_TIME);
               appCatalogSite = await SPOClient.getAppCatalogSite(spoToken);
               retry += 1;
             }
             if (appCatalogSite) {
               SPOClient.setBaseUrl(appCatalogSite);
+              ctx.logProvider?.info(
```
HuihuiWu-Microsoft, line 314: Will user get confused if the waiting time is too long and no hint for that since info message will not open output channel automatically?

nliu-ms: actually output channel is opened automatically in spfx deploy. line 212-214

HuihuiWu-Microsoft (approved):

PR #3554, opened 2022-01-06
perf(aad): update aad plugin for removing simple auth

KennethBWSong:


packages/fx-core/src/plugins/resource/aad/plugin.ts
```diff
@@ -402,15 +407,27 @@ export class AadAppForTeamsImpl {
...
+        redirectUris: [],
+      },
+      spa: {
+        redirectUris: [],
+      },
+    };
     if (frontendEndpoint) {
-      redirectUris.push(`${frontendEndpoint}/auth-end.html`);
+      redirectUris.web?.redirectUris?.push(`${frontendEndpoint}/auth-end.html`);
+      redirectUris.spa?.redirectUris?.push(`${frontendEndpoint}/auth-end.html/blank.html`);
```
blackchoey, line 423: appending a page after another page makes the url seems strange, can we remove `auth-end.html` and give the blank page a more meaningful name?

KennethBWSong: Good idea. Will update, thx！

KennethBWSong: Updated.

blackchoey (approved):

adashen (approved):

PR #3555, opened 2022-01-06
chore: add comments for function bicep files

blackchoey:
help users better understand the provision behavior of Azure Functions


packages/fx-core/tests/plugins/resource/function/unit/expectedBicepFiles/configWithKeyVaultPlugin.result.bicep
```diff
@@ -1,6 +1,8 @@
+// Get existing app settings and site config for merge
```
hund030, line 1: Do we really need to sync up comments changes?

hund030 (approved):

packages/fx-core/tests/plugins/resource/function/unit/expectedBicepFiles/configWithKeyVaultPlugin.result.bicep
blackchoey: The comment is visible to end users to help them better understand the bicep content. So if we didn't update the expected value in test case, the case will fail due to missing these comments.

PR #3557, opened 2022-01-06
chore: remove isMultiEnvEnabled from CLI

a1exwang:
code cleanup https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/12890204/

### DIFF
--- a/packages/cli/src/utils.ts
+++ b/packages/cli/src/utils.ts
@@ -35,14 +35,13 @@ import { ConfigNotFoundError, UserdataNotFound, EnvUndefined, ReadFileError } fr
 import AzureAccountManager from "./commonlib/azureLogin";
 import { FeatureFlags } from "./constants";
 import {
-  isMultiEnvEnabled,
   environmentManager,
   WriteFileError,
   localSettingsFileName,
+  FxCore,
+  isSPFxProject,
 } from "@microsoft/teamsfx-core";
 import { WorkspaceNotSupported } from "./cmds/preview/errors";
-import { FxCore } from "@microsoft/teamsfx-core";
-import { isSPFxProject } from "@microsoft/teamsfx-core";
 
 export type Json = { [_: string]: any };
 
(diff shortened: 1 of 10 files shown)


packages/cli/src/utils.ts
```diff
@@ -259,11 +246,7 @@ export async function readProjectSecrets(
   }
   const secretFile = secretFileResult.value;
   if (!fs.existsSync(secretFile)) {
-    if (isMultiEnvEnabled()) {
```
chagong, line 262: No problem with removing this check  to multi-env?

a1exwang: No problem because `isMultiEnvEnabled` is always `true` in current version.

packages/cli/src/utils.ts
```diff
@@ -328,16 +302,12 @@ export async function setSubscriptionId(
 export function isWorkspaceSupported(workspace: string): boolean {
```
chagong, line 302: I am not sure if there is an API in core to do this workspace checking. Did you check it when you are trying to clean it?

chagong: And for the following several utils like 'getTeamsAppId', 'getLocalTeamsAppId', I saw someone add comments like 'only used when multi-env is disabled'. I am not sure if these utils are properly used.

packages/cli/src/utils.ts
chagong: got it.

chagong (approved):

packages/cli/src/utils.ts
a1exwang: 1. It seems there is a `isValidProject` in core that is similar to `isWorkspaceSupported`. But this pr is mainly for removing `isMultiEnvEnabled`. We can cleanup that later. 2. Good point, I will cleanup these utils too

PR #3560, opened 2022-01-06
chore: prepare sideloading for checker

swatDong:
Move `getSideloadingStatus` to `core` to be used by vsc/cli/deps-checker.
Put in `common/tools` since it's account checker so may not only be used by local debug.

@1yefuwang1 @jayzhang @LongOddCode - Found that `common/telemetry` has dependency cycle which breaks unit tests, so I moved `core/index.ts - telemetryReporter` to `common/telemetry - TelemetryReporterInstance.telemetryReporter` to fix the deps cycle.

@1openwindow @HuihuiWu-Microsoft - There are vsc changes since code is moved from vsc to core.


kuojianlu (approved):

jayzhang (approved):

packages/fx-core/src/common/tools.ts
```diff
@@ -639,3 +647,52 @@ export function getAllowedAppIds(): string[] {
...
+            "UnknownValue",
+            `AppStudio response code: ${response.status}, body: ${response.data}`,
+            "M365Account"
+          )
+        );
+      }
+
+      return result;
+    } catch (error) {
+      sendTelemetryErrorEvent(
```
HuihuiWu-Microsoft, line 688: Don't forget to check the GDPR classify status.

HuihuiWu-Microsoft (approved):
LGTM

packages/fx-core/src/common/tools.ts
swatDong: Good point. Will check in RC release.

PR #3563, opened 2022-01-07
feat(sdk): support multiple sql databases

tecton:
Add an optional parameter to specify database name.


packages/sdk/src/core/defaultTediousConnectionConfiguration.ts
```diff
@@ -121,9 +123,13 @@ export class DefaultTediousConnectionConfiguration {
    * @returns Tedious connection configuration with username and password.
    * @internal
    */
-  private generateDefaultConfig(sqlConfig: SqlConfiguration): ConnectionConfig {
+  private generateDefaultConfig(
+    sqlConfig: SqlConfiguration,
+    databaseName?: string
+  ): ConnectionConfig {
+    const dbName = databaseName ?? sqlConfig.sqlDatabaseName;
```
SLdragon, line 130: What the expected behaviour when databaseName is empty? 

Be careful that `""??"test"` will result `""`

tecton: Because the developer explicitly pass the parameter, the empty string "" will be used in tedious config to connect database and depends on tedious behavior (probably runtime exception)

tecton: Yes, I added warning message.

SLdragon (approved):

PR #3572, opened 2022-01-07
chore: replace deps checker in cli

meifans:
Replace old deps checker in cli


swatDong (dismissed):

packages/cli/src/cmds/preview/depsChecker/cliUtils.ts
```diff
@@ -1,18 +1,35 @@
-// Copyright (c) Microsoft Corporation.
```
swatDong, line 1: Do not remove the header.

meifans: Fixed

swatDong (approved):

PR #3595, opened 2022-01-10
feat(dotnet): generate arm template for .NET scenario

hund030:


packages/fx-core/src/plugins/resource/frontend/dotnet/constants.ts
```diff
@@ -85,3 +88,14 @@ export class DotnetConfigInfo {
...
   static readonly projectFilePath = "projectFilePath";
 }
+
+export class WebappBicepFile {
+  static readonly configurationTemplateFileName = "webappConfiguration.template.bicep";
+  static readonly provisionTemplateFileName = "webappProvision.template.bicep";
+}
+
+export class WebappBicep {
+  static readonly webappEndpoint = "provisionOutputs.webappOutput.value.endpoint";
```
eriolchan, line 98: Is it possible to make it as a constant so that if the configuration is changed, we only update one place in the code.

packages/fx-core/src/plugins/resource/frontend/dotnet/plugin.ts
```diff
@@ -18,10 +25,16 @@ import { Logger } from "../utils/logger";
 import path from "path";
 import fs from "fs-extra";
 import {
+  generateBicepFromFile,
   getResourceGroupNameFromResourceId,
   getSiteNameFromResourceId,
   getSubscriptionIdFromResourceId,
+  getTemplatesFolder,
 } from "../../../..";
```
eriolchan, line 33: same here for the import path

packages/fx-core/src/plugins/resource/frontend/dotnet/plugin.ts
```diff
@@ -77,6 +90,56 @@ export class DotnetPluginImpl {
...
+        webappResourceId: WebappBicep.webappResourceId,
+        endpoint: WebappBicep.webappEndpoint,
+        domain: WebappBicep.webappDomain,
+      },
+    };
+    return ok(result);
+  }
+
   public async postProvision(ctx: PluginContext): Promise<TeamsFxResult> {
     return ok(undefined);
```
eriolchan, line 144: Do we have any post provision step for dotnet plugin?

packages/fx-core/src/plugins/resource/frontend/index.ts
```diff
@@ -44,20 +44,23 @@ export class FrontendPlugin implements Plugin {
   dotnetPluginImpl = new DotnetPluginImpl();
 
   private static setContext(ctx: PluginContext): void {
-    Logger.setLogger(ctx.logProvider);
-    TelemetryHelper.setContext(
-      ctx,
-      isVsCallingCli() ? DotnetPluginInfo.pluginName : FrontendPluginInfo.PluginName
-    );
+    const component = this.isVsPlatform(ctx)
```
eriolchan, line 47: Is it better to get implement instance from context as below, so that we don't need to use `this.isVsPlatform(ctx)?...` across many method such as scaffold/provision/deploy etc.

```
private static getImpl(ctx: PluginContext): ImplInterface {
  const impl = this.isVsPlatform(ctx) ? this.dotnetPluginImpl : this.frontendPluginImpl;
  Logger.setLogger(ctx.logProvider, impl.pluginName);
  TelemetryHelper.setContext(ctx, impl.pluginName);
  return impl;
}

  const impl = FrontendPlugin.getImpl(ctx);
  return this.runWithErrorHandling(ctx, TelemetryEvent.Scaffold, () => impl.scaffold(ctx));
```
ImplInterface (can be any more proper name) is the interface of frontendPluginImpl and dotnetPluginImpl.

packages/fx-core/tests/plugins/resource/dotnet/unit/deploy.test.ts
```diff
@@ -8,25 +8,24 @@ import fs from "fs-extra";
 import { PluginContext } from "@microsoft/teamsfx-api";
 import chaiAsPromised from "chai-as-promised";
 
-import { FrontendPlugin as BlazorPlugin } from "../../../../../src/plugins/resource/frontend";
+import { FrontendPlugin as WebappPlugin } from "../../../../../src/plugins/resource/frontend";
```
eriolchan, line 11: I am a little confused on is it WebappPlugin or dotnetPlugin?

hund030: I would like to distinguish dotnet plugin and webapp plugin. The former is for scaffolding and the latter is for provisioning and deploying.

packages/fx-core/src/plugins/resource/frontend/dotnet/constants.ts
```diff
@@ -65,6 +65,9 @@ export class DotnetPathInfo {
     framework = DotnetPluginInfo.defaultFramework,
     runtime = DotnetPluginInfo.defaultRuntime
   ): string => path.join(workingPath, "bin", "Release", framework, runtime, "publish");
+
+  static readonly bicepTemplateFolder = (templateFolder: string) =>
+    path.join(templateFolder, "plugins", "resource", "webapp", "bicep");
```
IvanJobs, line 70: Nit: Although checked the point in related comment below, I still recommend using `dotnet` instead of `webapp`, which is based on to simplify concept for readability and to follow existing style/contract, I mean, (bot/simpleauth/tab used (webapp/webapp/storage) for provision/deploy, but no webapp plugin/storage plugin existed since we may treat them as implementation choices).

packages/fx-core/src/plugins/resource/frontend/index.ts
```diff
@@ -44,91 +45,70 @@ export class FrontendPlugin implements Plugin {
   frontendPluginImpl = new FrontendPluginImpl();
   dotnetPluginImpl = new DotnetPluginImpl();
 
+  private getImpl(ctx: PluginContext): PluginImpl {
+    return FrontendPlugin.isVsPlatform(ctx) ? this.dotnetPluginImpl : this.frontendPluginImpl;
```
IvanJobs, line 49: Nit: Another direction may be considered is to use `Polymorphisn` if it's feasible (frontendPluginImpl shares the same interface with dotnetPluginImpl).

IvanJobs (dismissed):
Two NITs for reference, and overall lgtm.

packages/fx-core/src/plugins/resource/frontend/index.ts
hund030: I prefer to make FrontendPlugin stateless. Using polymorphisn make a state, that means pluginImpl either to be FrontendPluginImpl or DotnetPluginImpl

packages/fx-core/src/plugins/resource/frontend/dotnet/constants.ts
hund030: 1. In v3 design, there will be a dotnet plugin for scaffolding and building dotnet project, and a webapp plugin for provisioning resources and deploying code. They are mapping to current dotnet plugin's scaffold and provision, deploy.
I would like to distinguish them in advance.
2. We used to naming plugin by capability or component (bot/simpleauth/function), but `dotnet` is different from those. It is the runtime of the project. It is wired for me to see a dotnet/bicep folder.

packages/fx-core/src/plugins/resource/frontend/dotnet/plugin.ts
hund030: No. Just to align with FrontendPluginImpl.

packages/fx-core/src/plugins/resource/frontend/index.ts
hund030: Updated.

packages/fx-core/src/plugins/resource/frontend/dotnet/plugin.ts
hund030: Updated.

IvanJobs (approved):
lgtm

PR #3599, opened 2022-01-11
chore(local debug): prepare account checker

swatDong:
Prepare account checker for prerequisites validation.
For example, block debugging if no account found:
image


packages/vscode-extension/src/debug/prerequisitesHandler.ts
```diff
@@ -19,9 +39,27 @@ export async function checkAndInstall(): Promise<Result<any, FxError>> {
...
+    // login
+    const accountFailure = await checkM365Account();
+    if (accountFailure) {
+      failures.push(accountFailure);
+    }
+
+    // handle failures
+    if (failures.length > 0) {
+      await handleFailures(failures);
+      return err(
```
kuojianlu, line 55: `return` here will lose telemetry, `throw` instead?

swatDong: Good point. Updated to `return` and add checkers information to error message as well, thus we can know which ones are failed.

kuojianlu (approved):

swatDong:
@1openwindow @HuihuiWu-Microsoft - this PR has error type/message changes in vscode-extension. Please help review.

HuihuiWu-Microsoft (approved):

PR #3609, opened 2022-01-11
chore: set correct error code for arm deployment

blackchoey:
currently all telemetries set the error code to ArmDeploymentFailed, update the code to set correct error code in telemetry, like BicepInstallFailed.

### DIFF
--- a/packages/fx-core/src/plugins/solution/fx-solution/arm.ts
+++ b/packages/fx-core/src/plugins/solution/fx-solution/arm.ts
@@ -608,11 +608,11 @@ export async function deployArmTemplates(ctx: SolutionContext): Promise<Result<v
       result = err(error);
     } else if (error instanceof Error) {
       result = err(
-        returnUserError(error, SolutionSource, SolutionError.FailedToDeployArmTemplatesToAzure)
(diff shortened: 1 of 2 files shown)


packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -603,9 +604,21 @@ export async function deployArmTemplates(ctx: SolutionContext): Promise<Result<v
       );
     }
   } catch (error) {
-    result = err(
-      returnUserError(error, SolutionSource, SolutionError.FailedToDeployArmTemplatesToAzure)
```
blackchoey, line 607: @dilin-MS can you help recall why we set the error to user error here? If no special reason I will update the code to set it as system error

jayzhang (approved):

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
blackchoey: discussed offline, will change the error type to system error

jayzhang (approved):

dilin-MS (approved):

PR #3611, opened 2022-01-11
chore(local debug): prepare task runner in vsc

swatDong:
This is for `npm install` prerequisites check - to run `npm install` as vscode tasks.
Added `runTask` utility to execute vsc tasks, and watch for exit code.


packages/vscode-extension/src/debug/teamsfxTaskHandler.ts
```diff
@@ -135,6 +179,20 @@ async function checkCustomizedPort(component: string, componentRoot: string, che
...
+function onDidStartTaskHandler(event: vscode.TaskStartEvent): void {
+  const taskId = event.execution.task?.definition?.teamsfxTaskId;
+  if (taskId !== undefined) {
+    taskStartEventEmitter.fire(taskId as string);
+  }
+}
+
+function onDidEndTaskHandler(event: vscode.TaskEndEvent): void {
+  const taskId = event.execution.task?.definition?.teamsfxTaskId;
+  if (taskId !== undefined && trackedTasks.has(taskId as string)) {
```
kuojianlu, line 191: Seems missing `trackedTasks.add()` beforewards.

packages/vscode-extension/src/debug/teamsfxTaskHandler.ts
```diff
@@ -168,6 +226,12 @@ async function onDidEndTaskProcessHandler(event: vscode.TaskProcessEndEvent): Pr
   const task = event.execution.task;
   const activeTerminal = vscode.window.activeTerminal;
 
+  const taskId = task?.definition?.teamsfxTaskId;
+  if (taskId !== undefined) {
+    trackedTasks.delete(taskId as string);
+    taskEndEventEmitter.fire({ id: taskId as string, exitCode: event.exitCode });
```
kuojianlu, line 232: Curious what is the behavior when a task end listener is fired twice, i.e., one on onDidEndTaskProcessHandler and the other on onDidEndTaskHandler?

packages/vscode-extension/src/debug/teamsfxTaskHandler.ts
swatDong: Good catch. Added `trackedTasks.add` and `trackedTasks.delete` whenever end event is fired.

packages/vscode-extension/src/debug/teamsfxTaskHandler.ts
swatDong: Per vscode source code, `onDidEndTaskProcessHandler` will be called first. For normal case `trackedTasks` is updated so `onDidEndTaskHandler` will not fire event.

For error case that process has not been started, only `onDidEndTaskHandler` will fire event.

For corner case that end event is fired twice, updated `runTask` to dispose listener after receiving the first event. So the second event has no effect on `runTask`.

PR #3625, opened 2022-01-12
feat(corev3): local debug

jayzhang:


packages/fx-core/src/common/local/projectSettingsHelper.ts
```diff
@@ -12,20 +17,24 @@ export class ProjectSettingsHelper {
...
     );
 
-  public static includeFrontend = (projectSettings: ProjectSettings | undefined): boolean =>
-    !!(projectSettings?.solutionSettings as AzureSolutionSettings)?.activeResourcePlugins?.some(
-      (pluginName) => pluginName === ResourcePlugins.FrontendHosting
-    );
-
+  public static includeFrontend = (projectSettings: ProjectSettings | undefined): boolean => {
+    if (!projectSettings) return false;
+    const solutionSettings = projectSettings.solutionSettings as AzureSolutionSettings;
```
swatDong, line 22: Will `solutionSettings` or `capabilities` be undefined?

jayzhang: No, their are required

packages/fx-core/src/common/local/projectSettingsHelper.ts
```diff
@@ -12,20 +17,24 @@ export class ProjectSettingsHelper {
...
+    return solutionSettings.capabilities.includes(TabOptionItem.id);
```
swatDong, line 23: Consider backward compatibility - what if opening existing (before v3) project?
Is `solutionSettings.version` designed for determining whether to use `solutionSettings.activeResourcePlugins` or `solutionSettings.capabilities`?

jayzhang: Yes, but a better way is to use capabilities for both versions. Capabilities are the single truth.

PR #3627, opened 2022-01-12
chore: account offline

xiaolang124:
1. Do not clear account cache when offline.
2. TreeView will show m365 account name when user has logged in.


packages/vscode-extension/src/commonlib/codeFlowLogin.ts
```diff
@@ -356,3 +360,16 @@ export function ConvertTokenToJson(token: string): object {
   const buff = Buffer.from(array[1], "base64");
   return JSON.parse(buff.toString(UTF8));
 }
+
+export async function checkIsOnline(): Promise<boolean> {
+  try {
+    await new Promise<http.IncomingMessage>((resolve, reject) => {
+      const url = "https://login.microsoftonline.com/";
+      https.get(url, resolve).on("error", reject);
```
swatDong, line 368: Does `get` automatically follow redirect? A full `get` could be heavy, any `head` option?

packages/vscode-extension/src/commonlib/codeFlowLogin.ts
```diff
@@ -258,6 +259,9 @@ export class CodeFlowLogin {
                   error.message
                 )
             );
+            if (!(await checkIsOnline())) {
```
swatDong, line 262: Any change needed in CLI?

packages/vscode-extension/src/accountTree.ts
```diff
@@ -312,6 +312,16 @@ export async function registerAccountTreeHandler(): Promise<Result<Void, FxError
...
+      } else if (accountInfo !== undefined) {
+        const treeItem = {
+          commandId: "fx-extension.signinM365",
+          label: (accountInfo.upn as string) ? (accountInfo.upn as string) : "",
+          callback: signinM365Callback,
+          parent: TreeCategory.Account,
+          contextValue: "signedinM365",
+          icon: "M365",
+        };
+        tools.treeProvider?.add([treeItem]);
```
HuihuiWu-Microsoft, line 324: I think here should be 'refresh()'?

HuihuiWu-Microsoft: BTW, what about showing a warning to remind user he is offline, or user might get confused the account shows normally but error occurs

xiaolang124: I have tried with ```refresh```, the account info will not show.

image

image

packages/vscode-extension/src/commonlib/codeFlowLogin.ts
xiaolang124: Already add CLI logic.

packages/vscode-extension/src/commonlib/codeFlowLogin.ts
xiaolang124: Changed to ```head``` request.

packages/vscode-extension/src/accountTree.ts
HuihuiWu-Microsoft: It's weird, from the code logic, add() should be the one that doesn't work.

HuihuiWu-Microsoft: I got it, the item is added the first time extension started. The tree view also listened to SharepointLoginProvider, will spfx deploy affect the status?

xiaolang124: I think so, spfx login can also trigger treeView update user info.

HuihuiWu-Microsoft: Then is it needed to also update the sharepointLoginProvider getStatus() logic ?

xiaolang124: I update sharepoint logic, but it does not work. The treeProvider has not initialized when offline at this time.

image

xiaolang124: Move setStatus location to make sure notify will update treeView.

HuihuiWu-Microsoft (approved):

swatDong (approved):

Alive-Fish (approved):

PR #3651, opened 2022-01-13
chore(localdebug): provide resolve local cert api in local env manager

XiaofuHuang:
Detail: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/12863544


kuojianlu (approved):

packages/fx-core/src/common/local/localCertificateManager.ts
```diff
@@ -33,19 +26,17 @@ const confirmMessage =
 export interface LocalCertificate {
   certPath: string;
   keyPath: string;
-  isTrusted: boolean;
+  isTrusted?: boolean;
```
swatDong, line 29: Why changing this to nullable? What does `isTrusted === undefined` mean?

XiaofuHuang: It means we do not check the trusted status.
When provision local resources, now `isTrusted` is useless. So in the `getCertificate` method, only return the paths and do not return the `isTrusted`.

swatDong: If so, make `setupCertificate` return correct `isTrusted` value. E.g., for linux it's `undefined`, for window/mac it's `true/false`, for unexpected error it's `undefined`.

This is a good change that `validate-prerequisites` can block F5 when `isTrusted === false`, but need to make sure `undefined` is meaningful.
BTW, for provision local resources, keep it calling `setupCertificate` since we are not going to migration existing `tasks.json` so there could be project without `validation-prerequisites`.

PR #3652, opened 2022-01-13
feat: new sample gallery overview page layout

HuihuiWu-Microsoft:
image

### DIFF
--- a/packages/vscode-extension/src/controls/SampleGallery.tsx
+++ b/packages/vscode-extension/src/controls/SampleGallery.tsx
@@ -75,7 +75,7 @@ export default class SampleGallery extends React.Component<any, any> {
       <div className="sample-gallery">
         <div className="section" id="title">
           <div className="logo">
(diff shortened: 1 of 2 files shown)


packages/vscode-extension/src/controls/SampleGallery.tsx
```diff
@@ -274,3 +274,99 @@ class SampleAppCard extends React.Component<any, any> {
     });
   };
 }
+
+class SampleCard extends React.Component<any, any> {
```
1openwindow, line 278: looks like there is not any reference to SampleCard.

1openwindow (approved):

packages/vscode-extension/src/controls/SampleGallery.tsx
HuihuiWu-Microsoft: This is to avoid user from getting access to new card since the feature will not be release in next version

PR #3654, opened 2022-01-13
perf(collaboration): auto input default user email when grant permission

SLdragon:
When click grant permission, it will auto input default user email:

If user forget to update this default email, it will show tips:

### DIFF
--- a/packages/fx-core/src/plugins/solution/fx-solution/question.ts
+++ b/packages/fx-core/src/plugins/solution/fx-solution/question.ts
@@ -256,21 +256,31 @@ export const AskSubscriptionQuestion: FuncQuestion = {
   },
 };
 
-export const GetUserEmailQuestion: TextInputQuestion = {
-  name: "email",
-  type: "text",
-  title: "Add owner to Teams/AAD app for the account under the same M365 tenant (email)",
-  validation: {
-    validFunc: (input: string, previousInputs?: Inputs): string | undefined => {
-      if (!input || input.trim() === "") {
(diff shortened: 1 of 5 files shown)


KennethBWSong (approved):

1yefuwang1 (approved):

packages/fx-core/src/plugins/solution/fx-solution/question.ts
```diff
@@ -256,21 +256,31 @@ export const AskSubscriptionQuestion: FuncQuestion = {
...
-  name: "email",
-  type: "text",
-  title: "Add owner to Teams/AAD app for the account under the same M365 tenant (email)",
-  validation: {
-    validFunc: (input: string, previousInputs?: Inputs): string | undefined => {
-      if (!input || input.trim() === "") {
-        return "Email address cannot be null or empty";
-      }
+export function getUserEmailQuestion(currentUserEmail: string): TextInputQuestion {
+  const defaultUserEmail = "[UserName]@" + currentUserEmail.split("@")[1];
```
adashen, line 260: Do we need check whether currentUserEmail is null?

SLdragon: Updated

adashen (approved):

1yefuwang1 (dismissed):

KennethBWSong (approved):

1yefuwang1 (approved):

Alive-Fish (approved):

PR #3656, opened 2022-01-13
feat(sql): multiple database

xzf0587:
1. add multiple database bicep 
2. update adding database user for multiple database


xzf0587:
bicep snippet for database module.
image

xzf0587:
bicep snippet for provision
image

packages/fx-core/src/plugins/resource/sql/constants.ts
```diff
@@ -108,4 +109,7 @@ export class AzureSqlBicep {
 export class AzureSqlBicepFile {
   static readonly moduleTemplateFileName: string = "provision.template.bicep";
   static readonly ProvisionModuleTemplateFileName = "sqlProvision.template.bicep";
+  static readonly newDatabaseModuleTemplateFileName: string =
+    "newDatabase.provision.template.bicep";
+  static readonly newDatabaseProvisionModuleTemplateFileName = "newDatabase.template.bicep";
```
wenytang-ms, line 114: The two values and names don't seem to match

packages/fx-core/src/plugins/resource/sql/plugin.ts
```diff
@@ -249,9 +250,18 @@ export class SqlPluginImpl {
     managementClient: ManagementClient
   ): Promise<void> {
     let retryCount = 0;
+    const databases: { [key: string]: boolean } = {};
```
blackchoey, line 253: better naming? my understading is this indicates whether add user succeedded for each database?

packages/fx-core/src/plugins/resource/sql/plugin.ts
```diff
@@ -446,4 +493,20 @@ export class SqlPluginImpl {
...
+      if (k.startsWith(Constants.databaseName)) {
+        this.config.databases.push(v);
+      }
+    });
+  }
+
+  private removeDatabases(ctx: PluginContext) {
+    ctx.config.forEach((v: string, k: string) => {
+      if (k.startsWith(Constants.databaseName) && k !== Constants.databaseName) {
+        ctx.config.delete(k);
```
blackchoey, line 508: can we merge loadDatabases and removeDatabases into one function?

packages/fx-core/src/plugins/resource/sql/constants.ts
xzf0587: Sure. Updated it. Thanks.

packages/fx-core/src/plugins/resource/sql/plugin.ts
```diff
@@ -271,6 +281,9 @@ export class SqlPluginImpl {
   }
 
   public async generateArmTemplates(ctx: PluginContext): Promise<Result<any, FxError>> {
+    if (ctx.answers?.existingResources?.includes(Constants.pluginFullName)) {
```
blackchoey, line 284: minor, can we see which is better?
1. Check existingResources in plugin.ts as what we're doing here
2. Check existingResources in `generateArmTemplates` in index.ts, and call addNewDatabase() from there directly.

packages/fx-core/src/plugins/resource/sql/plugin.ts
xzf0587: Actually, loadDatabases is used in both pre-provision and post-provision.
removeDatabases is used in pre-provision stage before loadDatabases. It is used to remove possible outdated database name. It is an additional action.

packages/fx-core/src/plugins/resource/sql/plugin.ts
xzf0587: Thanks.
It is fixed.

packages/fx-core/src/plugins/resource/sql/plugin.ts
xzf0587: Sure. 
Move the check in index.ts.

packages/fx-core/src/plugins/resource/sql/plugin.ts
```diff
@@ -311,6 +324,40 @@ export class SqlPluginImpl {
     return ok(result);
   }
 
+  public async addNewDatabase(ctx: PluginContext): Promise<Result<any, FxError>> {
```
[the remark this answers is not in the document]
xzf0587: It is fixed.
Thanks.

blackchoey (approved):

PR #3661, opened 2022-01-14
feat(aad): implement aad plugin with api v3

jayzhang:


packages/fx-core/src/plugins/resource/aad/utils/common.ts
```diff
@@ -45,24 +53,66 @@ export class Utils {
...
+  public static async getCurrentTenantId(graphTokenProvider?: GraphTokenProvider): Promise<string> {
+    const tokenObject = await graphTokenProvider?.getJsonObject();
+    const tenantId: string = (tokenObject as any)?.tid;
     return tenantId;
   }
 
+  public static skipCreateAadForProvision(envInfo: v3.EnvInfoV3): boolean {
+    const envConfig: EnvConfig = envInfo.config as EnvConfig;
+    const envState: v3.AADApp = envInfo.state[BuiltInResourcePluginNames.aad] as v3.AADApp;
+    const objectId = envConfig.auth?.objectId ? envConfig.auth.objectId : envState.objectId;
```
KennethBWSong, line 77: I think we should not read state here:
`const objectId = envConfig.auth?.objectId`
If there is no value from config, we should just return false in line 97

packages/fx-core/src/plugins/resource/aad/utils/common.ts
```diff
@@ -45,24 +53,66 @@ export class Utils {
...
+    } else if (objectId || clientId || oauth2PermissionScopeId || clientSecret) {
+      throw ResultFactory.UserError(
+        GetSkipAppConfigError.name,
+        GetSkipAppConfigError.message(ConfigFilePath.LocalSettings)
+      );
+    } else {
+      return false;
+    }
+  }
   public static skipAADProvision(ctx: PluginContext, isLocalDebug = false): boolean {
```
KennethBWSong, line 116: Maybe we can remove this?

packages/fx-core/src/plugins/resource/aad/utils/configs.ts
```diff
@@ -213,7 +316,69 @@ export class SetApplicationInContextConfig {
...
+      throw ResultFactory.SystemError(
+        GetConfigError.name,
+        GetConfigError.message(Errors.GetConfigError(ConfigKeys.clientId, Plugins.pluginName))
+      );
+    }
+  }
+  public restoreConfigFromEnvInfo(ctx: v2.Context, envInfo: v3.EnvInfoV3): void {
+    const solutionSettings = ctx.projectSetting.solutionSettings as v3.TeamsFxSolutionSettings;
+    const aadResource = envInfo.state[BuiltInResourcePluginNames.aad] as v3.AADApp;
+    let frontendDomain = aadResource?.domain;
```
KennethBWSong, line 342: domain should be get from frontend plugin?

packages/fx-core/src/plugins/resource/aad/utils/configs.ts
```diff
@@ -213,7 +316,69 @@ export class SetApplicationInContextConfig {
...
+    if (frontendDomain) {
+      this.frontendDomain = format(frontendDomain as string, Formats.Domain);
+    }
 
+    const botModules = solutionSettings.modules.filter((m) =>
+      m.capabilities.includes(BotOptionItem.id)
+    );
+    let botId;
+    if (botModules.length > 0) {
+      const hostingPlugin = botModules[0].hostingPlugin;
```
KennethBWSong, line 364: Can one Teams App add multiple Bots? If not, may be we don't need to use array for botModules?

packages/fx-core/src/plugins/resource/aad/utils/configs.ts
```diff
@@ -235,6 +400,113 @@ export class PostProvisionConfig {
...
+      throw ResultFactory.SystemError(
+        GetConfigError.name,
+        GetConfigError.message(Errors.GetConfigError(ConfigKeys.clientId, Plugins.pluginName))
+      );
+    }
+  }
+  public restoreConfigFromEnvInfo(ctx: v2.Context, envInfo: v3.EnvInfoV3): void {
+    const solutionSettings = ctx.projectSetting.solutionSettings as v3.TeamsFxSolutionSettings;
+    const aadResource = envInfo.state[BuiltInResourcePluginNames.aad] as v3.AADApp;
+    let frontendEndpoint = aadResource?.endpoint;
```
KennethBWSong, line 448: Same as above. Will there be endpoint in aad state? Seems there is no code saving this.

nliu-ms (approved):

KennethBWSong (approved):

PR #3663, opened 2022-01-14
fix: potential fix to create project telemetry issue

HuihuiWu-Microsoft:
@chagong  @1openwindow @1yefuwang1 This is a potential fix to the create project issue since when 'openFolder' executed, the telemetry might not flashed.

### DIFF
--- a/packages/vscode-extension/src/commonlib/telemetry.ts
+++ b/packages/vscode-extension/src/commonlib/telemetry.ts
@@ -155,4 +155,8 @@ export class VSCodeTelemetryReporter extends vscode.Disposable implements Teleme
       this.reporter.sendTelemetryException(error, properties, measurements);
     }
   }
+
+  async dispose() {
+    await this.reporter.dispose();
+  }
 }
--- a/packages/vscode-extension/src/handlers.ts
+++ b/packages/vscode-extension/src/handlers.ts
@@ -351,6 +351,7 @@ export async function createNewProjectHandler(args?: any[]): Promise<Result<any,
   ExtTelemetry.sendTelemetryEvent(TelemetryEvent.CreateProjectStart, getTriggerFromProperty(args));
   const result = await runCommand(Stage.create);
   if (result.isOk()) {
+    await ExtTelemetry.dispose();
     commands.executeCommand("vscode.openFolder", result.value);
   }
   return result;
--- a/packages/vscode-extension/src/telemetry/extTelemetry.ts
+++ b/packages/vscode-extension/src/telemetry/extTelemetry.ts
@@ -159,4 +159,8 @@ export namespace ExtTelemetry {
 
     reporter.sendTelemetryException(error, properties, measurements);
   }
+
+  export async function dispose() {
+    await reporter.dispose();
+  }
 }
--- a/packages/vscode-extension/test/unit/extension/handlers.test.ts
+++ b/packages/vscode-extension/test/unit/extension/handlers.test.ts
@@ -64,6 +64,7 @@ suite("handlers", () => {
       sinon.stub(handlers, "core").value(new MockCore());
       sinon.stub(ExtTelemetry, "sendTelemetryEvent");
       sinon.stub(ExtTelemetry, "sendTelemetryErrorEvent");
+      sinon.stub(ExtTelemetry, "dispose");
       const createProject = sinon.spy(handlers.core, "createProject");
       sinon.stub(vscode.commands, "executeCommand");
 


packages/vscode-extension/src/commonlib/telemetry.ts
```diff
@@ -155,4 +155,8 @@ export class VSCodeTelemetryReporter extends vscode.Disposable implements Teleme
       this.reporter.sendTelemetryException(error, properties, measurements);
     }
   }
+
+  async dispose() {
```
swatDong, line 159: Is this duplicated with the constructor? Line 40 already added dispose callback in `super()`;

packages/vscode-extension/src/handlers.ts
```diff
@@ -351,6 +351,7 @@ export async function createNewProjectHandler(args?: any[]): Promise<Result<any,
   ExtTelemetry.sendTelemetryEvent(TelemetryEvent.CreateProjectStart, getTriggerFromProperty(args));
   const result = await runCommand(Stage.create);
   if (result.isOk()) {
+    await ExtTelemetry.dispose();
```
swatDong, line 354: Just curious that `ExtTelemetry.Reporter` has been added to vscode context's subscriptions, will vscode call dispose on reloading?

HuihuiWu-Microsoft: Yes, in theory it will and it works for my testing. But currently we have[ telemetry issue](https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/12974459?src=WorkItemMention&src-action=artifact_link) that many telemetry events are not sent and this only happens for create project scenario. So we want to manually flush the telemetry reporter to make sure it is sent before vscode reload. This is just a potential fix and we also would like to see whether this will have effect on the issue.

swatDong: I see. So in your test will the reported be disposed twice, any error if dispose twice?

packages/vscode-extension/src/commonlib/telemetry.ts
swatDong: Since dispose callback has already been added in `super()`, can I just call `this.dispose()` instead of adding a new one?

packages/vscode-extension/src/handlers.ts
HuihuiWu-Microsoft: Yes, vscode TelemetryReporter.dispose() is called twice and as far as I can see, that's no error occurred.

packages/vscode-extension/src/commonlib/telemetry.ts
HuihuiWu-Microsoft: Do you mean in this dispose function call 'await this.dispose()' instead of 'await this.reporter.dispose()'?

swatDong: I mean not to add a `dispose()` function, just calling `vscodeTelemetryReporter.dispose()` when needed. Since its constructor already added dispose callback to `super()`. So calling `vscodeTelemetryReporter.dispose()` is actually calling `vscode.Disposable.dispose()`.

HuihuiWu-Microsoft: Just to confirm with you: will vscode.Disposable.dispose() call the callOnDispose function with await if it's async function?

HuihuiWu-Microsoft: image
I'm afraid by default vscode.Disposable.dispose() will not await the function

swatDong: Oh, that may be the cause of missing telemetry - vscode only triggers the async dispose, but not wait for it.

HuihuiWu-Microsoft: Hope so~

swatDong: One more concern - will the added `VSCodeTelemetryReporter.dispose()` override `vscode.Disposable.dispose()`?

swatDong: Just to keep code clean / remove unused code : )

1openwindow:
Thanks @HuihuiWu-Microsoft to fix this. Before we ship it to production, we can do some test to verify if this potential fix really works.

chagong (approved):

HuihuiWu-Microsoft:
Already have another PR

PR #3681, opened 2022-01-17
feat(bot): implement generateArmTemplate in dotnet scenario

hund030:


IvanJobs (approved):
lgtm.
Despite the fact that ut cases are added to verify the rendered/expected bicep files, better to use bicep cli to provision these bicep files in real to check their effectiveness if feasible.

hund030:
> lgtm. Despite the fact that ut cases are added to verify the rendered/expected bicep files, better to use bicep cli to provision these bicep files in real to check their effectiveness if feasible.

In fact, these expected bicep files can not be compiled, solution does a lot to orchestrate those biceps. I agree that we should verify if these bicep files feasible but it can not be done by plugin level test, and e2e test should have covered it.

packages/fx-core/src/plugins/resource/bot/interface.ts
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { PluginContext } from "@microsoft/teamsfx-api";
+import { FxResult } from "./result";
+
+export interface PluginImpl {
```
eriolchan, line 7: Is this interface not provided by solution or FxCore?

PR #3685, opened 2022-01-17
fix(localdebug): use detect-port to detect if ports are in use

kuojianlu:
[detect-port](https://github.com/node-modules/detect-port) is used by react-scripts and many other projects.

To fix: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/13023066


packages/fx-core/package.json
```diff
@@ -63,6 +63,7 @@
     "axios": "^0.21.2",
     "cryptr": "^6.0.2",
     "dateformat": "^4.5.1",
+    "detect-port": "^1.3.0",
```
swatDong, line 66: `detect-port` has its type definitions - https://www.npmjs.com/package/@types/detect-port

packages/fx-core/tests/common/local/portChecker.test.ts
```diff
@@ -70,7 +70,7 @@ describe("portChecker", () => {
 
     it("happy path", async () => {
       const mockServer = new MockServer([]);
-      sinon.stub(net, "createServer").returns(mockServer as unknown as net.Server);
+      sinon.stub(net, "Server").returns(mockServer as unknown as net.Server);
```
swatDong, line 73: Is it possible to directly mock `detect`?

packages/fx-core/package.json
kuojianlu: Good catch, updated.

packages/fx-core/tests/common/local/portChecker.test.ts
kuojianlu: Yes, mocked `detect-port` instead.

packages/fx-core/package.json
```diff
@@ -84,6 +85,7 @@
     "node-ts-uuid": "^1.0.8",
     "openapi-types": "^7.2.3",
     "proper-lockfile": "^4.1.2",
+    "proxyquire": "^2.1.3",
```
swatDong, line 88: Move to `devDependencies` if it's only used by test.

swatDong (approved):

PR #3693, opened 2022-01-18
chore(sql): update axios error

xzf0587:
update axios error

### DIFF
--- a/packages/fx-core/src/plugins/resource/sql/managementClient.ts
+++ b/packages/fx-core/src/plugins/resource/sql/managementClient.ts
@@ -182,23 +182,23 @@ export class ManagementClient {
   }
 
   async addLocalFirewallRule(): Promise<void> {
-    const response = await axios.get(Constants.echoIpAddress);
-    const localIp: string = response.data;
-    const partials: string[] = localIp.split(".");
-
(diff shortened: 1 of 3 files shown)


xzf0587:
When the axios throws error , it looks like below
image

packages/fx-core/src/plugins/resource/sql/managementClient.ts
```diff
@@ -182,23 +182,23 @@ export class ManagementClient {
   }
 
   async addLocalFirewallRule(): Promise<void> {
-    const response = await axios.get(Constants.echoIpAddress);
-    const localIp: string = response.data;
-    const partials: string[] = localIp.split(".");
+    try {
+      const response = await axios.get(Constants.echoIpAddress);
```
adashen, line 186: maybe not related to the error fix. But one question is why we not use os networkInterfaces instead to get ip address. Is there any limitation?

xzf0587: We need to get the external IP address of the machine to config the firewall rule. The network interface in machine provides the internal  IP address which is not public IP address and usually is translated by NAT.

blackchoey (approved):

PR #3697, opened 2022-01-18
chore(localdebug): add v1 property in local debug telemetry

XiaofuHuang:
Add telemetry to track whether there are v1 F5 users


packages/fx-core/src/plugins/solution/fx-solution/debug/provisionLocal.ts
```diff
@@ -47,6 +47,7 @@ export async function setupLocalDebugSettings(
   const includeBot = ProjectSettingsHelper.includeBot(ctx.projectSetting);
   const includeAAD = ProjectSettingsHelper.includeAAD(ctx.projectSetting);
   const includeSimpleAuth = ProjectSettingsHelper.includeSimpleAuth(ctx.projectSetting);
+  const migrateFromV1 = ProjectSettingsHelper.isMigrateFromV1(ctx.projectSetting);
```
swatDong, line 50: Duplicated with `isMigrateFromV1`. Make sure `ProjectSettingsHelper.isMigrateFromV1` only called once.

XiaofuHuang: Got it

PR #3699, opened 2022-01-18
chore: template config sync up with template version change

wenytang-ms:


.github/scripts/fxcore-sync-up-version.js
```diff
@@ -0,0 +1,24 @@
+const path = require("path");
+const semver = require("semver");
+const fse = require("fs-extra");
+
+const templatePath = path.join(__dirname, "../../templates")
+const templateVersion = require(path.join(templatePath, "package.json")).version
+const templateName = require(path.join(templatePath, "package.json")).version
```
hund030, line 7: templateName is same with templateVersion?

wenytang-ms: sorry.

PR #3717, opened 2022-01-19
perf: enhance telemetry collection

1openwindow:
ADO: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/12974459


1openwindow:
base on this PR: https://github.com/OfficeDev/TeamsFx/pull/3663
add two more thing.
1. set time out
2. call dispose() when vs code deactivate() is trigged

swatDong (approved):

packages/vscode-extension/src/handlers.ts
```diff
@@ -351,7 +351,12 @@ export async function createNewProjectHandler(args?: any[]): Promise<Result<any,
   ExtTelemetry.sendTelemetryEvent(TelemetryEvent.CreateProjectStart, getTriggerFromProperty(args));
   const result = await runCommand(Stage.create);
   if (result.isOk()) {
-    commands.executeCommand("vscode.openFolder", result.value);
+    await ExtTelemetry.dispose();
+    // after calling dispose(), let reder process to while for a while instead of directly call "open folder"
+    // otherwise, the flush operation in dispose() will be interrupted due to shut down the render process.
+    setTimeout(() => {
+      commands.executeCommand("vscode.openFolder", result.value);
```
HuihuiWu-Microsoft, line 358: Can we make this duration more accurate if it works for less timeout?

1openwindow: it is hard to give a accurate duration. I did some test on 1000ms, it works on my local. But to be more safe, I like to set it as 3000ms. From the test result, we don't loss telemetry in 3000ms.

HuihuiWu-Microsoft (approved):

swatDong (approved):

PR #3721, opened 2022-01-19
perf: generate config files according to whether config content exist

wenytang-ms:


wenytang-ms:
this PR modify 3 test cases.
1. create a tab app without sa, so there will be no any config or teamsfx in templates.
2. create a tab app without sa, then add bot plugin, there will be add config and teamsfx.
3. create a tab app with sa, there must be config and teamsfx in templates, then add bot plugin, there the config added bot plugin, and teamsfx folder add bot config bicep.

packages/fx-core/tests/plugins/solution/solution.arm.test.ts
```diff
@@ -292,6 +239,10 @@ Mocked aad configuration orchestration content. Module path: './teamsFx/aadConfi
         AadParameter: `${TestFileContent.aadParameterValue}`,
       })
     );
+    expect(await fs.pathExists(path.join(projectArmTemplateFolder, TestFilePath.configFileName))).to
```
blackchoey, line 242: should we expect this to be true? The bot plugin requires config

wenytang-ms: 261 line will add bot plugin, current there has no any other plugins except frontend hosting. 
so there is no any config files

packages/fx-core/tests/plugins/solution/solution.arm.test.ts
```diff
@@ -390,15 +334,155 @@ Mocked bot provision orchestration content. Module path: './provision/botProvisi
...
+    TestHelper.mockedFehostGenerateArmTemplates(mocker);
+    TestHelper.mockedAadGenerateArmTemplates(mocker);
+    TestHelper.mockedIdentityGenerateArmTemplates(mocker);
+    TestHelper.mockedSimpleAuthGenerateArmTemplates(mocker);
+    TestHelper.mockedBotGenerateArmTemplates(mocker);
+    TestHelper.mockedFeHostUpdateArmTemplates(mocker);
+    TestHelper.mockedIdentityUpdateArmTemplates(mocker);
+    TestHelper.mockedSimpleAuthUpdateArmTemplates(mocker);
+    // Action
+    let result = await generateArmTemplate(mockedCtx, [
```
blackchoey, line 366: the second parameter is selected plugin, should it only conatins bot plugin in this test case?

blackchoey: for this test case, my understanding is we're testing the situation that when config.bicep and teamsfx folder already exists, adding another resource plugin that requires config will success with no error. So maybe we can generateArmTemplates(ctx,[tab,simpleauth]) first, then generateArmTemplates(ctx, [bot]) to test this scenario.

packages/fx-core/tests/plugins/solution/solution.arm.test.ts
blackchoey: thanks

packages/fx-core/tests/plugins/solution/solution.arm.test.ts
```diff
@@ -390,15 +334,155 @@ Mocked bot provision orchestration content. Module path: './provision/botProvisi
...
+      aadPlugin,
+      botPlugin,
```
blackchoey, line 368: the bot plugin should be fehostPlugin here?

wenytang-ms: thanks!

blackchoey (approved):

jayzhang (approved):

PR #3723, opened 2022-01-19
fix: select capability can go back

HuihuiWu-Microsoft:
Fix the P0 bug: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_sprints/taskboard/Teams%20Extensibility%20E2E%20Team/Microsoft%20Teams%20Extensibility/CY22-1.1?workitem=13023698

Root cause: actually there's an invisible question for creating new project. So the current step for select capability is 2 and back button is available.

### DIFF
--- a/packages/api/src/qm/visitor.ts
+++ b/packages/api/src/qm/visitor.ts
@@ -89,7 +89,7 @@ const questionVisitor = async function (
   totalSteps?: number
 ): Promise<Result<InputResult<any>, FxError>> {
   if (inputs[question.name] !== undefined) {
-    return ok({ type: "success", result: inputs[question.name] });
+    return ok({ type: "skip", result: inputs[question.name] });
   }
   if (question.type === "func") {
     try {


packages/api/src/qm/visitor.ts
```diff
@@ -89,7 +89,7 @@ const questionVisitor = async function (
   totalSteps?: number
 ): Promise<Result<InputResult<any>, FxError>> {
   if (inputs[question.name] !== undefined) {
-    return ok({ type: "success", result: inputs[question.name] });
+    return ok({ type: "skip", result: inputs[question.name] });
```
HuihuiWu-Microsoft, line 92: @jayzhang I was considering removing the create question from question model but has side effect since the download sample logic will also traverse the question model. So I have to keep the question model for create and make the question as 'skip' if the answer to the question is pre-set. Do you have any concern?

jayzhang (approved):

PR #3728, opened 2022-01-20
chore: adjust templates version and fix CD workflow condition

wenytang-ms:


.github/workflows/cd.yml
```diff
@@ -118,7 +118,7 @@ jobs:
           fi
       
       - name: sync up templates deps with sdk version for RC
-        if: ${{ contains(steps.version-change.outputs.CHANGED, '@microsoft/teamsfx') && !contains(steps.version-change.outputs.CHANGED, 'templates@') && github.event_name == 'workflow_dispatch' && github.ref == 'refs/heads/main' && github.event.inputs.preid == 'rc' }}
+        if: ${{ contains(steps.version-change.outputs.CHANGED, '@microsoft/teamsfx') && github.event_name == 'workflow_dispatch' && github.event.inputs.preid == 'rc' }}
```
qinezh, line 121: do we need to keep the check `github.ref == 'refs/heads/main'` for this step?

adashen (approved):

.github/workflows/cd.yml
wenytang-ms: the hotfix branch also needs, avoid too many condition, and these condition is validate at first

qinezh (approved):

PR #3729, opened 2022-01-20
perf(template): add connection timeout for fetching templates

hund030:
The timeout parameter in axios is a response timeout, which throw an error when server doesn't response in time. While what we need is a connection timeout, which throw an error when the request doesn't end in time.

## How was this change tested
1. Replace template url with 1GB size test file download link.
2. Run scaffold, plugin takes more than 2 minutes to download the test file.
```
[2022-01-20T03:01:26.038Z] [Info] - [fx-resource-frontend-hosting] Retrieving template from 'https://speed.hetzner.de/1GB.bin'.
[2022-01-20T03:03:59.687Z] [Info] - [fx-resource-frontend-hosting] Error: Invalid or unsupported zip format. No END header found
[2022-01-20T03:03:59.689Z] [Info] - [fx-resource-frontend-hosting] Failed to retrieve latest template from GitHub. Using local template instead.
```
3. Add the changes.
4. Run scaffold, plugin takes 30 seconds but request timeout, and fallback to using local templates.
```
[2022-01-19T08:53:27.252Z] [Info] - [fx-resource-frontend-hosting] Retrieving template from 'https://speed.hetzner.de/1GB.bin'.
[2022-01-19T08:53:57.254Z] [Info] - [fx-resource-frontend-hosting] Error: Request timeout
[2022-01-19T08:53:57.255Z] [Info] - [fx-resource-frontend-hosting] Failed to retrieve latest template from GitHub. Using local template instead.
```

### DIFF
--- a/packages/fx-core/src/common/template-utils/templatesActions.ts
+++ b/packages/fx-core/src/common/template-utils/templatesActions.ts
@@ -43,8 +43,8 @@ export interface ScaffoldAction {
   run: (context: ScaffoldContext) => Promise<void>;
 }
 
-const defaultTryLimits = 3;
-const defaultTimeoutInMs = 10000;
+const defaultTryLimits = 1;
+const defaultTimeoutInMs = 30000;
 
 const missKeyErrorInfo = (key: string) => `Missing ${key} in template action.`;
 
--- a/packages/fx-core/src/common/template-utils/templatesUtils.ts
+++ b/packages/fx-core/src/common/template-utils/templatesUtils.ts
@@ -2,7 +2,7 @@
 // Licensed under the MIT license.
 
 import AdmZip from "adm-zip";
-import axios, { AxiosResponse } from "axios";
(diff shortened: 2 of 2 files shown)


a1exwang (approved):
axios CancelToken is [deprecated after 0.22](https://axios-http.com/docs/cancellation). We are currently using 0.21. It would be better to fix that. But it is also okay to fix that later when we upgrading axios.

IvanJobs (commented):
Refer to this thread: https://github.com/axios/axios/issues/1739
Does the follow solution looks better?
image

hund030:
> axios CancelToken is [deprecated after 0.22](https://axios-http.com/docs/cancellation). We are currently using 0.21. It would be better to fix that. But it is also okay to fix that later when we upgrading axios.

Try to upgrade axios to 0.22 or 0.24, both get many compile errors. Let's stay at 0.21.

IvanJobs (approved):
lgtm

PR #3735, opened 2022-01-20
feat: run npm install automatically after scaffolding in vsc 

kuojianlu:
- add "Automatic NPM Install" setting
- add `automaticNpmInstall` state in core
- add automatic npm install (after scaffolding) handler

Wordins confirmed with Junjie.

This PR does not include:
- Run npm install automatically after adding resource/capability
- Automatic npm install for CLI

`automaticNpmInstallHandler` may be changed to support adding resource/capability afterwards.

image

image

image


packages/vscode-extension/src/debug/npmInstallHandler.ts
```diff
@@ -16,3 +32,47 @@ export async function runNpmInstallAll(projectRoot: string): Promise<void> {
...
+        StringResources.vsc.handlers.automaticNpmInstall,
+        false,
+        StringResources.vsc.handlers.disableAutomaticNpmInstall
+      ).then((selection) => {
+        if (
+          selection.isOk() &&
+          selection.value === StringResources.vsc.handlers.disableAutomaticNpmInstall
+        ) {
+          vscode.commands.executeCommand(
+            "workbench.action.openSettings",
```
swatDong, line 53: Which scope of settings is opened by default, user or workspace?

swatDong:
One potential conflict - what if user clicks F5 when the npm install tasks are still running?

packages/vscode-extension/package.json
```diff
@@ -807,6 +807,12 @@
             "type": "string",
             "description": "Set the default root directory for creating new teams app project. By default, the root directory is ${homeDir}/TeamsApps. (requries reload of VS Code)",
             "default": "${homeDir}/TeamsApps"
+          },
+          "fx-extension.automaticNpmInstall": {
```
swatDong, line 811: Any way to hide/disable this feature before releasing *validate-prerequisites* in `tasks.json`.
Current F5 still runs `npm install` which is a bit conflict with this feature.

packages/vscode-extension/src/debug/npmInstallHandler.ts
kuojianlu: User. Same behavior as change `Default Project Root Directory` setting.

kuojianlu:
> One potential conflict - what if user clicks F5 when the npm install tasks are still running?

The running npm install tasks have no side effect on F5. F5 will also trigger npm install tasks. The only releationship between them is that the automatic npm install tasks share the same start/end handler with thoses triggered by F5.

packages/vscode-extension/package.json
kuojianlu: Removing the setting here can simply disable this feature. Let me remove it.

swatDong:
> > One potential conflict - what if user clicks F5 when the npm install tasks are still running?
> 
> The running npm install tasks have no side effect on F5. F5 will also trigger npm install tasks. The only releationship between them is that the automatic npm install tasks share the same start/end handler with thoses triggered by F5.

That's great. Another conflict may be - what if user chooses to change project folder when the tasks are still running?

kuojianlu:
> > > One potential conflict - what if user clicks F5 when the npm install tasks are still running?
> > 
> > 
> > The running npm install tasks have no side effect on F5. F5 will also trigger npm install tasks. The only releationship between them is that the automatic npm install tasks share the same start/end handler with thoses triggered by F5.
> 
> That's great. Another conflict may be - what if user chooses to change project folder when the tasks are still running?

You mean to change the default project root dir in vscode settings? Changing this setting will just set the configuration, and it will take effect in next project creation.

swatDong (approved):

HuihuiWu-Microsoft (approved):

jayzhang (approved):

PR #3740, opened 2022-01-20
fix(local debug): remove simple auth related local settings when simple auth plugin is not activated

KennethBWSong:
related bug: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_queries/edit/13051456/?triage=true

### DIFF
--- a/packages/fx-core/tests/core/localSettings.test.ts
+++ b/packages/fx-core/tests/core/localSettings.test.ts
@@ -14,6 +14,7 @@ import {
   LocalSettingsBackendKeys,
   LocalSettingsBotKeys,
   LocalSettingsFrontendKeys,
+  LocalSettingsSimpleAuthKeys,
   LocalSettingsTeamsAppKeys,
 } from "../../src/common/localSettingsConstants";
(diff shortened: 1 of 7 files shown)


KennethBWSong:
E2E TEST: None

swatDong:
This simple auth key logic looks good to me, but have another question about the original auth keys, e.g., `clientSecret`.
My understanding is - after disabling simple auth plugin, `clientSecret` is not required for tab-only project. So is there any plan to **always generate the minimized local settings**? Correct me if I miss any auth case.

packages/fx-core/tests/core/localSettings.test.ts
```diff
@@ -187,6 +204,16 @@ describe("LocalSettings provider APIs", () => {
       chai.assert.isTrue(localSettings!.teamsApp?.has(key));
     }
 
+    // Verify Simple Auth related settings
+    if (hasSimpleAuth) {
```
blackchoey, line 208: we may also need to verify when simple auth is disabled, there's no simple auth related configs.

PR #3748, opened 2022-01-21
chore(localdebug): add log in port checker

XiaofuHuang:
Add log while detecting port.


kuojianlu (approved):

packages/fx-core/src/common/local/portChecker.ts
```diff
@@ -21,19 +21,23 @@ const botDebugPortRegex = /--inspect[\s]*=[\s"']*9239/im;
 const botDebugPorts = [9239];
 const botServicePorts = [3978];
 
-async function detectPortListening(port: number): Promise<boolean> {
+async function detectPortListening(port: number, logger?: LogProvider): Promise<boolean> {
   try {
+    logger?.info(`Start to detect port: ${port}`);
```
swatDong, line 26: Can we use `logger.debug` or any way to by-default hide the log?

XiaofuHuang: The debug level will always be ignored in vscode. Would you like to send the telemetry instead of showing the log to our users?

swatDong: I see. I'm ok with log or telemetry.

packages/fx-core/src/common/local/telemetry.ts
```diff
@@ -0,0 +1,48 @@
...
+
+import { FxError, TelemetryReporter, UserError } from "@microsoft/teamsfx-api";
+import { Component, TelemetryErrorType, TelemetryProperty, TelemetrySuccess } from "../telemetry";
+
+export enum TelemetryEvent {
+  DetectPortStart = "detect-port-start",
+  DetectPort = "detect-port",
+}
+
+export function sendTelemetryEvent(
```
swatDong, line 12: Why not directly use `sendTelemetryEvent` from "../telemetry"?

XiaofuHuang: ../telemetry use the static instance of TelemetryReporter, https://github.com/OfficeDev/TeamsFx/blob/6c049f00d86ca5b6fe4656ef9d64611642cec47e/packages/fx-core/src/common/telemetry.ts#L7
which is init in `FxCore`.
https://github.com/OfficeDev/TeamsFx/blob/6c049f00d86ca5b6fe4656ef9d64611642cec47e/packages/fx-core/src/core/index.ts#L186

`LocalEnvManager` is directly called from vsc / cli. So here an independent send telemetry method is implemented

swatDong: I think they are the same one. `LocalEnvManager` is always called after core is activated, so telemetry instance from core is the same one passed from vsc/cli.

It seems logProvider and ui are also duplicated from core, one further improvement could be using TOOLS from core and make `LocalEnvManager` static.

XiaofuHuang: Got it, changed to use `sendTelemetryEvent` in `../telemetry`

swatDong (approved):

kuojianlu (approved):

PR #3752, opened 2022-01-21
chore: support correlation-id

chagong:
chore: support correlation-id

https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/12924674

### DIFF
--- a/packages/server/src/serverConnection.ts
+++ b/packages/server/src/serverConnection.ts
@@ -4,7 +4,7 @@
 import { CancellationToken, MessageConnection } from "vscode-jsonrpc";
 
 import { FxError, Inputs, Void, Tools, Result } from "@microsoft/teamsfx-api";
-import { FxCore } from "@microsoft/teamsfx-core";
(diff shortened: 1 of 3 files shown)


packages/server/src/serverConnection.ts
```diff
@@ -55,47 +55,78 @@ export default class ServerConnection implements IServerConnection {
...
     token: CancellationToken
   ): Promise<Result<Void, FxError>> {
-    const res = await this.core.provisionResources(inputs);
+    const corrId = inputs.correlationId ? inputs.correlationId : "";
+    const res = await Correlator.runWithId(
+      corrId,
+      (params) => this.core.provisionResources(params),
+      inputs
+    );
+    console.log(res);
```
Alive-Fish, line 90: Can comment/delete this log? it makes this method more different from other methods

chagong: oh, sure. I forgot it.

Alive-Fish (approved):

PR #3757, opened 2022-01-24
feat(sdk): move configuration and auth to teamsfx class

tecton:
1. Remove global configuration and related APIs.
2. Add new class `TeamsFx` and integrate configuration and credential.
3. Modify other APIs to use TeamsFx instead of credential.

Please use file filter to ignore generated Markdown files.

- [x] API update
- [x] README update
- [x] Unit tests update

E2E TEST: enable later.


packages/sdk/src/core/teamsfx.browser.ts
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  public setCustomConfig(customConfig: Record<string, string>): TeamsFx {
+    for (const key of Object.keys(customConfig)) {
+      const value = customConfig[key];
+      if (value) {
+        this.configuration.set(key, value);
+      }
+    }
+    this.teamsUserCredential = undefined;
```
SLdragon, line 90: Why set teamsUserCredential to undefined

packages/sdk/src/core/teamsfx.browser.ts
```diff
@@ -0,0 +1,117 @@
...
+    this.loadFromEnv();
+  }
+
+  private loadFromEnv(): void {
+    let env: any;
+    if (window && (window as any).__env__) {
+      // testing purpose
+      env = (window as any).__env__;
+    } else {
+      env = process.env;
```
SLdragon, line 43: Did this work in pure browser environment? Because there is no process object

packages/sdk/src/core/teamsfx.browser.ts
SLdragon: This maybe introduce problem, for example:
```ts
const credential = teamsfx.getCredential();
teamsfx.setCustomConfig(config);
credential.login() // this credential may use old config
```

packages/sdk/src/core/teamsfx.ts
```diff
@@ -0,0 +1,217 @@
...
+   */
+  public setCustomConfig(customConfig: Record<string, string>): TeamsFx {
+    for (const key of Object.keys(customConfig)) {
+      const value = customConfig[key];
+      if (value) {
+        this.configuration.set(key, value);
+      }
+    }
+    this.oboUserCredential = undefined;
+    this.appCredential = undefined;
```
SLdragon, line 144: same here

packages/sdk/src/core/teamsfx.browser.ts
tecton: No, I think you mean the pure browser is importing SDK by <script> tag. This approach is not supported now so I assume SDK has the dependency of webpack to inject `process` object.

packages/sdk/src/core/teamsfx.browser.ts
tecton: Yes this could be a problem and what I think is to hide the direct usage of credential instance. Since `TeamsFx:getCredential()` returns `TokenCredential` which doesn't have `login()` function.
That's the reason to expose `TeamsFx:login()` to do user consent process. One possible fix is to keep the internal credential instance and update config of credential by adding a new interface like `TeamsUserCredential:updateConfig()`.

tecton: Fixed using second option.

SLdragon (dismissed):

SLdragon:
@tecton ,
Please ask wenyu help update CD pipeline to make sure that sdk version will not be updated automatically in template when daily build

packages/sdk/src/core/teamsfx.browser.ts
```diff
@@ -48,33 +48,33 @@ export class TeamsFx implements TeamsFxConfiguration {
...
       // testing purpose
-      env = (window as any).__env__;
+      const env = (window as any).__env__;
+      this.configuration.set("authorityHost", env.REACT_APP_AUTHORITY_HOST);
+      this.configuration.set("tenantId", env.REACT_APP_TENANT_ID);
+      this.configuration.set("clientId", env.REACT_APP_CLIENT_ID);
+      this.configuration.set("initiateLoginEndpoint", env.REACT_APP_START_LOGIN_PAGE_URL);
+      this.configuration.set("applicationIdUri", env.M365_APPLICATION_ID_URI);
+      this.configuration.set("apiEndpoint", env.REACT_APP_FUNC_ENDPOINT);
+      this.configuration.set("apiName", env.REACT_APP_FUNC_NAME);
```
SLdragon, line 60: Better extract these sentence to one function:

```
setConfigurationsFromEnv(env);
...
setConfigurationsFromEnv(process.env);
```

SLdragon (approved):

packages/sdk/src/core/teamsfx.browser.ts
tecton: Keeping `process.env.xxx` here to be replaced by webpack plugin.
https://webpack.js.org/plugins/environment-plugin/

PR #3762, opened 2022-01-24
chore: add telemetry for automatic npm install

kuojianlu:
@chagong Please help review the new telemetry events and properties.

### DIFF
--- a/packages/vscode-extension/src/debug/npmInstallHandler.ts
+++ b/packages/vscode-extension/src/debug/npmInstallHandler.ts
@@ -21,7 +21,13 @@ import { runTask } from "./teamsfxTaskHandler";
 import { createTask } from "./teamsfxTaskProvider";
 import VsCodeLogInstance from "../commonlib/log";
 import { ExtTelemetry } from "../telemetry/extTelemetry";
-import { TelemetryEvent } from "../telemetry/extTelemetryEvents";
+import {
+  TelemetryEvent,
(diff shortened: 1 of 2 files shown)


swatDong (approved):

packages/vscode-extension/src/debug/npmInstallHandler.ts
```diff
@@ -84,8 +104,24 @@ export async function automaticNpmInstallHandler(
...
           }
         });
+        const keys = tasks.keys();
+        const exitCodes = await Promise.all(tasks.values());
+        try {
+          const properties: { [key: string]: string } = {};
+          for (const exitCode of exitCodes) {
+            properties[keys.next().value] = exitCode + "";
+          }
+          ExtTelemetry.sendTelemetryEvent(TelemetryEvent.AutomaticNpmInstall, properties);
```
chagong, line 121: To me if this operation is always successful, there is no need to send out '-start' and 'exit' events.

kuojianlu: "start" and "exit" events are to record the event timestamp, which may be used afterwards in BI.
Let me make any non-zero exit code as error, and send error event instead. Thanks.

chagong (approved):

swatDong (approved):

HuihuiWu-Microsoft (approved):

PR #3769, opened 2022-01-25
chore: prepare custom local tasks

swatDong:
To be transparent, directly expose `npm run xxx` shell task in tasks.json, so extension only needs to handle such tasks, no need to provide/resolve such tasks.

### DIFF
--- a/packages/vscode-extension/src/debug/teamsfxTaskHandler.ts
+++ b/packages/vscode-extension/src/debug/teamsfxTaskHandler.ts
@@ -58,7 +58,6 @@ function isNpmInstallTask(task: vscode.Task): boolean {
 
 function isTeamsfxTask(task: vscode.Task): boolean {
   // teamsfx: xxx start / xxx watch
-  // teamsfx: dev / watch
   if (task) {
     if (
       task.source === ProductName &&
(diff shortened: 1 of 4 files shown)


packages/vscode-extension/src/debug/teamsfxTaskHandler.ts
```diff
@@ -75,11 +74,16 @@ function isTeamsfxTask(task: vscode.Task): boolean {
...
-          command.trim().toLowerCase() === "dev" ||
-          command.trim().toLowerCase() === "watch")
+          command.trim().toLocaleLowerCase().endsWith("watch"))
       );
     }
+
+    // dev:teamsfx and watch:teamsfx
+    const commandLine: string | undefined =
+      task.execution && (<vscode.ShellExecution>task.execution).commandLine;
+    if (commandLine !== undefined) {
```
kuojianlu, line 84: Is it possible that user uses command and args instead of full command line?

swatDong: Good catch. Yes it's possible, added `command` and `args` support.

kuojianlu (approved):

PR #3772, opened 2022-01-25
fix(localdebug): fix task handler

swatDong:
Two fixes:
1. Update ngrok task to `teamsfx` source. The original one somehow could be recognized as `source: shell` so not correctly killed by task handler.
2. Use string as the key of `allRunningTeamsfxTasks`. The original one `IRunningTeamsfxTask` does not have `===` overload so the map is keeping increasing.

### DIFF
--- a/packages/fx-core/src/plugins/solution/fx-solution/debug/util/tasks.ts
+++ b/packages/fx-core/src/plugins/solution/fx-solution/debug/util/tasks.ts
@@ -221,10 +221,7 @@ function startBot(): Record<string, unknown> {
 function startNgrok(): Record<string, unknown> {
   return {
     label: "start ngrok",
-    type: ProductName,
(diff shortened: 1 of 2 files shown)


kuojianlu:
Curious why `ngrok start` could be recognized as source shell since its type is `teamsfx`?

swatDong:
> Curious why `ngrok start` could be recognized as source shell since its type is `teamsfx`?

I noticed this when debugging the extension. Just guess - vscode gets confused since `ngrok start` is defined twice. One from taskProvider and another from tasks.json. Following code seems to define a task rather than referring a task:
``` json
        {
            "label": "Pre Debug Check",
            "type": "teamsfx",
            "command": "ngrok start",
            ...
        },
```

kuojianlu (approved):

PR #3773, opened 2022-01-25
refactor(frontend): refine log and progress

hund030:


packages/fx-core/src/plugins/resource/frontend/dotnet/plugin.ts
```diff
@@ -60,29 +56,27 @@ export interface DotnetPluginConfig {
...
-    this.config.subscriptionId = solutionConfig?.get(DependentPluginInfo.subscriptionId) as string;
+    config.resourceGroupName = solutionConfig?.get(DependentPluginInfo.resourceGroupName) as string;
+    config.subscriptionId = solutionConfig?.get(DependentPluginInfo.subscriptionId) as string;
 
-    this.config.webAppName = ctx.config.get(ConfigInfo.webAppName) as string;
-    this.config.appServicePlanName = ctx.config.get(ConfigInfo.appServicePlanName) as string;
-    this.config.projectFilePath = ctx.config.get(ConfigInfo.projectFilePath) as string;
+    config.webAppName = ctx.config.get(ConfigInfo.webAppName) as string;
+    config.appServicePlanName = ctx.config.get(ConfigInfo.appServicePlanName) as string;
+    config.projectFilePath = ctx.projectSettings?.pluginSettings?.projectFilePath as string;
```
IvanJobs, line 69: Nit: 
I'm curious about the change of getting projectFilePath. I guess both the two ways of getting projectFilePath are existing, but the older one is to be deprecated?

packages/fx-core/src/plugins/resource/frontend/dotnet/plugin.ts
```diff
@@ -60,29 +56,27 @@ export interface DotnetPluginConfig {
 }
 
 export class DotnetPluginImpl implements PluginImpl {
-  config: DotnetPluginConfig = {};
-
-  private syncConfigFromContext(ctx: PluginContext): void {
```
IvanJobs, line 65: Nit: Good change to remove the data member of config and sync/generate new config in runtime by needs.
It will get rid of the side-effect risks when plugins store their own states.

IvanJobs (approved):
lgtm

packages/fx-core/src/plugins/resource/frontend/dotnet/plugin.ts
hund030: No, pluginFilePath is irrelevant to environment. It's better to store it in projectSettings rather than envState.

packages/fx-core/src/plugins/resource/frontend/dotnet/plugin.ts
```diff
@@ -177,24 +178,25 @@ export class DotnetPluginImpl implements PluginImpl {
   }
 
   public async deploy(ctx: PluginContext): Promise<TeamsFxResult> {
-    Logger.info(Messages.StartDeploy(PluginInfo.displayName));
+    Logger.info(Messages.StartDeploy);
+    await ProgressHelper.startProgress(ctx, DeployProgress);
 
-    this.syncConfigFromContext(ctx);
```
eriolchan, line 182: When we create config as local variable, shall we remove method of syncConfigFromContext()?

packages/fx-core/src/plugins/resource/frontend/utils/progress-helper.ts
```diff
@@ -4,131 +4,30 @@
...
+    await this.progressHandler?.end(true);
 
-    this.postProvisionProgress = ctx.ui?.createProgressBar(
-      Messages.PostProvisionProgressTitle,
-      Object.entries(PostProvisionSteps).length
+    this.progressHandler = ctx.ui?.createProgressBar(
+      progress.title,
+      Object.entries(progress.steps).length
     );
-    await this.postProvisionProgress?.start(Messages.ProgressStart);
```
eriolchan, line 60: Good to see this kind of refactor to make code clean and simple, thanks for your effort!

eriolchan (approved):

packages/fx-core/src/plugins/resource/frontend/dotnet/plugin.ts
hund030: The naming could be updated, getConfigFromContext may be better?

PR #3776, opened 2022-01-25
chore(local debug): get env for func task

swatDong:
Add a new hidden command `fx-extension.append-func-path` to be used by tasks.json.

``` json
{
  "label": "start",
  "type": "shell",
  "command": "func start",
  "options": {
    "env": {
      "PATH": "${env:PATH}${command:fx-extension.append-func-path}"
    }
  }
}
```

### DIFF
--- a/packages/vscode-extension/package.json
+++ b/packages/vscode-extension/package.json
@@ -628,26 +628,11 @@
       {
         "type": "teamsfx",
         "required": [
-          "command",
-          "component"
+          "command"
         ],
         "properties": {
           "command": {
-            "type": "string",
-            "description": "The command to be executed on a teamsfx component.",
-            "enum": [
-              "dev",
-              "watch"
-            ]
-          },
-          "component": {
-            "type": "string",
-            "description": "The teamsfx component.",
-            "enum": [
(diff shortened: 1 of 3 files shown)


swatDong:
@1openwindow @HuihuiWu-Microsoft - this change is for local debug (tasks.json) but added a hidden command to extension. Please help review.

packages/vscode-extension/package.json
```diff
@@ -490,6 +490,11 @@
         "title": "Teams - Validate Local Prerequisites",
         "enablement": "never"
       },
+      {
+        "command": "fx-extension.append-func-path",
```
kuojianlu, line 494: I think `get-func-path` is better since this command is just to get the path of func managed/detected by our toolkit. The appending operation is done in tasks.json but not in this command.

swatDong: Good idea. Updated to `get-func-path`

kuojianlu (dismissed):

kuojianlu (approved):

HuihuiWu-Microsoft (approved):

PR #3778, opened 2022-01-25
feat: run npm install automatically after scaffolding in cli

kuojianlu:
- add automatic npm install handler
- add telemetry event and property for automatic npm install
- update new command to support automatic npm install

### DIFF
--- a/packages/cli/src/cmds/preview/npmInstallHandler.ts
+++ b/packages/cli/src/cmds/preview/npmInstallHandler.ts
@@ -1,0 +1,120 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+"use strict";
+
+import { LogLevel } from "@microsoft/teamsfx-api";
+import {
+  ITaskDefinition,
+  LocalEnvManager,
+  ProjectSettingsHelper,
+  TaskDefinition,
+} from "@microsoft/teamsfx-core";
+import { spawn } from "child_process";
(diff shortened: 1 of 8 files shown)


packages/cli/src/cmds/preview/npmInstallHandler.ts
```diff
@@ -0,0 +1,120 @@
...
+      return false; // TODO: make automatic-npm-install enabled by default
+    }
+    return config[automaticNpmInstallOption] !== CliConfigAutomaticNpmInstall.Off;
+  } catch (error: any) {
+    cliLogger.warning(`Getting automatic-npm-install setting failed: ${error}`);
+    return false; // TODO: make automatic-npm-install enabled by default
+  }
+}
+
+async function runTask(task: ITaskDefinition): Promise<number | null> {
```
swatDong, line 108: What's this different from `preview/task.ts`? Or, how about just reuse `preview/task.ts task.wait()`

kuojianlu: `preview/task.ts` is not that extensible to reuse... it hides stdout, which is required by auto npm install command to print npm logs. Considered to refactor `preview/task.ts` to support this case afterwards.

swatDong: I see, make sense.

swatDong (approved):

Alive-Fish (approved):

PR #3789, opened 2022-01-26
feat(api): make solution settings optional

jayzhang:
To support existing app feature, solution settings is now optional


LongOddCode (approved):

packages/fx-core/src/core/index.ts
```diff
@@ -726,7 +733,14 @@ export class FxCore implements v3.ICore {
   ): Promise<Result<Void, FxError>> {
     currentStage = Stage.provision;
     inputs.stage = Stage.provision;
-    if (!ctx || !ctx.solutionV2 || !ctx.contextV2 || !ctx.envInfoV2) {
+    if (!ctx || !ctx.projectSettings) {
```
SLdragon, line 736: how about simplify to `!ctx?.projectSettings`

jayzhang: OK

KennethBWSong (approved):

SLdragon (approved):

LongOddCode (approved):

nliu-ms (approved):

packages/fx-core/src/core/index.ts
```diff
@@ -265,52 +266,58 @@ export class FxCore implements v3.ICore {
...
-      );
-      if (createEnvResult.isErr()) {
-        return err(createEnvResult.error);
-      }
+      if (isCreatedFromExistingApp(inputs)) {
+        //TODO create from existing Tab or Bot/ME
+        // 1. call App Studio V3 API to create manifest with placeholder
+        // 2. create config.local.json to store existing App information
+      } else {
+        // there is no solution settings if created from existing app
```
dooriya, line 277: Looks like this comment is for the `if` case, which is created from existing app.

jayzhang: Yes, will be moved later

dooriya (approved):

PR #3797, opened 2022-01-27
chore: setup doctor output of vscode for new F5

meifans:


packages/vscode-extension/src/debug/depsChecker/doctorConstant.ts
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+export const doctorConstant = {
+  Tick: "√",
+  Cross: "×",
+  WhiteSpace: "   ",
```
swatDong, line 7: Why three spaces?

packages/vscode-extension/src/debug/depsChecker/doctorConstant.ts
```diff
@@ -0,0 +1,12 @@
...
+  NodeNotFound: `Cannot find Node.js. `,
```
swatDong, line 8: Is the suffix space redundant?

packages/vscode-extension/src/debug/depsChecker/doctorLogger.ts
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import commonlibLogger, { VsCodeLogProvider } from "../../commonlib/log";
+import { OutputChannel } from "vscode";
+import { DepsLogger } from "@microsoft/teamsfx-core";
+
+export class DoctorLogger implements DepsLogger {
```
swatDong, line 8: The log logic seems a bit duplicated. Two loggers are added, `vsc/DoctorLogger` and `core/EmptyLogger`, and the goal is to only output necessary logs. So could the two loggers be merged into one place, since there's already a `doctor` flag?
E.g., 
- Remove `vsc/DoctorLogger` and let `DepsManager` to control which logs to output via `doctor` flag.
- Remove `core/EmptyLooger` and let each checker support `undefined` logger.

packages/vscode-extension/src/debug/prerequisitesHandler.ts
```diff
@@ -224,41 +240,54 @@ async function checkDependencies(
...
     await localEnvManager.resolveLocalCertificate(trustDevCert);
-    return undefined;
-  } catch (error: any) {
-    return {
-      checker: "Local Certificate",
-      error: assembleError(error),
-    };
+  } catch (err: any) {
+    result = true;
+    error = assembleError(err);
```
XiaofuHuang, line 273: If trust certificate failes but debug continues, can we use a warning instead of success?
[!] Local Certification
error message ......

Or we can stop the debugger when trust cert failed.

packages/vscode-extension/src/debug/depsChecker/doctorConstant.ts
meifans: It's a proposal and maybe change to others after collecting feedback.

packages/vscode-extension/src/debug/depsChecker/doctorConstant.ts
meifans: Good catch, will fix it.

packages/vscode-extension/src/debug/depsChecker/doctorLogger.ts
meifans: Well, control log in each checker is fine. But have to handle `cpUtils / test` for `undefined` logger.
My consideration is that current F5 will be replaced with `new F5(with doctor)` and at that point we can just delete `core/EmptyLogger` and `core/VSCodeLogger`(maybe) without changing logic in each checker.

Further, if the output of depManager(and it's checkers) in VSCode and CLI and VS is same, we can move doctorLogger to DepsManager as you said. but so far, I am not sure about that.

packages/vscode-extension/src/debug/prerequisitesHandler.ts
meifans: This is my mistake, I didn't intend to change the logic here. we should stop the debugger like before.

meifans: Fixed.

packages/vscode-extension/src/debug/depsChecker/doctorLogger.ts
swatDong: I see, so `EmptyLogger` sounds good to reduce maintenance effort on each checker.

But `DoctorLogger` still looks weird. I noticed that the only usage of `DoctorLogger` is to append message, so why not directly use `vsc/depsChecker/VSCodeLogger`? My concern is, **since `DoctorLogger` does not implement log-level methods, it brings extra maintenance effort on `core/DepsManager`** - calling `this.logger.info/warning/error` has no output.

New F5 does use doctor, but for backward compatibility, we'll still support original F5 and keep `vsc/depsChecker/VSCodeLogger`, so I feel like no need to create new `DoctorLogger` and could just leverage `VSCodeLogger`.

meifans: I agree. and will remove DoctorLogger.

meifans: Removed.

swatDong (approved):

XiaofuHuang (approved):

PR #3799, opened 2022-01-27
feat(solution): provision for minimized teamsfx project

jayzhang:
1. feat: support provision of minimized teamsfx project without solutionSettings
2. refactor: env.state will not be converted to {output:{}, secret:{}} structure


packages/api/src/v3/resourceStates.ts
```diff
@@ -53,6 +53,7 @@ export interface AzureSolutionConfig extends Json {
   location: string;
   provisionSucceeded: boolean;
   needCreateResourceGroup: boolean;
+  teamsAppTenantId: string;
```
qinezh, line 56: From my understanding, the tenant id of Teams app should belong to M365 instead of Azure resource?

packages/api/src/v2/resourcePlugin.ts
```diff
@@ -114,17 +114,15 @@ export interface ResourcePlugin {
    *
    * @param {Context} ctx - plugin's runtime context shared by all lifecycles.
    * @param {ProvisionInputs} inputs - inputs injected by Toolkit runtime and solution.
-   * @param {DeepReadonly<EnvInfoV2>} envInfo - a readonly view of environment info modeled after (config|state).${env}.json
+   * @param {EnvInfoV2} envInfo - a readonly view of environment info modeled after (config|state).${env}.json
```
qinezh, line 117: Curious why the env info no longer need to be readonly?

packages/api/src/v2/resourcePlugin.ts
```diff
@@ -135,17 +133,17 @@ export interface ResourcePlugin {
...
+   * @returns {Json} resource provision output which will be persisted by the toolkit into envInfo's state.
    */
   configureResource?: (
     ctx: Context,
     inputs: ProvisionInputs,
-    envInfo: DeepReadonly<EnvInfoV2>,
+    envInfo: EnvInfoV2,
     tokenProvider: TokenProvider
-  ) => Promise<Result<ResourceProvisionOutput, FxError>>;
+  ) => Promise<Result<Void, FxError>>;
```
qinezh, line 146: the comment in L139 mentioned the return value is Json, but here is Void?

packages/fx-core/src/core/index.ts
```diff
@@ -41,6 +41,7 @@ import { assign } from "lodash";
 import * as path from "path";
 import { Container } from "typedi";
 import * as uuid from "uuid";
+import { resourceLimits } from "worker_threads";
```
qinezh, line 44: Seems the module is imported by accident?

packages/api/src/v3/resourceStates.ts
jayzhang: The naming of AzureSolutionConfig is somehow not accurate, you can ignore it for now.

packages/fx-core/src/plugins/resource/aad/v2/index.ts
```diff
@@ -52,7 +52,7 @@ export class AadPluginV2 implements v2.ResourcePlugin {
     inputs: v2.ProvisionInputs,
     envInfo: Readonly<v2.EnvInfoV2>,
     tokenProvider: TokenProvider
-  ): Promise<Result<v2.ResourceProvisionOutput, FxError>> {
+  ): Promise<Result<Json, FxError>> {
```
KennethBWSong, line 55: Need to change to void.
After offline discussion, @jayzhang will fix in another pr.

KennethBWSong (dismissed):
Approve for aad and simple auth. One change need to be fixed in another pr.

packages/api/src/v2/resourcePlugin.ts
jayzhang: Preview API, each plugin will return provisioned output, which will make the API v2 more complicated.
From practice, EnvInfoV2 is kept as both input and output (by reference) is more easy to work than value copy from return

packages/api/src/v3/resourceStates.ts
qinezh: Would it be better to add a comment here to avoid confusion from others while they read this part of code?

jayzhang: This key is redundant and just to keep compatibility to current state file.

packages/fx-core/src/core/middleware/envInfoWriter.ts
```diff
@@ -50,15 +50,15 @@ async function writeEnvInfo(ctx: CoreHookContext, skip: boolean) {
...
+  const state = envInfoV2.state;
+  if (state === undefined) return;
   // DO NOT persist local debug plugin config.
-  if (provisionOutputs[PluginNames.LDEBUG]) {
-    delete provisionOutputs[PluginNames.LDEBUG];
+  if (state[PluginNames.LDEBUG]) {
+    delete state[PluginNames.LDEBUG];
   }
-  const envState = flattenConfigJson(provisionOutputs);
+  // const envState = flattenConfigJson(provisionOutputs);
```
dooriya, line 59: Can we just remove this unused code instead of commenting it out?

hund030 (approved):
Approve for frontend and function plugin.

nliu-ms (dismissed):

xzf0587 (approved):

XiaofuHuang (approved):

XiaofuHuang:
Approved for APIM plugin

nliu-ms (approved):

xzf0587 (approved):

hund030 (approved):

KennethBWSong (approved):

IvanJobs (approved):
lgtm for the bot part, assume dev test is passed.

PR #3800, opened 2022-01-27
chore: add data loss test case

1openwindow:


HuihuiWu-Microsoft:
Will you add other test cases as well to make sure this issue not happen again?

1openwindow:
> 

this is other topic. Since we didn't have over E2E telemetry test, it will also introduce additional dependency, such as Data Query for different environment. I would like to consider it in E2E telemetry test.

HuihuiWu-Microsoft (approved):

PR #3802, opened 2022-01-27
feat: add vs platform scaffolding blazor project

JerryYangKai:
Including Tab and Bot capability

**Hint !!**

This pull request looks very big because it includes all the template files, I recommend reviewer to focus on the first several files in **packages** folder path, this is the logic of scaffolding, the following files which is in **template** path are the template files.


packages/fx-core/src/plugins/resource/frontend/dotnet/ops/scaffold.ts
```diff
@@ -0,0 +1,79 @@
...
+        if (action.name === ScaffoldActionName.FetchTemplatesUrlWithTag) {
+          Logger.info(Messages.getTemplateFrom(context.zipUrl ?? Constants.emptyString));
+        }
+      },
+      onActionError: async (action: ScaffoldAction, context: ScaffoldContext, error: Error) => {
+        Logger.info(error.toString());
+        switch (action.name) {
+          case ScaffoldActionName.FetchTemplatesUrlWithTag:
+          case ScaffoldActionName.FetchTemplatesZipFromUrl:
+            TelemetryHelper.sendScaffoldFallbackEvent(new TemplateManifestError(error.message));
```
eriolchan, line 58: Why do we send scaffold fallback event when fetching templates zip from url?

.github/scripts/template-zip-autogen.sh
```diff
@@ -12,6 +13,9 @@ TEMPLATE_LIST=(
     bot.default
     msgext.default
     bot-msgext.default
+    blazor-base.tab
```
eriolchan, line 16: We can offline discuss about the template name convention with @hund030

.github/scripts/template-zip-autogen.sh
```diff
@@ -35,22 +39,46 @@ for LANGUAGE in ${LANGUAGE_LIST[@]}; do
...
             echo "The folder ./templates/${SCOPE}/${LANGUAGE}/${SCENARIO} does not exist."
-            exit -1
+            continue
         fi        
         
-        # Generate code from Mustache templates
-        if [ ${SCOPE} == "bot" ]; then           
-            IS_BOT=true mo ${TEMPLATE_FILE_PREFIX}.${LANGUAGE}.mustache > ./templates/${SCOPE}/${LANGUAGE}/${SCENARIO}/teamsBot.${LANGUAGE}
+        # Generate code from Mustache templates for js and ts
+        if [ ${SCOPE} == "bot" ] && [ ${LANGUAGE} != "csharp" ]; then
```
eriolchan, line 46: I don't quite understand this script, can you help me learn more about it offline?

packages/fx-core/src/plugins/resource/frontend/dotnet/plugin.ts
```diff
@@ -93,6 +96,25 @@ export class DotnetPluginImpl implements PluginImpl {
   }
 
   public async scaffold(ctx: PluginContext): Promise<TeamsFxResult> {
+    Logger.info(Messages.StartScaffold(PluginInfo.displayName));
+
+    const selectedCapabilities = (ctx.projectSettings?.solutionSettings as AzureSolutionSettings).capabilities;
+    const includeTab = selectedCapabilities.includes(DotnetSupportCapability.tabCapability);
+    const includeBot = selectedCapabilities.includes(DotnetSupportCapability.botCapability);
+
+    if (includeTab) {
```
eriolchan, line 105: need to confirm with you the logic offline

packages/fx-core/src/plugins/resource/frontend/dotnet/resources/errors.ts
```diff
@@ -65,6 +65,47 @@ export class ProjectPathError extends DotnetPluginError {
   }
 }
 
+export class UnknownScaffoldError extends FrontendPluginError {
```
eriolchan, line 68: Is this possible to share the same error of frontend plugin instead of create the similar one?

templates/.eslintignore
```diff
@@ -0,0 +1 @@
+tab/csharp/default/*
```
eriolchan, line 1: why add this ignore file?

templates/blazor-base/csharp/tab/GettingStarted.txt
```diff
@@ -0,0 +1,40 @@
+Welcome to the Teams Toolkit! Here are some tips to get started building your app.
```
eriolchan, line 1: Looks like this GettingStarted.txt is the same as Bot one. Can we share them instead of duplicating them?

PR #3803, opened 2022-01-27
feat: add vs platform scaffolding blazor project

JerryYangKai:
Including Tab and Bot capability

**Hint !!**

This pull request looks very big because it includes all the template files, I recommend reviewer to focus on the first several files in **packages** folder path, this is the logic of scaffolding, the following files which is in **template** path are the template files.

E2E TEST: not available.


templates/mustache-templates/appsettings.Development.json.mustache
```diff
@@ -0,0 +1,31 @@
+{
+  "Logging": {
+    "LogLevel": {
+      "Default": "Information",
+      "Microsoft": "Warning",
+      "Microsoft.Hosting.Lifetime": "Information"
+    }
+  },
+  "AllowedHosts": "*"
+{{#IS_TAB}}
```
hund030, line 10: Can we write them in a same line?
```
"AllowedHosts": "*"{{#IS_TAB}},
```

.github/scripts/template-zip-autogen.sh
```diff
@@ -35,22 +35,26 @@ for LANGUAGE in ${LANGUAGE_LIST[@]}; do
...
         fi
 
         cd ./templates/${SCOPE}/${LANGUAGE}/${SCENARIO}
         zip -rq ../../../../${SCOPE}.${LANGUAGE}.${SCENARIO}.zip .
         cd -
     done
-done
\ No newline at end of file
+done
+
```
hund030, line 57: Extra blank lines.

hund030 (approved):

packages/fx-core/src/plugins/resource/frontend/dotnet/plugin.ts
```diff
@@ -93,6 +96,30 @@ export class DotnetPluginImpl implements PluginImpl {
...
+    const templateVariable: TemplateVariable = { BlazorAppServer: projectName };
+
+    if (includeTab) {
+      templateVariable.IS_TAB = "true";
+      const templateInfo = new TemplateInfo(ctx, Group.Tab, templateVariable);
+      await Scaffold.scaffoldFromZipPackage(ctx.root, templateInfo);
+    }
+    if (includeBot) {
+      templateVariable.IS_BOT = "true";
+      const templateInfo = new TemplateInfo(ctx, Group.Bot, templateVariable);
```
eriolchan, line 116: Is that expected that if we choose tab & bot, then in L116 templateVariable has both IS_TAB and IS_BOT is true? These two flags are not isolated according to this code. So it looks like templateVariable cannot be shared?

.github/detect/excludes.txt
```diff
@@ -3,9 +3,12 @@
 *.png
 *.gif
 *.svg
-*.png
-*.sppkg
 *.ico
+*.woff
+*.ttf
+*.otf
+*.eot
```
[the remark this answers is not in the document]
JerryYangKai: I checked with wenyu, this is the suggested path to add ignore pattern

.github/scripts/template-zip-autogen.sh
JerryYangKai: Fixed

packages/fx-core/src/plugins/resource/frontend/dotnet/ops/scaffold.ts
```diff
@@ -0,0 +1,79 @@
...
+import {
+  genTemplateRenderReplaceFn,
+  ScaffoldAction,
+  ScaffoldActionName,
+  ScaffoldContext,
+  scaffoldFromTemplates,
+} from "../../../../../common/template-utils/templatesActions";
+import { TemplateInfo } from "../resources/templateInfo";
+
+export type Manifest = {
```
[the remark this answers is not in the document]
JerryYangKai: I will removed it.

packages/fx-core/src/plugins/resource/frontend/dotnet/ops/scaffold.ts
```diff
@@ -0,0 +1,79 @@
...
+    [key: string]: {
+      [key: string]: {
+        version: string;
+        url: string;
+      }[];
+    };
+  };
+};
+
+export class DotnetScaffold {
```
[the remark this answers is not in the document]
JerryYangKai: Good catch!

templates/.prettierignore
```diff
@@ -0,0 +1 @@
+tab/csharp/default/*
```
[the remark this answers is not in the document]
JerryYangKai: I have checked with wenyu, if we do not want to check the static files like these templates, it's OK to add ignore here.

packages/fx-core/src/plugins/resource/frontend/dotnet/resources/errors.ts
```diff
@@ -65,6 +65,47 @@ export class ProjectPathError extends DotnetPluginError {
   }
 }
 
+export class UnknownScaffoldError extends DotnetPluginError {
```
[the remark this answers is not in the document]
JerryYangKai: For now, it is OK to share error. But for further, every plugin has its own error seems follow the core design.

templates/tab/csharp/default/wwwroot/css/open-iconic/ICON-LICENSE
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
```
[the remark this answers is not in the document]
JerryYangKai: This part is provided by PM side, I will add a task to double check the template files with John

packages/fx-core/src/plugins/resource/frontend/dotnet/plugin.ts
JerryYangKai: Sure, this templateVariable will add the parameters as we expected.

packages/fx-core/src/plugins/resource/frontend/dotnet/plugin.ts
```diff
@@ -93,6 +96,30 @@ export class DotnetPluginImpl implements PluginImpl {
   }
 
   public async scaffold(ctx: PluginContext): Promise<TeamsFxResult> {
+    Logger.info(Messages.StartScaffold(PluginInfo.displayName));
```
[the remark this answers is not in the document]
JerryYangKai: Looks the code maybe it is not supporting this way, this is just my thought. Waiting for Zhijie to confirm.

packages/fx-core/src/plugins/resource/frontend/dotnet/resources/templateInfo.ts
```diff
@@ -24,7 +19,33 @@ export class TemplateInfo {
...
+  const templateInfoList: TemplateInfo[] = [];
+
+  selectedCapabilities.forEach((capability) => {
+    if (capability in TemplateInfo.SupportCapability) {
+      templateInfoList.push(new TemplateInfo(ctx, capability.toLowerCase(), templateVariable));
+    }
+  });
+
+  // Generate templateInfo for base scenrio.
+  if (selectedCapabilities.includes(Capability.tab)) {
```
eriolchan, line 41: Just curious why we use "in" in L35, but use includes() here, given both SupportCapability and selectedCapabilities are List?

packages/fx-core/src/plugins/resource/frontend/dotnet/resources/templateInfo.ts
```diff
@@ -24,7 +19,33 @@ export class TemplateInfo {
     this.scenario = TemplateInfo.DefaultScenario;
   }
 
+  static readonly SupportCapability = [Capability.tab, Capability.bot];
```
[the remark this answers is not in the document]
JerryYangKai: Good catch, fixed.

packages/fx-core/src/plugins/resource/frontend/dotnet/resources/templateInfo.ts
JerryYangKai: Fixed. I changed all the code from 'in' to 'include' to make them align. Since 'include' is the typescript style.

eriolchan: "in" is not correct here. Actually Array.includes()checks for the existence of a value in an array, while the in operator checks for the existence of a key in an object (or index in the case of arrays). Please check [this link](https://stackoverflow.com/questions/48499181/what-is-the-difference-between-in-operator-and-includes-for-javascript-arr) for more details.
This could be validated by unit test case, so please add some unit test after this PR before we releasing it.

packages/fx-core/src/plugins/resource/frontend/dotnet/resources/templateInfo.ts
```diff
@@ -0,0 +1,51 @@
...
+    if (TemplateInfo.SupportedCapabilities.includes(capability)) {
+      templateInfoList.push(new TemplateInfo(ctx, capability.toLowerCase(), templateVariable));
+    }
+  });
+
+  // Generate templateInfo for base scenrio.
+  if (selectedCapabilities.includes(Capability.tab)) {
+    templateVariable.IS_TAB = "true";
+  }
+  if (selectedCapabilities.includes(Capability.bot)) {
```
eriolchan, line 44: For this method, it could be better to use map instead of if condition to make the code be more generic and extensible for future requirement (e.g. messaging extension support).
```
const supportedCapabilities = new Map<string, string>([
  ["Tab", "IS_TAB"],
  ["Bot", "IS_BOT"],
]);
...

  const variables: string[] = [];

  selectedCapabilities.forEach((capability) => {
    if (supportedCapabilities.has(capability)) {
      templateInfoList.push(new TemplateInfo(ctx, capability.toLowerCase(), templateVariable));
      variables.push(supportedCapabilities.get(capability));
    }
  });
  // add variables to template variable
  variables.forEach(v => templateVariable[v] = "true");
  templateInfoList.push(new TemplateInfo(ctx, TemplateInfo.BaseGroup, templateVariable);

  return templateInfoList;
```

eriolchan: In this way, there is not if condition about tab/bot capability so that it is more generic as framework code. If we need to support messaging extension in the future. The only code change is to add key-value pair to supportedCapabilities.
```
const supportedCapabilities = new Map<string, string>([
  ["Tab", "IS_TAB"],
  ["Bot", "IS_BOT"],
  ["MessagingExtension", "IS_MSG"],  <= only add one code here!
]);
```

packages/fx-core/src/plugins/resource/frontend/dotnet/resources/templateInfo.ts
JerryYangKai: Sure!

packages/fx-core/src/plugins/resource/frontend/dotnet/plugin.ts
```diff
@@ -87,6 +95,15 @@ export class DotnetPluginImpl implements PluginImpl {
   }
 
   public async scaffold(ctx: PluginContext): Promise<TeamsFxResult> {
+    Logger.info(Messages.StartScaffold);
+
+    const selectedCapabilities = (ctx.projectSettings?.solutionSettings as SolutionSettings)
+      .capabilities;
+    generateTemplateInfos(selectedCapabilities, ctx).forEach(async (templateInfo: TemplateInfo) => {
```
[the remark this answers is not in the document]
JerryYangKai: Great learning from this part.

eriolchan (approved):

PR #3812, opened 2022-01-28
fix: errorType should not be changed if statusCode = undefined

IvanJobs:
to finish: https://dev.azure.com/msazure/Microsoft%20Teams%20Extensibility/_workitems/edit/13026392

errorType shouldn't be changed if statusCode = undefined, instead, respect the original errorType.

+ update code owners also.
1. Refactor inferInnerError as a external function.
2. Infer Graph Error in CreateAADError and CreateAADSecretError by need.
3. Add UT and E2E cases.

E2E TEST: https://github.com/OfficeDev/TeamsFx/blob/b9b0d58ec78d29044008cd64b49cb6c2406b9ba5/packages/cli/tests/e2e/bot/CommandExecutionIsUserError.ts#L35

### DIFF
--- a/packages/fx-core/src/plugins/resource/bot/errors.ts
+++ b/packages/fx-core/src/plugins/resource/bot/errors.ts
@@ -57,7 +57,7 @@ export class PluginError extends Error {
     if (helpLink) this.helpLink = helpLink;
 
     const statusCode = this.innerError.response?.status;
-    if (!statusCode) return ;
+    if (!statusCode) return;
     if (
       statusCode >= Constants.statusCodeUserError &&
       statusCode < Constants.statusCodeServerError
--- a/packages/fx-core/src/plugins/resource/bot/errors.ts
+++ b/packages/fx-core/src/plugins/resource/bot/errors.ts
@@ -57,8 +57,8 @@ export class PluginError extends Error {
     if (helpLink) this.helpLink = helpLink;
 
     const statusCode = this.innerError.response?.status;
+    if (!statusCode) return ;
     if (
-      statusCode &&
       statusCode >= Constants.statusCodeUserError &&
       statusCode < Constants.statusCodeServerError
     ) {
--- a/packages/fx-core/src/plugins/resource/bot/errors.ts
+++ b/packages/fx-core/src/plugins/resource/bot/errors.ts
@@ -14,6 +14,25 @@ export enum ErrorType {
   System,
 }
 
+function InferInnerError(target: PluginError, helpLinkMap: Map<string, string>): void {
+  if (!target.innerError) return;
+
+  const errorCode = target.innerError.response?.data?.error?.code;
+  const helpLink = helpLinkMap.get(errorCode);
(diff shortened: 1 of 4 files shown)


packages/fx-core/src/plugins/resource/bot/errors.ts
```diff
@@ -14,6 +14,22 @@ export enum ErrorType {
...
+function InferInnerError(target: PluginError, helpLinkMap: Map<string, string>): void {
+  if (!target.innerError) return;
+
+  const errorCode = target.innerError.response?.data?.error?.code;
+  const helpLink = helpLinkMap.get(errorCode);
+  if (helpLink) target.helpLink = helpLink;
+
+  const statusCode = target.innerError.response?.status;
+  if (!statusCode) return;
+  if (statusCode >= Constants.statusCodeUserError && statusCode < Constants.statusCodeServerError) {
```
hund030, line 26: NIT: I don't think 400 and 500 are magic numbers here, and I prefer use them directly in code.
`statusCode >= 400 && statusCode < 500 `, which is more readable for me.

hund030 (approved):

packages/fx-core/src/plugins/resource/bot/errors.ts
```diff
@@ -112,12 +108,14 @@ export class AADAppCheckingError extends PluginError {
 export class CreateAADAppError extends PluginError {
   constructor(innerError?: any) {
     super(ErrorType.User, CreateAppError.name, CreateAppError.message(), [], innerError);
+    InferInnerError(this, GraphErrorCodes);
```
eriolchan, line 111: Maybe we could use "resolveInnerError"? the function name would be better starting with lower case

packages/fx-core/src/plugins/resource/bot/errors.ts
eriolchan: Both ways are fine to me, so I would leave it to Ivan to decide.

packages/fx-core/src/plugins/resource/bot/errors.ts
```diff
@@ -14,6 +14,22 @@ export enum ErrorType {
   System,
 }
 
+function InferInnerError(target: PluginError, helpLinkMap: Map<string, string>): void {
+  if (!target.innerError) return;
+
+  const errorCode = target.innerError.response?.data?.error?.code;
+  const helpLink = helpLinkMap.get(errorCode);
```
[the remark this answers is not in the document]
IvanJobs: I believe helpLink would be undefined.
Since it's legacy code, IMO, it's better to check errorCode anyway. Update it soon.

packages/fx-core/src/plugins/resource/bot/errors.ts
IvanJobs: `resolveInnerError` is okay for me, update it soon.

eriolchan (approved):

PR #3824, opened 2022-01-29
perf(aad): update AAD activate logic to allow Tab without AAD

KennethBWSong:
Now provision and deploy works with some changes to the tempalte:
1. remove aad from active plugins
2. remove webApplicationInfo in teams app manifest.
TODO：add aad plugin when adding function/apim in solution

### DIFF
--- a/packages/fx-core/src/plugins/resource/aad/index.ts
+++ b/packages/fx-core/src/plugins/resource/aad/index.ts
@@ -35,7 +35,13 @@ export class AadAppForTeamsPlugin implements Plugin {
     if (solutionSettings?.migrateFromV1) {
       return false;
     }
-    return solutionSettings.hostType === HostTypeOptionAzure.id;
+    return (
+      solutionSettings.hostType === HostTypeOptionAzure.id &&
+      (solutionSettings.activeResourcePlugins == undefined ||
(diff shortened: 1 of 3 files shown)


KennethBWSong:
E2E TEST: None

packages/fx-core/src/plugins/resource/aad/index.ts
```diff
@@ -35,7 +35,12 @@ export class AadAppForTeamsPlugin implements Plugin {
     if (solutionSettings?.migrateFromV1) {
       return false;
     }
-    return solutionSettings.hostType === HostTypeOptionAzure.id;
+    return (
+      solutionSettings.hostType === HostTypeOptionAzure.id &&
+      // For scaffold, activeResourecPlugins is undefined
```
blackchoey, line 40: Have we confirmed this is a contract between resource plugin and solution plugin / fx-core?

KennethBWSong: Yeah, I have confirmed with @jayzhang when updating Simple Auth plugin.

blackchoey (approved):

adashen (approved):

PR #3845, opened 2022-02-07
perf(local debug): update local debug to separate aad

KennethBWSong:


packages/fx-core/src/common/local/projectSettingsHelper.ts
```diff
@@ -41,8 +42,9 @@ export class ProjectSettingsHelper {
 
   public static includeAAD = (projectSettings: ProjectSettings | undefined): boolean =>
     !ProjectSettingsHelper.isMigrateFromV1(projectSettings) &&
-    (projectSettings?.solutionSettings as AzureSolutionSettings)?.hostType ===
-      HostTypeOptionAzure.id;
+    (projectSettings?.solutionSettings as AzureSolutionSettings)?.activeResourcePlugins?.includes(
```
swatDong, line 45: `ProjectSettingsHelper` follows the same logic as plugin activate.
I note that AAD plugin has comment "[// For scaffold, activeResourecPlugins is undefined](https://github.com/OfficeDev/TeamsFx/blob/dev/packages/fx-core/src/plugins/resource/aad/index.ts#L40)", so does here need the same logic? `ProjectSettingsHelper` is used by scaffold as well.

KennethBWSong: The comment in AAD plugin means that when activate() is called in scaffold, activeResourcePlugins is undefined. In this case, when ProjectSettingsHelper is called, activeResourcePlugins already have values.

swatDong: I see, make sense to me.

swatDong (approved):
Code logic looks good to me. UT fix needed since field changes.

swatDong (approved):

jayzhang (approved):

blackchoey:
add test case to test situation that AAD plugin is not enabled?

KennethBWSong:
> add test case to test situation that AAD plugin is not enabled?

Working on new test cases. Will add in another pr.

blackchoey (approved):

PR #3861, opened 2022-02-08
refactor: existing app manifest template

nliu-ms:
1. Define type ManifestCapability
2. Provide different manifest template for existing tab/bot.


nliu-ms:
E2E TEST: solution/AddCapabilities_TabAddBot

jayzhang (approved):

PR #3862, opened 2022-02-08
chore(vsc): enable automatic npm install

kuojianlu:
- add automatic npm install setting and set default value to false
- let prerequisitesHandler to wait current running npm install tasks

### DIFF
--- a/packages/vscode-extension/src/debug/prerequisitesHandler.ts
+++ b/packages/vscode-extension/src/debug/prerequisitesHandler.ts
@@ -372,21 +372,29 @@ async function checkNpmInstall(component: string, folder: string): Promise<Check
     if (!installed) {
       let exitCode: number | undefined;
 
-      let running = false;
-      for (const [key, value] of trackedTasks) {
-        if (value === `${component} npm install`) {
-          running = true;
-          break;
+      const checkNpmInstallRunning = () => {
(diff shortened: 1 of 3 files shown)


packages/vscode-extension/src/debug/prerequisitesHandler.ts
```diff
@@ -369,19 +370,39 @@ async function checkNpmInstall(component: string, folder: string): Promise<Check
...
+          break;
+        }
+      }
+      if (running) {
+        exitCode = await new Promise((resolve: (value: number | undefined) => void) => {
+          const endListener = taskEndEventEmitter.event((result) => {
+            if (result.name === `${component} npm install`) {
+              endListener.dispose();
+              resolve(result.exitCode);
+            }
```
swatDong, line 388: A very corner case is that task ended just now (between line 378 and line 384)...
Better to check running again here.
```
for (... trackedTasks) ...
if (!running) {
  endListener.dispose();
  resolve(undefined);
}
```

kuojianlu: Nice catch. Added double check.

swatDong (approved):

HuihuiWu-Microsoft (approved):
LGTM

PR #3863, opened 2022-02-08
test(local debug): add unit test for separate aad

KennethBWSong:


packages/fx-core/tests/plugins/solution/debug/solution.debug.scaffolding.test.ts
```diff
@@ -161,6 +161,43 @@ describe("solution.debug.scaffolding", () => {
...
+      it(`happy path: tab without function (${parameter.programmingLanguage}) and AAD`, async () => {
+        const projectSetting = {
+          appName: "",
+          projectId: uuid.v4(),
+          solutionSettings: {
+            name: "",
+            version: "",
+            hostType: "Azure",
+            capabilities: ["Tab"],
+            activeResourcePlugins: ["fx-resource-simple-auth"],
```
swatDong, line 174: Is it possible that simple auth is enabled but aad not?

KennethBWSong: We don't expect this scenario since simple auth has dependency on AAD.

swatDong: So... better to make this case more real - removing `"fx-resource-simple-auth"` here

packages/fx-core/tests/common/local/localSettingsHelper.test.ts
```diff
@@ -73,6 +73,20 @@ describe("localSettingsHelper", () => {
       chai.assert.equal(localEnvs["AUTH_TAB_APP_ENDPOINT"], "https://localhost:53000");
     });
 
+    it("happy path without AAD plugin", async () => {
+      await fs.ensureDir(projectPath);
+      await fs.emptyDir(projectPath);
+
+      const projectSettingsAll = cloneDeep(projectSettings0);
+      delete projectSettingsAll.solutionSettings.activeResourcePlugins[0];
```
blackchoey, line 81: can we find the index of aad plugin and remove it? this can avoid test failures when someone changes the content of activeResourcePlugins

packages/fx-core/tests/plugins/solution/debug/solution.debug.scaffolding.test.ts
```diff
@@ -327,6 +364,42 @@ describe("solution.debug.scaffolding", () => {
...
+        //assert output tasks.json
+        const tasksAll = fs.readJSONSync(expectedTasksFile);
+        const tasks: [] = tasksAll["tasks"];
+        chai.assert.equal(tasks.length, parameter.numTasks);
+
+        //assert output settings.json
+        const settings = fs.readJSONSync(expectedSettingsFile);
+        chai.assert.equal(Object.keys(settings).length, 1);
+
+        await assertLocalDebugLocalEnvs(v2Context, inputs, parameter.numLocalEnvs - 12);
```
blackchoey, line 401: add comments about why we need to `-12`?
By the way, can we define expected content for non-aad situations? So we don't need to have magic numbers here.

swatDong (dismissed):

packages/fx-core/tests/common/local/localSettingsHelper.test.ts
KennethBWSong: Fixed. Thanks!

packages/fx-core/tests/plugins/solution/debug/solution.debug.scaffolding.test.ts
KennethBWSong: Fixed. Thanks!

packages/fx-core/tests/plugins/solution/debug/solution.debug.scaffolding.test.ts
KennethBWSong: Fixed. Thanks!

swatDong (approved):

packages/fx-core/tests/common/local/localSettingsHelper.test.ts
```diff
@@ -25,7 +25,7 @@ describe("localSettingsHelper", () => {
         hostType: "Azure",
         azureResources: [] as string[],
         capabilities: ["Tab"],
-        activeResourcePlugins: ["fx-resource-aad-app-for-teams", "fx-resource-simple-auth"],
+        activeResourcePlugins: ["fx-resource-aad-app-for-teams"],
```
blackchoey, line 28: did we remove simple auth plugin from all test cases? Since we still support simple auth enabled project, we'd better keep some test cases for simple auth enabled scenarios to ensure changes won't break simple auth scenario

KennethBWSong: Added several tests. Thanks!

swatDong (approved):

blackchoey (approved):

PR #3872, opened 2022-02-08
chore(identity): remove useless code

xzf0587:


packages/fx-core/src/plugins/resource/identity/index.ts
```diff
@@ -38,77 +29,16 @@ export class IdentityPlugin implements Plugin {
...
-    "identity"
-  );
-  config: IdentityConfig = new IdentityConfig();
-
-  async provision(ctx: PluginContext): Promise<Result> {
-    return ok(undefined);
-  }
 
-  async postProvision(ctx: PluginContext): Promise<Result> {
+  async provision(ctx: PluginContext): Promise<Result<void, FxError>> {
```
blackchoey, line 33: remove provision and postProvision since they do nothing?

packages/fx-core/src/plugins/resource/identity/index.ts
```diff
@@ -157,86 +87,12 @@ export class IdentityPlugin implements Plugin {
     return ok(result);
   }
 
-  public async executeUserTask(func: Func, context: PluginContext): Promise<Result> {
+  public async executeUserTask(func: Func, context: PluginContext): Promise<Result<void, FxError>> {
```
blackchoey, line 90: No business logic in this function, can we remove it?

packages/fx-core/src/plugins/resource/identity/utils/telemetryUtil.ts
```diff
@@ -1,59 +0,0 @@
-import { PluginContext } from "@microsoft/teamsfx-api";
-import { Constants, Telemetry } from "../constants";
-
-export class TelemetryUtils {
```
blackchoey, line 4: we still need telemetry to collect usage data for identity plugin. We can check our existing dashboard and see what kind of telemetry is required to continue report the data in dashboard.

packages/fx-core/src/plugins/resource/identity/errors.ts
```diff
@@ -1,11 +0,0 @@
-export class ErrorMessage {
-  public static readonly IdentityLoadFileError = {
-    name: "IdentityLoadFileError",
-    message: () => "Failed to load Azure Resource Manager template files.",
```
blackchoey, line 4: Since this error is removed, what error will user get if our extension failed to read the bicep template file?

PR #3873, opened 2022-02-08
chore(sql-v3): sql plugin v3

jayzhang:


packages/fx-core/src/plugins/resource/sql/v3/index.ts
```diff
@@ -0,0 +1,434 @@
...
+    return ok([{ kind: "bicep", template: result }]);
+  }
+  @hooks([CommonErrorHandlerMW({ telemetry: { component: BuiltInFeaturePluginNames.sql } })])
+  async addFeature(
+    ctx: v3.ContextWithManifestProvider,
+    inputs: v2.InputsWithProjectPath
+  ): Promise<Result<v2.ResourceTemplate[], FxError>> {
+    const solutionSettings = ctx.projectSetting.solutionSettings as AzureSolutionSettings;
+    const activeResourcePlugins = solutionSettings.activeResourcePlugins;
+    const firstTime = activeResourcePlugins.includes(this.name);
```
xzf0587, line 141: Currently, we use existingResource in ctx.answer because the activeResourcePlugins contain SQL if SQL resource is selected.  
Here we use the activeResourcePlugins directly. The meaning of activeResourcePlugins has changed?

xzf0587 (dismissed):

xzf0587 (approved):

PR #3875, opened 2022-02-09
test: wrap mocha.it to support to inject test plan id

LongOddCode:


packages/extra-shot-mocha/src/it.ts
```diff
@@ -0,0 +1,87 @@
...
+  if (ctxOrFn) {
+    if (ctxOrFn instanceof Function) {
+      t = mocha.it(title, ctxOrFn);
+    } else {
+      t = mocha.it(title, async function (this) {
+        addContext(this, JSON.stringify(ctxOrFn));
+        await (fn as mocha.AsyncFunc).call(this);
+      });
+    }
+  } else {
```
1yefuwang1, line 83: It seems that the implementation doesn't handle edge cases like `it("should work", { caseId: 1 })`, which type-checks but throws at runtime.

1yefuwang1 (dismissed):

packages/extra-shot-mocha/tests/it.test.ts
```diff
@@ -0,0 +1,42 @@
...
+  it("should run as normal mocha.it with sync arrow function", () => {
+    expect(1).equals(1);
+  });
+
+  it("should run as normal mocha.it with async arrow function", async () => {
+    expect(1).equals(1);
+  });
+
+  it("should inject ctx with async function", { a: 1 }, async function () {
+    expect(1).equals(1);
```
HuihuiWu-Microsoft, line 28: Do we actually know whether the ctx is injected here?

LongOddCode: it relies on the reporter. Currently, only mochawesome can output the context. See https://www.npmjs.com/package/mochawesome

LongOddCode: image

HuihuiWu-Microsoft: Maybe check the injection when available.

HuihuiWu-Microsoft (approved):

PR #3877, opened 2022-02-09
chore: add auto npm install error handling for adding resource/cap

kuojianlu:
To fix https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/13143216.

### DIFF
--- a/packages/vscode-extension/src/handlers.ts
+++ b/packages/vscode-extension/src/handlers.ts
@@ -420,13 +420,22 @@ export async function addResourceHandler(args?: any[]): Promise<Result<null, FxE
     namespace: "fx-solution-azure",
     method: "addResource",
   };
-  const localEnvManager = new LocalEnvManager(VsCodeLogInstance, ExtTelemetry.reporter, VS_CODE_UI);
-  const projectSettings = await localEnvManager.getProjectSettings(ext.workspaceUri.fsPath);
-  const includeBackend = ProjectSettingsHelper.includeBackend(projectSettings);
+  let excludeBackend = true;
+  try {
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/handlers.ts
```diff
@@ -420,13 +420,22 @@ export async function addResourceHandler(args?: any[]): Promise<Result<null, FxE
...
-  const includeBackend = ProjectSettingsHelper.includeBackend(projectSettings);
+  let excludeBackend = true;
+  try {
+    const localEnvManager = new LocalEnvManager(
+      VsCodeLogInstance,
+      ExtTelemetry.reporter,
+      VS_CODE_UI
+    );
+    const projectSettings = await localEnvManager.getProjectSettings(ext.workspaceUri.fsPath);
+    excludeBackend = ProjectSettingsHelper.includeBackend(projectSettings);
```
HuihuiWu-Microsoft, line 431: The naming is confusing, does it mean include or exclude?

kuojianlu: When current project includes backend, then auto npm install for backend will be excluded because it has been done before. So I think using `excludeBackend` is better.

jayzhang (approved):

packages/vscode-extension/src/handlers.ts
HuihuiWu-Microsoft: It's clear to me once you explained. I just worry that others might get confused when they read code. Anyway, it's not a big problem

HuihuiWu-Microsoft (approved):

PR #3879, opened 2022-02-09
fix(vs): activate frontend plugin in VS

hund030:
E2E TEST: Will add e2e test later

### DIFF
--- a/packages/fx-core/src/plugins/resource/frontend/v2/index.ts
+++ b/packages/fx-core/src/plugins/resource/frontend/v2/index.ts
@@ -2,8 +2,6 @@
 // Licensed under the MIT license.
 
 import {
-  AzureAccountProvider,
-  AzureSolutionSettings,
   Func,
   FxError,
   Inputs,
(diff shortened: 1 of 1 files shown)


packages/fx-core/src/plugins/resource/frontend/v2/index.ts
```diff
@@ -48,8 +46,9 @@ export class FrontendPluginV2 implements ResourcePlugin {
   plugin!: FrontendPlugin;
 
   activate(projectSettings: ProjectSettings): boolean {
+    const activateInVS = projectSettings.programmingLanguage === TabLanguage.CSharp;
     const solutionSettings = projectSettings.solutionSettings as AzureSolutionSettings;
-    return this.plugin.activate(solutionSettings);
+    return this.plugin.activate(solutionSettings) || activateInVS;
```
IvanJobs, line 51: I'm not quite sure - do we need to make any change for v1?

Nit: The invocation of `this.plugin.activate` is meaningless  when `activateInVS` is true. Better to swap the position of `activateInVS` with the former `activate()`.

IvanJobs (approved):
lgtm

PR #3884, opened 2022-02-09
chore: merge local and remote config step 1

xiaolang124:
1.Add environment variable to control whether to turn on the merge config feature. Full control whether to generate ```localSettings.json``` or ```config.local.json``` will be updated later.
2.Generate ```config.local.json``` after scaffolding.
3.Generate ```state.local.json``` and ```local.userdata.json``` after local debug. Short name for local will be ```xxx-local-debug```.
4.Move solution.teamsTenantId to ```state.local.json```.

E2E test: https://github.com/OfficeDev/TeamsFx/actions/runs/1827778093


packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -966,6 +966,9 @@ async function persistBicepTemplates(
     const parameterEnvFolderPath = path.join(projectaPath, configsFolder);
     await fs.ensureDir(parameterEnvFolderPath);
     for (const env of envListResult.value) {
+      if (env === environmentManager.getLocalEnvName()) {
```
dooriya, line 969: In previous implementation, we only expect remote environments to be listed by using the  `listEnvConfigs` API.

Now, the `local` environment also be included in the result of this API, it may cause code break in other places that doesn't expect local environment to be in the list.

xiaolang124: Thanks for the advice, I will pay attention to this. For now, if we filter local out of ```listEnvConfigs``` result, it will cause ```checkEnvExist``` fail. And treeView already handles local scenario. I will check more in later prs.

dooriya: We can have two API: `listAllEnvConfig` and `listRemoteEnvConfig`, and replace the original `listEnvConfig` with `listRemoteEnvConfig`. For `checkEnvExist` API implementation, we can call `listAllEnvConfig`.

xiaolang124: Good idea, already updated.

dooriya (dismissed):

dooriya (dismissed):

packages/fx-core/src/plugins/solution/fx-solution/v2/provisionLocal.ts
```diff
@@ -23,12 +23,15 @@ import { environmentManager } from "../../../../core/environment";
 import { PermissionRequestFileProvider } from "../../../../core/permissionRequest";
 import { LocalSettingsTeamsAppKeys } from "../../../../common/localSettingsConstants";
 import { configLocalDebugSettings, setupLocalDebugSettings } from "../debug/provisionLocal";
+import { isConfigUnifyEnabled } from "../../../..";
```
dooriya, line 26:
```suggestion
import { isConfigUnifyEnabled } from "../../../../common/tools";

**Review by jayzhang — APPROVED** [review]

**Review by Alive-Fish — APPROVED** [review]

**Review by swatDong — APPROVED** [review]

File: packages/fx-core/src/core/index.ts [hunk 3]
```diff
@@ -1271,7 +1284,10 @@ export class FxCore implements v3.ICore {
     projectSettings: ProjectSettings,
     inputs: Inputs
   ): Promise<Result<Void, FxError>> {
-    const appName = projectSettings.appName;
+    let appName = projectSettings.appName;
+    if (targetEnvName === environmentManager.getLocalEnvName()) {
+      appName = appName + "-local-debug";
```
nliu-ms, line 1289: appName with suffix may exceed the manifest short name length limit of 30.

Suggest:
getLocalAppName() from "src/plugins/resource/appstudio/utils"

nliu-ms (approved):

packages/fx-core/src/core/index.ts
xiaolang124: Thanks for the advice, will update in next PR.

PR #3902, opened 2022-02-10
fix(dotnet): add fallback of local template

JerryYangKai:
Just for review, it needs merge after next release.
E2E TEST: no need.


hund030:
What if we don't exit process when failing to download template? So that we can merge this pr first and after stable template released, we can change back.

https://github.com/OfficeDev/TeamsFx/blob/552e79902eb39169785336220964c2b634388d47/packages/fx-core/scripts/download-templates-zip.js#L49

hund030 (approved):

PR #3906, opened 2022-02-11
refactor: add interface to core to provision teams app with user provided app package

nliu-ms:
add a new interface to avoid existing functions' middleware, like project setting loader.


nliu-ms:
For cli command implementation, just
1) get the app package path from params, add it to inputs
2) call core.provisionTeamsAppForCLI(); to create Teams app

nliu-ms:
E2E TEST: multienv/TestRemoteHappyPath

jayzhang (approved):

PR #3909, opened 2022-02-11
chore(corev3): fix create and add resource e2e flow

jayzhang:


hund030 (approved):

nliu-ms (approved):

packages/fx-core/src/core/index.ts
```diff
@@ -442,6 +445,9 @@ export class FxCore implements v3.ICore {
           }
         }
       }
+      if (ctx.projectSettings?.solutionSettings) {
```
HuihuiWu-Microsoft, line 448: I think we need to check input capability to decide whether it's Azure hosting?

jayzhang: I think in the future, we will ignore whether Azure hosting or not.

packages/fx-core/src/plugins/solution/fx-solution/v3/addFeature.ts
```diff
@@ -22,20 +22,31 @@ import { selectSingleFeatureQuestion } from "../../utils/questions";
...
 }
 
 export async function getQuestionsForAddFeature(
   ctx: v2.Context,
   inputs: v2.InputsWithProjectPath
 ): Promise<Result<QTreeNode | undefined, FxError>> {
+  const node = new QTreeNode({ type: "group" });
   const plugins = getAllFeaturePlugins();
   const featureNode = new QTreeNode(selectSingleFeatureQuestion);
+  if (!ctx.projectSetting.solutionSettings?.programmingLanguage) {
```
HuihuiWu-Microsoft, line 46: Is this question added twice(both in core and solution)?

packages/fx-core/src/core/index.ts
HuihuiWu-Microsoft: Currently hosting type is used in extension to check what UI to show.

xzf0587 (approved):

packages/fx-core/src/plugins/solution/fx-solution/v3/addFeature.ts
jayzhang: programmingLanguage is now core-level question, solution will not ask it.

PR #3933, opened 2022-02-14
chore(cd): generate zip to temporary folder in cd

hund030:
Release templates produces some temporary files, as below.

This PR is to generate the template zip files in a temporary folder so that we do not worry commit in those zip files by mistake.
There still be some js/ts files produced during generating template zip files, we will have another task to avoid those files.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -116,7 +116,7 @@ jobs:
           then
               git push -d origin $(git tag --points-at HEAD | grep templates | grep rc)
           fi
-      
+
       - name: sync up templates deps with sdk version for RC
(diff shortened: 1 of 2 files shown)


.github/workflows/cd.yml
```diff
@@ -168,34 +168,34 @@ jobs:
...
+          artifacts: ~/template-tags.txt
           name: 'Template Tag List'
           body: 'Release to maintain template tag list.'
           token: ${{ secrets.github_token }}
           tag: 'template-tag-list'
-          allowUpdates: true  
-    
+          allowUpdates: true
+
       - name: cleanup templates
```
wenytang-ms, line 199: please help remove this action, thanks!

wenytang-ms (approved):

PR #3946, opened 2022-02-15
fix(dotnet): remove create readme logic in solution for dotnet project

JerryYangKai:
E2E TEST: no need.

### DIFF
--- a/packages/fx-core/src/plugins/solution/fx-solution/v2/utils.ts
+++ b/packages/fx-core/src/plugins/solution/fx-solution/v2/utils.ts
@@ -54,6 +54,10 @@ export function isAzureProject(azureSettings: AzureSolutionSettings | undefined)
   return azureSettings !== undefined && HostTypeOptionAzure.id === azureSettings.hostType;
 }
 
+export function isVsPlatform(ctx: v2.Context): boolean {
+  return ctx.projectSetting.programmingLanguage === "csharp";
+}
+
 export function combineRecords<T>(records: { name: string; result: T }[]): Record<string, T> {
   const ret: Record<v2.PluginName, T> = {};
   for (const record of records) {
--- a/packages/fx-core/src/plugins/solution/fx-solution/v2/scaffolding.ts
+++ b/packages/fx-core/src/plugins/solution/fx-solution/v2/scaffolding.ts
@@ -24,6 +24,7 @@ import {
   getSelectedPlugins,
   fillInSolutionSettings,
   isAzureProject,
+  isVsPlatform,
 } from "./utils";
 import path from "path";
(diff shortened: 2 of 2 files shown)


packages/fx-core/src/plugins/solution/fx-solution/v2/utils.ts
```diff
@@ -54,6 +54,10 @@ export function isAzureProject(azureSettings: AzureSolutionSettings | undefined)
   return azureSettings !== undefined && HostTypeOptionAzure.id === azureSettings.hostType;
 }
 
+export function isVsPlatform(ctx: v2.Context): boolean {
```
eriolchan, line 57: @hund030  Just curious if there is any common place to put this isVsPlatform() logic? I am not sure if this method exist in both solution, tab and bot plugin?

hund030: Yes. Maybe we can put this in common/tools.ts.

JerryYangKai: Good idea！

jayzhang: we already have this function globally: 
image

kuojianlu:
Any effect on opening readme behavior after project creation? https://github.com/OfficeDev/TeamsFx/blob/dev/packages/fx-core/src/core/index.ts#L231

packages/fx-core/src/plugins/solution/fx-solution/v2/utils.ts
eriolchan: Thanks for sharing @jayzhang, @JerryYangKai can you help check if we can leverage this?

JerryYangKai:
> Any effect on opening readme behavior after project creation? https://github.com/OfficeDev/TeamsFx/blob/dev/packages/fx-core/src/core/index.ts#L231

This is only triggered when this project is created from VS, so I think no effect to VS code part. Any suggestions?

kuojianlu:
> > Any effect on opening readme behavior after project creation? https://github.com/OfficeDev/TeamsFx/blob/dev/packages/fx-core/src/core/index.ts#L231
> 
> This is only triggered when this project is created from VS, so I think no effect to VS code part. Any suggestions?

I think for VS, there is no need to scaffold readme, thus no need to set `openReadme` and `automaticNpmInstall` state in https://github.com/OfficeDev/TeamsFx/blob/dev/packages/fx-core/src/core/index.ts#L231. @jayzhang What do you think?

jayzhang (approved):

PR #3949, opened 2022-02-15
feat: manipulate manifest

nliu-ms:
We are adding a new package for manifest manipulation. 
Related task https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_sprints/taskboard/Teams%20Extensibility%20E2E%20Team/Microsoft%20Teams%20Extensibility/CY22-2.2?workitem=13487724

E2E TEST: not needed


packages/fx-core/src/plugins/resource/appstudio/manifest/src/index.ts
```diff
@@ -0,0 +1,40 @@
...
+  async load(filePath: string): Promise<TeamsAppManifest> {
+    return new TeamsAppManifest();
+  }
+
+  /**
+   * Save manifest to .json file
+   * @param filePath path to the manifest.json file
+   * @param manifest
+   */
+  async save(filePath: string, manifest: TeamsAppManifest): Promise<any> {}
```
1yefuwang1, line 20: Should we make it static since manifest is passed in. Or perhaps it's more OOP-ish to have this signature `save(filePath: string): Promise<void>`, which is more consistent with `fs.writeJSON`. 
Besides, we should probably add more Exception types, and document them in comments for bettern error handling

packages/manifest/src/index.ts
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+
+  /// strech goal
+
+  /**
+   * Provide an initial manifest file, without capabilities
+   * @param filePath path to the manifest.json file
+   */
+  static async init(filePath: string): Promise<void> {}
```
nliu-ms, line 51: appname: string

packages/manifest/src/index.ts
```diff
@@ -0,0 +1,74 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { TeamsAppManifest, IStaticTab, IConfigurableTab } from "@microsoft/teamsfx-api";
+import * as fs from "fs-extra";
+
+export class Manifest {
+  /**
```
nliu-ms, line 8: manifest： TeamsAppManifest

packages/manifest/src/index.ts
```diff
@@ -0,0 +1,74 @@
...
+   * Load manifest from manifest.json
+   * @param filePath path to the manifest.json file
+   * @throws FileNotFoundError - when file not found
+   * @throws InvalidManifestError - when file is not a valid json or can not be parsed to TeamsAppManifest
+   */
+  static async load(filePath: string): Promise<TeamsAppManifest> {
```
nliu-ms, line 14: remove static

packages/manifest/src/index.ts
```diff
@@ -0,0 +1,86 @@
...
+    manifest.name.full = appName;
+    this.manifest = manifest;
+  }
+  /**
+   * Load manifest from manifest.json
+   * @param filePath path to the manifest.json file
+   * @throws FileNotFoundError - when file not found
+   * @throws InvalidManifestError - when file is not a valid json or can not be parsed to TeamsAppManifest
+   */
+  async load(filePath: string): Promise<TeamsAppManifest> {
```
1yefuwang1, line 22: Maybe this should be static.

packages/manifest/src/index.ts
```diff
@@ -0,0 +1,86 @@
...
+
+
+  /// strech goal
+
+  /**
+   * Provide an initial manifest file, without capabilities
+   * @param appName short name for Teams app
+   * @param filePath path to the manifest.json file
+   */
+  async init(appName: string, filePath: string): Promise<void> {}
```
1yefuwang1, line 61: Perhaps this should be static and return TeamsAppManifest, given that we already have `save()` method.

packages/manifest/src/index.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  constructor(manifest?: TeamsAppManifest) {
+    if (manifest) {
+      this.manifest = manifest;
+    } else {
+      this.manifest = new TeamsAppManifest();
+    }
+  }
+
+  static async initWithAppName(appName: string): Promise<Manifest> {
```
igor-ribeiiro, line 18: Should we return `Manifest` instead of `TeamsAppManifest`?
This `Manifest` class seems to have some nice static functions, but I am not seeing the value in having it returned as what people really care about is the interface definition class (which is `TeamsAppManifest`).

packages/manifest/src/index.ts
```diff
@@ -0,0 +1,85 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { TeamsAppManifest, IStaticTab, IConfigurableTab } from "@microsoft/teamsfx-api";
+import * as fs from "fs-extra";
+
+export class Manifest {
+  private manifest: TeamsAppManifest;
```
igor-ribeiiro, line 8: Should this really be `private`?
I think there is value in providing a way to access the `TeamsAppManifest`. Having some `get` function at least would be nice

1yefuwang1: Good point. It seems you want to directly access `TeamsAppManifest`. How about instead of adding another layer of indirection, we just re-export the `TeamsAppManifest` type, and provide a few functions to work with it. e.g.
`loadFromPath(path: string): Promise<TeamsAppManifest>`
`save(manifest: TeamsAppManifest, path: string): Promise<void>`
`validate(manifest: TeamsAppManifest): Promise<string[]>`

1yefuwang1: @igor-ribeiiro Are there any other functions you feel like having?

igor-ribeiiro: Not for now, those seem great!

packages/manifest/src/index.ts
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Writes the manifest object to the given path.
+ * 
+ * @param path - Where to write
+ * @param manifest - Manifest object to be saved
+ * @throws Will propagate any error thrown by the fs-extra#writeJson.
+ * 
+ */
+export async function writeToPath(path: string, manifest: TeamsAppManifest): Promise<void> {
+  return fs.writeJson(path, manifest)
```
igor-ribeiiro, line 34: Can we add: `JSON.stringify(manifest, null, 4)` or something similar? Otherwise, the file created is messy

1yefuwang1: Sure, will do.

jayzhang (approved):

PR #3954, opened 2022-02-16
fix(localdebug): allow local debug for bot+backend project

kuojianlu:
- refactor lanuch(Next).ts for better maintenance
- remove cascadeTerminateToConfigurations in launch configurations
  because stopAll in compound is enough
- add backend launch configuration and task when adding function resource

To fix https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/13370300/

### DIFF
--- a/packages/fx-core/src/plugins/solution/fx-solution/debug/util/launch.ts
+++ b/packages/fx-core/src/plugins/solution/fx-solution/debug/util/launch.ts
@@ -21,199 +21,24 @@ export function generateConfigurations(
   const launchConfigurations: Record<string, unknown>[] = isMigrateFromV1
     ? []
     : [
-        {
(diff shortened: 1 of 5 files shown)


swatDong:
Not only tasks and launch, I remember that for local settings/envs we also checked `includeBackend` inside `includeFrontend`...

kuojianlu:
> Not only tasks and launch, I remember that for local settings/envs we also checked `includeBackend` inside `includeFrontend`...

Yes. Actually, function even hasn't been used in bot template yet... This PR is just to adjust tasks and launch configurations.

kuojianlu:
E2E TEST: skip; will add "bot+func" scenario to UI Test afterwards.

swatDong (approved):

PR #3956, opened 2022-02-16
chore: open read me happy path

1openwindow:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_sprints/taskboard/Teams%20Extensibility%20E2E%20Team/Microsoft%20Teams%20Extensibility/CY22-2.1?workitem=13133655


packages/vscode-extension/src/handlers.ts
```diff
@@ -1155,6 +1155,39 @@ async function openMarkdownHandler() {
   }
 }
 
+export async function openReadMeHandler(args: any[]) {
+  if (workspace.workspaceFolders && workspace.workspaceFolders.length > 0) {
```
HuihuiWu-Microsoft, line 1159: What if current project is not a valid teamsfx project?

packages/vscode-extension/src/handlers.ts
```diff
@@ -1155,6 +1155,39 @@ async function openMarkdownHandler() {
...
+        targetFolder = botFolder;
+      }
+    }
+    // When tab and bot coexist, readme file would reside in project root folder.
+    // Naming it README.md could accidently overwrite our users' own readme file.
+    // So we name it README-auto-generated.md here.
+    const autoGeneratedReadmePath = `${targetFolder}/${AutoGeneratedReadme}`;
+    const uri = (await fs.pathExists(autoGeneratedReadmePath))
+      ? Uri.file(autoGeneratedReadmePath)
+      : Uri.file(`${targetFolder}/README.md`);
```
HuihuiWu-Microsoft, line 1182: What if both path not exist?

packages/vscode-extension/src/handlers.ts
1openwindow: this PR is only for happy path. For this case, it is still under discussion. there should be other PR to take care this case.

HuihuiWu-Microsoft (approved):
Don't forget to update logic when discussion is done

PR #3960, opened 2022-02-16
chore(localdebug): fix azure function wording

XiaofuHuang:


swatDong (approved):

1openwindow (approved):

packages/fx-core/src/common/deps-checker/util/extensionInstaller.ts
```diff
@@ -41,7 +41,9 @@ export async function installExtension(
     }
   } catch (error) {
     await logger.printDetailLog();
-    await logger.error(`Failed to run backend extension install: error = '${error}'`);
+    await logger.error(
```
meifans, line 44: InstallExtension use a empty Logger. So it has no effect on doctor.

nliu-ms (approved):

meifans (approved):

PR #3962, opened 2022-02-16
chore(template): get alpha template when setting prerelease flag

hund030:


packages/fx-core/src/common/template-utils/templates.ts
```diff
@@ -8,8 +8,12 @@ export const templatesVersion = config.version;
 export const tagPrefix = config.tagPrefix;
 export const preRelease = process.env.TEAMSFX_TEMPLATE_PRERELEASE || "";
 export const tagListURL = config.tagListURL;
+export const alphaVersion = "templates-0.0.0-alpha";
 
 export function selectTag(tags: string[]): string | undefined {
+  if (preRelease === "alpha") {
```
eriolchan, line 14: Since alpha tag is changed, is it possible to make rc and alpha follow the same pattern, which could reduce special case in code?

hund030: No. We change the alpha tag prefix to make it different from rc and stable, so that when release a stable or rc, we can simly count commits from the latest `templates@` tag.
If rc uses the same prefix with alpha, then our cd pipeline may be more complex to distinguish them.

eriolchan (approved):

PR #3965, opened 2022-02-16
chore(v3): collaborator APIs v3

jayzhang:


packages/fx-core/src/core/collaborator.ts
```diff
@@ -0,0 +1,495 @@
...
+        });
+      }
+
+      message.push({ content: "\n", color: Colors.BRIGHT_WHITE });
+    }
+
+    if (inputs.platform === Platform.CLI) {
+      ctx.userInteraction.showMessage("info", message, false);
+    } else if (inputs.platform === Platform.VSCode) {
+      const hasSPFx = solutionSettings?.activeResourcePlugins?.includes(
```
SLdragon, line 168: Are there better ways to determine spfx project instead of using AAD? Because in the future, we may have tab app without AAD plugin but it is not SPFX project.

packages/fx-core/src/core/collaborator.ts
```diff
@@ -0,0 +1,495 @@
...
+          {
+            content: getStrings().solution.Collaboration.GrantPermissionResourceId,
+            color: Colors.BRIGHT_WHITE,
+          },
+          { content: `${permission.resourceId}`, color: Colors.BRIGHT_MAGENTA },
+        ];
+
+        ctx.userInteraction.showMessage("info", message, false);
+      }
+      const hasSPFx = solutionSettings?.activeResourcePlugins?.includes(
```
SLdragon, line 457: Same here

packages/fx-core/src/core/collaborator.ts
jayzhang: Here the name hasSPFx is not corrent, which realy means hasAAD. I just keep the same logic with v2

SLdragon (dismissed):

SLdragon (approved):

nliu-ms (approved):

PR #3970, opened 2022-02-17
feat: cicd scaffolding

IvanJobs:
to finish task: https://dev.azure.com/msazure/Microsoft%20Teams%20Extensibility/_workitems/edit/13249772

E2E TEST: NONE

PS:
1. Ignore the scaffolded README for now which will be updated in another PR.
2. Delete CLI experience to align with the plan of all `teamsfx add` commands.


jayzhang (dismissed):

packages/vscode-extension/src/handlers.ts
```diff
@@ -572,12 +572,26 @@ export async function publishHandler(args?: any[]): Promise<Result<null, FxError
...
-  const cicdGuideLink = "https://aka.ms/teamsfx-cicd-insider-guide";
+  const res = await runUserTask(func, TelemetryEvent.AddCICDWorkflows, true);
+  if (!res.isOk()) {
+    showError(res.error);
+    ExtTelemetry.sendTelemetryErrorEvent(TelemetryEvent.AddCICDWorkflows, res.error);
+    return err(res.error);
+  }
 
-  return await env.openExternal(Uri.parse(cicdGuideLink));
+  return ok(null);
```
HuihuiWu-Microsoft, line 594: Maybe also send success event for TelemetryEvent.AddCICDWorkflows here?

packages/vscode-extension/package.json
```diff
@@ -307,8 +307,8 @@
           "when": "false"
         },
         {
-          "command": "fx-extension.cicdGuide",
-          "when": "false"
+          "command": "fx-extension.migrateV1Project",
```
HuihuiWu-Microsoft, line 310: migrateV1Project should has been removed

packages/vscode-extension/src/handlers.ts
IvanJobs: Agree - updates come soon.

packages/vscode-extension/package.json
IvanJobs: Thanks - let my sync my branch with upstream.

packages/vscode-extension/src/treeview/deploymentView.tsx
```diff
@@ -80,6 +80,14 @@ export class DeploymentView extends React.Component<any, any> {
           disable={this.state.locked}
           command="fx-extension.publish"
         ></TreeItem>
+        <TreeItem
+          label="Add CI/CD Workflows"
```
1openwindow, line 84: why not reference the "addCICDWorkflowsTitle" in string.json as well?

packages/vscode-extension/src/treeview/deploymentView.tsx
```diff
@@ -80,6 +80,14 @@ export class DeploymentView extends React.Component<any, any> {
...
+          tooltip={StringResources.vsc.commandsTreeViewProvider.addCICDWorkflowsDescription}
```
1openwindow, line 85: the content of tooltip and label are same, which is not meaningful. Tooltip need to explain more detail.

IvanJobs: Let me update it with PM, thanks.

packages/vscode-extension/package.json
IvanJobs: Since there's no conflicts detected, I prefer to clean migrateV1Project automatically when merging.

packages/vscode-extension/src/treeview/deploymentView.tsx
IvanJobs: Good point, but I would not do this right now since that would be inconsistent if I reference string.json here but other items in the same level are not. Better to change them all in a batch, and I would like to leave this to later contributors to save efforts for this PR.

HuihuiWu-Microsoft: Actually the deploymentView here is not used in production code, so it's OK to keep it.

1openwindow (approved):

jayzhang (approved):

HuihuiWu-Microsoft (dismissed):

packages/fx-core/src/plugins/resource/cicd/questions.ts
```diff
@@ -0,0 +1,49 @@
+import { OptionItem } from "@microsoft/teamsfx-api";
```
hund030, line 1: Copyright header.

packages/fx-core/src/plugins/resource/cicd/index.ts
```diff
@@ -0,0 +1,190 @@
...
+  azdoOption,
+  jenkinsOption,
+  ciOption,
+  cdOption,
+  provisionOption,
+  publishOption,
+  questionNames,
+} from "./questions";
+import { Logger } from "./logger";
+import { environmentManager } from "../../..";
```
hund030, line 39: To avoid circular reference, we don't recommend import from folder.

packages/fx-core/src/plugins/resource/cicd/index.ts
```diff
@@ -0,0 +1,190 @@
...
+
+    const whichTemplate = new QTreeNode({
+      name: questionNames.Template,
+      type: "multiSelect",
+      staticOptions: [ciOption, cdOption, provisionOption, publishOption],
+      title: "Select template(s)",
+      default: [ciOption.id],
+    });
+
+    if (inputs.platform == Platform.VSCode) {
```
hund030, line 97: Better to use ===

hund030: Do we have any plan to support VS?

packages/fx-core/src/plugins/resource/cicd/index.ts
```diff
@@ -0,0 +1,190 @@
...
+    func: Func,
+    localSettings: Json,
+    envInfo: v2.EnvInfoV2,
+    tokenProvider: TokenProvider
+  ): Promise<Result<any, FxError>> {
+    if (func.method === "addCICDWorkflows") {
+      return await this.runWithExceptionCatching(
+        ctx,
+        envInfo,
+        () => this.addCICDWorkflows(ctx, inputs, envInfo),
```
hund030, line 136: This method also call runWithExceptionCatching, seems it will send same telemetry event twice?

packages/fx-core/src/plugins/resource/cicd/plugin.ts
```diff
@@ -0,0 +1,84 @@
...
+    // 1. Key inputs (envName, provider, template) x (hostingType, ).
+    if (!inputs.projectPath) {
+      throw new NoProjectOpenedError();
+    }
+    const projectPath = inputs.projectPath;
+    // By default(VSC), get env name from plugin's own `target-env` question.
+    let envName = inputs[questionNames.Environment];
+    if (inputs.platform == Platform.CLI) {
+      // In CLI, get env name from the default `env` question.
+      envName = envInfo.envName;
```
hund030, line 28: NIT: I think CLI also support question model, why are the flows different for CLI and VS Code.

packages/fx-core/src/plugins/resource/cicd/plugin.ts
```diff
@@ -0,0 +1,84 @@
...
+    }
+    const providerName = inputs[questionNames.Provider];
+    const templateNames = inputs[questionNames.Template] as string[];
+    if (!envName || !providerName || templateNames.length == 0) {
+      throw new InternalError("Some preconditions of inputs are not met.");
```
hund030, line 33: Better to indicate what precondition is not met.

packages/fx-core/src/plugins/resource/cicd/plugin.ts
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    telemetryHelper.sendSuccessEvent(context, envInfo, "collect-cicd-questions", {
+      env: envName,
+      provider: providerName,
+      template: templateNames.join(","),
```
hund030, line 39: Do we really need this event? We can set these properties in success event and fault event.

packages/fx-core/src/plugins/resource/cicd/plugin.ts
```diff
@@ -0,0 +1,84 @@
...
+      "Scaffolding workflow automation files",
+      templateNames.length
+    );
+
+    await progressBar.start(`Scaffolding workflow file for ${templateNames[0]}.`);
+    await providerInstance.scaffold(projectPath, templateNames[0], replacements);
+
+    //  3.2 Call scaffold.
+    for (const templateName of templateNames.slice(1)) {
+      await progressBar.next(`Scaffolding workflow file for ${templateName}.`);
```
hund030, line 70: How long does it take to scaffold from a template? Do we really need to make each of them a independent step?
Or could we make them parallel?

packages/fx-core/src/plugins/resource/cicd/index.ts
zhenyasav: i found this useful when i had to debug circularity once https://www.npmjs.com/package/madge

IvanJobs: It's auto generated, I believe. Let me refine it soon.

packages/fx-core/src/plugins/resource/cicd/index.ts
IvanJobs: > Better to use ===

Agree.

IvanJobs: > Do we have any plan to support VS?

Current status is just following the VS status - but theoretically there would not be much work as I described in the design spec, so will support VS soon when it's good to go.

packages/fx-core/src/plugins/resource/cicd/index.ts
IvanJobs: Good catch!

packages/fx-core/src/plugins/resource/cicd/plugin.ts
IvanJobs: It's pretty quick and it's unnecessary to make it parallel.
Another point to make it sequential is it's good for the progress bar.

packages/fx-core/src/plugins/resource/cicd/plugin.ts
IvanJobs: It's tricky since flows of CLI and VS Code are slightly different. We may chat a little bit offline.

packages/fx-core/src/plugins/resource/cicd/plugin.ts
IvanJobs: The error is kind of for robustness consideration, i.e., the error should never be throwed out since it's just about inner contract. So, I prefer to leave it as it is.

packages/fx-core/src/plugins/resource/cicd/plugin.ts
IvanJobs: To separate concerns, I believe we do. 
1. We have to join the two events to get a whole picture if we choose to set the properties by two events.
2. The telemetry event is purely to collect question answers and it doesn't care about the success or failure.

hund030 (approved):
Plugin LGTM.

packages/fx-core/src/plugins/resource/cicd/providers/azdoProvider.ts
```diff
@@ -0,0 +1,22 @@
...
+
+import { CICDProvider } from "./provider";
+
+export class AzDoProvider extends CICDProvider {
+  private static instance: AzDoProvider;
+  static getInstance() {
+    if (!AzDoProvider.instance) {
+      AzDoProvider.instance = new AzDoProvider();
+      AzDoProvider.instance.scaffoldTo = ".azure/pipelines";
+      AzDoProvider.instance.providerName = "azdo";
```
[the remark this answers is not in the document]
IvanJobs: Missed here - better to use predefined enum for consistency.

packages/fx-core/src/plugins/resource/cicd/providers/provider.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    // 1. Ensure the target path is existing.
+    const targetPath = path.join(projectPath, this.scaffoldTo);
+    try {
+      await fs.ensureDir(targetPath);
+    } catch (e) {
+      throw new FileSystemError(`Fail to create path: ${targetPath}`, e as Error);
+    }
+
+    // 2. Read README from local.
```
eriolchan, line 41: Maybe the function name could be "generate README file" which is more customer facing

packages/fx-core/src/plugins/resource/cicd/providers/provider.ts
```diff
@@ -0,0 +1,97 @@
...
+      );
+      const readmeContent = await this.readLocalFile(localReadMePath);
+      try {
+        await fs.writeFile(targetReadMePath, readmeContent);
+      } catch (e) {
+        throw new FileSystemError(`Fail to write file: ${targetReadMePath}`, e as Error);
+      }
+    }
+
+    // 3. Read template from local.
```
eriolchan, line 60: Same comment here

packages/fx-core/src/plugins/resource/cicd/index.ts
```diff
@@ -0,0 +1,182 @@
...
+  questionNames,
+} from "./questions";
+import { Logger } from "./logger";
+import { environmentManager } from "../../../core/environment";
+import { telemetryHelper } from "./utils/telemetry-helper";
+
+@Service(ResourcePluginsV2.CICDPlugin)
+export class CICDPluginV2 implements ResourcePlugin {
+  name = "fx-resource-cicd";
+  displayName = "CICD";
```
eriolchan, line 45: I see we already have a constant file packages/fx-core/src/plugins/resource/cicd/constants.ts, can we leverage the value of it instead of define it again?

packages/fx-core/src/plugins/resource/cicd/index.ts
```diff
@@ -0,0 +1,182 @@
...
+  publishOption,
+  questionNames,
+} from "./questions";
+import { Logger } from "./logger";
+import { environmentManager } from "../../../core/environment";
+import { telemetryHelper } from "./utils/telemetry-helper";
+
+@Service(ResourcePluginsV2.CICDPlugin)
+export class CICDPluginV2 implements ResourcePlugin {
+  name = "fx-resource-cicd";
```
eriolchan, line 44: Same here, looks like we haven't used all values in packages/fx-core/src/plugins/resource/cicd/constants.ts

packages/fx-core/src/plugins/resource/cicd/index.ts
```diff
@@ -0,0 +1,182 @@
...
+
+    const whichTemplate = new QTreeNode({
+      name: questionNames.Template,
+      type: "multiSelect",
+      staticOptions: [ciOption, cdOption, provisionOption, publishOption],
+      title: "Select template(s)",
+      default: [ciOption.id],
+    });
+
+    if (inputs.platform === Platform.VSCode) {
```
eriolchan, line 89: Why this only take effect on VSCode? How can CLI and VS use this feature?

packages/fx-core/src/plugins/resource/cicd/index.ts
```diff
@@ -0,0 +1,182 @@
...
+      if (!inputs.projectPath) {
```
eriolchan, line 90: Why we need to check the project path again when asking user question?

packages/fx-core/src/plugins/resource/cicd/index.ts
```diff
@@ -0,0 +1,182 @@
...
+        throw new NoProjectOpenedError();
+      }
+
+      const envProfilesResult = await environmentManager.listRemoteEnvConfigs(inputs.projectPath);
+      if (envProfilesResult.isErr()) {
+        throw new InternalError("Failed to list multi env.", envProfilesResult.error);
+      }
+
+      const whichEnvironment: SingleSelectQuestion = {
```
eriolchan, line 99: I am curious on how CLI command handle env question for scaffolding?

packages/fx-core/src/plugins/resource/cicd/plugin.ts
```diff
@@ -0,0 +1,84 @@
...
+
+export class CICDImpl {
+  public async addCICDWorkflows(
+    context: Context,
+    inputs: Inputs,
+    envInfo: v2.EnvInfoV2
+  ): Promise<Result<any, FxError>> {
+    // 1. Key inputs (envName, provider, template) x (hostingType, ).
+    if (!inputs.projectPath) {
+      throw new NoProjectOpenedError();
```
eriolchan, line 21: It looks like only check if projectPath is defined, why we throw error as "No project opened"? It make sense that throw internal error that project path is not defined?

packages/fx-core/src/plugins/resource/cicd/plugin.ts
```diff
@@ -0,0 +1,84 @@
...
+    }
+    const projectPath = inputs.projectPath;
+    // By default(VSC), get env name from plugin's own `target-env` question.
+    let envName = inputs[questionNames.Environment];
+    if (inputs.platform == Platform.CLI) {
```
eriolchan, line 26: use === instead of == please

packages/fx-core/src/plugins/resource/cicd/plugin.ts
```diff
@@ -0,0 +1,84 @@
...
+      // In CLI, get env name from the default `env` question.
+      envName = envInfo.envName;
+    }
+    const providerName = inputs[questionNames.Provider];
+    const templateNames = inputs[questionNames.Template] as string[];
+    if (!envName || !providerName || templateNames.length == 0) {
```
eriolchan, line 32: same here for ==

packages/fx-core/src/plugins/resource/cicd/plugin.ts
eriolchan: One question, does another plugin/solution send out telemetry for user question?
In my opinion, question is the part (input) of the final command, we can simply put input values as properties of add-cicd event to avoid sending out unnecessary data and correlate them again. What about your thought?

Another reason is that only input event without the following add operation is meaningless for our BI data. If user cancel the step and not trigger add cicd, this input event doesn't have too much help to our business analysis.

packages/fx-core/src/plugins/resource/cicd/plugin.ts
```diff
@@ -0,0 +1,84 @@
...
+        context.projectSetting.solutionSettings?.["azureResources"].includes("apim") ?? false,
+    };
+
+    const progressBar = context.userInteraction.createProgressBar(
+      "Scaffolding workflow automation files",
+      templateNames.length
+    );
+
+    await progressBar.start(`Scaffolding workflow file for ${templateNames[0]}.`);
+    await providerInstance.scaffold(projectPath, templateNames[0], replacements);
```
eriolchan, line 66: I don't quite understand what replacements is. It seems all logic of construct replacement is based on projectSetting, why don't we just pass context or projectSetting to scaffold()?

eriolchan: What's the reason for passing templateNames[0]? Why we separate the first one from others and don't include them in parallel or in one for each loop?

packages/fx-core/src/plugins/resource/cicd/providers/provider.ts
```diff
@@ -0,0 +1,97 @@
...
+    if (!(await fs.pathExists(targetReadMePath))) {
+      const localReadMePath = path.join(
+        getTemplatesFolder(),
+        "plugins",
+        "resource",
+        "cicd",
+        this.providerName,
+        "README.md"
+      );
+      const readmeContent = await this.readLocalFile(localReadMePath);
```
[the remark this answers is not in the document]
IvanJobs: Good catch! This part is not related to template rendering, so it should just copy the readme.

packages/fx-core/src/plugins/resource/cicd/providers/provider.ts
IvanJobs: I guess you mean the comment since there's no function in this code snippet.

packages/fx-core/src/plugins/resource/cicd/providers/provider.ts
IvanJobs: Agree

packages/fx-core/src/plugins/resource/cicd/index.ts
IvanJobs: Agree.

packages/fx-core/src/plugins/resource/cicd/index.ts
IvanJobs: VS is under active development and CLI's support for VS projects is not ready, so I prefer to not show any VS related stuff in this PR but add all VS stuff in one more PR in the future.

CLI's flow of questions with VS Code is slightly different caused by existing factors:
1. There already exists a question of `Select an environment` in middleware layer https://github.com/OfficeDev/TeamsFx/blob/306d41b05a4b73519374ed836fefb74fe7bf95bf/packages/fx-core/src/core/middleware/envInfoLoader.ts#L257
2. CICD Plugin returns all three questions including env question to the upper layer for VS Code, so the three questions are in one batch (can return to previous question). That is, the flow didn't reuse the env question in middleware to avoid the separation of env question with the other two ones.
3. CICD Plugin returns two questions (provider, template) to the upper layer for CLI, and leverage the env question from middleware since there's no `go back` experience in CLI.

packages/fx-core/src/plugins/resource/cicd/index.ts
IvanJobs: It's kind of for robustness consideration although the possibility to throw this kind of error is low since it may be guaranteed by inner contract.

packages/fx-core/src/plugins/resource/cicd/index.ts
IvanJobs: VS Code and CLI are slightly different in question experience for env question. Please see related comment above.

packages/fx-core/src/plugins/resource/cicd/utils/buildScripts.ts
```diff
@@ -0,0 +1,32 @@
...
+export function generateBuildScript(projectSettings: ProjectSettings): string {
+  const parts: string[] = [];
+
+  const solutionSettings = projectSettings?.solutionSettings;
+  const capabilities = solutionSettings?.["capabilities"];
+  const azureResources = solutionSettings?.["azureResources"];
+  const hostType = solutionSettings?.["hostType"];
+
+  if (capabilities?.includes("Tab")) {
+    if (hostType && hostType == "Azure") parts.push("cd tabs; npm install; npm run build; cd -;");
```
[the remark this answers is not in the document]
IvanJobs: Although undefined is not equal to "Azure", I prefer to separate 'undefined' related behaviors explicitly for readability and clarity.

packages/fx-core/src/plugins/resource/cicd/utils/buildScripts.ts
```diff
@@ -0,0 +1,32 @@
...
+    if (hostType && hostType == "SPFx") parts.push("cd SPFx; npm install; npm run build; cd -;");
+  }
+
+  if (capabilities?.includes("Bot") || capabilities.includes("MessagingExtension")) {
+    if (projectSettings.programmingLanguage == "typescript") {
+      parts.push("cd bot; npm install; npm run build; cd -;");
+    } else {
```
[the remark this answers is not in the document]
IvanJobs: See comment above - Simply, since CLI for VS is in an early stage and not ready, I prefer to add all VS related stuff in another PR in the future.

packages/fx-core/src/plugins/resource/cicd/plugin.ts
IvanJobs: I believe if projectPath is undefined, it has a good chance users didn't open any workspace in VSCode, so this error makes sense. This follows the existing pattern: https://github.com/OfficeDev/TeamsFx/blob/306d41b05a4b73519374ed836fefb74fe7bf95bf/packages/fx-core/src/core/middleware/envInfoLoader.ts#L70

packages/fx-core/src/plugins/resource/cicd/plugin.ts
IvanJobs: > I don't quite understand what replacements is. It seems all logic of construct replacement is based on projectSetting, why don't we just pass context or projectSetting to scaffold()?

Replacements are for template rendering (key value pairs - key is placeholder name, value is the target value to be replaced) since there're placeholders in templates.
I prefer not to pass projectSettings directly but put the placements construction out of scaffold:
1. one reason is to separate concerns - template rendering only cares about replacements, and doesn't care where the replacements come from.
2. to provide more flexibility - In the future, we may construct replacements from more than one place (not only from projectSettings).

IvanJobs: > What's the reason for passing templateNames[0]? Why we separate the first one from others and don't include them in parallel or in one for each loop?

Since progressBar's contract is `1xprogressBar.start + * x progressBar.next`, map the first templateNames[0]  to `progressBar.start`, then all the remaining can be put into one loop.

It's a pretty quick operation, so unnecessary to use parallel. Writing this in sequential order using a loop is also good for the progress bar experience, I mean, compare: 0/1 - 1/1 vs 0/4 - 4/4.

packages/vscode-extension/package.json
```diff
@@ -533,8 +533,8 @@
         "icon": "$(run)"
       },
       {
-        "command": "fx-extension.cicdGuide",
-        "title": "%teamstoolkit.commands.cicd.title%",
+        "command": "fx-extension.addCICDWorkflows",
+        "title": "Teams: Add CI/CD Workflows",
```
HuihuiWu-Microsoft, line 537: Please remove the title string to package.nls.json as other commands since I'm currently working on localization and this is needed for onboarding to localization team.

IvanJobs: Good to know that - let me check it.

IvanJobs: Done @HuihuiWu-Microsoft

jayzhang (dismissed):

packages/fx-core/src/plugins/resource/cicd/utils/buildScripts.ts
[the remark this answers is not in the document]
IvanJobs: ToDo: Add TODO for VS scenario.

packages/fx-core/src/plugins/resource/cicd/providers/provider.ts
```diff
@@ -49,15 +49,15 @@ export class CICDProvider {
         this.providerName,
         "README.md"
       );
-      const readmeContent = await this.readLocalFile(localReadMePath);
+
       try {
-        await fs.writeFile(targetReadMePath, readmeContent);
+        await fs.copyFile(localReadMePath, targetReadMePath);
```
hund030, line 54: According to the [document](https://nodejs.org/api/fs.html#fspromisescopyfilesrc-dest-mode). By default, targetReadMePath is overwritten if it already exists. Do we expect it?

packages/fx-core/src/plugins/resource/cicd/plugin.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    if (notScaffoldedTemplates.length > 0) {
+      context.userInteraction.showMessage(
+        "info",
+        `Workflow automation file(s) of ${notScaffoldedTemplates.join(
+          ","
+        )} for ${providerName} have been successfully added for your project. Follow the instructuons in Readme file to setup the workflow.`,
+        false
+      );
+    }
```
IvanJobs, line 97: Add log.

packages/fx-core/src/plugins/resource/cicd/providers/provider.ts
IvanJobs: Yeah, there's existence check before, so it doesn't influence anything here.

packages/fx-core/templates/plugins/resource/cicd/github/README.md
```diff
@@ -0,0 +1,45 @@
+# How to use pre-cooked workflows on GitHub
+
+## Prerequisites
+- Teams app projects are version controlled by GitHub.
+- An M365 account. If you do not have M365 account, apply one from [M365 developer program](https://developer.microsoft.com/en-us/microsoft-365/dev-program)
```
zyxiaoyuer, line 5: as my understanding, not all pipeline need M365 account. Only Provision and Publish need M365 account right?
we should make this clear to customer in the prerequsites.

packages/fx-core/templates/plugins/resource/cicd/github/README.md
```diff
@@ -0,0 +1,45 @@
...
+- An Azure service princial with necessary permissions.
```
zyxiaoyuer, line 6: Here is the same, let customer know if they want a provision/publish pipeline, they will need Azure Service principle.

packages/fx-core/templates/plugins/resource/cicd/github/README.md
zyxiaoyuer: And also let customer know in prerequsites section that they should remove two-factor verification for their M365 account.

packages/fx-core/templates/plugins/resource/cicd/github/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Steps
+After the pre-cooked workflows are scaffolded successfully, the following steps are expected to be performed:
+1. Configure necessary GitHub secrets for your workflows.
+1. Add the scaffolded yml files into GitHub by `git add`,`git commit` and `git push`.
+
```
zyxiaoyuer, line 12: I suggested describe the steps like the following:

1. Commit and push your project source code to github remote repository, including the CICD yml files.
2. Click Actions in remote repository. (better to have a screenshot) 
image
3. Select the CICD action that you want to run.
4. (optional) If the action you selected is Provision or Publish to Teams App, you will need to config Gihub Secrets for your github action. Please see below [gihub secretes](github secretes)

packages/fx-core/templates/plugins/resource/cicd/azdo/README.md
```diff
@@ -0,0 +1,47 @@
+# How to use pre-cooked pipelines on Azure DevOps
+
+## Prerequisites
+- Teams app projects are version controlled by Azure DevOps.
+- An M365 account. If you do not have M365 account, apply one from [M365 developer program](https://developer.microsoft.com/en-us/microsoft-365/dev-program)
```
zyxiaoyuer, line 5: as my understanding, not all pipeline need M365 account. Only Provision and Publish need M365 account right?
we should make this clear to customer in the prerequsites. And also let customer know in prerequsites section that they should remove two-factor verification for their M365 account.

packages/fx-core/templates/plugins/resource/cicd/azdo/README.md
```diff
@@ -0,0 +1,47 @@
...
+- An Azure service princial with necessary permissions.
```
zyxiaoyuer, line 6: Here is the same, let customer know if they want a provision/publish pipeline, they will need Azure Service principle.

packages/fx-core/templates/plugins/resource/cicd/azdo/README.md
```diff
@@ -0,0 +1,47 @@
...
+
```
zyxiaoyuer, line 7: better to include the prerequsites of AZDO pipeline.  e.x. make sure you are adminisitrator of your AZDO project to ensure you have the authority to create pipeline.
image

packages/fx-core/templates/plugins/resource/cicd/azdo/README.md
```diff
@@ -0,0 +1,47 @@
...
+## Steps
+After the pre-cooked pipelines are scaffolded successfully, the following steps are expected to be performed:
+1. Add the scaffolded yml files into Azure DevOps by `git add`,`git commit` and `git push`.
```
zyxiaoyuer, line 10: Commit and push your project repo to github remote repo, including the CICD yml files.

jayzhang (approved):

HuihuiWu-Microsoft (approved):

.github/detect/excludes.txt
```diff
@@ -31,4 +31,5 @@ packages/sdk/test/unit/node/appCredential.spec.ts
 packages/sdk/test/unit/node/core/onBehalfOfUserCredential.spec.ts
 packages/fx-core/tests/core/samples_v2.zip
 packages/cli/tests/e2e/manifest/appPackage.dev.zip
-packages/vscode-extension/src/debug/teamsfxDebugProvider.ts
\ No newline at end of file
+packages/fx-core/templates/plugins/resource/cicd/azdo/*.yml
```
eriolchan, line 34: why do we need to exclude these yaml file?

packages/fx-core/src/plugins/resource/cicd/constants.ts
```diff
@@ -0,0 +1,49 @@
...
+  public static readonly PLUGIN_NAME = "solution";
+  public static readonly REMOTE_TEAMS_APPID = "remoteTeamsAppId";
+}
+
+export class PluginCICD {
+  public static readonly PLUGIN_NAME = "fx-resource-cicd";
+}
+
+export class Suggestions {
+  public static readonly RETRY_THE_CURRENT_STEP = "Please retry the current step.";
```
eriolchan, line 45: FYI. Huihui is driving localization feature, we may also need to apply it into cicd plugin after this sprint. You can find Zhijie for more details on plugin

packages/fx-core/src/plugins/resource/cicd/plugin.ts
```diff
@@ -0,0 +1,91 @@
...
+      await progressBar.next(`Scaffolding workflow file for ${templateName}.`);
+      scaffolded = await providerInstance.scaffold(projectPath, templateName, envName, context);
+      scaffolded.isOk() && scaffoldedArr.push(scaffolded.value);
+    }
+
+    await progressBar.end(true);
+
+    // 4. Send notification messages.
+    let message = "";
+    if (scaffoldedArr.includes(false)) {
```
eriolchan, line 73: Since array.includes() will always go through all elements of array, this code happens to read the whole array twice when running array.filter(). Which is a little not performant.

packages/fx-core/src/plugins/resource/cicd/providers/provider.ts
```diff
@@ -0,0 +1,115 @@
...
+      try {
+        await fs.writeFile(targetTemplatePath, renderedContent);
+      } catch (e) {
+        throw new FileSystemError(`Fail to write file: ${targetTemplatePath}`, e as Error);
+      }
+    } else {
+      return ok(true);
+    }
+
+    return ok(false);
```
eriolchan, line 101: what does it mean for returning false?

packages/fx-core/src/plugins/resource/cicd/result.ts
```diff
@@ -0,0 +1,52 @@
+/**
+ * This file is used to wrap result type of fx-api for function plugin because of its instability.
```
eriolchan, line 2: What does this comment mean?

packages/fx-core/src/plugins/resource/cicd/providers/provider.ts
```diff
@@ -0,0 +1,115 @@
...
+  public scaffoldTo = "";
+  public providerName = "";
+  public sourceTemplateName?: (templateName: string) => string;
+  public targetTemplateName?: (templateName: string, envName: string) => string;
+  public async scaffold(
+    projectPath: string,
+    templateName: string,
+    envName: string,
+    context: Context
+  ): Promise<Result<boolean, FxError>> {
```
eriolchan, line 24: Why we don't use FxResult as return type, which is defined in packages/fx-core/src/plugins/resource/cicd/result.ts?

packages/fx-core/src/plugins/resource/cicd/plugin.ts
```diff
@@ -0,0 +1,91 @@
...
+      templateNames[0],
+      envName,
+      context
+    );
+    scaffolded.isOk() && scaffoldedArr.push(scaffolded.value);
+    //  3.2 Call the next scaffold.
+    for (const templateName of templateNames.slice(1)) {
+      await progressBar.next(`Scaffolding workflow file for ${templateName}.`);
+      scaffolded = await providerInstance.scaffold(projectPath, templateName, envName, context);
+      scaffolded.isOk() && scaffoldedArr.push(scaffolded.value);
```
eriolchan, line 66: I'm curious on what does scaffolded.isOk() behave? Can you explain to me more offline?

When will isOk() not return true? At that time, why we don't put value to scaffoldedArr? Will the count of scaffoldedArr is not same as the total file if some file fail to scaffold?

packages/fx-core/src/plugins/resource/cicd/plugin.ts
```diff
@@ -0,0 +1,91 @@
...
+    const providerInstance = CICDProviderFactory.create(providerName as ProviderKind);
+
+    // 3. Call instance.scaffold(template, replacements: any).
+    //  3.1 Call the initial scaffold.
+    const progressBar = context.userInteraction.createProgressBar(
+      "Scaffolding workflow automation files",
+      templateNames.length
+    );
+
+    const scaffoldedArr: boolean[] = [];
```
eriolchan, line 53: Since we have for loop for templateNames[], can we also get the scaffolded array in the same loop? Something like this
```
const created: string[] = [];
const skipped: string[] = [];

let scaffolded = await providerInstance.scaffold(projectPath, templateNames[0], ...);
if (scaffolded.isOk() && scaffolded.value) {
  created.push(templateNames[0]);
} else {
  skipped.push(templateNames[0]);
}

...

for (const templateName of templateNames.slice(1)) {
  scaffolded = await providerInstance.scaffold(projectPath, templateName, ...);
  if (scaffolded.isOk() && scaffolded.value) {
    created.push(templateName);
  } else {
    skipped.push(templateName);
  }
}
```
So the step 4 for constructing message will be easy as 
```
let message = "";
if (created.length > 0) {
  message += ...;
}
if (skipped.length > 0) {
  message += ...;
}
``

File: .github/detect/excludes.txt (hunk 38 above)
Comment (IvanJobs) [reply .github/detect/excludes.txt:34]: Because the code snippet `NodeTool@0` is recognized as sensitive password in the check progress.

File: packages/fx-core/src/plugins/resource/cicd/constants.ts (hunk 39 above)
Comment (IvanJobs) [reply packages/fx-core/src/plugins/resource/cicd/constants.ts:45]: Task created for tracking, thanks

File: packages/fx-core/src/plugins/resource/cicd/plugin.ts [hunk 46]
```diff
@@ -0,0 +1,91 @@
...
+      // In CLI, get env name from the default `env` question.
+      envName = envInfo.envName;
+    }
+    const providerName = inputs[questionNames.Provider];
+    const templateNames = inputs[questionNames.Template] as string[];
+    if (!envName || !providerName || templateNames.length === 0) {
+      throw new InternalError("Some preconditions of inputs are not met.");
+    }
+
+    telemetryHelper.sendSuccessEvent(context, envInfo, "collect-cicd-questions", {
```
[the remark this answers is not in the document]
IvanJobs: The telemetry is to collect question answers.

packages/fx-core/src/plugins/resource/cicd/plugin.ts
IvanJobs: Refined.

packages/fx-core/src/plugins/resource/cicd/providers/provider.ts
IvanJobs: false indicates the template is newly scaffoled, true is for existing case.

packages/fx-core/src/plugins/resource/cicd/result.ts
IvanJobs: Removed, legacy comment from function.

packages/fx-core/src/plugins/resource/cicd/providers/provider.ts
IvanJobs: Updated - use FxResult type instead.

packages/fx-core/src/plugins/resource/cicd/plugin.ts
IvanJobs: Good catch - as you told, corner case issues are here. updated.

1yefuwang1 (dismissed):
Changes to solution LGTM

packages/fx-core/src/plugins/resource/cicd/plugin.ts
IvanJobs: I just checked appstudio plugin, and let me align with it.

HuihuiWu-Microsoft (approved):

jayzhang (approved):

packages/fx-core/templates/plugins/resource/cicd/github/README.md
```diff
@@ -0,0 +1,45 @@
...
+## GitHub Secrets 
+Steps to create secrets by environment in GitHub:
+1. In the project `Settings` page, navigate to `Environments` section and click `New environment`.
```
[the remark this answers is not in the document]
IvanJobs: Yes, ui flow is updated for GitHub. I will align with the newest ui.

packages/fx-core/src/plugins/resource/cicd/plugin.ts
```diff
@@ -0,0 +1,101 @@
...
+
+    // 4. Send notification messages.
+    let message = "";
+    if (created.length > 0) {
+      message += `Workflow automation file(s) of ${created.join(
+        ","
+      )} for (${providerName}, ${envName}) have been successfully added for your project. Follow the instructions in Readme file to setup the workflow(s).`;
+    }
+    if (skipped.length > 0) {
+      message += `You have already created template(s) of ${skipped.join(
```
eriolchan, line 91: Not sure if we need to add "\n" between above and this message?

packages/fx-core/src/plugins/resource/cicd/result.ts
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+import { FxError, Result, SystemError, UserError, err, ok } from "@microsoft/teamsfx-api";
+
+import { Alias } from "./constants";
+
+export type FxResult = Result<any, FxError>;
```
eriolchan, line 7: Do we need to make FxResult as ```Result<boolean, FxError>```? Since we use ```  if (scaffolded.isOk() && !scaffolded.value)``` to try to get boolean value from scaffolded.value?

packages/fx-core/src/plugins/resource/cicd/plugin.ts
IvanJobs: It will not take effect, since vscode will strip \n.

packages/fx-core/src/plugins/resource/cicd/result.ts
IvanJobs: Unnecessary, dev test is passed and scaffolded.value will not always be true. To give further reason, Result type is generic type when creating obj by ok(true), the generic type is fixed to boolean.

PR #3985, opened 2022-02-18
chore(cli-preview): use dev:teamsfx to start services

kuojianlu:
- use dev:teamsfx script in package.json to start services

### DIFF
--- a/packages/cli/src/cmds/preview/preview.ts
+++ b/packages/cli/src/cmds/preview/preview.ts
@@ -22,6 +22,7 @@ import {
   FolderName,
   FxCore,
   ITaskDefinition,
+  loadTeamsFxDevScript,
   LocalEnvManager,
   ProjectSettingsHelper,
   TaskDefinition,
@@ -144,7 +145,7 @@ export default class Preview extends YargsCommand {
           }
           const spWorkbenchHttpsUrl = new URL("_layouts/workbench.aspx", spSite);
           this.sharepointSiteUrl = spWorkbenchHttpsUrl.toString();
-        } catch (error) {
+        } catch (error: any) {
           throw errors.InvalidSharePointSiteURL(error);
         }
       }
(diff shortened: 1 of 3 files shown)


packages/cli/src/cmds/preview/preview.ts
```diff
@@ -849,4 +875,42 @@ export default class Preview extends YargsCommand {
     }
     return { task: task, startCb: startCb, stopCb: stopCb };
   }
+
+  private prepareTaskNext(
+    taskDefinition: ITaskDefinition,
+    startMessage: string,
+    isWatchTask: boolean,
+    env?: { [key: string]: string }
```
swatDong, line 883: When is this `env` used?

packages/cli/src/cmds/preview/preview.ts
```diff
@@ -681,7 +707,7 @@ export default class Preview extends YargsCommand {
         backendWatchTask.stopCb,
         this.serviceLogWriter
       ),
-      await botStartTask?.task.waitFor(
+      botStartTask?.task.waitFor(
         constants.botStartPattern,
```
swatDong, line 711: For `teamsfx` script, it's possible the start pattern is changed. How about adding a default timeout (e.g., 30s) for those `teamsfx` tasks in case start pattern is changed?

packages/cli/src/cmds/preview/preview.ts
kuojianlu: Actually not used in `prepareTaskNext`, let me remove this parameter for simplification.

packages/cli/src/cmds/preview/preview.ts
kuojianlu: Good point. Added timeout for `waitFor` and only set 30s for bot start task.

swatDong (approved):

PR #3992, opened 2022-02-21
fix: generate teams manifest first if provision has done

HuihuiWu-Microsoft:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/13415034
E2E TEST: will add later

### DIFF
--- a/packages/fx-core/src/plugins/resource/appstudio/plugin.ts
+++ b/packages/fx-core/src/plugins/resource/appstudio/plugin.ts
@@ -446,14 +446,20 @@ export class AppStudioPluginImpl {
       (isLocalDebug ? "local" : ctx.envInfo.envName) +
       `.json`;
     if (!(await fs.pathExists(manifestFileName))) {
-      return err(
-        AppStudioResultFactory.UserError(
-          AppStudioError.FileNotFoundError.name,
-          AppStudioError.FileNotFoundError.message(manifestFileName) +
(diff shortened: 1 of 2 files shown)


packages/fx-core/src/plugins/resource/appstudio/plugin.ts
```diff
@@ -446,14 +446,20 @@ export class AppStudioPluginImpl {
...
     if (!(await fs.pathExists(manifestFileName))) {
-      return err(
-        AppStudioResultFactory.UserError(
-          AppStudioError.FileNotFoundError.name,
-          AppStudioError.FileNotFoundError.message(manifestFileName) +
-            " Run 'Provision in the cloud' first. Click Get Help to learn more about why you need to provision.",
-          HelpLinks.WhyNeedProvision
-        )
-      );
+      const isProvisionSucceeded = !!(ctx.envInfo.state
```
HuihuiWu-Microsoft, line 449: Check whether it's already provisioned. If the project has provisioned, generate the package first.

The "teamsAppUpdatedAt" field is in state file, so git clone will not discard the information. It works well to check the timestamp of dev portal and local when updating manifest file.

SLdragon: Looks good~

SLdragon (approved):

nliu-ms (approved):

PR #3996, opened 2022-02-21
chore: merge local remote 3

xiaolang124:
1.Update ```appStudio``` plugin to support merging local and remote configs.
2.Reading manifest file remains the same, needs to be updated later when all work has been done.
3.Update view logic to support env parameters, needs to be updated later when all work has been done.


packages/fx-core/src/plugins/solution/fx-solution/v2/provisionLocal.ts
```diff
@@ -178,6 +174,7 @@ export async function provisionLocalResource(
   }
 
   if (isConfigUnifyEnabled()) {
+    setPostDataForLocal(envInfo!, localSettings);
```
swatDong, line 177: Is `envInfo` modified in `configLocalEnvironment`? If so, move this line after calling `configLocalEnvironment`.

swatDong: Or... will `setPostDataForLocal` be deleted later?

xiaolang124: When I'm done not storing the ```localSettings``` file, ```setPostDataForLocal``` will be removed.

swatDong (dismissed):

dooriya (approved):

packages/fx-core/src/plugins/resource/appstudio/plugin.ts
```diff
@@ -1459,7 +1466,7 @@ export class AppStudioPluginImpl {
         outlineIconContent
       );
 
-      if (app.updatedAt && !isLocalDebug) {
+      if (app.updatedAt && !isLocalDebug && isConfigUnifyEnabled()) {
```
HuihuiWu-Microsoft, line 1469: Should here be || instead of &&?

xiaolang124: Thanks, fixed.

HuihuiWu-Microsoft (approved):

nliu-ms (approved):

PR #3999, opened 2022-02-22
chore(localdebug): improve local debug experience

kuojianlu:
- block debugging if there is already a debugging session
- terminate teamsfx tasks before validating local prerequisites

### DIFF
--- a/packages/vscode-extension/src/debug/prerequisitesHandler.ts
+++ b/packages/vscode-extension/src/debug/prerequisitesHandler.ts
@@ -54,7 +54,13 @@ import { vscodeLogger } from "./depsChecker/vscodeLogger";
 import { doctorConstant } from "./depsChecker/doctorConstant";
 import { runTask } from "./teamsfxTaskHandler";
 import { vscodeHelper } from "./depsChecker/vscodeHelper";
-import { taskEndEventEmitter, trackedTasks } from "./teamsfxTaskHandler";
+import {
+  taskEndEventEmitter,
+  trackedTasks,
+  allRunningDebugSessions,
+  allRunningTeamsfxTasks,
+  terminateAllRunningTeamsfxTasks,
+} from "./teamsfxTaskHandler";
 import { trustDevCertHelpLink } from "./constants";
 import AppStudioTokenInstance from "../commonlib/appStudioLogin";
 import { ProgressHandler } from "../progressHandler";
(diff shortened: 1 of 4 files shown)


packages/vscode-extension/src/handlers.ts
```diff
@@ -930,7 +930,6 @@ export async function validateSpfxDependenciesHandler(): Promise<string | undefi
 export async function validateLocalPrerequisitesHandler(): Promise<string | undefined> {
   const result = await localPrerequisites.checkAndInstall();
   if (result.isErr()) {
-    await debug.stopDebugging();
```
swatDong, line 933: What's this for?

packages/vscode-extension/src/debug/prerequisitesHandler.ts
```diff
@@ -157,6 +175,12 @@ export async function checkAndInstall(): Promise<Result<any, FxError>> {
       // ignore telemetry error
     }
 
+    // terminate all running teamsfx tasks
+    if (allRunningTeamsfxTasks.size > 0) {
+      VsCodeLogInstance.info("Terminate all running teamsfx tasks.");
+      terminateAllRunningTeamsfxTasks();
```
swatDong, line 181: I remember we have a setting `fx-extension.stopTeamsToolkitTasksPostDebug`. Any side effect if always terminating all running tasks? Or, what if removing `fx-extension.stopTeamsToolkitTasksPostDebug`?

packages/vscode-extension/src/handlers.ts
kuojianlu: If there is already a debug session, this line will stop it. Removed this line since `exit 1` for this task is enough to terminate the debugging.

packages/vscode-extension/src/debug/prerequisitesHandler.ts
kuojianlu: Good catch here. I prefer removing this setting due to:
- The only (?) scenario is that user starts services manually then debug the attach configurations. Since we already provide the way to start services via npm scripts, user has no need to worry about the services being terminated after debug.
- This setting introduces difficulty for us to distinguish the running tasks, i.e. whether the tasks are held after last successful debug or not terminated as expected after last unsuccessful debug.
- This setting introduces difficulty for us to check existing ports. Actually current checking ports logic do not consider this setting...

How do you think about?

swatDong: Make sense to me. Agree to removing the setting.

kuojianlu: Sure, removed.

packages/vscode-extension/src/debug/prerequisitesHandler.ts
```diff
@@ -149,6 +155,18 @@ async function checkPort(
 }
 
 export async function checkAndInstall(): Promise<Result<any, FxError>> {
+  // skip debugging if there is already a debug session
+  if (allRunningDebugSessions.size > 0) {
+    VsCodeLogInstance.warning("SKip debugging because there is already a debug session.");
+    return err(
+      returnUserError(
+        new Error("SKip debugging because there is already a debug session."),
```
HuihuiWu-Microsoft, line 163: Will this error displayed in extension notification? If yes, maybe move it to strings.json for later localization.

kuojianlu: No, it will only shows in the output channel.

HuihuiWu-Microsoft (approved):
LGTM

PR #2023, opened 2021-08-23
feat(frontend): add migrate v1 project user task

XiaofuHuang:
Move the project files from `.archive` folder into `tabs`

PR #4000, opened 2022-02-22
refactor(frontend): remove migrate v1 in frontend plugin

XiaofuHuang:
Remove migrate v1 in frontend plugin


XiaofuHuang:
E2E TEST: None

packages/fx-core/src/plugins/resource/frontend/plugin.ts
```diff
@@ -235,32 +222,6 @@ export class FrontendPluginImpl implements PluginImpl {
     );
   }
 
-  public async executeUserTask(func: Func, ctx: PluginContext): Promise<TeamsFxResult> {
```
swatDong, line 238: Just realized - is `executeUserTask` part of plugin interface? Totally removing this method or just let it do nothing?

hund030: Both ok for me. If we do not need this interface now, just remove it.

XiaofuHuang: I addded `executeUserTask ` in PR: https://github.com/OfficeDev/TeamsFx/pull/2023
Before that frontend plugin does not contain the `executeUserTask ` lifecycle. @hund030 What do you think about it?

hund030 (approved):

PR #4002, opened 2022-02-22
refactor(localdebug): remove migrate v1 in solution/debug & localdebug plugin

XiaofuHuang:
Remove migrate v1 in solution/debug & localdebug plugin


XiaofuHuang:
E2E TEST: None

packages/fx-core/src/plugins/solution/fx-solution/solution.ts
```diff
@@ -341,11 +340,6 @@ export class TeamsAppSolution implements Solution {
     const solutionSettings = settingsRes.value;
     const selectedPlugins = await this.reloadPlugins(solutionSettings);
 
-    const scaffoldLocalDebugSettingResult = await scaffoldLocalDebugSettingsV1(ctx);
-    if (scaffoldLocalDebugSettingResult.isErr()) {
-      return scaffoldLocalDebugSettingResult;
-    }
-
     const results: Result<any, FxError>[] = await Promise.all<Result<any, FxError>>(
```
swatDong, line 343: When to remove the whole `migrate` method?

XiaofuHuang: In the next PR, it contains removing the migrate v1 logic in core & solution.

packages/fx-core/tests/plugins/solution/solution.migrate.test.ts
```diff
@@ -146,9 +146,6 @@ describe("Solution migrate()", async () => {
     expect(mockedSolutionCtx.envInfo.state.get(GLOBAL_CONFIG)).is.not.undefined;
     const lang = mockedSolutionCtx.projectSettings?.programmingLanguage;
     expect(lang).equals("javascript");
-    expect(fileContent.has("./.vscode/launch.json")).to.be.true;
```
swatDong, line 149: If the test case is still needed, change this to `to.be.false`, otherwise just removing the whole test case.

swatDong (approved):

packages/fx-core/tests/plugins/solution/solution.migrate.test.ts
XiaofuHuang: I try to remove the shole test case but failed because of the test coverage. It will be removed in the  core & solution PR too.

1yefuwang1 (approved):

PR #4020, opened 2022-02-23
chore: remove unnecessary bot plugin registration code

Siglud:


packages/fx-core/tests/plugins/resource/bot/unit/index.test.ts
```diff
@@ -2,13 +2,13 @@
 // Licensed under the MIT license.
 import "mocha";
 import * as chai from "chai";
-import fs from "fs-extra";
+import fs = require("fs-extra");
```
hund030, line 5: What's the difference?

packages/fx-core/src/plugins/resource/bot/plugin.ts
```diff
@@ -135,18 +135,9 @@ export class TeamsBotImpl implements PluginImpl {
     );
     await handler?.start(ProgressBarConstants.PROVISION_STEP_START);
 
-    // 0. Check Resource Provider
-    const azureCredential = await this.getAzureAccountCredenial();
```
eriolchan, line 139: Just confirm if there is any method not used by other code (e.g. createResourceProviderClient(), ensureResourceProvider()), we can safely remove it.

packages/fx-core/src/plugins/resource/bot/plugin.ts
```diff
@@ -135,18 +135,9 @@ export class TeamsBotImpl implements PluginImpl {
...
-      azureCredential,
-      this.config.provision.subscriptionId!
-    );
-    await factory.ensureResourceProvider(rpClient, AzureConstants.requiredResourceProviders);
-
     // 1. Do bot registration.
     await handler?.next(ProgressBarConstants.PROVISION_STEP_BOT_REG);
-    const botAuthCreds = await this.createOrGetBotAppRegistration();
-
+    await this.createOrGetBotAppRegistration();
```
eriolchan, line 140: Just a reminder, if there is no other place require GetBotAppRegistration() logic, may be we can refactor it as just createXXX()

packages/fx-core/tests/plugins/resource/bot/unit/index.test.ts
eriolchan: require looks like JavaScript style, and in TypeScript the recommendation is to use `import ... from`. Can you help check this?

packages/fx-core/tests/plugins/resource/bot/unit/index.test.ts
```diff
@@ -2,13 +2,13 @@
...
 import * as sinon from "sinon";
-import { default as chaiAsPromised } from "chai-as-promised";
-import AdmZip from "adm-zip";
-import path from "path";
+//import { default as chaiAsPromised } from "chai-as-promised";
```
eriolchan, line 7: Why we comment this? if it is not used, we'd better to remove it instead of comment it

packages/fx-core/tests/plugins/resource/bot/unit/index.test.ts
```diff
@@ -166,16 +165,17 @@ describe("Teams Bot Resource Plugin", () => {
       const fakeCreds = testUtils.generateFakeTokenCredentialsBase();
 
       let item: any = { registrationState: "Unregistered" };
-      const namespace = ["ut"];
       const fakeRPClient: any = {
+        // eslint-disable-line no-unused-vars
```
eriolchan, line 169: If we don't use rpClient any more, does it mean we don't need to mock it and remove unused code?

packages/fx-core/tests/plugins/resource/bot/unit/index.test.ts
Siglud: TsLint notice me: TS1192: Module 'fs-extra' has no default export. so, we can change that to
`import * as fs from "fs-extra";`
or use require

packages/fx-core/tests/plugins/resource/bot/unit/index.test.ts
Siglud: nod. but still want to check I will not call some API anymore, so I used 
`sinon.assert.notCalled`
check that behavior.

packages/fx-core/tests/plugins/resource/bot/unit/index.test.ts
eriolchan: Let's use `import * as fs from "fs-extra"` which align with other code

packages/fx-core/tests/plugins/resource/bot/unit/index.test.ts
```diff
@@ -166,16 +162,15 @@ describe("Teams Bot Resource Plugin", () => {
...
 
       let item: any = { registrationState: "Unregistered" };
-      const namespace = ["ut"];
       const fakeRPClient: any = {
-        get: (namespace: string) => item,
-        register: (namespace: string) => {
+        get: (_namespace: string) => item,
+        register: (_namespace: string) => {
           item = {};
           item = { ...item, $namespace: { registrationState: "Registered" } };
```
eriolchan, line 169: Do we need to update $namespace to $_namespace here?

packages/fx-core/src/plugins/resource/bot/plugin.ts
Siglud: done

packages/fx-core/src/plugins/resource/bot/plugin.ts
Siglud: bot/index.js has used this method

packages/fx-core/src/plugins/resource/bot/plugin.ts
```diff
@@ -509,7 +505,7 @@ export class TeamsBotImpl implements PluginImpl {
     } else {
       Logger.info(Messages.ProvisioningBotRegistration);
       botAuthCreds = await AADRegistration.registerAADAppAndGetSecretByGraph(
-        token!,
+        token,
```
eriolchan, line 508: Good to see this kind of refactor to validate value and avoid using !. Nice job

eriolchan (approved):

packages/fx-core/tests/plugins/resource/bot/unit/index.test.ts
```diff
@@ -551,6 +531,9 @@ describe("Teams Bot Resource Plugin", () => {
       sinon
         .stub(mockedTokenProvider.azureAccountProvider, "getAccountCredentialAsync")
         .resolves(testUtils.generateFakeTokenCredentialsBase());
+      fs.mkdirSync(path.join(inputs.projectPath, CommonStrings.BOT_WORKING_DIR_NAME), {
```
eriolchan, line 534: why do we need to create folder in unit test? can we mock fs operation?

eriolchan (approved):

PR #4022, opened 2022-02-23
fix(solution): add m365 and azure account checking in v2 deploy

1yefuwang1:
Fixes bug https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_sprints/taskboard/Teams%20Extensibility%20E2E%20Team/Microsoft%20Teams%20Extensibility/CY22-2.2?workitem=13441379
E2E TEST: not needed.

### DIFF
--- a/packages/fx-core/src/plugins/solution/fx-solution/v2/deploy.ts
+++ b/packages/fx-core/src/plugins/solution/fx-solution/v2/deploy.ts
@@ -9,6 +9,8 @@ import {
   returnUserError,
   v2,
   Void,
+  returnSystemError,
+  SystemError,
 } from "@microsoft/teamsfx-api";
 import { isUndefined } from "lodash";
 import * as util from "util";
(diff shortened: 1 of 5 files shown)


packages/fx-core/src/plugins/solution/fx-solution/v2/deploy.ts
```diff
@@ -52,6 +53,25 @@ export async function deploy(
     );
   }
 
+  const appStudioTokenJson = await tokenProvider.appStudioToken.getJsonObject();
```
HuihuiWu-Microsoft, line 56: What if user didn't sign in here?

HuihuiWu-Microsoft (approved):

packages/fx-core/src/plugins/solution/fx-solution/v2/deploy.ts
1yefuwang1: In that case, a UserError will be thrown. Returning undefined only occurs on rare cases like networking failed.

xiaolang124 (approved):

nliu-ms (approved):

jayzhang (approved):

PR #4026, opened 2022-02-24
chore: release server bits to github page

wenytang-ms:


.github/workflows/cd.yml
```diff
@@ -296,6 +298,44 @@ jobs:
...
+            sleep 20
+            cd ./packages/server
+            npx vsce package
+
+      - name: release daily server bits to github
+        if: ${{ contains(steps.version-change.outputs.CHANGED, '@microsoft/teamsfx-server') && github.ref == 'refs/heads/dev'  }}
+        uses: ncipollo/release-action@v1.7.3
+        with:
+          token: ${{ secrets.CD_PAT }}
+          tag: 'vscode-extension-${{ steps.version-change.outputs.SERVER_VERSION_NUM }}'
```
qinezh, line 318: why the tag has prefix `vscode-extension-` for the release of server bits?

.github/workflows/cd.yml
```diff
@@ -296,6 +298,44 @@ jobs:
...
+        if: ${{ contains(steps.version-change.outputs.CHANGE, '@microsoft/teamsfx-server') }}
+        uses: nick-invision/retry@v2
+        with:
+          timeout_minutes: 10
+          max_attempts: 10
+          retry_on: error
+          command: |
+            sleep 20
+            cd ./packages/server
+            npx vsce package
```
Alive-Fish, line 311: This package is not a VSIX package, so this may be "npm run package".

wenytang-ms: thanks!

.github/workflows/cd.yml
wenytang-ms: because alpha version should not contains ****@, if not, once dev merge to main, the conventional-commits get the wrong start point.

wenytang-ms: thanks! fixed!

Alive-Fish (approved):

LongOddCode (approved):

PR #4031, opened 2022-02-24
chore: template sync up strategy

wenytang-ms:
this PR change the template release strategy.
before:     SDK and adaptive-card changed, the templates always bump up except skip on CD.
now:     template dep change need manual work, set the target dep as alpha version, then this sprint CD will always sync up the dependencies version.
if no dep change or dep change as a stable version, the sync up will not happened.

this PR also fix:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/13473890

And simpleauth and function release pipeline goes on AZDO, so remove relative except download simpleauth on CD.


templates/package.json
```diff
@@ -6,7 +6,11 @@
     "scripts": {
         "check-sensitive": "npx eslint --plugin 'no-secrets' --cache --ignore-pattern 'package.json' --ignore-pattern 'package-lock.json'",
         "precommit": "npm run check-sensitive && lint-staged",
-        "version": "../.github/scripts/pkg-version.sh core-template"
+        "version": "bash ../.github/scripts/pkg-version.sh template-sync && ../.github/scripts/pkg-version.sh core-template"
```
adashen, line 9: what is  core-template for? didn't find the option in pkg-version.sh

.github/scripts/sync-templates.js
```diff
@@ -0,0 +1,37 @@
...
+    const arr = fse.readdirSync(dir);
+    arr.forEach(function (item) {
+        if (item === "node_modules") return list;
+        const fullpath = path.join(dir, item);
+        const stats = fse.statSync(fullpath);
+        if (stats.isDirectory()) {
+            recursivelyListPackageJsonFilePath(fullpath, list);
+        } else if (item === "package.json") {
+            list.push(fullpath);
+        }
```
adashen, line 17: seems function (item) doesn't always have return value. Only return list when it is "node_modules".

adashen: Besides, the code here to iterate the path to find template package.json is not very readable. Can we delegate this to Template? Template package needs tell CD what are package.json needs to sync? As for currenly, it is only several templates needs to handle right?

templates/package.json
wenytang-ms: the core-template do sync up config to download templates version in fx-core.
this triggered once templates version change.

.github/scripts/sync-templates.js
wenytang-ms: this function updated.
I think it's better to go through the whole dir and get all the package.json file.
And this method already works fine

tecton (approved):

templates/package.json
```diff
@@ -6,12 +6,25 @@
...
         "precommit": "npm run check-sensitive && lint-staged",
-        "version": "bash ../.github/scripts/pkg-version.sh template-sync && ../.github/scripts/pkg-version.sh core-template"
-    },
-    "dependencies": {
-        "@microsoft/adaptivecards-tools": "^0.1.2",
-        "@microsoft/teamsfx": "^0.5.1"
+        "version": "bash ../.github/scripts/pkg-version.sh template-sync && bash ../.github/scripts/pkg-version.sh core-template"
     },
+    "templates": [
+        "blazor-base/default",
```
hund030, line 12: blazor-base/csharp/default?

templates/package.json
```diff
@@ -6,12 +6,25 @@
...
+        "bot/csharp/default",
+        "bot/js/default",
+        "bot/ts/default",
+        "bot-msgext/js/default",
+        "bot-msgext/ts/default",
+        "function-base/js/default",
+        "function-base/ts/default",
+        "function-trigger/js/default",
```
hund030, line 20: function-triggers/js/HTTPTrigger

templates/package.json
wenytang-ms: thanks! Upodated.

.github/scripts/sync-templates.js
```diff
@@ -0,0 +1,37 @@
...
+    const pkgVersion = require(pkgPath).version;
+    templatesDeps[`${pkgName}`] = pkgVersion;
+}
+for(let file of depPkgs) {
+    const pkg_ = fse.readJsonSync(file);
+    const dep_ = pkg_.dependencies;
+    let fileChange = false;
+    for(let [key,value] of Object.entries(templatesDeps)){
+        if(dep_[key] && semver.prerelease(semver.minVersion(dep_[key]))) {
+            dep_[key] = `^${value}`;
```
hund030, line 29: What if `value` is an alpha version? `^3.4.1-alpha.b2569834.0`?

templates/package.json
wenytang-ms: thanks!
updated.

.github/scripts/sync-templates.js
wenytang-ms: value ref from target specific version number of package.json file. 
Such a version number is not possible

templates/package.json
```diff
@@ -6,8 +6,25 @@
...
+        "blazor-base/csharp/default",
+        "bot/csharp/default",
+        "bot/js/default",
+        "bot/ts/default",
+        "bot-msgext/js/default",
+        "bot-msgext/ts/default",
+        "function-base/js/default",
+        "function-base/ts/default",
+        "function-trigger/js/HTTPTrigger",
+        "function-trigger/ts/HTTPTrigger",
```
hund030, line 21: function-trigger**s**
Image I add a new template but I add a wrong path to this list. Can we setup test to prevent this kind of error?

wenytang-ms: You own the templates, please add test cases, thanks!

SLdragon (dismissed):

hund030 (approved):

tecton (approved):

SLdragon (approved):

LongOddCode (approved):

qinezh (approved):

PR #4041, opened 2022-02-25
chore: update login get status

xiaolang124:
1. Login get status fail will not trigger open login page.


swatDong (approved):
Just curious - any stable repro step for this issue?

PR #4042, opened 2022-02-25
chore(dotnet): v3 dotnet plugin add instance & add bicep

hund030:
1. In addInstance, dotnet plugin updates the manifest to refer its config.
2. In addBicep, dotnet plugin generate biceps file for hosting web app and bot service for bot hosting.

E2E TEST: cli does not support dotnet scenario yet.


packages/fx-core/src/plugins/resource/dotnet/index.ts
```diff
@@ -0,0 +1,232 @@
...
+      Reference: WebappBicep.Reference,
+    };
+    return result;
+  }
+
+  private async updateWebAppBicep(pluginCtx: {
+    plugins: string[];
+    capabilities: string[];
+  }): Promise<v3.BicepTemplate> {
+    const configModule = await generateBicepFromFile(PathInfo.webappConfigModulePath, pluginCtx);
```
JerryYangKai, line 200: Just a  question, why this function named updateXXX, but the logic is call generateXXX and return the result

hund030: Good question.

PR #4059, opened 2022-02-28
perf(app studio): update app studio plugin to allow scaffold non-sso tab project

KennethBWSong:


nliu-ms (approved):

packages/fx-core/src/plugins/resource/appstudio/plugin.ts
```diff
@@ -1645,14 +1659,18 @@ export async function createManifest(
   hasFrontend: boolean,
   hasBot: boolean,
   hasMessageExtension: boolean,
-  isSPFx: boolean
+  isSPFx: boolean,
+  hasAad = true
 ): Promise<TeamsAppManifest | undefined> {
-  if (!hasBot && !hasMessageExtension && !hasFrontend) {
+  if (!hasBot && !hasMessageExtension && !hasFrontend && !hasAad) {
```
blackchoey, line 1665: will there be cases that we have no bot/tab/message extension, but have AAD ?

packages/fx-core/src/plugins/resource/appstudio/plugin.ts
```diff
@@ -1645,14 +1659,18 @@ export async function createManifest(
...
     throw new Error(`Invalid capability`);
   }
-  if (!isSPFx || hasBot || hasMessageExtension) {
-    const manifestString = TEAMS_APP_MANIFEST_TEMPLATE_FOR_MULTI_ENV;
+  if (!isSPFx || hasBot || hasMessageExtension || hasAad) {
```
blackchoey, line 1668: same to this condition

packages/fx-core/src/plugins/resource/appstudio/plugin.ts
KennethBWSong: I think we do not have such scenario now. Here I try to make it clear since it won't have any influence on current logic, but may benefit if we have such scenario in the future.

blackchoey (approved):

PR #4063, opened 2022-02-28
fix(dotnet): update readme content

JerryYangKai:
Update getting started in template.

@therealjohn  Please help review. Thanks

E2E TEST: no need.


tecton (approved):

therealjohn (commented):
Instead of instructing to delete the launchURL, could we include two profiles? 

1 for launching in Teams and 1 for launching in the browser, but not in Teams.

JerryYangKai:
Sure, I will make it in next pr.

PR #4065, opened 2022-02-28
chore(existing app): ensure package.json after add capability for existing app

jayzhang:


packages/fx-core/src/core/index.ts
```diff
@@ -833,6 +833,15 @@ export class FxCore implements v3.ICore {
           ctx.envInfoV2,
           this.tools.tokenProvider
         );
+        //兼容existing app
```
swatDong, line 836: better in English

swatDong (approved):

PR #4067, opened 2022-02-28
chore(localdebug): improve local debug telemetry

kuojianlu:
- add starting and stopping events for env cheker
- add starting and stopping events for service tasks
- add starting and stopping events for pre-debug-check
- block debugging when there is already a debug session
- adjust local debug correlation id


packages/vscode-extension/src/handlers.ts
```diff
@@ -874,14 +877,37 @@ function checkCoreNotEmpty(): Result<null, SystemError> {
 }
 
 export async function validateAzureDependenciesHandler(): Promise<string | undefined> {
+  // skip debugging if there is already a debug session
+  if (allRunningDebugSessions.size > 0) {
```
swatDong, line 881: Consider make this a utility to avoid duplicated code

packages/vscode-extension/src/debug/commonUtils.ts
```diff
@@ -240,16 +240,20 @@ export async function loadPackageJson(path: string): Promise<any> {
 }
 
 // Helper functions for local debug correlation-id, only used for telemetry
-let localDebugCorrelationId: string | undefined = undefined;
+// Use a 2-element tuple to handle concurrent F5
+const localDebugCorrelationIds: [string, string] = ["no-session-id", "no-session-id"];
+let current = -1;
```
swatDong, line 245: :P init `current` to `0` then no need to check `current >= 0`

packages/vscode-extension/src/handlers.ts
kuojianlu: Good advice, updated.

packages/vscode-extension/src/debug/commonUtils.ts
kuojianlu: Updated.

swatDong (dismissed):

packages/vscode-extension/src/handlers.ts
```diff
@@ -874,14 +874,34 @@ function checkCoreNotEmpty(): Result<null, SystemError> {
 }
 
 export async function validateAzureDependenciesHandler(): Promise<string | undefined> {
+  if (commonUtils.checkAndSkipDebugging()) {
+    // return non-zero value to let task "exit ${command:xxx}" to exit
+    return "1";
+  }
+
+  try {
+    ExtTelemetry.sendTelemetryEvent(TelemetryEvent.DebugEnvCheckStart);
```
HuihuiWu-Microsoft, line 883: Do we need to try catch sendTelemetryEvent?

packages/vscode-extension/src/handlers.ts
```diff
@@ -891,10 +911,31 @@ export async function validateAzureDependenciesHandler(): Promise<string | undef
...
+  } catch {
+    // ignore telemetry error
+  }
+
   const vscodeDepsChecker = new VSCodeDepsChecker(vscodeLogger, vscodeTelemetry);
   const shouldContinue = await vscodeDepsChecker.resolve([DepsType.SpfxNode, DepsType.Ngrok]);
+
+  try {
+    ExtTelemetry.sendTelemetryEvent(TelemetryEvent.DebugEnvCheck, {
```
HuihuiWu-Microsoft, line 929: Generally we use sendTelemetryErrorEvent api to attach extra info for error situation

packages/vscode-extension/src/handlers.ts
kuojianlu: We add try catch in case any exception... what is the recommended way to use `sendTelemetryEvent`?

packages/vscode-extension/src/handlers.ts
kuojianlu: If error occurs, there will be detailed error telemetries in deps checker logic. Here is just to record whether deps checking passes. I do not use `sendTelemetryErrorEvent` since it needs to construct an error object, which will be actually not useful for BI.

packages/vscode-extension/src/handlers.ts
HuihuiWu-Microsoft: Confirm with @chagong , no need to try catch, shouldn't throw error inside sendTelemetryEvent.(Just a suggestion)

swatDong (approved):

HuihuiWu-Microsoft (approved):

PR #4069, opened 2022-03-01
refactor(localdebug): remove useless script insallAll

XiaofuHuang:
To support existing app

### DIFF
--- a/packages/fx-core/src/plugins/solution/fx-solution/debug/scaffolding.ts
+++ b/packages/fx-core/src/plugins/solution/fx-solution/debug/scaffolding.ts
@@ -29,8 +29,6 @@ import { TelemetryEventName, TelemetryUtils } from "./util/telemetry";
 import { ScaffoldLocalDebugSettingsError } from "./error";
 import { isConfigUnifyEnabled } from "../../../../common/tools";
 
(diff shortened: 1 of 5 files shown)


XiaofuHuang:
E2E TEST: https://github.com/OfficeDev/TeamsFx/blob/384d07519ac1e79041231ad7895e13486eb90262/packages/cli/tests/e2e/scaffold/AzureAppScaffold.tests.ts#L31

packages/fx-core/src/plugins/solution/fx-solution/debug/scaffolding.ts
```diff
@@ -29,8 +29,6 @@ import { TelemetryEventName, TelemetryUtils } from "./util/telemetry";
 import { ScaffoldLocalDebugSettingsError } from "./error";
 import { isConfigUnifyEnabled } from "../../../../common/tools";
 
-const PackageJson = require("@npmcli/package-json");
```
swatDong, line 32: Shall we remove `@npmcli/package-json` from dependencies?

XiaofuHuang: Thx!

swatDong (approved):

PR #4072, opened 2022-03-01
chore: compatible with 0.10.0 azure account extension

xiaolang124:
1.Status change logic when version<0.10.0
Initializing->loggedIn/loggedOut

Status change logic when version==0.10.0
Initializing->loggingIn->loggedIn/loggedOut

2.```azureAccount.waitForSubscriptions()``` will wait for login finish when version==0.10.0.

### DIFF
--- a/packages/vscode-extension/src/commonlib/azureLogin.ts
+++ b/packages/vscode-extension/src/commonlib/azureLogin.ts
@@ -21,6 +21,7 @@ import { LoginFailureError } from "./codeFlowLogin";
 import * as vscode from "vscode";
 import * as identity from "@azure/identity";
 import {
+  initializing,
   loggedIn,
   loggedOut,
   loggingIn,
@@ -379,8 +380,11 @@ export class AzureAccountManager extends login implements AzureAccountProvider {
 
   async getStatus(): Promise<LoginStatus> {
     const azureAccount = this.getAzureAccount();
-    // add this to make sure Azure Account Extension has fully initialized
-    await azureAccount.waitForSubscriptions();
+    if (this.checkVersion()) {
+      // add this to make sure Azure Account Extension has fully initialized
+      // this will wait for login finish when version >= 0.10.0, so loggingIn status will be ignored
+      await azureAccount.waitForSubscriptions();
+    }
     if (azureAccount.status === loggedIn) {
       const credential = await this.doGetAccountCredentialAsync();
       const token = await credential?.getToken();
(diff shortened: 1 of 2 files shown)


packages/vscode-extension/src/commonlib/azureLogin.ts
```diff
@@ -379,8 +380,11 @@ export class AzureAccountManager extends login implements AzureAccountProvider {
 
   async getStatus(): Promise<LoginStatus> {
     const azureAccount = this.getAzureAccount();
-    // add this to make sure Azure Account Extension has fully initialized
-    await azureAccount.waitForSubscriptions();
+    if (this.checkVersion()) {
+      // add this to make sure Azure Account Extension has fully initialized
+      // this will wait for login finish when version >= 0.10.0, so loggingIn status will be ignored
+      await azureAccount.waitForSubscriptions();
```
swatDong, line 386: Will this be a long waiting? Any need to add timeout?

packages/vscode-extension/src/commonlib/azureLogin.ts
```diff
@@ -565,6 +596,19 @@ export class AzureAccountManager extends login implements AzureAccountProvider {
...
     }
   }
+
+  checkVersion(): boolean {
+    try {
+      const version: string =
+        vscode.extensions.getExtension<AzureAccount>("ms-vscode.azure-account")!.packageJSON
+          .version;
+      const versionDetail = version.split(".");
+      return parseInt(versionDetail[1]) < 10;
```
swatDong, line 606: To be safe, check major version as well... In case it publishes `1.x.x`

packages/vscode-extension/src/commonlib/azureLogin.ts
xiaolang124: Before 0.10.0, this will be quick. And after 0.10.0 this line will not be executed.

packages/vscode-extension/src/commonlib/azureLogin.ts
xiaolang124: Thanks for the advice, fixed.

packages/vscode-extension/src/commonlib/azureLogin.ts
xiaolang124: Before 0.10.0, this will be quick. And after 0.10.0 this line will not be executed.

swatDong: Make sense.
BTW, better to rename `checkVersion` to `isLegacyVersion`. I was confused by the name so thought this line is for 0.10.0...

swatDong (approved):

packages/vscode-extension/src/commonlib/azureLogin.ts
xiaolang124: Done.

PR #4075, opened 2022-03-01
refactor: scaffold one template

nliu-ms:
Scaffold only one template.


packages/fx-core/src/plugins/resource/appstudio/manifestTemplate.ts
```diff
@@ -34,35 +34,57 @@ import {
...
+    let localAppName = appName;
+    if (suffix.length + appName.length <= TEAMS_APP_SHORT_NAME_MAX_LENGTH) {
+      localAppName = localAppName + suffix;
+    }
+    localManifestString = replaceConfigValue(localManifestString, "appName", localAppName);
+    const localManifest = JSON.parse(localManifestString);
+    await saveManifest(projectRoot, localManifest, true);
 
-  let remoteManifestString = TEAMS_APP_MANIFEST_TEMPLATE_V3;
-  remoteManifestString = replaceConfigValue(remoteManifestString, "appName", appName);
```
nliu-ms, line 63: this step is not required, so just remove it.

packages/fx-core/src/plugins/resource/appstudio/manifestTemplate.ts
```diff
@@ -34,35 +34,57 @@ import {
...
-    localAppName = localAppName + suffix;
-  }
-  localManifestString = replaceConfigValue(localManifestString, "appName", localAppName);
-  const localManifest = JSON.parse(localManifestString);
-  await saveManifest(projectRoot, localManifest, true);
+  if (isConfigUnifyEnabled()) {
+    const manifestString = TEAMS_APP_MANIFEST_TEMPLATE_V3;
+    const manifest = JSON.parse(manifestString);
+    if (existingApp) {
+      manifest.developer = {
```
HuihuiWu-Microsoft, line 65: Is this all needed for existing app?

nliu-ms:
E2E TEST: AzureAppScaffold

packages/fx-core/src/plugins/resource/appstudio/manifestTemplate.ts
nliu-ms: For existing app, it has different developer config:
image

HuihuiWu-Microsoft: When will tab content updated to the template?

jayzhang (approved):

PR #4086, opened 2022-03-01
refactor: get capability from manifest, rather than project settings

nliu-ms:
Test with miniapp, provision succeed.


nliu-ms:
E2E TEST: TestRemoteHappyPath

jayzhang (approved):

Issue #4044, opened 2022-02-25
bug report: Cannot read property 'provisionSucceeded' of undefined


creage:
**Describe the bug**
The Teams Toolkit throws an error in VSCode output
```log
[2022-02-25T08:49:35.001Z] [Info] - Teams Toolkit extension is now active!
[2022-02-25T08:49:36.387Z] [Info] - [core] start task:getProjectConfig , API v3: false
[2022-02-25T08:49:38.705Z] [Info] - [core] finish task:getProjectConfig , time: 2318 ms
[2022-02-25T08:49:38.722Z] [Error] - code:Ext.TypeError, message: Cannot read property 'provisionSucceeded' of undefined
```

**Screenshots**
![image](https://user-images.githubusercontent.com/910967/155685259-6e4a5d1d-8bb8-40c9-95d1-37fa6d4e3271.png)

**VS Code Extension Information (please complete the following information):**
 - OS: Windows10
 - Version 3.4.0

**CLI Information (please complete the following information):**
 - OS: Windows10
 - Version 0.11.0

adashen:
close as the issue is fixed.

PR #4103, opened 2022-03-02
chore: fix load provision

xiaolang124:
1.Fix get provisionSucceeded when ```state``` file exists but ```state.solution``` does not exist.


xiaolang124:
Fix: https://github.com/OfficeDev/TeamsFx/issues/4044

HuihuiWu-Microsoft (approved):

PR #4108, opened 2022-03-02
chore: add is localDebug flag

xiaolang124:
1. Add ```isLocalDebug``` for manifest merge.


nliu-ms (approved):

packages/fx-core/src/plugins/resource/utils4v2.ts
```diff
@@ -404,6 +404,9 @@ export async function configureLocalResourceAdapter(
   if (isConfigUnifyEnabled() && envInfo) {
     pluginContext.envInfo.state = objectToMap(envInfo!.state);
   }
+  if (envInfo?.config.isLocalDebug) {
```
nliu-ms, line 407: both provisionLocal "& configureLocal should add this value.

xiaolang124: I will add ```provisionLocal``` later. Thanks

PR #4127, opened 2022-03-03
chore: update build local zip, update local manifest

xiaolang124:
1. Support build local zip and update local manifest.
2. ```codeLens``` for publish environment need to update.


dooriya:
From scenario perspective, do we really need to support publish the app which is running in a local environment?

xiaolang124:
> From scenario perspective, do we really need to support publish the app which is running in a local environment?

Not really publish Teams, it is used for publishing in manifest file. Publish will only work for remote environments. Any additions? @HuihuiWu-Microsoft 
image

HuihuiWu-Microsoft:
> > From scenario perspective, do we really need to support publish the app which is running in a local environment?
> 
> My bad title was wrong. Not really publish Teams, it is used for updating local manifest. Publish will only work for remote environments. Any additions? @HuihuiWu-Microsoft image

Yes, this is to align with our existing feature: user can update manifest manually to dev portal. Will not affect publish feature

HuihuiWu-Microsoft (approved):
LGTM

nliu-ms (approved):

PR #4142, opened 2022-03-07
chore(flag): add feature flag for incremental app building

qinezh:
Add experiential feature flag to enable to init a teamsfx app and unify the local & remote configs.

Preview:

### DIFF
--- a/packages/fx-core/src/common/tools.ts
+++ b/packages/fx-core/src/common/tools.ts
@@ -378,7 +378,11 @@ export function isBicepEnvCheckerEnabled(): boolean {
 }
 
 export function isConfigUnifyEnabled(): boolean {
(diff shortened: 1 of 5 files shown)


packages/fx-core/src/common/tools.ts
```diff
@@ -378,7 +378,7 @@ export function isBicepEnvCheckerEnabled(): boolean {
 }
 
 export function isConfigUnifyEnabled(): boolean {
```
qinezh, line 380: The method will be renamed in another separated PR to avoid too many changes.

xiaolang124 (approved):

HuihuiWu-Microsoft (approved):

PR #4148, opened 2022-03-07
chore: review bot help and change error handle

Siglud:


packages/cli/tests/e2e/bot/CommandExecutionIsUserError.ts
```diff
@@ -53,7 +53,7 @@ describe("Error type should be expected", function () {
       await CliHelper.deployProject(ResourceToDeploy.Bot, projectPath);
     } catch (e) {
       chai.assert.isTrue(e instanceof PluginError);
-      chai.assert.isTrue(e.ErrorType == ErrorType.User);
+      chai.assert.isTrue(e.ErrorType == "User");
```
IvanJobs, line 56: Nit: Can you share some insights for this change (use plain string to replace enum)?

packages/cli/tests/e2e/bot/CommandExecutionIsUserError.ts
```diff
@@ -29,7 +29,7 @@ describe("Error type should be expected", function () {
   const env = environmentManager.getDefaultEnvName();
 
   after(async () => {
-    await cleanUp(appName, projectPath, true, true, false, true);
+    await cleanUp(appName, projectPath, true, true, false, "dev");
```
IvanJobs, line 32: Good catch! since there's a variable `env` above, how about using the variable instead?

packages/cli/tests/e2e/bot/CommandExecutionIsUserError.ts
Siglud: typescript's Enum is not recommended by every TS book,  Their behavior is weird and strange.
see more in [this](https://blog.logrocket.com/why-typescript-enums-suck/) , more weird action: https://pbs.twimg.com/media/FMHHwd9VUAICzro?format=jpg&name=large
using literal types is more recommended in TS

packages/cli/tests/e2e/bot/CommandExecutionIsUserError.ts
Siglud: done

IvanJobs (dismissed):
lgtm

packages/fx-core/src/plugins/resource/bot/errors.ts
```diff
@@ -68,7 +106,7 @@ export class PluginError extends Error {
 
 export class PreconditionError extends PluginError {
   constructor(message: string, suggestions: string[]) {
-    super(ErrorType.User, ErrorNames.PRECONDITION_ERROR, message, suggestions);
+    super("User", ErrorNames.PRECONDITION_ERROR, message, suggestions);
```
eriolchan, line 109: If there are many place use "User" or "System", we could make them as constant if you don't like string-based enum. This would help us easy to update the string value in one place if there is any change.

packages/fx-core/src/plugins/resource/bot/errors.ts
```diff
@@ -78,10 +116,14 @@ export class SomethingMissingError extends PreconditionError {
   }
 }
 
-export function CheckThrowSomethingMissing(name: string, value: any): void {
+export function CheckThrowSomethingMissing<T>(
```
eriolchan, line 119: Do we still need this method? Is there any possible to leverage below method because there are two methods with similar name which may confused people.

packages/fx-core/src/common/projectSettingsHelper.ts
```diff
@@ -126,6 +126,6 @@ export function newProjectSettings(): ProjectSettings {
   };
   return projectSettings;
 }
-export function isVSProject(projectSettings: ProjectSettings): boolean {
-  return projectSettings.programmingLanguage === "csharp";
+export function isVSProject(projectSettings?: ProjectSettings): boolean {
```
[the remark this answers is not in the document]
Siglud: I did change the behavior of this method, so, it should be working fine with the other code.

packages/fx-core/src/plugins/resource/bot/errors.ts
Siglud: this method use type guard like:
```
if (CheckThrowSomethingMissing("name", value)) {
  // then value will not be undifined
}
```

the second one can be used like 
```
// value will not be undifined
let value = CheckAndThrowIfMissing("name", value);
```

different action, the same effect

packages/cli/tests/e2e/bot/CommandExecutionIsUserError.ts
Siglud: changed to other code type

eriolchan (dismissed):

eriolchan (approved):

PR #4151, opened 2022-03-07
chore(notification): add notification feature flag

XiaofuHuang:
Detail: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/13762945


a1exwang (commented):
Do we need to add a vscode settings like other flags in `vscode-extension/src/utils/commonUtils.ts:206`

swatDong (approved):

swatDong:
> Do we need to add a vscode settings like other flags in `vscode-extension/src/utils/commonUtils.ts:206`

IMO, no. This is for preview test and no need to expose to public release.

a1exwang (approved):

PR #4162, opened 2022-03-08
chore(core): localization support for fx-core

jayzhang:


packages/fx-core/src/common/localizeUtils.ts
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { Locale } from "../core/globalVars";
+import { getResourceFolder } from "../folder";
+import * as path from "path";
+import fs from "fs-extra";
+import * as util from "util";
+
+export function getLocalizedString(key: string, ...params: any[]): string {
```
HuihuiWu-Microsoft, line 10: Maybe allow a default value if no key defined in package.nls.json

jayzhang: We can just return empty string or throw an exception if the key does not exist. I don't think it is meaningful that developer provide a default value, if the value does not exist, he should add it in the package.nls.json file

HuihuiWu-Microsoft: Yes, that's a good way to make sure developer added the string in package.nls.json. My concern is the UI might get broken if something wrong with the json file.

jayzhang: I think UI will not broken with an empty message.

jayzhang:
Change points:
1. replace resource/strings.json by resource/package.nls.json
2. component call getLocalizedString() to get the string value instead of directly import strings.json
3. replace some error creation function by error class

hund030 (approved):
Approve for frontend plugin

wenytang-ms (approved):
LGTM on migrate and arm

Alive-Fish (approved):
Approve for cli part.

packages/fx-core/src/common/localizeUtils.ts
HuihuiWu-Microsoft: Please make sure all strings are extracted to json file when reviewing PR.

HuihuiWu-Microsoft (approved):

packages/fx-core/src/core/middleware/envInfoWriter.ts
```diff
@@ -20,7 +19,7 @@ export function EnvInfoWriterMW(skip = false): Middleware {
     try {
       await next();
     } catch (e) {
-      if ((e as any)["name"] === getStrings().solution.CancelProvision) throw e;
+      if ((e as any)["name"] === "CancelProvision") throw e;
```
a1exwang, line 22: Is it better to extract the constant (or to an error class) into `solution/fx-solution/constants.ts` instead of hard coding in different places?

a1exwang (approved):

meifans (approved):
LGTM on ProjectMigrator

packages/fx-core/src/core/middleware/envInfoWriter.ts
jayzhang: Sure. There exists the constant, I will improve later.

SLdragon (approved):

LongOddCode (approved):

1yefuwang1 (approved):

PR #4182, opened 2022-03-10
fix: app name does not show when local debug

HuihuiWu-Microsoft:
Before:
image

After:
image

E2E TEST: only for extension


packages/fx-core/src/plugins/resource/appstudio/plugin.ts
```diff
@@ -181,6 +181,7 @@ export class AppStudioPluginImpl {
     }
 
     const view = {
+      config: ctx.envInfo.config,
```
HuihuiWu-Microsoft, line 184: This should work for both situations(isConfigUnified = true/false)

nliu-ms (approved):

PR #4186, opened 2022-03-10
chore: setup new workflow for cli F5

meifans:
Happy path:

happy-path-gif

Nothing: 
image

E2E TEST: not needed since unit test cases are added.


packages/cli/src/cmds/preview/commonUtils.ts
```diff
@@ -93,7 +93,7 @@ export function createTaskStartCb(
             color: Colors.BRIGHT_GREEN,
           },
         ];
-        cliLogger.necessaryLog(LogLevel.Info, getColorizedString(message));
+        await cliLogger.log(LogLevel.Debug, getColorizedString(message));
```
swatDong, line 96: How will this impact `npm run dev:teamsfx` tasks?

packages/cli/src/cmds/preview/preview.ts
```diff
@@ -532,7 +602,7 @@ export default class Preview extends YargsCommand {
 
   private async startNgrok(
     workspaceFolder: string,
-    depsChecker: CliDepsChecker
```
swatDong, line 535: Any code clean up in the old deps checker?

packages/cli/src/cmds/preview/commonUtils.ts
meifans: The location of detail log is hidden for INFO level.
When users debug `preview`,  they can set DEBUG level to get the location of detail log for tasks.

kuojianlu: I prefer showing the localtion of log for INFO level. Otherwise it is not convenient for user to see the service status and troubleshot.

packages/cli/src/cmds/preview/preview.ts
meifans: Removed and add UT

packages/cli/src/cmds/preview/commonUtils.ts
meifans: Well.. It's a little inconvenient.
But, service status can be displayed by a progress bar. I just think it's normal for users to use `teamsfx preview --debug` for troubleshooting.  
And we have shown how to use --debug.
image

kuojianlu: In my understanding, `--debug` is to debug the CLI, for example if some unexpected error happens, CLI users can provide these debug information then CLI developers can know what happens. However, the service status is necessary for users to debug their apps.

meifans: @kuojianlu What is your concern?
Users can get location of detail log with `teamsfx preview --debug`, as shown below:
image

image

kuojianlu: Another scenario is that users may also want to see their output contents in codes. Suppose users write many `console.log(xxx)` codes, and they want to know which outputs are printed when debugging the app.

meifans: In that scenario, we do need give user the location of services log. 
Added.

kuojianlu (approved):

swatDong (approved):

PR #4188, opened 2022-03-10
feat(local-debug): add local debug flow for m365 apps 

kuojianlu:
- add install-app-in-teams command that will be used in tasks.json
- prompt window to let user install app in Teams
- call graph api to get the internal id of an installed app
- resolve the internal id in debug provider

image
image


kuojianlu:
E2E TEST: N/A.

packages/vscode-extension/src/debug/teamsAppInstallation.ts
```diff
@@ -0,0 +1,88 @@
...
+      return internalId === undefined ? await showInstallAppInTeamsMessage(true) : true;
+    } else if (result.value === StringResources.vsc.localDebug.installApp.cancel) {
+      return false;
+    }
+  }
+  return false;
+}
+
+export async function getTeamsAppInternalId(): Promise<string | undefined> {
+  const loginStatus = await graphLoginInstance.getStatus();
```
swatDong, line 58: Since we added new permission scope `TeamsAppInstallation.ReadForUser`, do we need to update graph login?
@xiaolang124

swatDong:
> * add install-app-in-teams command that will be used in tasks.json
> * prompt window to let user install app in Teams
> * call graph api to get the internal id of an installed app
> * resolve the internal id in debug provider
> 
> image image

What if including teams app installation in `pre-debug-check` command?

kuojianlu:
> > * add install-app-in-teams command that will be used in tasks.json
> > * prompt window to let user install app in Teams
> > * call graph api to get the internal id of an installed app
> > * resolve the internal id in debug provider
> > 
> > image image
> 
> What if including teams app installation in `pre-debug-check` command?

At that point, services are not started, users will see a empty page after added app, I think it introduces extra risk. But it depends on the UX design. Let me ask Junjie for the answer.

packages/vscode-extension/src/debug/teamsAppInstallation.ts
kuojianlu: From my practice, adding this scope seems good enough.

packages/vscode-extension/src/debug/teamsAppInstallation.ts
```diff
@@ -0,0 +1,88 @@
...
+  }
+  const debugConfig = await commonUtils.getDebugConfig(true);
+  if (debugConfig?.appId === undefined) {
+    throw returnUserError(
+      new Error("Debug config not found"),
+      ExtensionSource,
+      ExtensionErrors.GetTeamsAppInstallationFailed
+    );
+  }
+  const url = `https://graph.microsoft.com/v1.0/users/${loginStatus.accountInfo.upn}/teamwork/installedApps?$expand=teamsApp,teamsAppDefinition&$filter=teamsApp/externalId eq '${debugConfig?.appId}'`;
```
swatDong, line 74: `upn` may contains special character. (See this [doc](https://docs.microsoft.com/en-us/graph/api/user-get?view=graph-rest-1.0&tabs=http#http-request)). Consider to use user id.

packages/vscode-extension/src/debug/teamsAppInstallation.ts
swatDong: It's weird... In [code](https://github.com/OfficeDev/TeamsFx/blob/dev/packages/vscode-extension/src/commonlib/graphLogin.ts#L28) we only have `Application.ReadWrite.All` scope, so how could `TeamsAppInstallation.ReadForUser` work as well...

packages/vscode-extension/src/debug/teamsAppInstallation.ts
kuojianlu: Good point. `oid` in token is the user id. Can we assume every token contains `oid` field?

packages/vscode-extension/src/debug/teamsAppInstallation.ts
swatDong: One guess is that `Application.ReadWrite.All` includes `TeamsAppInstallation.ReadForUser` but have not found any official doc.
To be safety, should add `TeamsAppInstallation.ReadForUser` to our graph scopes

packages/vscode-extension/src/debug/teamsAppInstallation.ts
swatDong: Of course. And it's recommended to use `oid` instead of `upn`, by [doc](https://docs.microsoft.com/en-us/azure/active-directory/develop/access-tokens#validate-the-user-has-permission-to-access-this-data).

packages/vscode-extension/src/debug/teamsAppInstallation.ts
kuojianlu: What I mean is adding `TeamsAppInstallation.ReadForUser` to the graph scopes will work well. But not sure if there are some others works not considered.

swatDong: Oh I see... just not included in this PR, right?

kuojianlu: Yes, will add afterwards since this PR do not affect the current behavior.

packages/vscode-extension/src/debug/teamsAppInstallation.ts
kuojianlu: Updated to use `oid`.

swatDong (approved):

kuojianlu:
@swatDong Confirmed with Junjie and he prefers installing app after services started.

swatDong (approved):

packages/vscode-extension/src/debug/teamsAppInstallation.ts
```diff
@@ -0,0 +1,88 @@
...
+import { ExtensionErrors, ExtensionSource } from "../error";
+import * as StringResources from "../resources/Strings.json";
+
+import axios from "axios";
+import { returnSystemError, returnUserError } from "@microsoft/teamsfx-api";
+
+export async function showInstallAppInTeamsMessage(detected: boolean): Promise<boolean> {
+  const message = `${
+    detected
+      ? StringResources.vsc.localDebug.installApp.detection
```
HuihuiWu-Microsoft, line 16: Please extract the strings to package.nls.json and use localize() API to get string for later localization.

kuojianlu: Updated.

HuihuiWu-Microsoft (approved):

swatDong (approved):

PR #4194, opened 2022-03-11
chore: set eslint to lerna

wenytang-ms:
ESLint set to be pkg of lerna, and cut the lock file deps from eslint-cofig and prettier-config, and reduce the vulnerable impact from eslint and prettier-config
later job:
1. set npm version to 7 to support the peerdependencies
2. lerna + workspace


jayzhang (approved):

tecton (approved):

Alive-Fish (approved):
Look good to cli and server package

PR #4201, opened 2022-03-11
fix: empty strings for vsce package

HuihuiWu-Microsoft:
Issue:
MicrosoftTeams-image (9)

E2E TEST: only happens for vsce package


packages/vscode-extension/src/extension.ts
```diff
@@ -55,7 +55,7 @@ import { loadLocalizedStrings, localize } from "./utils/localizeUtils";
 export let VS_CODE_UI: VsCodeUI;
 
 export async function activate(context: vscode.ExtensionContext) {
-  VsCodeLogInstance.info(localize("teamstoolkit.common.activate"));
+  VsCodeLogInstance.info("Teams Toolkit extension is now active!");
```
HuihuiWu-Microsoft, line 58: Can't get localized string since extension variables haven't been loaded. Also I think no need to localize here.

packages/vscode-extension/src/utils/localizeUtils.ts
```diff
@@ -71,15 +71,15 @@ function getLocalizedString(key: string, defValue?: string): string {
 export function loadLocalizedStrings(): void {
   loadedLocale = parseLocale();
 
-  const nlsFile = path.join(EXTENSION_ROOT_DIR, "..", `package.nls.${loadedLocale}.json`);
+  const nlsFile = path.join(ext.context.extensionPath, `package.nls.${loadedLocale}.json`);
```
HuihuiWu-Microsoft, line 74: Use extension context instead of __dirname since the folder structure changed after webpack. It works when local debug because we use tsc for local debug and webpack is just for the webviews.

packages/vscode-extension/src/extension.ts
tecton: If the string is not used any more, you might also remove it from `package.nls.json`.

HuihuiWu-Microsoft: Updated

tecton (approved):

PR #4208, opened 2022-03-14
chore(local-debug): add localhost:port to manifest valid domains

kuojianlu:
- add localhost:port to local manifest valid domains to support m365 apps

### DIFF
--- a/packages/fx-core/src/plugins/resource/appstudio/plugin.ts
+++ b/packages/fx-core/src/plugins/resource/appstudio/plugin.ts
@@ -1413,6 +1413,9 @@ export class AppStudioPluginImpl {
     if (tabDomain) {
       validDomains.push(tabDomain);
     }
+    if (tabEndpoint && isLocalDebug) {
+      validDomains.push(tabEndpoint.slice(8));
+    }
 
     if (botId) {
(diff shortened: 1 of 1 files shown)


packages/fx-core/src/plugins/resource/appstudio/plugin.ts
```diff
@@ -1413,6 +1413,9 @@ export class AppStudioPluginImpl {
     if (tabDomain) {
       validDomains.push(tabDomain);
     }
+    if (tabEndpoint && isLocalDebug) {
```
swatDong, line 1416: Do we need to remove `localhost` if adding `localhost:port`?
Or any side-effect if keep both `localhost` and `localhost:port`?

kuojianlu: From my practice, there is no side-effect to keep both. The [manifest schema doc](https://docs.microsoft.com/en-us/microsoftteams/platform/resources/schema/manifest-schema#validdomains) does not have any useful information on the format of host:port...

swatDong (approved):

nliu-ms (approved):

PR #4219, opened 2022-03-14
refactor: build package for local

nliu-ms:
Interactive and non-interactive
image

work item: 13786618


nliu-ms:
E2E TEST: https://github.com/OfficeDev/TeamsFx/actions/runs/1984510515

Alive-Fish (approved):

PR #4220, opened 2022-03-14
perf(frontend): add non-sso template

KennethBWSong:


templates/tab/js/non-sso/package.json
```diff
@@ -0,0 +1,52 @@
...
+    "version": "0.1.0",
+    "private": true,
+    "dependencies": {
+        "@fluentui/react-northstar": "^0.54.0",
+        "@microsoft/mgt-element": "2.3.2-next.teamsfx.f5fc7d2",
+        "@microsoft/mgt-react": "2.3.2-next.teamsfx.f5fc7d2",
+        "@microsoft/mgt-teamsfx-provider": "2.3.2-next.teamsfx.f5fc7d2",
+        "@microsoft/microsoft-graph-client": "^3.0.1",
+        "@microsoft/teams-js": "^1.9.0",
+        "@microsoft/teamsfx": "^0.5.1",
```
SLdragon, line 12: Use latest 0.6.0?

templates/tab/js/non-sso/src/index.css
```diff
@@ -0,0 +1,13 @@
+* {
+  box-sizing: border-box;
+  /* outline: 1px dashed #f00; */
```
SLdragon, line 3: If it is not used, you can remove it

templates/tab/ts/non-sso/package.json
```diff
@@ -0,0 +1,54 @@
+{
+    "name": "teamsfx-template-tab",
+    "version": "0.1.0",
+    "private": true,
+    "dependencies": {
+        "@fluentui/react-northstar": "^0.54.0",
+        "@microsoft/teams-js": "^1.9.0",
+        "@microsoft/teamsfx": "^0.5.1",
```
SLdragon, line 8: same here

templates/tab/js/non-sso/package.json
KennethBWSong: Updated. Thx!

templates/tab/ts/non-sso/package.json
KennethBWSong: Updated. Thx!

templates/tab/js/non-sso/src/index.css
KennethBWSong: Here I prefer to keep the same as existing template.

SLdragon: Sure

SLdragon (approved):

kuojianlu:
Seems missing `.env` file.

Curious about: will this non-sso template replace the sso template as the default tab template? If so, when?

KennethBWSong:
> Seems missing `.env` file.
> 
> Curious about: will this non-sso template replace the sso template as the default tab template? If so, when?

Seems `.env` file is ignored. Is it required? I tried local debug & provision, and seems it works fine.
The non-sso template will replace current template after we finish the AAD manifest feature before GA. Currently we have a feature flag for this feature.

templates/tab/js/non-sso/src/components/sample/EditCode.jsx
```diff
@@ -0,0 +1,27 @@
+import React from "react";
+
+var functionName = process.env.REACT_APP_FUNC_NAME || "myFunc";
```
hund030, line 3: Remove function related code?

hund030 (approved):

PR #4221, opened 2022-03-14
chore: refactor bot plugin to support getQuestions

a1exwang:
- add support to `getQuestionsForUserTask`(for adding capabilities) and `getQuestionsForScaffolding`(for creating project) in bot plugin
- UT added

We need this because we need to ask user questions about triggers/hostType after the user selects notification capability.

The is the preparation of https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/13777227

### DIFF
--- a/packages/fx-core/src/plugins/resource/bot/index.ts
+++ b/packages/fx-core/src/plugins/resource/bot/index.ts
@@ -18,7 +18,7 @@ import {
 import { FxBotPluginResultFactory as ResultFactory, FxResult } from "./result";
 import { TeamsBotImpl } from "./plugin";
 import { ProgressBarFactory } from "./progressBars";
-import { LifecycleFuncNames, ProgressBarConstants } from "./constants";
+import { CustomizedTasks, LifecycleFuncNames, ProgressBarConstants } from "./constants";
 import {
   ErrorType,
   InnerError,
@@ -36,7 +36,7 @@ import "./v2";
 import "./v3";
 import { DotnetBotImpl } from "./dotnet/plugin";
 import { PluginImpl } from "./interface";
-import { isVSProject, BotHostTypes } from "../../../common";
+import { isVSProject, BotHostTypes, isBotNotificationEnabled } from "../../../common";
 import { FunctionsHostedBotImpl } from "./functionsHostedBot/plugin";
 import { ScaffoldConfig } from "./configs/scaffoldConfig";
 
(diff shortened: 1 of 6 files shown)


packages/fx-core/src/plugins/resource/bot/index.ts
```diff
@@ -185,6 +191,42 @@ export class TeamsBot implements Plugin {
...
   }
 
+  public async getQuestions(
+    stage: Stage,
+    context: PluginContext
+  ): Promise<Result<QTreeNode | undefined, FxError>> {
+    Logger.setLogger(context.logProvider);
+
+    if (stage === Stage.create) {
+      const result = await TeamsBot.runWithExceptionCatching(
```
Siglud, line 201: just return it, the local variable is redundant

packages/fx-core/src/plugins/resource/bot/index.ts
```diff
@@ -185,6 +191,42 @@ export class TeamsBot implements Plugin {
...
+    }
+  }
+
+  public async getQuestionsForUserTask(
+    func: Func,
+    context: PluginContext
+  ): Promise<Result<QTreeNode | undefined, FxError>> {
+    Logger.setLogger(context.logProvider);
+
+    const result = await TeamsBot.runWithExceptionCatching(
```
Siglud, line 220: redundant local variable result

a1exwang: fixed

packages/fx-core/src/plugins/resource/bot/index.ts
a1exwang: fixed

packages/fx-core/src/plugins/resource/bot/v2/index.ts
```diff
@@ -131,6 +140,23 @@ export class BotPluginV2 implements ResourcePlugin {
     return await deployAdapter(ctx, inputs, envInfo, tokenProvider, this.plugin);
   }
 
+  async getQuestionsForUserTask(
+    ctx: v2.Context,
+    inputs: Inputs,
+    func: Func,
+    envInfo: v2.DeepReadonly<v2.EnvInfoV2>,
+    tokenProvider: TokenProvider
+  ): Promise<Result<QTreeNode | undefined, FxError>> {
```
IvanJobs, line 149: Nit: since there's a FxResult alias in result.ts, use it as return type? align with other lifecycle methods.

IvanJobs (approved):
lgtm (offline synced with aocheng for the context)

PR #4224, opened 2022-03-14
chore: cd install wine64 need update

wenytang-ms:


.github/workflows/cd.yml
```diff
@@ -61,6 +61,7 @@ jobs:
 
       - name: Install wine64
         run: |
+          sudo apt update
```
IvanJobs, line 64: `sudo apt update -y`?

wenytang-ms: no need
https://github.com/OfficeDev/TeamsFx/runs/5538024005?check_suite_focus=true 
this pipeline run success.

IvanJobs (approved):

PR #4227, opened 2022-03-15
feat(m365-func): add m365 func template

kuojianlu:
Add m365 func template: `tempates/function-base/js/m365`, `templates/function-base/ts/m365`. Based on current func template, but update:
1. @microsoft/teamsfx version to 0.6.0-beta.0

The templates will be used in function plugin as a new scenario.

### DIFF
--- a/templates/function-base/js/m365/.funcignore
+++ b/templates/function-base/js/m365/.funcignore
@@ -1,0 +1,11 @@
+*.js.map
+*.ts
+.git*
+.vscode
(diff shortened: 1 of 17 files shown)


kuojianlu:
E2E TEST: N/A

swatDong (approved):

kuojianlu:
Make this PR as draft since we may disable "Add cloud resources" command for M365 apps.

PR #4241, opened 2022-03-15
chore(sql): update for localization

xzf0587:


jayzhang (approved):

packages/fx-core/src/plugins/resource/sql/utils/message.ts
```diff
@@ -28,24 +24,4 @@ export class Message {
 
   public static readonly adminName = (name: string) =>
```
blackchoey, line 25: do we need to localize `checkAadAdmin`, `connectDatabase`, `adminName` messages?

blackchoey (approved):
Please help confirm with @jayzhang whether we need to localize messages printed to VS Code output window before merge.

xzf0587:
> Please help confirm with @jayzhang whether we need to localize messages printed to VS Code output window before merge.

synced with @jayzhang, the message in output is not high priority. I will update it using another pr.

PR #4248, opened 2022-03-15
chore(identity): remove useless code

xzf0587:
1. remove useless code
2. add telemetry for template generate/update


xzf0587:
provision method in index.ts should be kept, otherwise the core will call the provision method and fail.

packages/fx-core/src/plugins/resource/identity/index.ts
```diff
@@ -1,114 +1,47 @@
...
-      return err(error);
-    }
-
-    ctx.config.set(Constants.identityName, this.config.identityName);
-    ctx.config.set(Constants.identityClientId, this.config.identityClientId);
-    ctx.config.set(Constants.identityResourceId, this.config.identityResourceId);
-    TelemetryUtils.sendEvent(Telemetry.stage.provision, true);
-    ctx.logProvider?.info(Message.endProvision);
+  async provision(ctx: PluginContext): Promise<Result<any, FxError>> {
     return ok(undefined);
```
blackchoey, line 39: can we remove this function? since it does nothing

blackchoey (approved):

PR #4257, opened 2022-03-16
chore(vsc): clean up useless command isMultiEnvEnabled

qinezh:


packages/vscode-extension/package.json
```diff
@@ -103,8 +103,7 @@
         },
         {
           "id": "teamsfx-environment",
-          "name": "Environment",
-          "when": "fx-extension.isMultiEnvEnabled"
```
HuihuiWu-Microsoft, line 107: Will this view be displayed when current project is not valid teamsfx project?

qinezh: No, it won't. It will display the welcome view.

HuihuiWu-Microsoft (approved):

PR #4259, opened 2022-03-16
chore(core): get default locale string

jayzhang:


blackchoey (approved):
minor: there're some duplicate logic with `getLocalizedString`, we may extract them to a common function.

PR #4260, opened 2022-03-16
chore: upgrade to consolidate

xiaolang124:
1. Add consolidate upgrade feature.
- Add ```config.local.json``` for local debug.
- Add ```manifest.template.json``` (copy remote or create new manifest for Azure, create new manifest for SPFX).
- Add items to ```.gitignore```.
2. Add consolidate test cases.
3. Add consolidate telemetry.


jayzhang (approved):

packages/fx-core/src/core/middleware/consolidateLocalRemote.ts
```diff
@@ -0,0 +1,326 @@
...
+  TOOLS?.logProvider.warning(
+    `[core] Upgrade success! Old localSettings.json, manifest.local.template.json and manifest.remote.template.json have been backed up to the .backup folder and you can delete it.`
+  );
+
+  if (inputs.platform === Platform.VSCode) {
+    await globalStateUpdate(showConsolidateLogsFlag, true);
+    sendTelemetryEvent(Component.core, TelemetryEvent.ProjectConsolidateGuide, {
+      [TelemetryProperty.Status]: ProjectMigratorGuideStatus.Reload,
+    });
+    await TOOLS?.ui.reload?.();
```
qinezh, line 298: why do we need to reload the window after migration is done?

packages/fx-core/src/core/middleware/consolidateLocalRemote.ts
```diff
@@ -0,0 +1,326 @@
...
+  const backupPath = path.join(inputs.projectPath as string, backupFolder);
+  sendTelemetryEvent(Component.core, TelemetryEvent.ProjectConsolidateBackupConfigStart);
+  const localSettingsFile = path.join(
+    inputs.projectPath as string,
+    ".fx",
+    "configs",
+    "localSettings.json"
+  );
+  if (await fs.pathExists(localSettingsFile)) {
+    await fs.move(localSettingsFile, path.join(backupPath, ".fx", "configs", "localSettings.json"));
```
qinezh, line 239: nit:
* `fx` => `ConfigFolderName`
* `configs` => `InputConfigsFolderName`

packages/fx-core/src/core/middleware/consolidateLocalRemote.ts
```diff
@@ -0,0 +1,326 @@
...
+        inputs.projectPath as string,
+        "templates",
+        "appPackage",
+        "template.remote.manifest.json"
+      );
+      const remoteManifestExist = await fs.pathExists(remoteManifestFile);
+      if (remoteManifestExist) {
+        sendTelemetryEvent(Component.core, TelemetryEvent.ProjectConsolidateCopyAzureManifestStart);
+        const manifestString = (await fs.readFile(remoteManifestFile)).toString();
+        manifest = JSON.parse(manifestString);
```
qinezh, line 195: why do we need to parse the manifest when remote manifest template exists? copy is enough?

packages/fx-core/src/core/middleware/consolidateLocalRemote.ts
```diff
@@ -0,0 +1,326 @@
...
+        sendTelemetryEvent(Component.core, TelemetryEvent.ProjectConsolidateCopyAzureManifest);
+      } else {
+        sendTelemetryEvent(Component.core, TelemetryEvent.ProjectConsolidateAddAzureManifestStart);
```
qinezh, line 198: if the remote manifest template doesn't exist, do we need to generate a new one? Seems it's not the responsibility for migration logic...

packages/fx-core/src/core/middleware/consolidateLocalRemote.ts
xiaolang124: Just to be consistent with upgrade scenario. Removed now.

packages/fx-core/src/core/middleware/consolidateLocalRemote.ts
xiaolang124: If migration does not do this, which part will handle this part? Or just ignore this?

packages/fx-core/src/core/middleware/consolidateLocalRemote.ts
xiaolang124: Just to be consistent with create new manifest code.

qinezh: The problem is the content of original manifest template will be modified, which is unnecessary. For example, the indent might be updated from 2 to 4...

packages/fx-core/src/core/middleware/consolidateLocalRemote.ts
qinezh: Yeah, I think so, we can safely ignore it if remote template not found.

packages/fx-core/src/core/middleware/consolidateLocalRemote.ts
```diff
@@ -0,0 +1,332 @@
...
+import { TOOLS } from "../globalVars";
+import { getLocalizedString } from "../../common/localizeUtils";
+import { createManifest } from "../../plugins/resource/appstudio/plugin";
+import { getManifestTemplatePath } from "../../plugins/resource/appstudio/manifestTemplate";
+import { getTemplatesFolder } from "../../folder";
+import { loadProjectSettings } from "./projectSettingsLoader";
+import { addPathToGitignore, needMigrateToArmAndMultiEnv } from "./projectMigrator";
+
+const upgradeButton = "Upgrade";
+let fromReloadFlag = false;
```
qinezh, line 48: The variable can be removed as reload windows is not required.

xiaolang124: This is used to prevent users from asking multiple times after canceling the upgrade. Not for reload VS Code.

qinezh: Got it. How about renaming the variable to avoid confusion here?

packages/fx-core/src/core/middleware/consolidateLocalRemote.ts
```diff
@@ -0,0 +1,315 @@
...
+    const appDir = `${ctx.root}/${APP_PACKAGE_FOLDER_FOR_MULTI_ENV}`;
+    if (isSPFxProject(projectSettings)) {
+      sendTelemetryEvent(Component.core, TelemetryEvent.ProjectConsolidateAddSPFXManifestStart);
+      const templateManifestFolder = path.join(templatesFolder, "plugins", "resource", "spfx");
+      const manifestFile = path.resolve(
+        templateManifestFolder,
+        "./solution/manifest_multi_env.json"
+      );
+      const manifestString = (await fs.readFile(manifestFile)).toString();
+      manifest = JSON.parse(manifestString);
```
qinezh, line 178: No need to parse the manifest template either?

packages/fx-core/src/core/middleware/consolidateLocalRemote.ts
```diff
@@ -0,0 +1,315 @@
...
+  sendTelemetryEvent(Component.core, TelemetryEvent.ProjectConsolidateGuideStart);
+  await updateGitIgnore(projectPath, TOOLS.logProvider, backupFolder);
+
+  TOOLS?.logProvider.warning(
+    `[core] Upgrade success! Old localSettings.json, manifest.local.template.json and manifest.remote.template.json have been backed up to the .backup folder and you can delete it.`
+  );
+
+  if (inputs.platform === Platform.VSCode) {
+    sendTelemetryEvent(Component.core, TelemetryEvent.ProjectConsolidateGuide, {
+      [TelemetryProperty.Status]: ProjectMigratorGuideStatus.Reload,
```
qinezh, line 280: No longer need to send the `Reload` event?

packages/fx-core/src/core/middleware/consolidateLocalRemote.ts
```diff
@@ -0,0 +1,315 @@
...
+import { getManifestTemplatePath } from "../../plugins/resource/appstudio/manifestTemplate";
+import { getTemplatesFolder } from "../../folder";
+import { loadProjectSettings } from "./projectSettingsLoader";
+import { addPathToGitignore, needMigrateToArmAndMultiEnv } from "./projectMigrator";
+
+const upgradeButton = "Upgrade";
+let fromReloadFlag = false;
+const backupFolder = ".backup";
+
+export const ProjectConsolidateMW: Middleware = async (
```
qinezh, line 51: are we going to upgrade the project version in `projectSettings.json`?

packages/fx-core/src/core/middleware/consolidateLocalRemote.ts
xiaolang124: AppStudio plugin uses the same logic for adding manifest, it prevents manifest format change after packing.

packages/fx-core/src/core/middleware/consolidateLocalRemote.ts
xiaolang124: Fixed.

packages/fx-core/src/core/middleware/consolidateLocalRemote.ts
xiaolang124: Fixed.

packages/fx-core/src/core/middleware/consolidateLocalRemote.ts
xiaolang124: For now, the project version has not been changed when creating a new project or upgrade an old project.

packages/fx-core/src/core/middleware/consolidateLocalRemote.ts
qinezh: Got it, that makes sense.

packages/fx-core/src/core/middleware/consolidateLocalRemote.ts
qinezh: To be consistent with Go Prod migration (https://github.com/OfficeDev/TeamsFx/pull/2988) , we may need do it as well, and this can also help us simplify the logic to detect whether current project needs to be migrated, e.g. only project version (2.0.0<= v < 3.0.0) needs to be upgraded.

Any thoughts? @xiaolang124 @jayzhang @kimizhu

packages/fx-core/src/core/middleware/consolidateLocalRemote.ts
```diff
@@ -0,0 +1,312 @@
...
+  sendTelemetryEvent(Component.core, TelemetryEvent.ProjectConsolidateUpgrade);
+
+  postConsolidate(inputs.projectPath as string, ctx, inputs, backupFolder);
+  return true;
+}
+
+function checkMethod(ctx: CoreHookContext): boolean {
+  const methods: Set<string> = new Set(["getProjectConfig", "checkPermission"]);
+  if (ctx.method && methods.has(ctx.method) && userCancelFlag) return false;
+  userCancelFlag = ctx.method != undefined && methods.has(ctx.method);
```
qinezh, line 261: A little bit confused about the `userCancelFlag` variable, seems the value is not from user cancel operation but the fx-core methods checking?

xiaolang124: When opening new fx project, it will enter ```consolidateLocalRemote``` logic multi times so if user cancel we need to save it temporarily. When provision or local debug, we need to ask user whether to upgrade every time. Check method is used to separate different enter point.

qinezh: > ... if user cancel we need to save it temporarily.

Seems there's no logic to save the 'user cancel' info according to L66-L72?

xiaolang124: Save this not when the user cancels but when the question is first asked during VsCode opens ```teamsfx``` project. May be ```user cancel``` is not a good name..

packages/fx-core/src/core/middleware/consolidateLocalRemote.ts
xiaolang124: For incremental sprint, project settings version will not be upgraded. V3 API sprint will update project settings version.

packages/fx-core/src/core/middleware/consolidateLocalRemote.ts
```diff
@@ -0,0 +1,312 @@
...
+      const templateManifestFolder = path.join(templatesFolder, "plugins", "resource", "spfx");
+      const manifestFile = path.resolve(
+        templateManifestFolder,
+        "./solution/manifest_multi_env.json"
+      );
+      const manifestString = (await fs.readFile(manifestFile)).toString();
+      manifest = JSON.parse(manifestString);
+      await fs.ensureDir(appDir);
+      const manifestTemplatePath = await getManifestTemplatePath(inputs.projectPath as string);
+      await fs.writeFile(manifestTemplatePath, JSON.stringify(manifest, null, 4));
```
HuihuiWu-Microsoft, line 181: The SPFx manifest template need to be updated for configurationUrl and contentUrl

packages/fx-core/src/core/middleware/consolidateLocalRemote.ts
qinezh: I see, thanks for the confirmation.

packages/fx-core/src/core/middleware/consolidateLocalRemote.ts
xiaolang124: Change adding manifest logic for SPFX to copy remote manifest and update ```contentUrl```, ```configurationUrl```. Because post scaffold has done a lot of work for SPFX manifest, like change ```color.png``` name to ```${componentId}_color.png```. Just adding a new manifest cannot qualify the demand.

HuihuiWu-Microsoft (dismissed):
LGTM

packages/fx-core/resource/package.nls.json
```diff
@@ -40,6 +40,9 @@
   "core.migrationToArmAndMultiEnv.SuccessMessage": "Upgrade succeeded. Your project is upgraded to new configuration files.",
   "core.migrationToArmAndMultiEnv.ErrorMessage": "Upgrade Failed. Please check the error message in Output window to troubleshoot and fix. You can also click Learn More for the FAQ and how to upgrade manually.",
   "core.migrationToArmAndMultiEnv.PreCheckErrorMessage": "Upgrade Failed. '%s' doesn't exist or is not in json format. Please fix it and try again by running command (Teams: Upgrade project).\nYou can also click Learn More for the FAQ.",
+  "core.consolidateLocalRemote.Message": "Teams Toolkit will upgrade your project configuration files to support the latest features. The upgrade process will not change your custom code and create the backup files in case revert is needed.\nNotice this upgrade is a must-have to continue to use current version Teams Toolkit. If you are not ready to upgrade and want to continue to use the old version, please find Teams Toolkit in Extension and install the version <= ????",
```
qinezh, line 43: Manifest template is migrated as well, let's be more accurate. We may update it to:
```suggestion
  "core.consolidateLocalRemote.Message": "Teams Toolkit will upgrade your project's local config and manifest template to support the latest features. The upgrade process will not change your custom code and create the backup files in case revert is needed.\nNotice this upgrade is a must-have to continue to use current version Teams Toolkit. If you are not ready to upgrade and want to continue to use the old version, please find Teams Toolkit in Extension and install the version <= ????",
```

packages/fx-core/resource/package.nls.json
```diff
@@ -232,6 +235,8 @@
   "error.ProjectEnvNotExistError": "Environment %s not found. Make sure the config.%s.json file exist.",
   "error.InvalidEnvNameError": "Environment name can only contain letters, digits, _ and -.",
   "error.UpgradeCanceledError": "If you don't want to upgrade your project, please install another version of Teams Toolkit (version <= 2.10.0).",
+  "error.ConsolidateCanceledError": "If you don't want to upgrade your project, please install another version of Teams Toolkit (version <= ????).",
+  "error.OperationNotSupportedForExistingAppError": "Operation is not supported for existing app: %s.",
```
qinezh, line 239: Looks the error message is not used and can be removed?

packages/fx-core/src/core/middleware/consolidateLocalRemote.ts
```diff
@@ -0,0 +1,335 @@
...
+  return false;
+}
+
+function outputCancelMessage(ctx: CoreHookContext) {
+  TOOLS?.logProvider.warning(`[core] Upgrade cancelled.`);
+
+  const inputs = ctx.arguments[ctx.arguments.length - 1] as Inputs;
+  if (inputs.platform === Platform.VSCode) {
+    TOOLS?.logProvider.warning(
+      `[core] Notice upgrade to new configuration files is a must-have to continue to use current version Teams Toolkit. If you are not ready to upgrade and want to continue to use the old version Teams Toolkit, please find Teams Toolkit in Extension and install the version <= ????`
```
qinezh, line 132: do we need to move these warning message to `package.nls.json`?

packages/fx-core/src/core/middleware/consolidateLocalRemote.ts
```diff
@@ -0,0 +1,335 @@
...
+      "templates",
+      "appPackage",
+      "manifest.remote.template.json"
+    );
+    const remoteManifestExist = await fs.pathExists(remoteManifestFile);
+    if (remoteManifestExist) {
+      if (isSPFxProject(projectSettings)) {
+        sendTelemetryEvent(Component.core, TelemetryEvent.ProjectConsolidateAddSPFXManifestStart);
+        const manifestTemplatePath = await getManifestTemplatePath(inputs.projectPath as string);
+        await fs.copyFile(remoteManifestFile, manifestTemplatePath);
```
qinezh, line 187: Looks no need extra step to copy from `remoteManifestFile` to `manifestTemplatePath` and you could load JSON from `remoteManifestFile` directly.

packages/fx-core/resource/package.nls.json
xiaolang124: My bad, add import to ```ConsolidateCanceledError```.

packages/fx-core/src/core/middleware/consolidateLocalRemote.ts
xiaolang124: For output channel, we do not need to move it to ```package.nls.json```.

packages/fx-core/src/core/middleware/consolidateLocalRemote.ts
xiaolang124: Fixed.

qinezh (approved):

jayzhang (approved):

PR #4261, opened 2022-03-16
fix: need to show error

nliu-ms:
Need to showError(), rather than only returning it.
image


nliu-ms:
E2E TEST: no test for extension

HuihuiWu-Microsoft (approved):

PR #4262, opened 2022-03-16
perf(aad-manifest): add aad app manifest manager

SLdragon:


packages/fx-core/src/plugins/resource/aad/aadAppManifestManager.ts
```diff
@@ -0,0 +1,88 @@
...
+    }
+  }
+
+  export async function updateAadApp(
+    graphToken: string,
+    manifest: AADManifest
+  ): Promise<AADManifest> {
+    const instance = initAxiosInstance(graphToken);
+    const aadApp = AadManifestHelper.manifestToApplication(manifest);
+    delete aadApp.id;
```
KennethBWSong, line 45: Shall we check whether all the keys exists?

KennethBWSong: Close according to offline discussion.
Will add validation in the future.

KennethBWSong (approved):

PR #4273, opened 2022-03-16
feat(m365): add create m365 app command

kuojianlu:
- add a new command to create a m365 app, this command will be only enabled if m365 feature flag is set
- add create new m365 project handler, set isM365 in inputs to let FxCore know this is to create a m365 project
- add TeamsAppInstallation.ReadForUser to graph scopes

create-m365-app

### DIFF
--- a/packages/vscode-extension/src/handlers.ts
+++ b/packages/vscode-extension/src/handlers.ts
@@ -371,7 +371,10 @@ export async function createNewProjectHandler(args?: any[]): Promise<Result<any,
 }
 
 export async function createNewM365ProjectHandler(args?: any[]): Promise<Result<any, FxError>> {
-  ExtTelemetry.sendTelemetryEvent(TelemetryEvent.CreateProjectStart, getTriggerFromProperty(args));
+  ExtTelemetry.sendTelemetryEvent(TelemetryEvent.CreateProjectStart, {
+    ...getTriggerFromProperty(args),
+    [TelemetryProperty.IsM365]: "true",
+  });
   const inputs = getSystemInputs();
   inputs.isM365 = true;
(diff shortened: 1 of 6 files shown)


kuojianlu:
E2E TEST: N/A

swatDong (dismissed):

packages/vscode-extension/src/handlers.ts
```diff
@@ -370,6 +370,17 @@ export async function createNewProjectHandler(args?: any[]): Promise<Result<any,
   return result;
 }
 
+export async function createNewM365ProjectHandler(args?: any[]): Promise<Result<any, FxError>> {
+  ExtTelemetry.sendTelemetryEvent(TelemetryEvent.CreateProjectStart, getTriggerFromProperty(args));
```
HuihuiWu-Microsoft, line 374: Use CreateM365ProjectStart event to distinguish the operations?

kuojianlu: Creating M365 app will have two entry points:
1. This new command.
2. The original creating command (just be informed from PM today. I will be added later).
I am not sure how to distinguish creating M365 apps from creating the original Teams app. What is your idea?

HuihuiWu-Microsoft: If user trigger from the new command, we can easily distinguish since different handlers are triggered. If user trigger from original creating command, I think we can just distinguish from the question model telemetry.

kuojianlu: Added is-m365 property to creating project event when creating M365 app. Thanks~

HuihuiWu-Microsoft (approved):

swatDong (approved):

PR #4282, opened 2022-03-17
chore(cli): support local preview for minimal app

qinezh:
Support local preview for minimal app by CLI:
* register Teams app with `core.localDebug`.
* no need to check dependency as there's no app code.
* launch the browser.

### DIFF
--- a/packages/cli/src/cmds/preview/preview.ts
+++ b/packages/cli/src/cmds/preview/preview.ts
@@ -10,12 +10,15 @@ import {
   assembleError,
   AzureSolutionSettings,
   Colors,
+  ConfigFolderName,
   err,
   FxError,
+  InputConfigsFolderName,
   Inputs,
   LogLevel,
   ok,
   Platform,
+  ProjectSettingsFileName,
   Result,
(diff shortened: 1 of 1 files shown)


packages/cli/src/cmds/preview/preview.ts
```diff
@@ -557,6 +573,53 @@ export default class Preview extends YargsCommand {
...
+    const coreResult = await activate();
+    if (coreResult.isErr()) {
+      return err(coreResult.error);
+    }
+    const core = coreResult.value;
+
+    const inputs: Inputs = {
+      projectPath: workspaceFolder,
+      platform: Platform.CLI,
+      env: environmentManager.getLocalEnvName(),
```
kuojianlu, line 590: No need to check isConfigUnifyEnabled?

qinezh: No need, since minimal app is only supported (init/preview...) when `isConfigUnifyEnabled` is enabled.

kuojianlu (approved):

PR #4291, opened 2022-03-17
perf(solution): add 'add sso' in execute user task

KennethBWSong:


packages/fx-core/src/plugins/solution/fx-solution/v2/executeUserTask.ts
```diff
@@ -704,3 +708,133 @@ export type ParamForRegisterTeamsAppAndAad = {
...
+      name: TeamsAppSolutionNameV2,
+      version: "1.0.0",
+      hostType: "Azure",
+      capabilities: [],
+      azureResources: [],
+      activeResourcePlugins: [],
+    };
+    ctx.projectSetting.solutionSettings = solutionSettings;
+    //aad need this file
+    await fs.writeJSON(`${inputs.projectPath}/permissions.json`, DEFAULT_PERMISSION_REQUEST, {
```
SLdragon, line 789: This file should not need when there is a aad.template.json file

packages/fx-core/src/plugins/solution/fx-solution/v2/executeUserTask.ts
```diff
@@ -704,3 +708,133 @@ export type ParamForRegisterTeamsAppAndAad = {
...
+      spaces: 4,
+    });
+  }
+
+  // Check whether can add sso
+  const canProceed = canAddSso(solutionSettings, ctx.telemetryReporter);
```
SLdragon, line 795: Do we need to put this check at the beginning?

packages/fx-core/src/plugins/solution/fx-solution/v2/executeUserTask.ts
KennethBWSong: Good catch. Will update this, thx！

jayzhang (approved):

SLdragon (dismissed):

packages/fx-core/src/plugins/solution/fx-solution/v2/executeUserTask.ts
1yefuwang1: It seems having `canAddSso()` here can cause dangling AddSsoStart  event without a subsequent AddSso success/failure event due to early return.

1yefuwang1 (approved):

PR #4309, opened 2022-03-18
perf(aad-manifest): add validate and preprocess manifest function

SLdragon:


packages/fx-core/src/plugins/resource/aad/utils/aadManifestHelper.ts
```diff
@@ -152,4 +156,104 @@ export class AadManifestHelper {
...
+      warningMsg += AadManifestErrorMessage.RequiredResourceAccessIsMissing;
+    }
+
+    // if manifest doesn't contain oauth2Permissions or oauth2Permissions length is 0
+    if (!manifest.oauth2Permissions || manifest.oauth2Permissions.length === 0) {
+      warningMsg += AadManifestErrorMessage.Oauth2PermissionsIsMissing;
+    }
+
+    // if manifest doesn't contain preAuthorizedApplications
+    if (!manifest.preAuthorizedApplications) {
```
KennethBWSong, line 184: pre auth apps should also contains outlook, etc.?
image

packages/fx-core/src/plugins/resource/aad/utils/aadManifestHelper.ts
```diff
@@ -152,4 +156,104 @@ export class AadManifestHelper {
...
+    return warningMsg;
+  }
+
+  public static processRequiredResourceAccessInManifest(manifest: AADManifest): void {
+    const map = getPermissionMap();
+    manifest.requiredResourceAccess.forEach((requiredResourceAccessItem) => {
+      const resourceIdOrName = requiredResourceAccessItem.resourceAppId;
+      let resourceId = resourceIdOrName;
+      if (!isUUID(resourceIdOrName)) {
+        resourceId = map[resourceIdOrName].id;
```
KennethBWSong, line 223: Will `map[resourceIdOrName]` be undefined?

SLdragon: Update and add a test case

KennethBWSong (approved):

PR #4314, opened 2022-03-18
chore(m365): add launch remote configurations for m365 

kuojianlu:
- add launch remote configurations for m365
- update m365 launch url
- let tenant in account-hint only applies for teams

image

### DIFF
--- a/packages/fx-core/src/plugins/solution/fx-solution/debug/constants.ts
+++ b/packages/fx-core/src/plugins/solution/fx-solution/debug/constants.ts
@@ -12,3 +12,22 @@ export class HubName {
   public static readonly outlook: string = "Outlook";
   public static readonly office: string = "Office";
 }
+
+export class LaunchUrl {
+  public static readonly teamsLocal: string =
+    "https://teams.microsoft.com/l/app/${localTeamsAppId}?installAppPackage=true&webjoin=true&${account-hint}";
+  public static readonly teamsRemote: string =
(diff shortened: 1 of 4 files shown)


packages/fx-core/src/plugins/solution/fx-solution/debug/constants.ts
```diff
@@ -12,3 +12,22 @@ export class HubName {
...
+  public static readonly outlookLocalTab: string =
+    "https://outlook.office.com/host/${localTeamsAppInternalId}?${account-hint}";
+  public static readonly outlookRemoteTab: string =
+    "https://outlook.office.com/host/${teamsAppInternalId}?${account-hint}";
+  public static readonly outlookLocalBot: string =
+    "https://outlook.office.com/mail?${account-hint}";
+  public static readonly outlookRemoteBot: string =
+    "https://outlook.office.com/mail?${account-hint}";
+  public static readonly officeLocalTab: string =
+    "https://www.office.com/m365apps/${localTeamsAppInternalId}?auth=2&${account-hint}";
```
swatDong, line 30: What does `auth=2` mean?

swatDong (approved):

packages/fx-core/src/plugins/solution/fx-solution/debug/constants.ts
kuojianlu: Actually I do not know either. `auth=2` will be automatically appended to the launch url. This url is confirmed by Office PM. I think we can just use it currently.

PR #4318, opened 2022-03-20
fix: invalid state file when create spfx project

HuihuiWu-Microsoft:
ADO:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/13862621
E2E TEST: not 100% reproducible

### DIFF
--- a/packages/fx-core/src/common/globalState.ts
+++ b/packages/fx-core/src/common/globalState.ts
@@ -6,7 +6,7 @@ import * as path from "path";
 import * as fs from "fs-extra";
 import crypto from "crypto";
 import { ConfigFolderName, ProductName } from "@microsoft/teamsfx-api";
-import * as properLock from "proper-lockfile";
+import properLock from "proper-lockfile";
 import { waitSeconds } from "./tools";
 
 const GlobalStateFileName = "state.json";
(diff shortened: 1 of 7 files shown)


packages/fx-core/src/common/globalState.ts
```diff
@@ -14,16 +17,37 @@ const GlobalStateFileName = "state.json";
...
 
-  const config = fs.readJSONSync(filePath);
-  let value = config[key];
-  if (value === undefined) {
-    value = defaultValue;
+  const lockFileDir = getLockFolder(filePath);
+  const lockfilePath = path.join(lockFileDir, `${ConfigFolderName}.lock`);
+  await fs.ensureDir(lockFileDir);
+
+  const retryNum = 10;
```
tecton, line 28: Could consider leveraging `retries` option when calling `lock()` later.

tecton (approved):

packages/fx-core/src/common/globalState.ts
HuihuiWu-Microsoft: Will refine in dev branch later,  thanks!

swatDong (approved):

PR #4331, opened 2022-03-22
feat(m365): add teamsfx new m365 command

kuojianlu:
- add teamsfx new m365 command if m365 feature flag enabled

image
image
teamsfx-new-m365


kuojianlu:
E2E TEST: N/A

Alive-Fish (approved):

PR #4333, opened 2022-03-22
chore(vsc): refresh env tree when project settings updated

qinezh:
refresh env tree and update the `Preview` button to `Local Debug` button when project settings is updated (for example, a component is added to a mini app).

### DIFF
--- a/packages/vscode-extension/src/handlers.ts
+++ b/packages/vscode-extension/src/handlers.ts
@@ -50,6 +50,7 @@ import {
   Void,
   VsCodeEnv,
   IProgressHandler,
+  ProjectSettingsFileName,
 } from "@microsoft/teamsfx-api";
 import {
   CollaborationState,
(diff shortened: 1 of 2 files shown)


packages/vscode-extension/src/handlers.ts
```diff
@@ -244,6 +245,10 @@ export async function activate(): Promise<Result<Void, FxError>> {
 
         await refreshEnvTreeOnFileChanged(workspacePath, files);
       });
+
+      workspace.onDidSaveTextDocument(async (event) => {
+        await refreshEnvTreeOnFileContentChanged(workspacePath, event.uri.fsPath);
```
HuihuiWu-Microsoft, line 250: Is this manually triggered when user save in vscode?

HuihuiWu-Microsoft (approved):
LGTM

packages/vscode-extension/src/handlers.ts
qinezh: It will be triggered automatically when the file is saved, no matter by Teams toolkit or User.

PR #4347, opened 2022-03-23
fix: add back projectId property for create-project event

HuihuiWu-Microsoft:
ADO: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/13900320
E2E TEST: will add UT later


HuihuiWu-Microsoft:
Raised another one to commit to hofix branch

PR #4348, opened 2022-03-23
chore(m365): prompt window to install app for remote

kuojianlu:
- prompt window to let user install app in Teams for launching remote in
  Outlook/Office
- add retrying logic for calling graph api to get internal id
- reload cache in graph getStatus

### DIFF
--- a/packages/vscode-extension/src/commonlib/graphLogin.ts
+++ b/packages/vscode-extension/src/commonlib/graphLogin.ts
@@ -159,6 +159,7 @@ export class GraphLogin extends login implements GraphTokenProvider {
   }
 
   async getStatus(): Promise<LoginStatus> {
+    await GraphLogin.codeFlowInstance.reloadCache();
     if (GraphLogin.codeFlowInstance.account) {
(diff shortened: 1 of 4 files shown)


packages/vscode-extension/src/commonlib/graphLogin.ts
```diff
@@ -159,6 +159,7 @@ export class GraphLogin extends login implements GraphTokenProvider {
   }
 
   async getStatus(): Promise<LoginStatus> {
+    await GraphLogin.codeFlowInstance.reloadCache();
```
swatDong, line 162: Reminder - CLI may have the same issue...

swatDong (approved):

packages/vscode-extension/src/commonlib/graphLogin.ts
kuojianlu: Good point. Let me update it in later PR.

HuihuiWu-Microsoft (approved):

PR #4352, opened 2022-03-23
chore(cicd): throw error while adding cicd to minimal app

qinezh:
For minimal app, there's no need to add CI/CD workflows for minimal app as it doesn't contain app source code.


IvanJobs (approved):
lgtm

jayzhang (approved):

PR #4362, opened 2022-03-23
perf(aad-manifest): support scaffold aad manifest

SLdragon:
Add scaffold functon in aad plugin:
1. Add aad.template.json file to templates/appPackage/aad.template.json

### DIFF
--- a/packages/fx-core/src/plugins/resource/aad/plugin.ts
+++ b/packages/fx-core/src/plugins/resource/aad/plugin.ts
@@ -1,7 +1,13 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
-import { FxError, LogProvider, PluginContext, Result } from "@microsoft/teamsfx-api";
+import {
+  AzureSolutionSettings,
+  FxError,
+  LogProvider,
+  PluginContext,
+  Result,
+} from "@microsoft/teamsfx-api";
 import { AadResult, ResultFactory } from "./results";
 import {
   CheckGrantPermissionConfig,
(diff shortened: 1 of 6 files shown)


packages/fx-core/src/plugins/resource/aad/plugin.ts
```diff
@@ -587,4 +598,20 @@ export class AadAppForTeamsImpl {
 
     return requiredResourceAccessList;
   }
+
+  public async scaffold(ctx: PluginContext): Promise<AadResult> {
+    if (isAadManifestEnabled() && !isSPFxProject(ctx.projectSettings)) {
```
KennethBWSong, line 603: I remember that only scaffold() of plugins that activated will be called. AAD plugin should not activate for spfx projects. Maybe we can remove the check? You may confirm with @jayzhang

packages/fx-core/src/plugins/resource/aad/plugin.ts
```diff
@@ -587,4 +598,20 @@ export class AadAppForTeamsImpl {
...
+      const templatesFolder = getTemplatesFolder();
+      const appDir = `${ctx.root}/${Constants.appPackageFolder}`;
+      const solutionSettings: AzureSolutionSettings = ctx.projectSettings
+        ?.solutionSettings as AzureSolutionSettings;
+      if (isAADEnabled(solutionSettings)) {
```
KennethBWSong, line 608: same for here

packages/fx-core/src/plugins/resource/aad/plugin.ts
```diff
@@ -587,4 +598,20 @@ export class AadAppForTeamsImpl {
...
+        const aadManifestTemplate = `${templatesFolder}/${Constants.aadManifestTemplateFolder}/${Constants.aadManifestTemplateName}`;
+        await fs.ensureDir(appDir);
+        await fs.copy(aadManifestTemplate, `${appDir}/${Constants.aadManifestTemplateName}`);
+        solutionSettings.capabilities.push(Constants.aadCapabilityName);
```
KennethBWSong, line 612: capability should be added by solution?

SLdragon: Updated

KennethBWSong (approved):

KennethBWSong (approved):

PR #4379, opened 2022-03-24
feat(frontend): support customize install command

hund030:
E2E TEST: https://github.com/OfficeDev/TeamsFx/actions/runs/2032521640

### DIFF
--- a/packages/fx-core/src/plugins/resource/frontend/ops/deploy.ts
+++ b/packages/fx-core/src/plugins/resource/frontend/ops/deploy.ts
@@ -65,6 +65,11 @@ export class FrontendDeployment {
         async () => (await fs.readJSON(path.join(componentPath, PathInfo.NodePackageFile))).scripts
       )) ?? [];
 
+    if (!("install:teamsfx" in scripts)) {
+      // * Track legacy projects
+      TelemetryHelper.sendGeneralEvent(TelemetryEvent.InstallScriptNotFound);
+    }
+
     await progressHandler?.next(DeployProgress.steps.NPMInstall);
     await runWithErrorCatchAndThrow(new NpmInstallError(), async () => {
       await Utils.execute(
@@ -109,6 +114,11 @@ export class FrontendDeployment {
         (error) => new FileIOError(error.message),
         async () => (await fs.readJSON(path.join(componentPath, PathInfo.NodePackageFile))).scripts
       )) ?? [];
+
+    if (!("install:teamsfx" in scripts)) {
+      // * Track legacy projects
+      TelemetryHelper.sendGeneralEvent(TelemetryEvent.InstallScriptNotFound);
+    }
 
     await progress?.next(DeployProgress.steps.NPMInstall);
     await runWithErrorCatchAndThrow(new v3error.NpmInstallError(), async () => {
--- a/packages/fx-core/src/plugins/resource/frontend/ops/deploy.ts
+++ b/packages/fx-core/src/plugins/resource/frontend/ops/deploy.ts
@@ -60,7 +60,6 @@ export class FrontendDeployment {
 
     const progressHandler = ProgressHelper.progressHandler;
 
-    await progressHandler?.next(DeployProgress.steps.NPMInstall);
     const scripts = await runWithErrorCatchAndWrap(
       (error) => new FileIOError(error.message),
       async () => {
@@ -68,6 +67,8 @@ export class FrontendDeployment {
         return JSON.parse(pack).scripts;
       }
     );
+
+    await progressHandler?.next(DeployProgress.steps.NPMInstall);
     await runWithErrorCatchAndThrow(new NpmInstallError(), async () => {
       await Utils.execute(
         "install:teamsfx" in scripts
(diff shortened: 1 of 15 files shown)


packages/fx-core/src/plugins/resource/frontend/v3/index.ts
```diff
@@ -113,7 +110,7 @@ export class NodeJSTabFrontendPlugin implements v3.PluginV3 {
         }
       },
     });
-    await progress.end(true);
+    await ProgressHelper.endProgress(true);
```
IvanJobs, line 113: It looks the PR is not only related to customize install but involve refactor of the current code, right?

packages/fx-core/src/plugins/resource/frontend/ops/deploy.ts
```diff
@@ -51,60 +52,104 @@ export class FrontendDeployment {
...
 
+    const scripts =
+      (await runWithErrorCatchAndWrap(
+        (error) => new FileIOError(error.message),
+        async () => (await fs.readJSON(path.join(componentPath, PathInfo.NodePackageFile))).scripts
+      )) ?? [];
+
+    if (!("install:teamsfx" in scripts)) {
+      // * Track legacy projects
+      TelemetryHelper.sendGeneralEvent(TelemetryEvent.InstallScriptNotFound);
```
IvanJobs, line 70: How about adding one extra field into common properties instead? I encountered a same comment developing CICD plugin recently as you may remember.

packages/fx-core/src/plugins/resource/frontend/ops/deploy.ts
```diff
@@ -51,60 +52,104 @@ export class FrontendDeployment {
...
     if (!(await FrontendDeployment.needBuild(componentPath, envName))) {
       return FrontendDeployment.skipBuild();
     }
 
     const progressHandler = ProgressHelper.progressHandler;
 
+    const scripts =
+      (await runWithErrorCatchAndWrap(
+        (error) => new FileIOError(error.message),
+        async () => (await fs.readJSON(path.join(componentPath, PathInfo.NodePackageFile))).scripts
```
IvanJobs, line 65: Nit: Better to wrap the exception of file's nonexistence here to provide user-friendly error message.

packages/fx-core/src/plugins/resource/frontend/ops/deploy.ts
```diff
@@ -51,60 +52,104 @@ export class FrontendDeployment {
...
+    if (!("install:teamsfx" in scripts)) {
+      // * Track legacy projects
+      TelemetryHelper.sendGeneralEvent(TelemetryEvent.InstallScriptNotFound);
+    }
+
     await progressHandler?.next(DeployProgress.steps.NPMInstall);
     await runWithErrorCatchAndThrow(new NpmInstallError(), async () => {
-      await Utils.execute(Commands.InstallNodePackages, componentPath);
+      await Utils.execute(
+        "install:teamsfx" in scripts
```
IvanJobs, line 76: How about the following?
`
    execute(scripts["install:teamsfx"] ?? DefaultScript)
`

packages/fx-core/src/plugins/resource/frontend/ops/deploy.ts
eriolchan: Which is better for this case, send separate event, or put it as property of existing event (maybe deploy event)?

packages/fx-core/src/plugins/resource/frontend/v3/index.ts
hund030: Yes. I did this refactor because I want to unify plugin and v3 plugin, rather than maintain these duplicated code. But soon I found they are based on different error handling, so I stop at refactoring the progress helper.

packages/fx-core/src/plugins/resource/frontend/ops/deploy.ts
hund030: A little different. 
```
await Utils.execute(
        "install:teamsfx" in scripts
          ? Commands.InstallNodePackages
          : Commands.DefaultInstallNodePackages,
        componentPath
      );
```
This output logs: `Execute command: npm run install:teamsfx` so that user knows how to customize install script by reading logs.
```
execute(scripts["install:teamsfx"] ?? DefaultScript)
```
This output: `Execute command: npm install` by default. User may not get how to customize it.

packages/fx-core/src/plugins/resource/frontend/ops/deploy.ts
hund030: This property may be not fit to deploy event. Offline discussed with @eriolchan, let's keep this event.

packages/fx-core/src/plugins/resource/frontend/ops/deploy.ts
hund030: Do you mean `FileIOError` -> `FileNotFoundError`?  `fs.readJSON` may fail because of invalid json file format, so I think non existence may not be accurate.

eriolchan (approved):

PR #4392, opened 2022-03-25
chore(env-tree): refresh env tree after capabilities added

qinezh:
Refresh env tree automatically after new capabilities added.

### DIFF
--- a/packages/vscode-extension/src/handlers.ts
+++ b/packages/vscode-extension/src/handlers.ts
@@ -136,6 +136,7 @@ import * as uuid from "uuid";
 import { automaticNpmInstallHandler } from "./debug/npmInstallHandler";
 import { showInstallAppInTeamsMessage } from "./debug/teamsAppInstallation";
 import { localize } from "./utils/localizeUtils";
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/handlers.ts
```diff
@@ -584,10 +585,18 @@ export async function addCapabilityHandler(args: any[]): Promise<Result<null, Fx
...
+    return err(result.error);
   }
+
+  // refresh env tree after adding capability.
+  const refreshResult = await registerEnvTreeHandler();
+  if (refreshResult.isErr()) {
+    return err(refreshResult.error);
+  }
+
+  await globalStateUpdate("automaticNpmInstall", true);
```
HuihuiWu-Microsoft, line 598: If addCapability succeed but refresh tree view failed, the global state will be affected. What about move it above envTree refresh?

qinezh: Updated~

HuihuiWu-Microsoft (approved):
LGTM

PR #4402, opened 2022-03-25
chore(vsc): improve UX to prevent user performing concurrent operations

tecton:
1. Show running icon and change the title of sidebar command.
2. Show tooltip for other sidebar commands when running another blocking one. (VSC doesn't support disabling other buttons now)
3. Show warning message if user clicks other sidebar command or trigger from Command Palette.
4. Skip the invocation of fx-core API in VSC extension if there's already a running command.

- [x] localized string

image

Related work item: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/13827711


packages/vscode-extension/src/treeview/treeViewManager.ts
```diff
@@ -106,282 +241,200 @@ class TreeViewManager {
...
   }
 
-  public dispose() {
-    this.treeviewMap.forEach((value) => {
-      value.dispose();
-    });
+  private async registerTreeViewsForNonTeamsFxProject() {
+    const disposables: vscode.Disposable[] = [];
+
+    this.registerAccount(disposables);
```
HuihuiWu-Microsoft, line 430: Remove this if not needed?

tecton: Thanks, removed.

HuihuiWu-Microsoft (approved):

PR #4405, opened 2022-03-25
perf(solution): add non sso tab in 'add capability'

KennethBWSong:


packages/fx-core/src/plugins/solution/fx-solution/v2/executeUserTask.ts
```diff
@@ -297,9 +299,39 @@ export async function addCapability(
...
+        SolutionError.InvalidInput,
+        getLocalizedString("core.addSsoFiles.canNotAddTabWhenSsoNotEnabled"),
+        SolutionSource
+      );
+      return err(e);
+    }
+
+    if (toAddTabNonSso) {
+      const index = capabilitiesAnswer.indexOf(TabNonSsoItem.id);
+      capabilitiesAnswer.splice(index);
```
1yefuwang1, line 330: `Array.splice()` here will remove all items within range [index, capabilitiesAnswer.length).

1yefuwang1: Besides, I'm a little bit confused about why you want to remove `TabNonSsoItem.id` and manually add `TabOptionItem.id` here when `toAddTabNonSso` is true.

packages/fx-core/src/plugins/solution/fx-solution/v2/executeUserTask.ts
```diff
@@ -398,11 +482,37 @@ export async function addCapability(
       pluginNamesToScaffold.add(ResourcePluginsV2.FrontendPlugin);
       if (!alreadyHasBot && !alreadyHasME) {
         pluginNamesToArm.add(ResourcePluginsV2.BotPlugin);
+
+        if (isAadManifestEnabled() && alreadyHasSso) {
+          const createAuthFilesRes = await createAuthFiles(inputsNew, false, true, true);
+          if (createAuthFilesRes.isErr()) {
```
SLdragon, line 488: Seems return error logic are all same, can we extract to a function to make it simple

packages/fx-core/src/plugins/solution/fx-solution/v2/executeUserTask.ts
```diff
@@ -876,3 +1004,58 @@ export async function addSso(
 
   return ok(undefined);
 }
+
+export async function createAuthFiles(
+  input: Inputs,
+  needTab: boolean,
+  needBot: boolean,
+  isVsProject = false
```
SLdragon, line 1012: isVsProject is not used, we can add a todo for it

packages/fx-core/src/plugins/solution/fx-solution/v2/executeUserTask.ts
KennethBWSong: Both tab and non sso tab activates fe plugin, and fe plugin will handle the logic whether to scaffold sso or non sso template. What we need to do is to add a tab. By remove the non sso tab and add the tab option, we can keep the logic below same as current :
image

packages/fx-core/src/plugins/solution/fx-solution/v2/executeUserTask.ts
KennethBWSong: Sure, will do

packages/fx-core/src/plugins/solution/fx-solution/v2/executeUserTask.ts
1yefuwang1: OK, that's a little bit tricky.

SLdragon (approved):

1yefuwang1 (approved):

PR #4414, opened 2022-03-28
perf(aad-manifest): provision using aad manifest

SLdragon:


packages/fx-core/src/plugins/resource/aad/errors.ts
```diff
@@ -151,6 +157,17 @@ export const ListCollaboratorError: AadError = {
...
 
+export const AadManifestNotFoundError: AadError = {
+  name: "AadManifestNotFoundError",
+  message: () => "Aad manifest file not found",
+};
+
+export const AadManifestLoadError: AadError = {
+  name: "AadManifestLoadError",
+  message: (manifestPath: string, reason: string) =>
+    util.format("Failed to load manifest file from %s, due to %s", manifestPath, reason),
```
KennethBWSong, line 168: use getLocalizatedString?

SLdragon: Thanks, will update

KennethBWSong (approved):

PR #4416, opened 2022-03-28
perf(solution): can only add sso for app service hosted bot

KennethBWSong:


packages/fx-core/src/plugins/solution/fx-solution/v2/executeUserTask.ts
```diff
@@ -852,6 +859,21 @@ export function canAddSso(
     );
   }
 
+  // Can only add sso for bot with host type == 'app-service'
```
a1exwang, line 862: Better to clarify this comment. `hostType === 'azure-functions'` is func hosted bot. `hostType === 'app-service'` or `undefined` is app service hosted bot. `undefined` is for legacy projects. Do you support legacy bot projects (for example bot project scaffolded before GA release)?

KennethBWSong: @a1exwang Will migration add host type for previous projects?

a1exwang: No. host type exists only for newly created projects.

SLdragon (dismissed):

packages/fx-core/src/plugins/solution/fx-solution/v2/executeUserTask.ts
KennethBWSong: OK, then I will just update the comment. Thx！

1yefuwang1 (approved):

PR #4417, opened 2022-03-28
feat: impl teamsfx add cicd

IvanJobs:
to finish: https://dev.azure.com/msazure/Microsoft%20Teams%20Extensibility/_workitems/edit/13882675

1. `teamsfx add cicd` interactively.
2. `teamsfx add cicd --env dev --provider github --template ci cd --interactive false` non-interactively.

E2E TEST: https://github.com/OfficeDev/TeamsFx/blob/838ca43f447155ddd124a187977ae50c27e2d3ca/packages/cli/tests/e2e/cicd/VerifyGeneratedTemplates.tests.ts#L19


MuyangAmigo (approved):
CLI Signature and Description Looks Good to Me.

wenytang-ms:
just rebase or merge dev branch, the audit problem will fixed.

Alive-Fish (dismissed):

Alive-Fish (approved):

packages/fx-core/src/plugins/resource/cicd/index.ts
```diff
@@ -69,7 +69,7 @@ export class CICDPluginV2 implements ResourcePlugin {
     tokenProvider: TokenProvider
   ): Promise<FxResult> {
     // add CI CD workflows for minimal app is not supported.
-    if (ctx.projectSetting && isPureExistingApp(ctx.projectSetting)) {
+    if (inputs.platform !== Platform.CLI_HELP && isPureExistingApp(ctx.projectSetting)) {
```
eriolchan, line 72: What does this if condition mean?

packages/cli/src/cmds/add.ts
```diff
@@ -0,0 +1,96 @@
...
+    CliTelemetry.withRootFolder(rootFolder).sendTelemetryEvent(TelemetryEvent.AddCICDStart);
+
+    const result = await activate(rootFolder);
+    if (result.isErr()) {
+      CliTelemetry.sendTelemetryErrorEvent(TelemetryEvent.AddCICD, result.error);
+      return err(result.error);
+    }
+    const core = result.value;
+    let inputs: Inputs;
+    {
```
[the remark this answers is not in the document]
IvanJobs: I just copied the pattern from existing code and the same pattern exists in `cmds/validate.ts` and other. 
For me, it seems unnecessary, let me try to remove it next time.

packages/fx-core/src/plugins/resource/cicd/index.ts
IvanJobs: The original code is added by @qinezh to give a nice error in minimal app scenario. 
I found a bug (when CLI calls this getQuestions for CLI `--help`, it should not throw this error) when I implementing the CLI, so fix it along the way.
BTW, ctx.projectSetting is not an undefined type, so remove the useless if (ctx.projectSetting).

PR #4423, opened 2022-03-28
feat(apiconnector): add bearer auth provider

dilin-MS:
Add `IAuthProvider`, `BearerAuthProvider` and `createApiClient()` implementation.

E2E TEST: https://github.com/OfficeDev/TeamsFx/blob/6307bb97cec947405551685e38dbcc2813e4f2ec/packages/sdk/test/e2e/node/apiClient.spec.ts#L8


packages/sdk/src/authentication/apiClient.ts
```diff
@@ -0,0 +1,32 @@
...
+ * @param apiEndpoint - Base url of the API
+ * @param authProvider - Auth provider that injects authentication info to each request
+ * @returns axios instance configured with specfic auth provider
+ *
+ * @example
+ * ```typescript
+ * const client = createApiClient("https://kudos.microsoft.com/api", new BasicAuthProvider("xxx","xxx"));
+ * ```
+ */
+export function createApiClient(apiEndpoint: string, authProvider: IAuthProvider): AxiosStatic {
```
blackchoey, line 19: Should we return `AxiosInstance` instead of `AxiosStatic`? Seems `AxiosStatic` is expected to be a global shared instance.

packages/sdk/src/authentication/apiClient.ts
```diff
@@ -0,0 +1,32 @@
...
+  // Add a request interceptor
+  axios.interceptors.request.use(
+    async function (config) {
+      config.baseURL = apiEndpoint;
+      await authProvider.AddAuthenticationInfo(config);
+      return config;
+    },
+    function (error) {
```
blackchoey, line 27: is it possible to not set the error handling function and use axios's default error handling logic? We didn't bring extra benifit to users on the error handling part.

packages/sdk/src/authentication/apiClient.ts
```diff
@@ -0,0 +1,32 @@
...
+      return Promise.reject(error);
+    }
+  );
+  return axios;
```
blackchoey, line 31: Are we returning a shared instance here? If yes please create a new axios instance since we need to support different apis.

packages/sdk/src/authentication/apiClient.ts
```diff
@@ -0,0 +1,32 @@
...
+import axios, { AxiosStatic } from "axios";
+import { IAuthProvider } from "./iAuthProvider";
+
+/**
+ * Initializes new Axios instance with specific auth provider
+ *
+ * @param apiEndpoint - Base url of the API
+ * @param authProvider - Auth provider that injects authentication info to each request
+ * @returns axios instance configured with specfic auth provider
+ *
```
blackchoey, line 13: please add @beta for this function

packages/sdk/src/authentication/bearerAuthProvider.ts
```diff
@@ -0,0 +1,31 @@
...
+
+import { AxiosRequestConfig } from "axios";
+import { IAuthProvider } from "./iAuthProvider";
+
+/**
+ * Provider that handles Bearer Token authentication
+ *
+ * @beta
+ */
+export class BearerAuthProvider implements IAuthProvider {
```
blackchoey, line 12: Can we rename it to `BearerTokenAuthProvider` since bearer token is a well known term

packages/sdk/src/authentication/iAuthProvider.ts
```diff
@@ -0,0 +1,15 @@
...
+
+// Defines method that injects authentication info to http requests
+export interface IAuthProvider {
+  /**
+   * Adds authentication info to http requests
+   *
+   * @param config - Contains all the request information and can be updated to include extra authentication info.
+   * Refer https://axios-http.com/docs/req_config for detailed document.
+   */
+  AddAuthenticationInfo: (config: AxiosRequestConfig) => Promise<void>;
```
blackchoey, line 14: can we follow axios's interface to return AxiosRequestConfig? There should be some reason that we don't know that drives axios return a config instance in interceptor.

packages/sdk/src/authentication/iAuthProvider.ts
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { AxiosRequestConfig } from "axios";
+
+// Defines method that injects authentication info to http requests
+export interface IAuthProvider {
```
blackchoey, line 7: I happened to find that ts does not recommend use `I` as interface prefix. Let's just use `AuthProvider` as interface name.
https://stackoverflow.com/questions/31876947/confused-about-the-interface-and-class-coding-guidelines-for-typescript

blackchoey (commented):
can we change the folder name of the source code to something like `apiClient`? authentication covers too much concept

packages/sdk/src/authentication/bearerAuthProvider.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  /**
+   * Adds authentication info to http requests
+   *
+   * @param config - Contains all the request information and can be updated to include extra authentication info.
+   * Refer https://axios-http.com/docs/req_config for detailed document.
+   */
+  public async AddAuthenticationInfo(config: AxiosRequestConfig): Promise<void> {
+    const token = await this.getToken();
+    config.headers = {
```
blackchoey, line 27: will it overwrite other request headers?

packages/sdk/src/authentication/bearerAuthProvider.ts
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Provider that handles Bearer Token authentication
+ *
+ * @beta
+ */
+export class BearerAuthProvider implements IAuthProvider {
+  private getToken: () => Promise<string>;
+
+  constructor(getToken: () => Promise<string>) {
```
tecton, line 15: There could be some comments or code snippets to show how to leverage credentials provided by `TeamsFx` since we are accepting a `getToken` function here.

packages/sdk/src/authentication/apiClient.ts
```diff
@@ -0,0 +1,32 @@
...
+ * @example
+ * ```typescript
+ * const client = createApiClient("https://kudos.microsoft.com/api", new BasicAuthProvider("xxx","xxx"));
```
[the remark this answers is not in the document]
dilin-MS: Fixed.

packages/sdk/src/authentication/apiClient.ts
dilin-MS: Fixed.

packages/sdk/src/authentication/bearerAuthProvider.ts
dilin-MS: Fixed.

packages/sdk/src/authentication/iAuthProvider.ts
dilin-MS: Fixed.

packages/sdk/src/authentication/bearerAuthProvider.ts
dilin-MS: Fixed.

packages/sdk/src/authentication/bearerAuthProvider.ts
dilin-MS: Fixed.

dilin-MS:
> can we change the folder name of the source code to something like `apiClient`? authentication covers too much concept

Fixed.

packages/sdk/src/apiClient/bearerTokenAuthProvider.ts
```diff
@@ -0,0 +1,40 @@
...
+   * Refer https://axios-http.com/docs/req_config for detailed document.
+   *
+   * @beta
+   */
+  public async AddAuthenticationInfo(config: AxiosRequestConfig): Promise<AxiosRequestConfig> {
+    const token = await this.getToken();
+    if (!config.headers) {
+      config.headers = {};
+    }
+    config.headers["Authorization"] = `Bearer ${token}`;
```
blackchoey, line 37: I just realized that we need to handle the case that Authorization header already exists. Let's chat offline on the expected behavior.

dilin-MS: Fixed.

packages/sdk/test/unit/apiClient/apiClient.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+import axios from "axios";
+import { createApiClient } from "../../../src";
+import MockAdapter from "axios-mock-adapter";
+
+describe("ApiClient Tests - Node", () => {
+  it("createApiClient should return axios instance with base url and config updated", async function () {
+    // Arrange
+    const apiBaseUrl = "https://fake-api-endpoint";
+    const mockAuthProvider = new MockAuthProvider();
+    const axiosMockAdapter = new MockAdapter(axios);
```
blackchoey, line 15: not sure whether this is a special case, but seems changes to axios also impacts the new axios instance we created. From feature perspective, we may expect the new instance is a brand new instance that do not have any customization that made by developers somewhere else.

We don't need to resolve this issue in this PR. Let's add a task to double confirm the behavior and see whether we need some changes.

packages/sdk/test/e2e/node/apiClient.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+    const bearerTokenAuthProvider = new BearerTokenAuthProvider(async () => "test-bearer-token");
+    const apiClient = createApiClient(apiBaseUrl, bearerTokenAuthProvider);
+
+    // Act
+    const res = await apiClient.get("/foo");
+
+    // Assert
+    assert.equal(res.config.baseURL, apiBaseUrl);
+    assert.equal(res.config.url, "/foo");
+    assert.equal(res.config.headers!["Authorization"], "Bearer test-bearer-token");
```
blackchoey, line 40: I saw the test server does not set response headers, why we can get the response headers here?

blackchoey: by the way, can we rename the file to bearerTokenAuthProvider.spec.ts? Since we mainly tests the auth provider, not createApiClient function here.

dilin-MS: I was referring to the `res.config.headers` which according to [axios API: response schema](https://axios-http.com/docs/res_schema) ,  `config` is the config that was provided to `axios` for the request.

dilin-MS: I was thinking having several `it()` function to test different auth provider e2e test in this same test file. How do you think?

dilin-MS: Fiexed by verifying authentication info with request header from server side.

blackchoey (approved):

packages/sdk/src/apiClient/apiClient.ts
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Initializes new Axios instance with specific auth provider
+ *
+ * @param apiEndpoint - Base url of the API
+ * @param authProvider - Auth provider that injects authentication info to each request
+ * @returns axios instance configured with specfic auth provider
+ *
+ * @example
+ * ```typescript
+ * const client = createApiClient("https://my-api-endpoint-base-url", new BasicAuthProvider("xxx","xxx"));
```
tecton, line 16: Maybe we can use `BearerTokenAuthProvider` here as example because we don't provide BaiscAuthProvider now.

tecton (approved):

wenytang-ms (approved):

packages/sdk/src/apiClient/apiClient.ts
dilin-MS: Will soon add the BasicAuthProvider. Thx

PR #4427, opened 2022-03-29
chore(vsc): add tree registration for non-teamsfx project

tecton:
Register tree views since environment tree depends on TreeViewProvider in non-teamsfx project and causes exception otherwise.


HuihuiWu-Microsoft (approved):
Please update to not call env related logic for non-teamsfx project later.

PR #4431, opened 2022-03-29
fix: condition overwitten in solution/bot plugin

a1exwang:
- fix a bug which was worked around in solution (should not skip getQuestionsForScaffolding for command bot)
- fix qm condition overwritten issue


a1exwang:
E2E TEST: n/a

jayzhang (approved):

PR #4437, opened 2022-03-29
chore: support customize local debug cert,bot endpoint

xiaolang124:
Add
```
    "frontend" : {
		"sslCertFile": "",
		"sslKeyFile": ""
    },
	"bot": {
		"siteEndpoint": ""
	}
```
to support customize local cert and bot endpoint.


packages/fx-core/src/common/local/localEnvManager.ts
```diff
@@ -29,6 +30,7 @@ import { ProjectSettingsHelper } from "./projectSettingsHelper";
 import { LocalCertificate, LocalCertificateManager } from "./localCertificateManager";
 import { DepsManager } from "../deps-checker/depsManager";
 import { LocalStateProvider } from "../localStateProvider";
+import { EnvInfoV2 } from "@microsoft/teamsfx-api/build/v2";
```
swatDong, line 33: `v2.EnvInfoV2` or `EnvInfoV2`? May save one import here.

swatDong: "@microsoft/teamsfx-api/build/v2" may not be a good style since the `build` output folder can be changed.
Could `EnvInfoV2` be imported directly from "@microsoft/teamsfx-api"? Otherwise `v2.EnvInfoV2` is also acceptable.

swatDong (approved):

packages/fx-core/src/common/local/localEnvManager.ts
xiaolang124: Thanks for the advice, changed to ```v2.EnvInfoV2```.

KennethBWSong (approved):

LongOddCode (approved):

PR #4442, opened 2022-03-30
Update "Azure Resources" string to "Azure resources".

pierceboggan:


HuihuiWu-Microsoft:
The Strings.json file has been deprecated and now we use the package.nls.json for to-be-localized strings. I'll remove the file later

PR #4445, opened 2022-03-30
chore: add telemetry for spfx

HuihuiWu-Microsoft:
1. Add isSPFx property to all lifecycle in extension
2. Improve 'create' telemetry event in solution so for create project we can query how many spfx projects created
3. Attach spfx dependency checker to events
4. Send error code in spfx plugin so we can query how many users get dependency installation failed error
5. Fix error due to wrong conflict resolve for main merge back to dev

### DIFF
--- a/packages/fx-core/src/plugins/solution/fx-solution/v2/scaffolding.ts
+++ b/packages/fx-core/src/plugins/solution/fx-solution/v2/scaffolding.ts
@@ -46,6 +46,9 @@ export async function scaffoldSourceCode(
   ctx: v2.Context,
   inputs: Inputs
 ): Promise<Result<Void, FxError>> {
+  ctx.telemetryReporter?.sendTelemetryEvent(SolutionTelemetryEvent.CreateStart, {
+    [SolutionTelemetryProperty.Component]: SolutionTelemetryComponentName,
+  });
   if (inputs.projectPath === undefined) {
     return err(
       new SystemError("Solution", SolutionError.InternelError, "projectPath is undefined")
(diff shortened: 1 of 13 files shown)


packages/fx-core/src/plugins/solution/fx-solution/v2/scaffolding.ts
```diff
@@ -46,6 +46,9 @@ export async function scaffoldSourceCode(
   ctx: v2.Context,
   inputs: Inputs
 ): Promise<Result<Void, FxError>> {
+  ctx.telemetryReporter?.sendTelemetryEvent(SolutionTelemetryEvent.CreateStart, {
```
HuihuiWu-Microsoft, line 49: There was no CreateStart event for solution before

packages/vscode-extension/src/utils/commonUtils.ts
```diff
@@ -115,8 +115,8 @@ export function getProjectId(): string | undefined {
   }
 }
 
-export async function isSPFxProject(workspacePath: string): Promise<boolean> {
-  if (await fs.pathExists(`${workspacePath}/SPFx`)) {
+export function isSPFxProject(workspacePath: string): boolean {
+  if (fs.pathExistsSync(`${workspacePath}/SPFx`)) {
```
tecton, line 119: Why is async version replaced with sync version?

xiaolang124 (approved):

packages/vscode-extension/src/utils/commonUtils.ts
HuihuiWu-Microsoft: Because sendTelemetryEvent API doesn't have a async version, so I can't use await in this function.

tecton: Ok, I think this value should be kept in memory instead of reading from disk every time when we send telemetry event.

tecton (approved):
VSC looks good and telemetry part could be improved later.

packages/vscode-extension/src/utils/commonUtils.ts
HuihuiWu-Microsoft: I've considered about this solution, but I'm afraid the value might change. For example, user first init an app, then add spfx and at this time, it'll change to spfx project.

jayzhang (approved):

PR #4448, opened 2022-03-30
perf: create bot related files when add sso

KennethBWSong:
image

TODO: test js project after bot template is ready


packages/fx-core/src/plugins/solution/fx-solution/v2/executeUserTask.ts
```diff
@@ -1038,18 +1052,64 @@ export async function createAuthFiles(
...
+        fs.writeFile(readmeTargetPath, readme);
+
+        // Sample Code
+        const sampleSourceFolder = path.join(botTemplateFolder, languageFolderName);
+        const sampleZip = new AdmZip();
+        sampleZip.addLocalFolder(sampleSourceFolder);
+        await unzip(sampleZip, botFolder);
+      }
+    } catch (error) {
+      const e = new SystemError(
```
SLdragon, line 1084: A suggestion: if we failed to add resources, we'd better to remove all the file we added

1yefuwang1 (approved):

PR #4461, opened 2022-03-31
chore: refine brand wording

IvanJobs:
to fix: https://dev.azure.com/msazure/Microsoft%20Teams%20Extensibility/_workitems/edit/13964774

image


packages/fx-core/src/plugins/resource/cicd/questions.ts
```diff
@@ -41,10 +41,22 @@ export const provisionOption: OptionItem = {
 
 export const publishOption: OptionItem = {
   id: "publish",
-  label: "Publish",
+  label: "Publish to Teams",
   detail: getLocalizedString("plugins.cicd.publishOption.detail"),
 };
 
+export function templateIdToLabel(templateId: string): string {
```
eriolchan, line 48: Since it behaves like a Map, is it possible to use Map in TypeScript for more readability and performance?

IvanJobs: Updated.

eriolchan: Good to see this change, BTW, can we define the Map as const outside of function so that this map won't be created again and again when called.

eriolchan (approved):

PR #4464, opened 2022-03-31
perf: backup logic on api connector

wenytang-ms:


packages/fx-core/src/plugins/resource/apiconnector/plugin.ts
```diff
@@ -35,14 +42,61 @@ export class ApiConnectorImpl {
...
+
+  private async scaffoldInComponent(
+    projectPath: string,
+    componentItem: string,
+    config: ApiConnectorConfiguration,
+    languageType: string
+  ): Promise<ApiConnectorResult> {
+    const componentPath = path.join(projectPath, componentItem);
+    const backupFolderName = generateTempFolder();
+    const sampleFileName = getSampleCodeFileName(config.APIName, languageType);
```
blackchoey, line 58: we do not need to backup the sample file, since we should throw an error if sample file already exists.

blackchoey:
If user selects multiple component, can we roll back all the component's file when error occurred? The current logic may make the project to a state that one component is updated to call existing API but another component is not updated.

packages/fx-core/src/plugins/resource/apiconnector/plugin.ts
wenytang-ms: sure

packages/fx-core/src/plugins/resource/apiconnector/plugin.ts
```diff
@@ -41,52 +36,55 @@ export class ApiConnectorImpl {
...
-  ): Promise<ApiConnectorResult> {
-    const componentPath = path.join(projectPath, componentItem);
+    // backup relative files.
     const backupFolderName = generateTempFolder();
-    const sampleFileName = getSampleCodeFileName(config.APIName, languageType);
-    await this.backupExistingFiles(componentPath, sampleFileName, backupFolderName);
+    const componentFolders: string[] = [];
+    config.ComponentPath.forEach((item) => componentFolders.push(path.join(projectPath, item)));
+    componentFolders.forEach(async (componentPath) => {
+      await this.backupExistingFiles(componentPath, backupFolderName);
```
blackchoey, line 44: some functions take `componentPath` as input and some functions take `projectPath` and `componentItem` as input. Since `componentPath`=`projectPath`+`componentItem`, can we make some small refractors to align the input parameters?

packages/fx-core/src/plugins/resource/apiconnector/plugin.ts
```diff
@@ -34,15 +36,65 @@ export class ApiConnectorImpl {
...
+    });
 
+    try {
+      config.ComponentPath.forEach(async (componentItem) => {
+        await this.scaffoldInComponent(projectPath, componentItem, config, languageType);
+      });
+    } catch (err) {
+      componentFolders.forEach(async (componentPath) => {
+        await fs.move(path.join(componentPath, backupFolderName), componentPath, {
```
xzf0587: need to remove sample code?

wenytang-ms: sure, thanks!

xzf0587 (dismissed):

blackchoey (approved):

packages/fx-core/src/plugins/resource/apiconnector/plugin.ts
```diff
@@ -34,15 +42,88 @@ export class ApiConnectorImpl {
...
+      })
+    );
 
+    try {
+      await Promise.all(
+        config.ComponentPath.map(async (component) => {
+          await this.scaffoldInComponent(projectPath, component, config, languageType);
+        })
+      );
+    } catch (err) {
```
xzf0587, line 59: minor case: the err caught maybe a systemError/userError type, which can be used directly.

xzf0587 (approved):

packages/fx-core/src/plugins/resource/apiconnector/plugin.ts
wenytang-ms: thank you, will resolve in the next pr

PR #4466, opened 2022-03-31
fix(bot): skip deployment after *.ts file is changed

XiaofuHuang:
Fix the bug during calculate whether need to redeploy.
1. Move the generation of `lastUpdateTime` after `npm build`
2. Do not check the file which is ignored in both `.funcignore` and `.gitignore`.
PS: `*.ts` is ignored in  `.funcignore` bot not in `.gitignore`

### DIFF
--- a/packages/fx-core/src/plugins/resource/bot/functionsHostedBot/deployMgr.ts
+++ b/packages/fx-core/src/plugins/resource/bot/functionsHostedBot/deployMgr.ts
@@ -40,11 +40,17 @@ export class FuncHostedDeployMgr {
     }
   }
 
-  public async needsToRedeploy(ignoreRules: string[]): Promise<boolean> {
+  public async needsToRedeploy(): Promise<boolean> {
     try {
       const lastDeployTime = await this.getLastDeployTime();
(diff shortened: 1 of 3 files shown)


XiaofuHuang:
E2E TEST: None

packages/fx-core/src/plugins/resource/bot/functionsHostedBot/deployMgr.ts
```diff
@@ -40,11 +40,17 @@ export class FuncHostedDeployMgr {
     }
   }
 
-  public async needsToRedeploy(ignoreRules: string[]): Promise<boolean> {
+  public async needsToRedeploy(): Promise<boolean> {
```
swatDong, line 43: Is it possible to always run `build` stage before checking changes?
In case certain ignored file impacts the build output, e.g. some specific settings in `.npmrc`.

XiaofuHuang: If we run `build` stage first, the output files in `dist/` will be updated. So the project will always be deployed.

swatDong (approved):

Siglud (approved):

PR #4477, opened 2022-03-31
chore(vsc): add command controller to execute command

tecton:
Split the controller from TreeViewManager so that TreeViews are no longer registered in non-teamsfx project.


HuihuiWu-Microsoft (approved):
LGTM

PR #4499, opened 2022-04-02
test: add test to make sure create project event always has project id

HuihuiWu-Microsoft:
Add test for https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/13900320


tecton (commented):
This test checks that projectId is passed to `createProject`, it there any test to ensure that new-project-id is defined in telemetry?

HuihuiWu-Microsoft:
> This test checks that projectId is passed to `createProject`, it there any test to ensure that new-project-id is defined in telemetry?

Nice catch. Actually new-project-id will be attached with the value of project-id and this bug is introduced because the removal of projectId(I thought there's no relationship with new-project-id). I can update the test later to also verify new-project-id property.

tecton (approved):

PR #4503, opened 2022-04-02
perf(solution): remove added files when failure

KennethBWSong:


packages/fx-core/src/plugins/solution/fx-solution/v2/executeUserTask.ts
```diff
@@ -1098,7 +1098,7 @@ export async function createAuthFiles(
         await unzip(sampleZip, tabFolder);
       }
     } catch (error) {
-      // TODO: remove added code
+      await fs.remove(tabFolder);
```
SLdragon, line 1101: check whether folder exist before remove

KennethBWSong: The folder is created above, it should exist
image

SLdragon (dismissed):

packages/fx-core/src/plugins/solution/fx-solution/v2/executeUserTask.ts
1yefuwang1: Actually, it's possible that OS fails to create the folder and throw an error. So, the folder doesn't necessarily exist here.

KennethBWSong: @1yefuwang1 Error handling here improved. Thx!

1yefuwang1 (approved):

PR #4509, opened 2022-04-02
chore(m365): add outlook channel prompt in F5 flow for bot app

kuojianlu:
For launch page:
image
For messaging extension:
image


packages/vscode-extension/src/debug/commonUtils.ts
```diff
@@ -238,6 +239,25 @@ export async function getLocalTeamsAppId(): Promise<string | undefined> {
   }
 }
 
+export async function getLocalBotId(): Promise<string | undefined> {
+  try {
+    if (isConfigUnifyEnabled()) {
```
swatDong, line 244: Do we have the case that an M365 app with unify config disabled?

swatDong:
Since currently hard to detect outlook channel, do we have any expected error to give some hints to user? E.g., if you see XXX error/behavior on outlook, that means you have not connected your bot to outlook channel yet.

packages/vscode-extension/src/debug/commonUtils.ts
kuojianlu: At current stage, unify config is disabled by default, so M365 feature also needs to check this config.

swatDong (approved):

kuojianlu:
> Since currently hard to detect outlook channel, do we have any expected error to give some hints to user? E.g., if you see XXX error/behavior on outlook, that means you have not connected your bot to outlook channel yet.

Good idea to give some hints to user. If not connected to Outlook channel, the messaging extension will show "Something went wrong". Maybe we can put this information in README or help doc.
image

HuihuiWu-Microsoft (approved):

PR #4513, opened 2022-04-02
chore: update local debug telemetry

xiaolang124:
1. Fix local debug telemetry.


packages/fx-core/src/common/local/localEnvManager.ts
```diff
@@ -121,7 +121,7 @@ export class LocalEnvManager {
 
   public async getLocalEnvInfo(
     projectPath: string,
-    cryptoOption?: { projectId: string }
+    cryptoOption: { projectId: string }
```
swatDong, line 124: Why make this required? I remember for some cases just ignoring crypto to improve performance.

packages/cli/src/utils.ts
```diff
@@ -389,15 +408,28 @@ export function getLocalTeamsAppId(rootfolder: string | undefined): any {
...
-    } catch (error) {
-      return undefined;
+    if (isConfigUnifyEnabled()) {
+      const result = readLocalStateJsonFile(rootfolder);
+      if (result.isErr()) {
+        return undefined;
+      }
+      const localState = result.value;
+      try {
+        return localState["fx-resource-appstudio"].teamsAppId;
```
kuojianlu, line 418: Use `PluginNames.APPST`.

xiaolang124: Fixed.

packages/fx-core/src/common/local/localEnvManager.ts
xiaolang124: Because this part uses ```environmentManager.loadEnvInfo``` to load envInfo, it needs ```crypto``` as a parameter.

swatDong: I remember the `crypto` in `loadEnvInfo` can be undefined. Does it still support undefined now?

xiaolang124: image
I have tried with undefined ```crypto```, it will fail in the end.

swatDong: Oh, I see. The original `LocalSettingsProvider` supports empty crypto, but the `LocalStateProvider` does not.

swatDong (approved):

Alive-Fish (approved):

PR #4517, opened 2022-04-05
fix: spfx scaffold failed due to yo not recognized

HuihuiWu-Microsoft:
ADO:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/14000279

1> Get bin folder according to npm version since npm v7 has different behavior with npm v6
2> Show instruction to tell user suggested versions for npm and node.(Maybe add help link to tell user how to troubleshooting later in dev branch)

image

E2E TEST: will add later

### DIFF
--- a/packages/fx-core/src/plugins/resource/spfx/plugin.ts
+++ b/packages/fx-core/src/plugins/resource/spfx/plugin.ts
@@ -406,7 +406,7 @@ export class SPFxPluginImpl {
         (error as any).message.includes("'yo' is not recognized as an internal or external command")
       ) {
         ctx.logProvider?.error(
-          "NPM v6.x is recommended for spfx scaffolding and later development. You can use correct version and try again."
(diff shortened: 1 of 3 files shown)


nliu-ms (approved):

packages/fx-core/src/plugins/resource/spfx/plugin.ts
```diff
@@ -401,6 +401,14 @@ export class SPFxPluginImpl {
           `Teams Toolkit recommends using ${yoInfo.displayName} ${genInfo.displayName}`
         );
       }
+      if (
+        (error as any).message &&
+        (error as any).message.includes("'yo' is not recognized as an internal or external command")
+      ) {
+        ctx.logProvider?.error(
+          "NPM v6.x with Node.js v12.13.0+ (Erbium) or Node.js v14.15.0+ (Fermium) is recommended for spfx scaffolding and later development. You can use correct version and try again."
```
nliu-ms, line 409: do we need localization here?

HuihuiWu-Microsoft: Logs in output window is not required.

PR #4520, opened 2022-04-06
chore(apicon): improve code for apicon

xzf0587:


packages/fx-core/src/plugins/resource/apiconnector/utils.ts
```diff
@@ -25,3 +28,14 @@ export async function removeFileIfExist(file: string) {
     await fs.remove(file);
   }
 }
+
+export function checkInputEmpty(inputs: Inputs, ...keys: string[]) {
+  for (const key of keys) {
+    if (!inputs[key]) {
+      throw ResultFactory.SystemError(
+        ErrorMessage.ApiConnectorInputError.name,
+        ErrorMessage.ApiConnectorInputError.message(key)
```
blackchoey, line 37: The key is internal strings. Can we provide some strings that displayed on UI so users can know which input is empty?
This can be a low priority thing since we can add checks when asking user input. Feel free to add a task and resolve in another PR.

blackchoey (approved):

packages/fx-core/src/plugins/resource/apiconnector/utils.ts
xzf0587: We have validated the inputs in the question model already. 
Here is the intern check for plugin owner to guarantee the input in self-controlled code. Usually, it will not be triggered.
Users could create an issue for system error.

PR #4523, opened 2022-04-06
refactor(vsc): change environment tree to native provider to enhance UX

tecton:
1. Postpone the data update time when refreshing tree view.
2. Leverage the progress indicator of VSCode like "clock" and running progress bar.
3. Remove the registration of environment tree in non-teamsfx project.

E2E TEST: N/A


packages/vscode-extension/src/treeview/environmentTreeItem.ts
```diff
@@ -0,0 +1,233 @@
...
+        const subscriptionNode = new SubscriptionNode(this.identifier, subscriptionInfo);
+        children.push(subscriptionNode);
+      }
+    }
+    return children;
+  }
+
+  public async getTreeItem(): Promise<vscode.TreeItem> {
+    const envInfo = await this.getCurrentEnvInfo(this.identifier);
+    const isSpfxProject = await isSPFxProject(ext.workspaceUri.fsPath);
```
qinezh, line 68: seems the `await` is unnecessary.

qinezh: do we need to check if `ext.workspaceUri` is valid here?

packages/vscode-extension/src/treeview/environmentTreeItem.ts
```diff
@@ -0,0 +1,233 @@
...
+    } else {
+      this.resourceGroupNode = undefined;
+      this.collapsibleState = vscode.TreeItemCollapsibleState.None;
+    }
+
+    return this;
+  }
+}
+
+class ResourceGroupNode extends DynamicNode {
```
qinezh, line 219: where's the collaboration related node?

packages/vscode-extension/src/treeview/environmentTreeItem.ts
tecton: Thanks, fixed this `await` code.
The `ext.workspaceUri` is guaranteed in `EnvironmentTreeViewProvider:reloadEnvironments()` so it should always be valid here.

packages/vscode-extension/src/treeview/environmentTreeItem.ts
tecton: Do you mean `generateCollaboratorNode()` and `generateCollaboratorWarningNode()`? I didn't find any usage so I just removed these 2 functions.
Other collaboration functions seem to be provided by "view item context" and they are updated in `extension.ts`.

packages/vscode-extension/src/treeview/environmentTreeViewProvider.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  public static getInstance() {
+    if (!EnvironmentTreeViewProvider.instance) {
+      EnvironmentTreeViewProvider.instance = new EnvironmentTreeViewProvider();
+    }
+    return EnvironmentTreeViewProvider.instance;
+  }
+
+  public async reloadEnvironments(): Promise<Result<Void, FxError>> {
+    if (!ext.workspaceUri) {
```
HuihuiWu-Microsoft, line 34: Maybe check whether it's teamsfx project here so we don't need to continue for non-teamsfx projects?

packages/vscode-extension/src/treeview/environmentTreeItem.ts
qinezh: @SLdragon @KennethBWSong may take a look at the usage of collaboration node.

SLdragon: Thanks, we don't add collaboration node to the tree view any more, so we can safely remove this two function

SLdragon (approved):

packages/vscode-extension/src/treeview/environmentTreeViewProvider.ts
tecton: Thanks, fixed.

packages/vscode-extension/src/treeview/environmentTreeItem.ts
tecton: This makes sense and it's fixed.

qinezh (approved):

HuihuiWu-Microsoft (approved):

PR #4532, opened 2022-04-06
chore: add ConversationBot back to sdk

swatDong:
Add `ConversationBot` back to provide simpler user code, e.g.,
``` typescript
const conversationBot = new ConversationBot({
  command: { enable: true },
  notification: { enable: true },
});
conversationBot.xxx();
conversationBot.command.xxx();
conversationBot.notification.xxx();
```


swatDong:
@dooriya - per feedback discussion, this is to provide more *default* APIs.

@tecton - please help review since there're *export* changes in `index.ts`.

packages/sdk/src/conversation/conversation.ts
```diff
@@ -0,0 +1,111 @@
...
+   *
+   * @param options - initialize options
+   *
+   * @beta
+   */
+  public constructor(options: ConversationOptions) {
+    if (options.adapter) {
+      this.adapter = options.adapter;
+    } else {
+      this.adapter = new BotFrameworkAdapter({
```
dooriya, line 97: According to Zhenya's feedback, we may also provide the default `onTurnError` callback for the default adapter.

packages/sdk/src/conversation/interface.ts
```diff
@@ -138,3 +139,73 @@ export interface TeamsFxBotCommandHandler {
...
+   *
+   * @beta
+   */
+  command: {
+    /**
+     * Whether to enable command or not.
+     *
+     * @beta
+     */
+    enable: boolean;
```
dooriya, line 181: nit: `enable` -> `enabled`

packages/sdk/src/conversation/conversation.ts
swatDong: Yep, I'd like to consolidate class and constructor definitions first. Default error handling and request handler will be in later PRs.

BTW, any suggestion on the input options? I considered two ways for enablement, not sure which one is better.
a. An `enabled` flag for each functionality, like `options.command.enabled` + `options.notification.enabled` + `options.xxx.enabled`
b. Use enum flags, like `options.enablement = Command | Notification | Xxx`

dooriya: The former one (in your current PR) looks good to me, since we can organize the `ConversationOptions` in the same way (hierarchy) as the underlying implementation.

dooriya (dismissed):

dooriya (approved):

tecton (approved):

PR #4542, opened 2022-04-06
chore: refine the trigger pattern for command-response bot

dooriya:
Refine current `TeamsFxBotCommandHandler` to support:
- A set of trigger patterns, it can be `string`, `RegExp` or `(string | RegExp)[]`.
- Also pass the capture groups that matches the provided regular expression(s). This is abstracted in the `CommandMessage` interface:
    * the `commandMessage.text` property represents the user input text
    * the `commandMessage.matches` property represents the captures groups if it has.


packages/sdk/src/conversation/middleware.ts
```diff
@@ -120,33 +132,44 @@ export class CommandResponseMiddleware implements Middleware {
...
     return ActivityType.Unknown;
   }
 
-  private isCommandReceived(activity: Activity): boolean {
-    if (this.commandHandlers) {
-      const commandText = this.getActivityText(activity);
-      const handlers = this.filterCommandHandler(commandText, this.commandHandlers);
-      return handlers.length > 0;
+  private matchPattern(pattern: string | RegExp, text: string): boolean | RegExpMatchArray {
+    const normalizedText = text?.trim().toLocaleLowerCase();
```
swatDong, line 143: The same as line 164. Since both are private functions, text can be normalized at the very beginning.

packages/sdk/src/conversation/interface.ts
```diff
@@ -115,6 +115,26 @@ export interface NotificationOptions {
   storage?: NotificationTargetStorage;
 }
 
+/**
+ * The trigger pattern used to trigger a {@link TeamsFxBotCommandHandler} instance.
+ */
+export type TriggerPatterns = string | RegExp | (string | RegExp)[];
```
swatDong, line 121: Would it be simpler if just use `(string | RegExp)[]`? Support single element array as well.

packages/sdk/src/conversation/middleware.ts
```diff
@@ -120,33 +132,44 @@ export class CommandResponseMiddleware implements Middleware {
...
+
+    if (typeof pattern === "string") {
+      const test = new RegExp(pattern as string, "i");
+      if (normalizedText && normalizedText.match(test)) {
+        return true;
+      }
+    } else if (pattern instanceof RegExp) {
+      const test = pattern as RegExp;
+      if (normalizedText && normalizedText.match(test)) {
+        const matches = normalizedText.match(test);
```
swatDong, line 153: Is it possible to only call `match` once?

packages/sdk/src/conversation/middleware.ts
dooriya: make sense, all these normalization can be removed since this is already handled in `getActivityText()`.

packages/sdk/src/conversation/middleware.ts
dooriya: updated.

packages/sdk/src/conversation/interface.ts
dooriya: From developers' perspective, I think this would be more flexible, clear and straightforward if we also support `string` and `RegExp` in the surface, like the [hears() API in BotKit](https://github.com/howdyai/botkit/blob/main/packages/docs/reference/core.md#hears), although single element in array can also cover the case. 

Since this would not bring too much effort in our SDK development, why not just support it  :)

packages/sdk/src/conversation/middleware.ts
```diff
@@ -120,33 +132,38 @@ export class CommandResponseMiddleware implements Middleware {
...
+  private matchPattern(pattern: string | RegExp, text: string): boolean | RegExpMatchArray {
+    if (text) {
+      if (typeof pattern === "string") {
+        const regExp = new RegExp(pattern as string, "i");
+        return regExp.test(text);
+      }
+
+      if (pattern instanceof RegExp) {
+        const matches = text.match(pattern as RegExp);
+        return matches ?? true;
```
swatDong, line 151: `matches ?? false`? Null `matches` means not matched.

dooriya: Yep. Should be a bug and already fixed.

swatDong (approved):

packages/sdk/src/conversation/interface.ts
```diff
@@ -125,18 +145,18 @@ export interface TeamsFxBotCommandHandler {
   /**
    * The command name or RegExp pattern that can trigger this handler.
```
tecton, line 146: "command name" can be removed in comment.

swatDong (approved):

tecton (approved):

packages/sdk/src/conversation/interface.ts
dooriya: Good catch, fixed.

tecton (approved):

swatDong (approved):

PR #4546, opened 2022-04-07
chore(env-tree): display accurate warning

qinezh:
Since Azure account is not required for existing tab scenario, this PR is to fix the inaccurate warning under the env tree:

Before:

After:


packages/vscode-extension/src/treeview/environmentTreeItem.ts
```diff
@@ -144,11 +158,18 @@ export class EnvironmentNode extends DynamicNode {
...
 class WarningNode extends DynamicNode {
-  constructor(public identifier: string, warnings: string[]) {
+  constructor(public identifier: string, accountStatus: accountStatus) {
     super(identifier, vscode.TreeItemCollapsibleState.None);
-    this.label = `Sign in with your correct Azure / M365 account`;
+    if (accountStatus.isAzureAccountLogin === false && !accountStatus.isM365AccountLogin) {
+      this.label = `Sign in with your correct Azure / M365 account`;
+    } else if (!accountStatus.isM365AccountLogin) {
+      this.label = `Sign in with your correct M365 account`;
+    } else if (!accountStatus.isAzureAccountLogin === false) {
```
HuihuiWu-Microsoft, line 167: Should be "accountStatus.isAzureAccountLogin === false" here?

qinezh: nice catch!!

HuihuiWu-Microsoft (dismissed):
LGTM

packages/vscode-extension/src/treeview/environmentTreeItem.ts
```diff
@@ -144,11 +158,18 @@ export class EnvironmentNode extends DynamicNode {
 }
 
 class WarningNode extends DynamicNode {
-  constructor(public identifier: string, warnings: string[]) {
+  constructor(public identifier: string, accountStatus: accountStatus) {
     super(identifier, vscode.TreeItemCollapsibleState.None);
-    this.label = `Sign in with your correct Azure / M365 account`;
+    if (accountStatus.isAzureAccountLogin === false && !accountStatus.isM365AccountLogin) {
+      this.label = `Sign in with your correct Azure / M365 account`;
```
tecton, line 164: Do we need to localize these strings?

qinezh: I think so, let me move it to package.nls.json, thanks!

tecton (approved):

PR #4549, opened 2022-04-07
perf: add notification to api connector

wenytang-ms:


packages/fx-core/src/plugins/resource/apiconnector/constants.ts
```diff
@@ -33,6 +33,8 @@ export class Constants {
   public static readonly pkgLockFile = "package-lock.json";
   public static readonly sdkName = "@microsoft/teamsfx";
   public static readonly sdkJsName = "@microsft/teams-js";
+  public static readonly READ_MORE = "Read more";
```
blackchoey, line 36: we may also need to localize this string

packages/fx-core/src/plugins/resource/apiconnector/utils.ts
```diff
@@ -39,3 +41,69 @@ export function checkInputEmpty(inputs: Inputs, ...keys: string[]) {
     }
   }
 }
+
+export class Notification {
+  public static GetBasiString(apiName: string, components: string[], languageType: string): string {
```
blackchoey, line 46: We can keep current design. In the future, we can see how to let each auth type's (and programming language's) template provider to provide the message. Since the content highly depends on the template.

blackchoey (approved):

packages/fx-core/src/plugins/resource/apiconnector/constants.ts
wenytang-ms: sure, will fix in the later PR.

packages/fx-core/src/plugins/resource/apiconnector/utils.ts
wenytang-ms: sure

PR #4559, opened 2022-04-07
feat: add default behaviors to conversation bot

swatDong:
- Add default error handling
- Add default web request handler
- Make constructor option's fields optional


swatDong:
E2E TEST: N/A (by unit test)

packages/sdk/src/conversation/conversation.ts
```diff
@@ -94,18 +95,78 @@ export class ConversationBot {
     if (options.adapter) {
       this.adapter = options.adapter;
     } else {
-      this.adapter = new BotFrameworkAdapter({
-        appId: process.env.BOT_ID,
-        appPassword: process.env.BOT_PASSWORD,
-      });
+      this.adapter = this.createDefaultAdapter(process.env.BOT_ID, process.env.BOT_PASSWORD);
+    }
```
dooriya, line 99: Can we also allow user to pass in the `appId` and `appPassword` to create the default adapter?

packages/sdk/src/conversation/conversation.ts
```diff
@@ -94,18 +95,78 @@ export class ConversationBot {
...
+   * ```
+   *
+   * @beta
+   */
+  public async requestHandler(
+    req: WebRequest,
+    res: WebResponse,
+    logic?: (context: TurnContext) => Promise<any>
+  ): Promise<void> {
+    if (logic === undefined) {
```
dooriya, line 165: What do you think of providing a default `TeamsActivityHandler` for the `ConversationBot` users? If they want to handler [other event/activity](https://docs.microsoft.com/en-us/microsoftteams/platform/bots/bot-basics?tabs=javascript#bot-logic), they can just implement the default handler's `onXXX` method.

packages/sdk/src/conversation/conversation.ts
swatDong: Good suggestion. Added `adapterConfig` as well.

packages/sdk/src/conversation/conversation.ts
swatDong: IMO, different from `Adapter`, `ConversationBot` has nothing to do with `TeamsActivityHandler`. If providing activity handler, it may confuse developer that *conversation bot requires an activity handler for balabala...*

My suggestion is to add `TeamsActivityHandler` in template or sample code if needed. This `requestHandler` does support to integrate with activity handler:
``` typescript
server.post("api/messages", async (req, res) => {
  await conversationBot.requestHandler(req, res, async (context) => {
    await activityHandler.run(context);
  });
});
```

BTW, providing default `TeamsActivityHandler` may make the SSO scenario complex...

dooriya:
>  *   - send notification to varies targets (e.g., member, channel, incoming wehbook).

`wehhook` -> `webhook`

---
Refers to: packages/sdk/src/conversation/conversation.ts:12 in 0d9e8bc. [](commit_id = 0d9e8bc34f63cea3a130b5c4229c9a628c471c3d, deletion_comment = False)

dooriya (approved):

tecton (approved):

PR #4564, opened 2022-04-07
chore: always use unify config model

xiaolang124:
1. Always use unify config model.
2. Update unit test cases to support unify config.
3. Remove unify config vscode setting.


packages/fx-core/tests/plugins/resource/appstudio/spfx-resources/appPackage/manifest.template.json
```diff
@@ -0,0 +1,56 @@
+{
```
HuihuiWu-Microsoft, line 1: What about removing the legacy manifest.local/remote.template.json files if unused anymore?

xiaolang124: Thanks for the advice, I will work on this in the next pr.

HuihuiWu-Microsoft (approved):

KennethBWSong (approved):
approve for simple auth

jayzhang (approved):

PR #4566, opened 2022-04-07
feat: manifest auto completion

nliu-ms:
1. When user type "" in manifest template file, we can provide an auto completion. But it duplicates with language server. 
image

image

3. Another way is to provide default snippet in schema file. But we cannot update the manifest schema. 
https://code.visualstudio.com/Docs/languages/json#_define-snippets-in-json-schemas
image

And $schema defined in json is prior to schema defined in package.json
https://code.visualstudio.com/api/references/contribution-points#contributes.jsonValidation


nliu-ms:
E2E TEST: no test for extension

PR #4568, opened 2022-04-07
feat: add cert auth provider for API support

blackchoey:
E2E TEST: https://github.com/OfficeDev/TeamsFx/blob/chyuan/api-cert-auth-provider/packages/sdk/test/e2e/node/certificateAuthProvider.spec.ts


packages/sdk/src/apiClient/certificateAuthProvider.ts
```diff
@@ -0,0 +1,127 @@
...
+  private certOption: SecureContextOptions;
+
+  /**
+   *
+   * @param { SecureContextOptions } certOption - information about the cert used in http requests
+   *
+   * @beta
+   */
+  constructor(certOption: SecureContextOptions) {
+    this.certOption = certOption;
```
wenytang-ms, line 27: will the cerOption will be empty object?

blackchoey: It's hard to validate whether the certOption is valid. So I only validates user input in `createPemCertOption` and `createPfxCertOption`, since we have clear scenario for these 2 helper functions.

If advanced user wants to pass their own certOption (which is not generated by our helper function) here, they need to ensure the option is correct. And we rely on Axios to find problems in the certOption.

packages/sdk/src/index.browser.ts
```diff
@@ -30,6 +30,11 @@ export {
 export { createApiClient } from "./apiClient/apiClient";
 export { AuthProvider } from "./apiClient/authProvider";
 export { BearerTokenAuthProvider } from "./apiClient/bearerTokenAuthProvider";
+export {
+  CertificateAuthProvider,
+  createPemCertOption,
+  createPfxCertOption,
+} from "./apiClient/certificateAuthProvider.browser";
```
dilin-MS, line 37: Do we want to have separate certificateAuthProvider file for browser? Or we just leverage the same file?

blackchoey: The logic for browser/node is not same, so we need different files

packages/sdk/test/unit/apiClient/certificateAuthProvider.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+  ErrorCode,
+  ErrorWithCode,
+} from "../../../src";
+import * as chaiPromises from "chai-as-promised";
+
+chaiUse(chaiPromises);
+
+describe("CertificateAuthProvider Tests - Node", () => {
+  it("should add PEM cert to https agent", async function () {
+    const bearerTokenAuthProvider = new CertificateAuthProvider({
```
tecton, line 18: bearerTokenAuthProvider -> certificateAuthProvider?

packages/sdk/test/unit/apiClient/certificateAuthProvider.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+    expect(() => createPfxCertOption(""))
+      .to.throw("Parameter pfx is empty")
+      .that.has.property("code")
+      .that.equals(ErrorCode.InvalidParameter);
+  });
+});
+
+describe("createPemCertOption Tests - Node", () => {
+  it("should initialize SecureContextOption with correct property", async function () {
+    it("should initialize SecureContextOption with correct property", async function () {
```
tecton, line 121: Is this duplicated with above line?

packages/sdk/test/unit/apiClient/certificateAuthProvider.spec.ts
blackchoey: thank you, updated

packages/sdk/test/unit/apiClient/certificateAuthProvider.spec.ts
blackchoey: thank you, updated

tecton (approved):

wenytang-ms (approved):

PR #4569, opened 2022-04-07
fix: add missing await

qinezh:
E2E TEST: N/A for extension

### DIFF
--- a/packages/vscode-extension/test/unit/extension/handlers.test.ts
+++ b/packages/vscode-extension/test/unit/extension/handlers.test.ts
@@ -81,14 +81,13 @@ suite("handlers", () => {
       const clock = sinon.useFakeTimers();
 
       sinon.stub(handlers, "core").value(new MockCore());
-      sinon.stub(commonUtils, "isExistingTabApp").returns(Promise.resolve(false));
       const sendTelemetryEventFunc = sinon.stub(ExtTelemetry, "sendTelemetryEvent");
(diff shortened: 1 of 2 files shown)


packages/vscode-extension/test/unit/extension/handlers.test.ts
```diff
@@ -81,6 +81,7 @@ suite("handlers", () => {
       const clock = sinon.useFakeTimers();
 
       sinon.stub(handlers, "core").value(new MockCore());
+      sinon.stub(commonUtils, "isExistingTabApp").returns(Promise.resolve(false));
```
swatDong, line 84: The root cause should be at line 91 - real `await` hangs if using fake timer, as the clock is stopped.
See [doc](https://sinonjs.org/releases/latest/fake-timers/#examples) as well.

Correct fix should be
``` typescript
const promise = xxx(); // do not await here
clock.tick(x);
await promise();
```

or
``` typescript
xxx(); // do not await here
await clock.tickAsync(x);
```

qinezh: Tried the second approach, but still hang... 

Considering the case mainly tests the timing to trigger `vscode.openFolder`, I'll mock the `isExistingTabApp` to unblock this case.

swatDong (approved):

HuihuiWu-Microsoft (approved):

PR #4571, opened 2022-04-07
perf: fix bug in add sso

KennethBWSong:
fix bug:
1. show correct error message
2. add correct auth files according to capability
3. scaffold aad plugin when adding sso tab

### DIFF
--- a/packages/fx-core/src/plugins/solution/fx-solution/v2/executeUserTask.ts
+++ b/packages/fx-core/src/plugins/solution/fx-solution/v2/executeUserTask.ts
@@ -336,18 +336,18 @@ export async function addCapability(
   if (isAadManifestEnabled()) {
     if (alreadyHasTabSso && toAddTabNonSso) {
       const e = new SystemError(
+        SolutionSource,
         SolutionError.InvalidInput,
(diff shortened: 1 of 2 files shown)


packages/fx-core/src/plugins/solution/fx-solution/v2/executeUserTask.ts
```diff
@@ -822,9 +823,9 @@ export function canAddSso(
   // Can not add sso if feature flag is not enabled
   if (!isAadManifestEnabled()) {
     const e = new SystemError(
+      SolutionSource,
       SolutionError.NeedEnableFeatureFlag,
-      getLocalizedString("core.addSso.needEnableFeatureFlag"),
-      SolutionSource
+      getLocalizedString("core.addSso.needEnableFeatureFlag")
     );
```
SLdragon, line 829: tip: can we hide add sso command palette in package.json file? So that we do not need this check any more

KennethBWSong: Actually it is already disabled in the command palette. Here it is only to ensure user we won't enable this feature by mistake.

SLdragon (approved):

1yefuwang1 (approved):

PR #4573, opened 2022-04-08
perf: add confirm before execute api connector user task

wenytang-ms:
Animation

Cancel


packages/fx-core/src/plugins/resource/apiconnector/index.ts
```diff
@@ -40,6 +41,19 @@ export class ApiConnectorPluginV2 implements ResourcePlugin {
...
   ): Promise<ApiConnectorResult> {
+    const res = await ctx.userInteraction?.showMessage(
+      "warn",
+      getLocalizedString("plugins.apiconnector.ExecuteUserTask.Message"),
+      true,
+      "OK"
+    );
+    const answer = res?.isOk() ? res.value : undefined;
+    if (!answer || answer != "OK") {
+      throw ResultFactory.UserError(
```
blackchoey, line 52: Will this error be displayed to user? Since the cancel button is clicked by user, we may not want to show cancel error to user again. But let's ensure we have telemetry to record how many users clicked cancel.

packages/fx-core/resource/package.nls.json
```diff
@@ -250,6 +250,7 @@
   "error.cicd.FailedToListMultiEnv": "Failed to list multi env.",
   "error.cicd.PreconditionNotMet": "Some preconditions of inputs are not met.",
   "error.cicd.NoProjectOpened.details": "No project opened.",
+  "plugins.apiconnector.ExecuteUserTask.Message": "Teams Toolkit is going to generate new sample files in your bot or function project. This process will not change your custom code and create the backup files in case revert is needed.",
```
blackchoey, line 253: Can we change the word to following?
`Teams Toolkit is going to generate new sample file, update package.json and .env.teamsfx.local files in the selected component. The changes will be rollback if unexpected error occurred. Do you want to continue?`
The key thing we want to deliver is which files will be changed. In the meantime, I added a question to the end to let users feel more smooth.

@MuyangAmigo do you have any suggestions on the wording? The messages will be displayed to user before we start updating the project. Just like the confirmation we displayed during provision.

packages/fx-core/src/plugins/resource/apiconnector/index.ts
wenytang-ms: this will display in the output, like cancel any other command pallette with clicking Esc

packages/fx-core/resource/package.nls.json
MuyangAmigo: When and how will this message be prompted?

wenytang-ms: once user trigger the command, before asking user questions, toolkit will pop up the prompt.
And we need user's approval to continue.

PR #4575, opened 2022-04-08
chore(m365): add bot m365 template

kuojianlu:
M365 templates are the same as the default templates. Adding M365 scenario is mainly for updating README files later.


swatDong (approved):
How to treat the default "welcome" and "learn" command? Remove them later or add to README that command cannot be previewed in outlook?

kuojianlu:
> How to treat the default "welcome" and "learn" command? Remove them later or add to README that command cannot be previewed in outlook?

Prefer letting users know this in README. Since bot and messaging extension share the same template, removing the bot part may be not a good choice.

Siglud (approved):

PR #4576, opened 2022-04-08
fix: add capability notification/commandbot manifest

a1exwang:
- Fix the bug 'manifest is not correct when adding notification/commandresponse capability'

notification:
img

command response:
img

https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/13942934/


a1exwang:
E2E TEST: n/a

nliu-ms (approved):

PR #4578, opened 2022-04-08
chore(apicon): add aad for existing app

xzf0587:


packages/fx-core/templates/plugins/resource/apiconnector/sample/js/aad.template
```diff
@@ -1,7 +1,15 @@
 const teamsfxSdk = require("@microsoft/teamsfx");
 
 // Loads current app's configuration
+{{#if (equals config.AuthConfig.ReuseTeamsApp true)}}
 const teamsFx = new teamsfxSdk.TeamsFx(teamsfxSdk.IdentityType.App);
+{{else}}
+const teamsFx = new teamsfxSdk.TeamsFx(teamsfxSdk.IdentityType.App, {
+  tenantId: process.env.TEAMSFX_API_{{capitalName}}_TENANT_ID,
```
blackchoey, line 8: we may also need to set authority host. we can hard code the value since most of the time people will use the global AAD endpoint

xzf0587: It is updated. Thanks.

blackchoey (approved):

PR #4584, opened 2022-04-08
fix(env-tree): update the collapsible status condition

qinezh:
Update the condition to collapse the env tree node according to its children number.

E2E TEST: UI related.

### DIFF
--- a/packages/vscode-extension/src/treeview/environmentTreeItem.ts
+++ b/packages/vscode-extension/src/treeview/environmentTreeItem.ts
@@ -44,11 +44,17 @@ const subscriptionIcon = new vscode.ThemeIcon("key");
 const resourceGroupIcon = new vscode.ThemeIcon("symbol-method");
 
 export class EnvironmentNode extends DynamicNode {
+  private _children: DynamicNode[] | undefined;
+
   constructor(public identifier: string) {
     super(identifier, vscode.TreeItemCollapsibleState.None);
   }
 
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/treeview/environmentTreeItem.ts
```diff
@@ -69,13 +69,14 @@ export class EnvironmentNode extends DynamicNode {
 
   public async getTreeItem(): Promise<vscode.TreeItem> {
     const envInfo = await this.getCurrentEnvInfo(this.identifier);
-    const isSpfxProject = ext.workspaceUri ? isSPFxProject(ext.workspaceUri.fsPath) : false;
-
     this.iconPath = envInfo === EnvInfo.ProvisionedRemoteEnv ? provisionedIcon : nonProvisionedIcon;
+
+    const children = await this.getChildren();
```
tecton, line 74: I think we'd better not calling `getChildren()` directly because VSCode will call this function if collapsibleState is Expanded. So the function would be called twice. The time cost is relatively small here but it's better to cache the children data.

qinezh: Thanks! It makes sense, let me take a further look at it.

qinezh: Updated, please take a look.

tecton (approved):

PR #4591, opened 2022-04-08
chore(cli-m365): update m365 preview wording

kuojianlu:


swatDong (approved):
The code logic looks good to me, but the PR title does not match the changes.

PR #4601, opened 2022-04-10
perf: skip arm deployment when adding sso

KennethBWSong:


packages/fx-core/src/plugins/resource/aad/plugin.ts
```diff
@@ -424,18 +425,24 @@ export class AadAppForTeamsImpl {
...
-          ),
-          ConstantString.UTF8Encoding
-        )
-      ),
-    };
+    const solutionSettings = ctx.projectSettings?.solutionSettings as AzureSolutionSettings;
+    const capabilities = solutionSettings.capabilities;
+    let result: ArmTemplateResult | undefined = undefined;
+
+    if (capabilities.includes(TabOptionItem.id) || capabilities.includes(BotOptionItem.id)) {
```
SLdragon, line 432: Except bot and tab, do we need to consider other azure resource type such as sql, function?

KennethBWSong: Currently there won't be any azure resource except aad that exists without Tab or Bot capability.
We plan to use the hasAzureResource function, but this function includes AAD. I will confirm with Huajie with this, and will not include this change in this pr.

SLdragon: Sure, I created another pr to update hasAzureResource function~

packages/fx-core/package-lock.json
```diff
@@ -2192,7 +2192,7 @@
     },
     "archy": {
       "version": "1.0.0",
-      "resolved": "https://registry.npmjs.org/archy/-/archy-1.0.0.tgz",
+      "resolved": "https://msazure.pkgs.visualstudio.com/9660fff2-2363-48b0-9e15-64df2283e932/_packaging/MODS-NPM-Registry/npm/registry/archy/-/archy-1.0.0.tgz",
```
SLdragon, line 2195: Seems the registry is not correct.

KennethBWSong: It is automatically changed when I tried to fix the audit error in CI. I will check with wenyu.

packages/fx-core/src/plugins/solution/fx-solution/arm.ts
```diff
@@ -877,6 +877,12 @@ async function doGenerateArmTemplate(
   );
   const moduleProvisionFiles = new Map<string, string>();
   const moduleConfigFiles = new Map<string, string>();
+
+  const templateFolderPath = path.join(ctx.root, templatesFolder);
+  if (!(await fs.pathExists(path.join(templateFolderPath, bicepOrchestrationFileName)))) {
```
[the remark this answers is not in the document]
KennethBWSong: Updated, thx!

SLdragon (approved):

1yefuwang1 (approved):

PR #4607, opened 2022-04-11
feat: codelens to show real value

nliu-ms:
show real values for manifest template (local & dev)
image


nliu-ms:
E2E TEST: n/a

packages/vscode-extension/src/codeLensProvider.ts
```diff
@@ -98,16 +116,62 @@ export class ManifestTemplateCodeLensProvider implements vscode.CodeLensProvider
...
+
+        const localValue = getPropertyByPath(localEnvInfo, key);
+        title = `${title} ${environmentManager.getLocalEnvName()}: ${localValue}`;
+
+        const defaultValue = getPropertyByPath(defaultEnvInfo, key);
+        title = `${title}, ${defaultEnvInfo.envName}: ${defaultValue}`;
+
+        lens.command = {
+          title: title,
+          command: "fx-extension.openConfigState",
```
HuihuiWu-Microsoft, line 167: Does this feature show real value for config?  If so, should pass 'config' as type here.

HuihuiWu-Microsoft:
If we show local&default env values for codelens, will users know he can also click on the codelens to check values for other env?

nliu-ms:
> If we show local&default env values for codelens, will users know he can also click on the codelens to check values for other env?

We will add a hover to show values for other environments.

packages/vscode-extension/src/codeLensProvider.ts
nliu-ms: Updated, thanks!

HuihuiWu-Microsoft (approved):

PR #4609, opened 2022-04-11
chore(apicon): add custom auth sample

xzf0587:
Here is the custom auth provider sample code.

Will add other content after the sample code is confirmed.


packages/fx-core/tests/plugins/resource/apiconnector/expectedFiles/js/custom.js
```diff
@@ -0,0 +1,53 @@
...
+  AddAuthenticationInfo = async (config) => {
+    if (!config.headers) {
+      config.headers = {};
+    }
+    config.headers[this.customProperty] = this.customValue;
+    return config;
+  };
+}
+
+// Loads current app's configuration and use app for auth.
```
blackchoey, line 24: Let's remove `use app for auth` to avoid distracting users, since it's not AAD auth. Please also update the ts part

blackchoey (approved):

PR #4614, opened 2022-04-11
perf(cli): add 'add sso' for cli

KennethBWSong:
Will add e2e test in another pr.


Alive-Fish (dismissed):

packages/cli/src/cmds/add.ts
```diff
@@ -70,12 +71,65 @@ export class AddCICD extends YargsCommand {
...
+  public readonly commandHead = `sso`;
+  public readonly command = this.commandHead;
+  public readonly description = "Add Single Sign On for your project.";
+
+  public builder(yargs: Argv): Argv<any> {
+    this.params = HelpParamGenerator.getYargsParamForHelp("addSso");
+    return yargs.version(false).options(this.params);
+  }
+
+  public modifyArguments(args: { [argName: string]: any }): { [argName: string]: any } {
```
SLdragon, line 84: seems this function does nothing?

KennethBWSong: Followed code structure in addCICD above, not quite sure whether it is required.

SLdragon: You can ask @Alive-Fish whether it required, if not required, you can remove it

SLdragon (approved):

PR #4627, opened 2022-04-12
feat: add hover for manifest template

nliu-ms:
When user hover over the placeholder, it will show values from all environments, with the button to open state/config file.
image

### DIFF
--- a/packages/vscode-extension/src/hoverProvider.ts
+++ b/packages/vscode-extension/src/hoverProvider.ts
@@ -1,0 +1,75 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import * as vscode from "vscode";
+import { getPropertyByPath } from "@microsoft/teamsfx-core";
+import { manifestConfigDataRegex, manifestStateDataRegex } from "./constants";
+import { core, getSystemInputs } from "./handlers";
+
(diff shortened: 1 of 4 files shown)


nliu-ms:
E2E TEST: n/a

packages/vscode-extension/src/hoverProvider.ts
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  private async generateHoverMessage(key: string): Promise<vscode.MarkdownString> {
+    const inputs = getSystemInputs();
+    const getConfigRes = await core.getProjectConfigV3(inputs);
+    if (getConfigRes.isErr()) throw getConfigRes.error;
+    const projectConfigs = getConfigRes.value;
+
+    let message = "";
+    if (projectConfigs && projectConfigs.envInfos) {
```
HuihuiWu-Microsoft, line 51: If provision hasn't been done for one environment, will it show undefined and error pops up when user click view state file?

nliu-ms: yes, exactly. When user click view state file, and the environment has not been provisioned, it will show error like before.

HuihuiWu-Microsoft (approved):

PR #4637, opened 2022-04-12
chore: fix question model for CLI

a1exwang:
test
image
"

### DIFF
--- a/packages/fx-core/src/plugins/solution/fx-solution/question.ts
+++ b/packages/fx-core/src/plugins/solution/fx-solution/question.ts
@@ -40,9 +40,10 @@ export const NotificationOptionItem: OptionItem = {
 };
 
 export const CommandAndResponseOptionItem: OptionItem = {
-  id: "CommandBot",
+  // For default option, id and cliName must be the same
+  id: "command-bot",
   label: getLocalizedString("core.CommandAndResponseOption.label"),
-  cliName: "command-and-response",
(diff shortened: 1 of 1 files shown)


a1exwang:
E2E TEST: n/a

packages/fx-core/src/plugins/solution/fx-solution/question.ts
```diff
@@ -40,9 +40,10 @@ export const NotificationOptionItem: OptionItem = {
 };
 
 export const CommandAndResponseOptionItem: OptionItem = {
-  id: "CommandBot",
+  // For default option, id and cliName must be the same
+  id: "command-bot",
   label: getLocalizedString("core.CommandAndResponseOption.label"),
-  cliName: "command-and-response",
+  cliName: "command-bot",
```
kuojianlu, line 46: The capability name used in adding capability is `command-and-response`, better to keep consistency?

Alive-Fish (approved):

jayzhang (approved):

packages/fx-core/src/plugins/solution/fx-solution/question.ts
kuojianlu: Curious why here is still `command-and-response`, https://github.com/OfficeDev/TeamsFx/blob/dev/packages/cli/src/cmds/capability.ts#L302

PR #4638, opened 2022-04-12
fix: app studio api response type changed

nliu-ms:


nliu-ms:
E2E TEST: publish is disabled temporarily for account issue.

PR #4641, opened 2022-04-12
chore(apicon): cert and api-key auth

xzf0587:
Add cert and api-key sample code for preview.
Will add the rest content after the code is confirmed.


packages/fx-core/tests/plugins/resource/apiconnector/expectedFiles/ts/cert.ts
```diff
@@ -0,0 +1,34 @@
...
+  createApiClient,
+  CertificateAuthProvider,
+  createPemCertOption,
+} from "@microsoft/teamsfx";
+
+// Loads current app's configuration.
+const teamsFx = new TeamsFx();
+// Initializes a new axios instance to call fake API.
+const authProvider = new CertificateAuthProvider(
+  // Please replace '<your-cert>' with the actual cert chain in PEM format and
```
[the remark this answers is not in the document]
xzf0587: It is fixed. Thanks.

packages/fx-core/src/plugins/resource/apiconnector/plugin.ts
```diff
@@ -155,13 +164,14 @@ export class ApiConnectorImpl {
 
   private getAuthConfigFromInputs(inputs: Inputs): AuthConfig {
     let config: AuthConfig;
-    if (inputs[Constants.questionKey.apiType] === AuthType.BASIC) {
+    const apiType = inputs[Constants.questionKey.apiType];
+    if (apiType === AuthType.BASIC) {
```
blackchoey, line 168: is it possible to use switch to avoid multiple if else?

packages/fx-core/resource/package.nls.json
```diff
@@ -266,6 +269,8 @@
   "plugins.apiConnector.reuseAppOption.placeholder": "Use the AAD app registration created for current Teams app",
   "plugins.apiConnector.anotherAppOption.title": "Use another app registration",
   "plugins.apiConnector.anotherAppOption.placeholder": "Provide information for your AAD app registration",
+  "plugins.apiConnector.headerLocationOption.title": "Request header",
```
blackchoey, line 272: can we rename it to something like plugins.apiConnector.`requestHeaderOption`.title? the current name seems a little confusing since it's not location for header. Please also help see how to better describe `plugins.apiConnector.queryLocationOption.title`

packages/fx-core/src/plugins/resource/apiconnector/plugin.ts
```diff
@@ -368,4 +395,26 @@ export class ApiConnectorImpl {
...
+  public buildAPIKeyAuthQuestion(): QTreeNode {
+    const node = new QTreeNode({
+      name: Constants.questionKey.apiAPIKeyLocation,
+      type: "singleSelect",
+      staticOptions: [headerLocationOption, queryLocationOption],
+      title: getLocalizedString("plugins.apiConnector.getQuestion.apiKeyLocation.title"),
+    });
+    node.condition = { equals: APIKeyAuthOption.id };
+
+    const headerKeyNameQuestionNode = new QTreeNode(buildAPIKeyNameQuestion(KeyLocation.Header));
```
blackchoey, line 408: we may need to pass localized string `plugins.apiConnector.headerLocationOption.title` here, otherwise the text displayed to user will have non localized part

PR #4665, opened 2022-04-13
chore: update unify config doc

xiaolang124:
1. Add collaborate cautions and benefits to doc.
2. Add help link to upgrade question.

### DIFF
--- a/packages/fx-core/resource/unify-config-change-logs.md
+++ b/packages/fx-core/resource/unify-config-change-logs.md
@@ -2,11 +2,12 @@
 
 We have updated the configuration files so that your project is compatible with the latest Teams Toolkit features, including a consistent configuration file schema and a single Teams app manifest template across local and remote environments.
 
-# Collaborate cautions
-If you collaborate on this project with your co-workers, they need to update Teams toolkit extension to the latest version.
+> Important Notes: If you collaborate on this project with your co-workers, please ensure your team members update the Teams Toolkit extension to the latest version after committing the changes with this upgrade.
 
-## Know about the upgrade benefits
-We combine local and remote manifest configs into one style. You can configure once for bot local and remote environments.
+## Why upgrade
+Teams Toolkit continues to improve your Teams application development experience. We are upgrading the Teams app project structure so that you can:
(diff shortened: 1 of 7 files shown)


packages/fx-core/src/core/middleware/consolidateLocalRemote.ts
```diff
@@ -38,6 +38,8 @@ import * as util from "util";
 import { ManifestTemplate } from "../../plugins/resource/spfx/utils/constants";
 
 const upgradeButton = "Upgrade";
+const LearnMore = "Learn More";
+const LearnMoreLink = "https://github.com/OfficeDev/TeamsFx/wiki/Upgrade-to-support-unify-config";
```
kimizhu, line 42: Use aka link instead.

packages/fx-core/resource/unify-config-change-logs.md
```diff
@@ -2,6 +2,13 @@
 
 We have updated the configuration files so that your project is compatible with the latest Teams Toolkit features, including a consistent configuration file schema and a single Teams app manifest template across local and remote environments.
 
+> Important Notes: If you collaborate on this project with your co-workers, please ensure your team members update the Teams Toolkit extension to the latest version after committing the changes with this upgrade.
```
kimizhu, line 5: @MuyangAmigo can you help to review this?

packages/fx-core/src/core/middleware/consolidateLocalRemote.ts
xiaolang124: Already changed to ```https://aka.ms/teamsfx-unify-config-guide```.

packages/fx-core/resource/unify-config-change-logs.md
MuyangAmigo: This is what I have updated. Do you have any particular comments?

kimizhu: I saw it was different from what in the wiki page, so was wonder which one we should use. 
Now I see wiki page is also updated with this change, so no more question here. Thanks @MuyangAmigo and @xiaolang124

nliu-ms (approved):

PR #4670, opened 2022-04-14
feat(vsc): UI change for GA features

tecton:
1. Add configuration to enable GA features.
2. Use "Add features" to add capabilities and resources.

E2E TEST: N/A

WIT: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/14100028


jayzhang (approved):

packages/fx-core/src/common/featureFlags.ts
```diff
@@ -13,6 +13,20 @@ export function isFeatureFlagEnabled(featureFlagName: string, defaultValue = fal
   }
 }
 
+/**
+ * Update feature flags related to GA.
+ */
+export function initializeGAFeatureFlags(): void {
+  if (isFeatureFlagEnabled(FeatureFlagName.GAPreview, false)) {
```
HuihuiWu-Microsoft, line 20: What if user changed one of the feature flags after fxcore initialized?

packages/vscode-extension/package.json
```diff
@@ -463,6 +463,11 @@
         "title": "%teamstoolkit.commands.addCapability.title%",
         "enablement": "!fx-extension.isSPFx && !fx-extension.isM365"
       },
+      {
+        "command": "fx-extension.addFeature",
+        "title": "Teams: Add Feature",
```
HuihuiWu-Microsoft, line 468: Move to package.nls.json?

packages/vscode-extension/src/handlers.ts
```diff
@@ -624,6 +624,83 @@ export async function addCapabilityHandler(args?: any[]): Promise<Result<null, F
   return result;
 }
 
+export async function addFeatureHandler(args?: any[]): Promise<Result<null, FxError>> {
+  // TODO: move the question model to fx-core addFeature API
```
HuihuiWu-Microsoft, line 628: Send telemetry here?

packages/vscode-extension/src/handlers.ts
```diff
@@ -624,6 +624,83 @@ export async function addCapabilityHandler(args?: any[]): Promise<Result<null, F
...
+      },
+      {
+        id: "cicd",
+        label: `${localize("teamstoolkit.handlers.addFeature.cicd.label")}`,
+        detail: localize("teamstoolkit.handlers.addFeature.cicd.detail"),
+      },
+    ],
+  };
+  const stack = [addFeatureConfig];
+  while (stack.length > 0) {
```
HuihuiWu-Microsoft, line 674: Why not ask the questions in fx-core and leverage existing question model(maybe move these questions to getQuestionsForAddFeature())? You can also benefit from question model for other platforms.

packages/fx-core/src/common/featureFlags.ts
tecton: If user changes configuration in VSC, they are instructed to reload window to take effect. This function is exported and called by VSC and CLI during activation.

packages/vscode-extension/package.json
tecton: Thanks, will fix this.

packages/vscode-extension/src/handlers.ts
tecton: This function delegates call to other existing handlers and they have their own telemetry events. I'd like to add telemetry for "Add feature" once the API in core is ready.

packages/vscode-extension/src/handlers.ts
tecton: Yes, this logic will be moved to fx-core later. I will change to invoking user task after it's ready.

HuihuiWu-Microsoft (approved):

jayzhang (approved):

IvanJobs (approved):
lgtm for bot plugin

PR #4672, opened 2022-04-14
chore: update call function sample code

blackchoey:
The latest SDK can help create API client. Update sample code to use this feature to call function api.


wenytang-ms (dismissed):

hund030 (dismissed):

blackchoey:
I will not update M365 template for now since it requires a special SDK version which has different release cycle. We can update them later.

hund030 (approved):

PR #4684, opened 2022-04-14
fix: some string constants not affected by feature flag

a1exwang:
Constants are initialized before syncFeatureFlags() in vscode extension. So we cannot use feature flags for them. Move the logic into places where the constants are actually used.

Tested without environment variable and only vscode settings.
image


a1exwang:
E2E TEST: n/a

packages/fx-core/src/plugins/solution/fx-solution/question.ts
```diff
@@ -14,14 +14,17 @@ import { getLocalizedString } from "../../../common/localizeUtils";
...
-  detail: isBotNotificationEnabled()
-    ? getLocalizedString("core.TabOption.detailNew")
-    : getLocalizedString("core.TabOption.detail"),
+  description: getLocalizedString("core.TabOption.description"),
+  detail: getLocalizedString("core.TabOption.detail"),
+};
+
+export const TabNewUIOptionItem: OptionItem = {
+  id: "Tab",
+  label: "SSO enabled launch page",
```
tecton, line 25: Do these labels need to be localized?

a1exwang: Updated

tecton (approved):

1yefuwang1 (approved):

PR #4685, opened 2022-04-14
chore(apicon): add telemetry

xzf0587:
1. add sql telemetry
2. add api-connector telemetry
3. update api-connector env

### DIFF
--- a/packages/fx-core/src/plugins/resource/apiconnector/plugin.ts
+++ b/packages/fx-core/src/plugins/resource/apiconnector/plugin.ts
@@ -451,7 +451,7 @@ export class ApiConnectorImpl {
   public getTelemetryProperties(config: ApiConnectorConfiguration): { [key: string]: string } {
     const properties = {
       [Telemetry.properties.authType]: config.AuthConfig.AuthType.toString(),
-      [Telemetry.properties.componentPath]: config.ComponentPath.join(","),
+      [Telemetry.properties.componentType]: config.ComponentPath.join(","),
     };
 
     switch (config.AuthConfig.AuthType) {
--- a/packages/fx-core/src/plugins/resource/apiconnector/plugin.ts
+++ b/packages/fx-core/src/plugins/resource/apiconnector/plugin.ts
@@ -11,6 +11,7 @@ import {
   UserError,
   ok,
   Platform,
+  FxError,
 } from "@microsoft/teamsfx-api";
 import { Context } from "@microsoft/teamsfx-api/build/v2";
 import {
@@ -58,6 +59,7 @@ import { isAADEnabled } from "../../../common";
 import { getAzureSolutionSettings } from "../../solution/fx-solution/v2/utils";
 import { DepsHandler } from "./depsHandler";
 import { checkEmptySelect } from "./checker";
+import { Telemetry, TelemetryUtils } from "./telemetry";
 export class ApiConnectorImpl {
   public async scaffold(ctx: Context, inputs: Inputs): Promise<ApiConnectorResult> {
     if (!inputs.projectPath) {
@@ -69,6 +71,14 @@ export class ApiConnectorImpl {
     const projectPath = inputs.projectPath;
     const languageType: string = ctx.projectSetting!.programmingLanguage!;
     const config: ApiConnectorConfiguration = this.getUserDataFromInputs(inputs);
+
+    TelemetryUtils.init(ctx.telemetryReporter);
+    const telemetryProperties = this.exactTelemetryProperties(config);
+    TelemetryUtils.sendEvent(
+      Telemetry.stage.scaffold + Telemetry.startSuffix,
+      undefined,
+      telemetryProperties
+    );
     // backup relative files.
     const backupFolderName = generateTempFolder();
     await Promise.all(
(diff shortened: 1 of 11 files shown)


packages/fx-core/src/plugins/resource/apiconnector/plugin.ts
```diff
@@ -426,4 +443,26 @@ export class ApiConnectorImpl {
...
+    switch (config.AuthConfig.AuthType) {
+      case AuthType.AAD:
+        const aadAuthConfig = config.AuthConfig as AADAuthConfig;
+        properties[Telemetry.properties.reuseTeamsApp] = aadAuthConfig.ReuseTeamsApp
+          ? Telemetry.valueYes
+          : Telemetry.valueNo;
+        break;
+      case AuthType.APIKEY:
+        const authConfig = config.AuthConfig as APIKeyAuthConfig;
+        properties[Telemetry.properties.keyInHeader] =
```
blackchoey, line 460: can we just add a api key location property who's value might be query parameter or request header? keyInHeader = false may not be enough to tell people where is the key

packages/fx-core/src/plugins/resource/sql/plugin.ts
```diff
@@ -217,6 +219,15 @@ export class SqlPluginImpl {
   }
 
   public async generateArmTemplates(ctx: PluginContext): Promise<Result<any, FxError>> {
+    TelemetryUtils.init(ctx.telemetryReporter);
```
blackchoey, line 222: similar to another comment, it may be better to initialize telemetry util globally. Otherwise we may forgot init somewhere

packages/fx-core/src/plugins/resource/apiconnector/plugin.ts
```diff
@@ -58,6 +59,7 @@ import { isAADEnabled } from "../../../common";
 import { getAzureSolutionSettings } from "../../solution/fx-solution/v2/utils";
 import { DepsHandler } from "./depsHandler";
 import { checkEmptySelect } from "./checker";
+import { Telemetry, TelemetryUtils } from "./telemetry";
 export class ApiConnectorImpl {
   public async scaffold(ctx: Context, inputs: Inputs): Promise<ApiConnectorResult> {
     if (!inputs.projectPath) {
```
wenytang-ms, line 65: shall we send telemetry event on the start and end scaffold

packages/fx-core/src/plugins/resource/apiconnector/plugin.ts
```diff
@@ -69,6 +71,14 @@ export class ApiConnectorImpl {
...
     const languageType: string = ctx.projectSetting!.programmingLanguage!;
     const config: ApiConnectorConfiguration = this.getUserDataFromInputs(inputs);
+
+    TelemetryUtils.init(ctx.telemetryReporter);
+    const telemetryProperties = this.exactTelemetryProperties(config);
+    TelemetryUtils.sendEvent(
+      Telemetry.stage.scaffold + Telemetry.startSuffix,
+      undefined,
+      telemetryProperties
+    );
```
wenytang-ms, line 81: shall we send with inputs?

packages/fx-core/src/plugins/resource/apiconnector/plugin.ts
wenytang-ms, line 65: project path empty need to send telemetry?

packages/fx-core/src/plugins/resource/sql/plugin.ts
xzf0587: It is hard to initialize it as the plugin instance is created by core. As there is no context passed, we can't initialize telemetry reporter.
And the core can call every exposed method of plugin instance. We can only do this inside the method.

blackchoey: ok， thanks

packages/fx-core/src/plugins/resource/apiconnector/plugin.ts
```diff
@@ -112,21 +123,27 @@ export class ApiConnectorImpl {
...
-        throw err;
-      } else {
-        throw ResultFactory.SystemError(
+      if (!(err instanceof SystemError) && !(err instanceof UserError)) {
+        err = ResultFactory.SystemError(
           ErrorMessage.generateApiConFilesError.name,
           ErrorMessage.generateApiConFilesError.message(err.message)
         );
       }
+      const thrownErr = err as FxError;
```
wenytang-ms, line 132: can we extract line 132 -136 as a util function?

xzf0587: It is fixed. Thanks.

wenytang-ms (approved):
approve for api connector

packages/fx-core/src/plugins/resource/apiconnector/plugin.ts
```diff
@@ -444,10 +448,12 @@ export class ApiConnectorImpl {
     return node;
   }
 
-  public exactTelemetryProperties(config: ApiConnectorConfiguration): { [key: string]: string } {
+  public getTelemetryProperties(config: ApiConnectorConfiguration): { [key: string]: string } {
     const properties = {
       [Telemetry.properties.authType]: config.AuthConfig.AuthType.toString(),
+      [Telemetry.properties.componentPath]: config.ComponentPath.join(","),
```
blackchoey, line 454: let's remove the path, since it may contain sensitive info

blackchoey: we can add component type (bot/api) instead, since they're constant defined by us

xzf0587: Here the path is api/bot, not the real file path.
Will update the property key to avoid confuse.

blackchoey (approved):

PR #4688, opened 2022-04-15
fix: change the log level to reduce num of logs

nliu-ms:
1) hover only works for isConfigUnifyEnabled()
2) change the log level to reduce num of logs


nliu-ms:
E2E TEST: n/a

jayzhang (approved):

HuihuiWu-Microsoft (approved):

PR #4699, opened 2022-04-15
perf: update msg

wenytang-ms:


packages/fx-core/src/plugins/resource/apiconnector/errors.ts
```diff
@@ -50,29 +59,29 @@ export class ErrorMessage {
...
-      getLocalizedString(`error.apiConnector.${ErrorMessage.NoValidCompoentExistError.name}`),
+    message: (): [string, string] =>
+      ErrorMessage.getMessages(`error.apiConnector.${ErrorMessage.NoValidCompoentExistError.name}`),
   };
 
   public static readonly NoActivePluginsExistError = {
     name: "NoActivePluginsExistError",
-    message: () =>
-      getLocalizedString(`error.apiConnector.${ErrorMessage.NoActivePluginsExistError.name}`),
+    message: (): [string, string] =>
```
blackchoey, line 68: can we make the return type an object? maybe something like:
```
{
  localizedErrorMessage: 'xxx',
  defaultErrorMessage: 'yyy'
}
```

This would make the code more clear and avoid possible errors

blackchoey (dismissed):

packages/fx-core/src/plugins/resource/apiconnector/errors.ts
wenytang-ms: updated

blackchoey (approved):

PR #4703, opened 2022-04-15
chore: update consolidate telemetry and frontend debug

xiaolang124:
1.Update consolidate telemetry.
2.Add ```localEnvs.HTTPS=true``` when it is undefined.

### DIFF
--- a/packages/fx-core/src/common/local/localStateHelper.ts
+++ b/packages/fx-core/src/common/local/localStateHelper.ts
@@ -91,6 +91,9 @@ export async function convertToLocalEnvs(
     localEnvs[LocalEnvFrontendKeys.Https] = frontendConfigs?.get(
       LocalStateFrontendKeys.Https
     ) as string;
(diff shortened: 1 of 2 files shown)


packages/fx-core/src/common/local/localStateHelper.ts
```diff
@@ -91,6 +91,9 @@ export async function convertToLocalEnvs(
     localEnvs[LocalEnvFrontendKeys.Https] = frontendConfigs?.get(
       LocalStateFrontendKeys.Https
     ) as string;
+    if (!localEnvs[LocalEnvFrontendKeys.Https]) {
```
kuojianlu, line 94: In what case will `https` be `undefined`?

xiaolang124: In some samples, like ```hello world tab withbackend```. Normally this will not happen.

kuojianlu (approved):

nliu-ms (approved):

LongOddCode (approved):

PR #4707, opened 2022-04-15
chore: improve wordings for CLI

a1exwang:
Display notification host type in `option.label` in CLI because CLI does not have `option.description`.
image


a1exwang:
E2E TEST: n/a covered by UT

Siglud (approved):

PR #4713, opened 2022-04-15
perf: api connection type

wenytang-ms:


packages/fx-core/src/plugins/resource/apiconnector/depsHandler.ts
```diff
@@ -48,6 +49,16 @@ export class DepsHandler {
     }
     if (needUpdate) {
       await fs.writeFile(localPkgPath, JSON.stringify(pkgContent, null, 4));
+      const telemetryProperties = {
+        component: this.componentType,
+        dependency: JSON.stringify(pkgConfig),
```
blackchoey, line 54: may be we do not need to record this? since the pkgConfig is fixed

packages/fx-core/src/plugins/resource/apiconnector/depsHandler.ts
```diff
@@ -48,6 +49,16 @@ export class DepsHandler {
...
+      };
+
+      TelemetryUtils.sendEvent(
```
blackchoey, line 57: can we name the event to something like `update-package`? `scaffold-installed` is not so straightforward for people to understand what's the meaning.

packages/fx-core/tests/plugins/resource/apiconnector/scaffold.test.ts
```diff
@@ -46,6 +46,29 @@ describe("Api Connector scaffold sample code", async () => {
...
+    const fakeInputs: Inputs = { ...inputs, ...expectInputs };
+    const apiConnector: ApiConnectorImpl = new ApiConnectorImpl();
+    try {
+      await apiConnector.scaffold(context, fakeInputs);
+    } catch (err) {
+      expect(err instanceof UserError).to.be.true;
+      chai.assert.strictEqual(err.source, "api-connector");
+      chai.assert.strictEqual(
+        err.displayMessage,
+        "Component fx-resource-function not exist, please add first"
```
blackchoey, line 68: can we should "api" instead of "fx-resource-function" in display message? It would be easier for users to understand the message.

blackchoey (approved):

blackchoey (approved):

PR #4720, opened 2022-04-18
fix(bot): use async deployment api

XiaofuHuang:
A limit of app service is that the deploy API will response 500 (Request Timed Out) after approx 230s. Now the first-time deployment may spend about 8 mins. Our solution is to use the [async zip deployment API]((https://github.com/projectkudu/kudu/wiki/Deploying-from-a-zip-file-or-url)) to check the deployment status.

Detail: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/14103329/

### DIFF
--- a/packages/fx-core/src/plugins/resource/bot/azureOps.ts
+++ b/packages/fx-core/src/plugins/resource/bot/azureOps.ts
@@ -10,10 +10,14 @@ import {
   ZipDeployError,
   MessageEndpointUpdatingError,
   RestartWebAppError,
+  DeployStatusError,
+  DeployTimeoutError,
 } from "./errors";
 import { CommonStrings, ConfigNames } from "./resources/strings";
 import * as utils from "./utils/common";
 import { default as axios } from "axios";
+import { waitSeconds } from "../../../common";
+import { DeployStatus } from "./constants";
 
 export class AzureOperations {
   public static async UpdateBotChannelRegistration(
@@ -132,7 +136,7 @@ export class AzureOperations {
     zipDeployEndpoint: string,
     zipBuffer: Buffer,
     config: any
-  ): Promise<void> {
+  ): Promise<string> {
     let res = undefined;
     try {
       res = await axios.post(zipDeployEndpoint, zipBuffer, config);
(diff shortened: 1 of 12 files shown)


XiaofuHuang:
E2E TEST: https://github.com/OfficeDev/TeamsFx/blob/82fc22ff53ad04bdb40ee59fb06346fbf35415a5/packages/cli/tests/e2e/bot/ProvisionCommandAndResponseBot.tests.ts#L27

hund030 (approved):
LGTM. Thanks for the improvement. 
BTW, I think we need to investigate how to improve the performance of bot deployment.

packages/fx-core/src/plugins/resource/bot/azureOps.ts
```diff
@@ -132,17 +136,42 @@ export class AzureOperations {
...
     }
+
+    return res.headers.location;
+  }
+
+  public static async CheckDeployStatus(location: string, config: any): Promise<void> {
+    let res = undefined;
+    for (let i = 0; i < DeployStatus.RETRY_TIMES; ++i) {
+      try {
+        res = await axios.get(location, config);
```
swatDong, line 158: Is this `axios` the default instance?
Creating an axios instance for deployment only to avoid unexpected default settings (auth/header/proxy...).

packages/fx-core/src/plugins/resource/bot/constants.ts
```diff
@@ -146,6 +146,11 @@ export class Retry {
   public static readonly BACKOFF_TIME_MS = 5000;
 }
 
+export class DeployStatus {
+  public static readonly RETRY_TIMES = 60;
```
swatDong, line 150: According your experienced behavior, the longest deploy time is 8 min, so better to have more buffer here - let the whole retry be 2*8 = 16 min.

swatDong: Or, is there any hard timeout line from Kudu?

Siglud (approved):

swatDong (approved):
The code logic looks good to me.
Consider comments as further improvements.

PR #4724, opened 2022-04-18
chore(sql): update sql message

xzf0587:
1.add detailed error message


blackchoey (approved):
I may suggest use an object for the error message array, since it's hard to understand what's the meaning of each item in the array. We can refractor this later.

PR #4725, opened 2022-04-18
chore: fix vscode terminal open login page break

xiaolang124:
1.Fix open login page in vscode terminal.

### DIFF
--- a/packages/cli/src/commonlib/codeFlowLogin.ts
+++ b/packages/cli/src/commonlib/codeFlowLogin.ts
@@ -181,6 +181,7 @@ export class CodeFlowLogin {
     try {
       await this.startServer(server, serverPort!);
       this.pca!.getAuthCodeUrl(authCodeUrlParameters).then(async (url: string) => {
+        url += "&";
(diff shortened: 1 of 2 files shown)


xiaolang124:
Fix: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/14124941/

packages/cli/src/commonlib/codeFlowLogin.ts
```diff
@@ -181,6 +181,7 @@ export class CodeFlowLogin {
     try {
       await this.startServer(server, serverPort!);
       this.pca!.getAuthCodeUrl(authCodeUrlParameters).then(async (url: string) => {
+        url += "&";
```
swatDong, line 184: Docs `#` work as well?
For security, `#` is better than `&` since parameters after `#` are not sent to server.

xiaolang124: Thanks for the advice, changed to ‘#’.

swatDong (approved):

PR #4729, opened 2022-04-18
fix(bot): set deploy timeout to 20 min

XiaofuHuang:
fix the comments in PR: https://github.com/OfficeDev/TeamsFx/pull/4720
If `always on` is enabled, the deployment will never be timeout. 
If `always on` is closed, the deployment will be timeout after 20 min. https://github.com/projectkudu/kudu/wiki/Investigating-issues#deployment-process-got-terminated-due-to-idle-timeout

### DIFF
--- a/packages/fx-core/src/plugins/resource/bot/azureOps.ts
+++ b/packages/fx-core/src/plugins/resource/bot/azureOps.ts
@@ -20,6 +20,8 @@ import { waitSeconds } from "../../../common";
 import { DeployStatus } from "./constants";
 
 export class AzureOperations {
+  private static readonly axiosInstance = axios.create();
+
   public static async UpdateBotChannelRegistration(
     botClient: AzureBotService,
     resourceGroup: string,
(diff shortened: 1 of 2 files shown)


XiaofuHuang:
E2E TEST: https://github.com/OfficeDev/TeamsFx/blob/82fc22ff53ad04bdb40ee59fb06346fbf35415a5/packages/cli/tests/e2e/bot/ProvisionCommandAndResponseBot.tests.ts#L27

packages/fx-core/src/plugins/resource/bot/azureOps.ts
```diff
@@ -137,9 +137,10 @@ export class AzureOperations {
     zipBuffer: Buffer,
     config: any
   ): Promise<string> {
+    const axiosInstance = axios.create();
```
swatDong, line 140: Could this be a static one instead of creating every deploy?

XiaofuHuang: Got it.

swatDong (approved):

Siglud (approved):

PR #4744, opened 2022-04-19
perf: update msg

wenytang-ms:
https://dev.azure.com/msazure/Microsoft%20Teams%20Extensibility/_workitems/edit/14171495

fx-core already sendTelemetryEvent once, and showErrorMsg, no need to do again on extension handler

### DIFF
--- a/packages/vscode-extension/src/handlers.ts
+++ b/packages/vscode-extension/src/handlers.ts
@@ -716,8 +716,6 @@ export async function connectExistingApiHandler(args?: any[]): Promise<Result<nu
 
   const res = await runUserTask(func, TelemetryEvent.ConnectExistingApi, true);
   if (!res.isOk()) {
-    showError(res.error);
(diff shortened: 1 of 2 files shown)


packages/vscode-extension/src/handlers.ts
```diff
@@ -716,8 +716,6 @@ export async function connectExistingApiHandler(args?: any[]): Promise<Result<nu
 
   const res = await runUserTask(func, TelemetryEvent.ConnectExistingApi, true);
   if (!res.isOk()) {
-    showError(res.error);
-    ExtTelemetry.sendTelemetryErrorEvent(TelemetryEvent.ConnectExistingApi, res.error);
```
blackchoey, line 720: Can we keep the telemetry here? since it makes us easier to query our telemetries.

blackchoey (approved):

packages/vscode-extension/src/handlers.ts
wenytang-ms: this code for vscode, vscode already done this job.

tecton (approved):

PR #4750, opened 2022-04-19
fix: add @types/node hint

swatDong:
Fix https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/14171601

There are build failures when using @types/fs-extra 9 and @types/node <= 12.
Force command/notification templates to use @types/node 14 to avoid such errors. (Here uses 14 since our toolkit recommends Node 14)


swatDong:
E2E TEST: Covered by UI test

dooriya (approved):

Siglud (approved):

PR #4751, opened 2022-04-19
chore: update m365 check tenant message and add local config when not exist

xiaolang124:
1.Update check m365 tenant message to support unify config.
2.Add local config when not exist.


xiaolang124:
Fix:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/14142176
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/14171519

packages/fx-core/src/core/environment.ts
```diff
@@ -82,9 +86,28 @@ class EnvironmentManager {
     }
 
     envName = envName ?? this.getDefaultEnvName();
-    const configResult = await this.loadEnvConfig(projectPath, envName);
+    let configResult = await this.loadEnvConfig(projectPath, envName);
     if (configResult.isErr()) {
-      return err(configResult.error);
+      if (envName === "local") {
```
HuihuiWu-Microsoft, line 91: Maybe only generate the local config file when it doesn't exist, or user customized configs will be lost.

xiaolang124: Move to check local config file location.

HuihuiWu-Microsoft (dismissed):
LGTM

HuihuiWu-Microsoft (approved):

nliu-ms (approved):

PR #4754, opened 2022-04-19
chore: check cert only for tabs

meifans:
Fix [bug](https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/14124691)

### DIFF
--- a/packages/vscode-extension/src/debug/prerequisitesHandler.ts
+++ b/packages/vscode-extension/src/debug/prerequisitesHandler.ts
@@ -257,12 +257,14 @@ export async function checkAndInstall(): Promise<Result<any, FxError>> {
     checkResults.push(accountResult);
 
     // local cert
-    const localCertResult = await resolveLocalCertificate(
-      localEnvManager,
-      `(${currentStep++}/${totalSteps})`
-    );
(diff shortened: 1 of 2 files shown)


packages/vscode-extension/src/debug/prerequisitesHandler.ts
```diff
@@ -257,12 +257,14 @@ export async function checkAndInstall(): Promise<Result<any, FxError>> {
     checkResults.push(accountResult);
 
     // local cert
-    const localCertResult = await resolveLocalCertificate(
-      localEnvManager,
-      `(${currentStep++}/${totalSteps})`
-    );
-    await progressHelper.end(Checker.LocalCertificate);
-    checkResults.push(localCertResult);
+    if (ProjectSettingsHelper.includeFrontend(projectSettings)) {
```
swatDong, line 260: Use `checkers.includes(Checker.LocalCertificate)` to keep the condition only one source of truth.

meifans: Added

swatDong (approved):

PR #4759, opened 2022-04-19
perf(aad-manifest): fix bot endpoint not correct issue

SLdragon:


packages/fx-core/src/plugins/resource/aad/utils/configs.ts
```diff
@@ -537,7 +537,13 @@ export class SetApplicationInContextConfig {
     if (isAadManifestEnabled()) {
       ConfigUtils.checkAndSaveConfig(ctx, ConfigKeys.botId, botId);
```
KennethBWSong, line 538: Also check whether botid exists?

KennethBWSong (approved):

PR #4766, opened 2022-04-20
fix(notification): check azurite port (10000, 10001, 10002)

XiaofuHuang:
Detail: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/14135370/


XiaofuHuang:
E2E TEST: none

swatDong (changes requested):
Azurite is not hard bound with our toolkit, so prefer to update Doc/FAQ, not hardcode azurite ports.

PR #4772, opened 2022-04-20
perf(aad-manifest): partially disable aad manifest json intellisense

SLdragon:
Fix this bug: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_sprints/taskboard/AuthAndData/Microsoft%20Teams%20Extensibility/CY22-4.2?workitem=14171623

Due to limitation of VSCode, we can only partially disable intellisense using empty JSON schema, see this issue for details: https://github.com/microsoft/vscode/issues/63270

Use empty JSON schema in VSCode settings.json file:
```
{
    "json.schemas": [
        {
            "fileMatch": [
                "/aad.*.json"
            ],
            "schema": {}
        }
    ]
}
```

### DIFF
--- a/packages/fx-core/src/plugins/solution/fx-solution/debug/scaffolding.ts
+++ b/packages/fx-core/src/plugins/solution/fx-solution/debug/scaffolding.ts
@@ -115,15 +115,6 @@ export async function _scaffoldLocalDebugSettings(
             EOL: os.EOL,
           }
         );
-
-        await fs.writeJSON(
-          `${inputs.projectPath}/.vscode/settings.json`,
-          Settings.generateSettings(false),
-          {
-            spaces: 4,
-            EOL: os.EOL,
-          }
-        );
       } else {
         const launchConfigurations = isM365
           ? LaunchNext.generateM365Configurations(includeFrontend, includeBackend, includeBot)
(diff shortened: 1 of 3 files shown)


KennethBWSong (dismissed):

swatDong (dismissed):

packages/fx-core/src/plugins/solution/fx-solution/debug/scaffolding.ts
```diff
@@ -115,15 +115,6 @@ export async function _scaffoldLocalDebugSettings(
             EOL: os.EOL,
           }
         );
-
-        await fs.writeJSON(
```
blackchoey, line 119: seems we only updated local debug's logic. is there any other logic that generates vscode settings?

blackchoey (approved):

blackchoey:
By the way, can we add a UT for this change?

packages/fx-core/src/plugins/solution/fx-solution/debug/scaffolding.ts
SLdragon: Nope, the generate vscode settings.json file only exist in local debug scaffold stage

SLdragon:
> By the way, can we add a UT for this change?

Thanks, will update UT in this PR

swatDong (approved):

KennethBWSong (approved):

blackchoey (approved):

PR #4774, opened 2022-04-20
chore: add bot troubleshooting

IvanJobs:
Show message and action button for bot trouble shooting after deploy.

https://dev.azure.com/msazure/Microsoft%20Teams%20Extensibility/_workitems/edit/13321566
WeChat Screenshot_20220420120058
WeChat Screenshot_20220420120035

### DIFF
--- a/packages/fx-core/src/plugins/solution/fx-solution/v2/deploy.ts
+++ b/packages/fx-core/src/plugins/solution/fx-solution/v2/deploy.ts
@@ -37,10 +37,9 @@ import { executeConcurrently, NamedThunk } from "./executor";
 import {
   extractSolutionInputs,
   getAzureSolutionSettings,
-  getBotTroubleShootMessages,
+  getBotTroubleShootMessage,
   getSelectedPlugins,
-  isAzureProject,
-  isBotProject,
+  isAzureProject
 } from "./utils";
 
 export async function deploy(
@@ -55,8 +54,7 @@ export async function deploy(
   });
   const provisionOutputs: Json = envInfo.state;
   const inAzureProject = isAzureProject(getAzureSolutionSettings(ctx));
-  const inBotProject = isBotProject(getAzureSolutionSettings(ctx));
-  const botTroubleShootMsg = getBotTroubleShootMessages(getAzureSolutionSettings(ctx));
+  const botTroubleShootMsg = getBotTroubleShootMessage(getAzureSolutionSettings(ctx));
   const provisioned =
     (provisionOutputs[GLOBAL_CONFIG][SOLUTION_PROVISION_SUCCEEDED] as boolean) ||
     inputs[Constants.DEPLOY_AAD_FROM_CODELENS] === "yes";
@@ -207,7 +205,7 @@ export async function deploy(
         getLocalizedString("core.deploy.successNotice", ctx.projectSetting.appName) +
         botTroubleShootMsg.textForLogging;
       ctx.logProvider.info(msg);
-      if (inBotProject) {
+      if (botTroubleShootMsg.textForLogging) {
         // Show a `Learn more` action button for bot trouble shooting.
         ctx.userInteraction
           .showMessage(
--- a/packages/fx-core/src/plugins/solution/fx-solution/v2/deploy.ts
+++ b/packages/fx-core/src/plugins/solution/fx-solution/v2/deploy.ts
@@ -39,7 +39,7 @@ import {
   getAzureSolutionSettings,
   getSelectedPlugins,
   isAzureProject,
-  IsBotProject,
+  isBotProject,
 } from "./utils";
 
 export async function deploy(
@@ -54,7 +54,7 @@ export async function deploy(
   });
   const provisionOutputs: Json = envInfo.state;
   const inAzureProject = isAzureProject(getAzureSolutionSettings(ctx));
-  const inBotProject = IsBotProject(getAzureSolutionSettings(ctx));
+  const inBotProject = isBotProject(getAzureSolutionSettings(ctx));
   const provisioned =
     (provisionOutputs[GLOBAL_CONFIG][SOLUTION_PROVISION_SUCCEEDED] as boolean) ||
     inputs[Constants.DEPLOY_AAD_FROM_CODELENS] === "yes";
@@ -203,18 +203,16 @@ export async function deploy(
     "https://aka.ms/teamsfx-bot-help#how-can-i-troubleshoot-issues-when-teams-bot-isnt-responding-on-azure";
   const botTroubleShootDesc = getLocalizedString("core.deploy.botTroubleShoot");
   const botTroubleShootLearnMore = getLocalizedString("core.deploy.botTroubleShoot.learnMore");
+  const botTroubleShootMsg = `${botTroubleShootDesc} ${botTroubleShootLearnMore}: ${botTroubleShootLink}.`;
   if (result.kind === "success") {
     if (inAzureProject) {
-      let msg = getLocalizedString("core.deploy.successNotice", ctx.projectSetting.appName);
+      const msg = getLocalizedString("core.deploy.successNotice", ctx.projectSetting.appName);
       // Append a new sentence for bot trouble shootting.
-      if (inBotProject) {
-        msg += ` ${botTroubleShootDesc}`;
-      }
-      ctx.logProvider.info(msg + ` ${botTroubleShootLearnMore}: ${botTroubleShootLink}.`);
+      ctx.logProvider.info(inBotProject ? `${msg} ${botTroubleShootMsg}` : msg);
       if (inBotProject) {
         // Show a `Learn more` action button for bot trouble shooting.
         ctx.userInteraction
-          .showMessage("info", msg, false, botTroubleShootLearnMore)
+          .showMessage("info", `${msg} ${botTroubleShootDesc}`, false, botTroubleShootLearnMore)
           .then((result) => {
             const userSelected = result.isOk() ? result.value : undefined;
             if (userSelected === botTroubleShootLearnMore) {
@@ -227,11 +225,8 @@ export async function deploy(
     }
     return ok(Void);
   } else {
-    let msg = getLocalizedString("core.deploy.failNotice", ctx.projectSetting.appName);
-    if (inBotProject) {
-      msg += ` ${botTroubleShootDesc} ${botTroubleShootLearnMore}: ${botTroubleShootLink}.`;
-    }
-    ctx.logProvider.info(msg);
+    const msg = getLocalizedString("core.deploy.failNotice", ctx.projectSetting.appName);
+    ctx.logProvider.info(inBotProject ? `${msg} ${botTroubleShootMsg}` : msg);
     return err(
       sendErrorTelemetryThenReturnError(
         SolutionTelemetryEvent.Deploy,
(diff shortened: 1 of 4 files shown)


MuyangAmigo (approved):
LGTM

packages/fx-core/src/plugins/solution/fx-solution/v2/deploy.ts
```diff
@@ -197,15 +199,38 @@ export async function deploy(
...
+            }
+          });
+      } else {
+        ctx.userInteraction.showMessage("info", msg, false);
+      }
     }
     return ok(Void);
   } else {
-    const msg = getLocalizedString("core.deploy.failNotice", ctx.projectSetting.appName);
+    let msg = getLocalizedString("core.deploy.failNotice", ctx.projectSetting.appName);
```
eriolchan, line 230: I am a little confused about the user experience, why it doesn't show dialog when deploy fails?

packages/fx-core/src/plugins/solution/fx-solution/v2/deploy.ts
```diff
@@ -197,15 +199,38 @@ export async function deploy(
...
     if (inAzureProject) {
-      const msg = getLocalizedString("core.deploy.successNotice", ctx.projectSetting.appName);
-      ctx.logProvider.info(msg);
-      ctx.userInteraction.showMessage("info", msg, false);
+      let msg = getLocalizedString("core.deploy.successNotice", ctx.projectSetting.appName);
+      // Append a new sentence for bot trouble shootting.
+      if (inBotProject) {
+        msg += ` ${botTroubleShootDesc}`;
+      }
+      ctx.logProvider.info(msg + ` ${botTroubleShootLearnMore}: ${botTroubleShootLink}.`);
```
eriolchan, line 213: What about just msg + link? why don't add Lean more message again even when it already has troubleshoot desc?

packages/fx-core/src/plugins/solution/fx-solution/v2/deploy.ts
```diff
@@ -197,15 +199,38 @@ export async function deploy(
...
+      if (inBotProject) {
+        // Show a `Learn more` action button for bot trouble shooting.
+        ctx.userInteraction
```
eriolchan, line 216: Is it possible to only add button in this step and then call showMessage together?
e.g.
```
if (!botMsg) {
  // add bot button
}
ctx.userInteraction.showMessage("info", msg, false);
```

packages/fx-core/src/plugins/solution/fx-solution/v2/deploy.ts
```diff
@@ -197,15 +199,38 @@ export async function deploy(
   ctx.logProvider.info(getLocalizedString("core.deploy.startNotice", PluginDisplayName.Solution));
   const result = await executeConcurrently(thunks, ctx.logProvider);
 
+  const botTroubleShootLink =
```
eriolchan, line 202: It looks not easy to maintain bot code. Is it possible to have a method to formalize message for Bot, so that other logic keep the same. e.g.
```
const botMsg: string = getBotTroubleShootMsg(); <= if it is not bot project, the value is empty string;
```
Then in L208, it can simply be 
```
const msg = getLocalizedString("core.deploy.successNotice", ctx.projectSetting.appName) + botMsg;
```
Then you can use if !botMsg to decide if it is a bot project and add additional button.

packages/fx-core/src/plugins/solution/fx-solution/v2/deploy.ts
```diff
@@ -39,6 +39,7 @@ import {
   getAzureSolutionSettings,
   getSelectedPlugins,
   isAzureProject,
+  IsBotProject,
```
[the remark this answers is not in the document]
IvanJobs: Good catch.

packages/fx-core/src/plugins/solution/fx-solution/v2/deploy.ts
IvanJobs: Although I'm not the owner to answer the question, I guess a specific error will popped up.

packages/fx-core/src/plugins/solution/fx-solution/v2/utils.ts
```diff
@@ -61,6 +61,14 @@ export function isAzureProject(azureSettings: AzureSolutionSettings | undefined)
   return azureSettings !== undefined && HostTypeOptionAzure.id === azureSettings.hostType;
 }
 
+export function IsBotProject(azureSettings: AzureSolutionSettings | undefined): boolean {
+  return (
+    azureSettings !== undefined &&
+    (azureSettings.capabilities.includes(BotOptionItem.id) ||
```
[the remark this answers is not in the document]
IvanJobs: Yes, the condition covers command/notification bot because they reused the same cap here.

packages/fx-core/src/plugins/solution/fx-solution/v2/deploy.ts
IvanJobs: As you can see from the screenshots listed in desc, ${botTroubleShootLearnMore} is just the text of `Learn more` which is not duplicated. `Learn more` is placed ahead to guide the real link which is kind of mapping the content of `Learn more` button in the UI flow.

packages/fx-core/src/plugins/solution/fx-solution/v2/deploy.ts
IvanJobs: It seems not possible because of the API contract of `showMessage` as you can see how it apply the `Learn more` clicking action.

packages/fx-core/src/plugins/solution/fx-solution/v2/deploy.ts
IvanJobs: Good idea, but not that easy. Since the message displayed in Teams Toolkit is slightly different from the one in output channel, I mean, in logging, we should print `Learn more: https://xxx`, but in UI we don't have to.
It's not just about the flag of isBotProject, but also the output channel of messages.

IvanJobs: Thanks for pointing this out so that reminding me to do a bit of refactor to improve readability.

jayzhang (approved):

packages/fx-core/src/plugins/solution/fx-solution/v2/deploy.ts
```diff
@@ -199,13 +203,36 @@ export async function deploy(
...
+        getLocalizedString("core.deploy.successNotice", ctx.projectSetting.appName) +
+        botTroubleShootMsg.textForLogging;
       ctx.logProvider.info(msg);
-      ctx.userInteraction.showMessage("info", msg, false);
+      if (inBotProject) {
+        // Show a `Learn more` action button for bot trouble shooting.
+        ctx.userInteraction
+          .showMessage(
+            "info",
+            `${getLocalizedString("core.deploy.successNotice", ctx.projectSetting.appName)} ${
```
eriolchan, line 215: Is the value of L215 the same as msg?

packages/fx-core/src/plugins/solution/fx-solution/v2/utils.ts
```diff
@@ -61,6 +61,38 @@ export function isAzureProject(azureSettings: AzureSolutionSettings | undefined)
...
 
+export function isBotProject(azureSettings: AzureSolutionSettings | undefined): boolean {
+  return (
+    azureSettings !== undefined &&
+    (azureSettings.capabilities?.includes(BotOptionItem.id) ||
+      azureSettings.capabilities?.includes(MessageExtensionItem.id))
+  );
+}
+
+export interface IBotTroubleShootMessage {
```
eriolchan, line 72: It is a good idea to have this object to hold all message.

eriolchan: The interface name could be BotTroubleShootMessage and we don't need to add I for it because Iinterface is only the convention for C#

packages/fx-core/src/plugins/solution/fx-solution/v2/utils.ts
```diff
@@ -61,6 +61,38 @@ export function isAzureProject(azureSettings: AzureSolutionSettings | undefined)
...
+  troubleShootLink: string;
+  textForLogging: string;
+  textForMsgBox: string;
+  textForActionButton: string;
+}
+
+export function getBotTroubleShootMessages(
+  azureSettings: AzureSolutionSettings | undefined
+): IBotTroubleShootMessage {
```
eriolchan, line 81: Could we simply return BotTroubleShootMessage | undefined for this method, so that we don't need to expose isBotProject() to external

packages/fx-core/src/plugins/solution/fx-solution/v2/utils.ts
```diff
@@ -61,6 +61,38 @@ export function isAzureProject(azureSettings: AzureSolutionSettings | undefined)
...
+}
+
+export interface IBotTroubleShootMessage {
+  troubleShootLink: string;
+  textForLogging: string;
+  textForMsgBox: string;
+  textForActionButton: string;
+}
+
+export function getBotTroubleShootMessages(
```
eriolchan, line 79: Not sure if this need to be => getBotTroubleShootMessage without "s"?

packages/fx-core/src/plugins/solution/fx-solution/v2/deploy.ts
IvanJobs: Not the same - msg contains `Learn more: <link>`, but this line should not.

packages/fx-core/src/plugins/solution/fx-solution/v2/utils.ts
IvanJobs: Agree, thanks for reminding. I will align this with TypeScript.

packages/fx-core/src/plugins/solution/fx-solution/v2/utils.ts
IvanJobs: Good idea. It potentially will make it simpler.

packages/fx-core/src/plugins/solution/fx-solution/v2/utils.ts
IvanJobs: Agree.

packages/fx-core/src/plugins/solution/fx-solution/v2/deploy.ts
```diff
@@ -199,13 +203,36 @@ export async function deploy(
 
   if (result.kind === "success") {
     if (inAzureProject) {
-      const msg = getLocalizedString("core.deploy.successNotice", ctx.projectSetting.appName);
+      const msg =
+        getLocalizedString("core.deploy.successNotice", ctx.projectSetting.appName) +
+        botTroubleShootMsg.textForLogging;
       ctx.logProvider.info(msg);
-      ctx.userInteraction.showMessage("info", msg, false);
+      if (inBotProject) {
```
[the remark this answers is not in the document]
IvanJobs: Yes, exactly we should use `if (botTroubleShootMsg.textForLogging)` without `!`. Will update.

packages/fx-core/src/plugins/solution/fx-solution/v2/utils.ts
IvanJobs: After a second thought, it seems we can't do this. In the `undefined` case, we can't get the msg.textForLogging so will break the logic in `deploy.ts`.

eriolchan (approved):

PR #4783, opened 2022-04-20
fix(adaptivecards-tools): change @type/react from devDependency to dependency

yukun-dong:
Some of the declaration files in adaptivecards-tools has error. This can be fixed by changing @type/react from devDependency to dependency.


yukun-dong:
E2E TEST: n/a

eriolchan (approved):

PR #4796, opened 2022-04-21
fix: update wording and log

nliu-ms:
1) Context menu and command:
image
image

2) Deloy message
image

3) Enhance log, app id in error


nliu-ms:
E2E TEST: n/a

HuihuiWu-Microsoft (approved):

PR #4803, opened 2022-04-22
feat: add customized icon for capability list

HuihuiWu-Microsoft:
Add customized icon for capability quick pick items 

UX design:
https://www.figma.com/file/qqVegbnWWBsqHhR8Qijs4O/Teams-Toolkit---GA?node-id=3006%3A262164

Screenshot:
customized_icons

E2E TEST: NA


HuihuiWu-Microsoft:
@jaeyonglee8687 ,I download the svgs from Figama and the ratio of these svgs are not the same. Not sure if this is designed, you  can check the screenshot for visual effects.

packages/fx-core/src/plugins/solution/fx-solution/question.ts
```diff
@@ -120,15 +120,15 @@ export const TabNonSsoItem: OptionItem = {
 
 export const M365SsoLaunchPageOptionItem: OptionItem = {
   id: "M365SsoLaunchPage",
-  label: "SSO enabled launch page",
+  label: `$(toolkit-tab) SSO enabled launch page`,
```
tecton, line 123: Maybe we can localize the string as well?

packages/vscode-extension/package.json
```diff
@@ -74,6 +74,36 @@
     "onView:teamsfx-help-and-feedback"
   ],
   "contributes": {
+    "icons": {
+      "toolkit-commandBot": {
```
tecton, line 78: What is the scope of these icons? If it's used globally in VS Code, maybe we could name it more specific like `teamsfx-toolkit-command-bot` to prevent conflicts.

HuihuiWu-Microsoft: They can only be used inside Teams Toolkit extension and dependent extensions.

packages/fx-core/src/plugins/solution/fx-solution/question.ts
HuihuiWu-Microsoft: Yes, will do

HuihuiWu-Microsoft: Updated

tecton (approved):

jayzhang (approved):

jaeyonglee8687:
> @jaeyonglee8687 ,I download the svgs from Figama and the ratio of these svgs are not the same. Not sure if this is designed, you can check the screenshot for visual effects. I'll merge this one first and please let me know if there're any updates for these icons.

All icons used in the screens are available codicon for now. Even when you 'add features' azure icon should be a part of the codicon.

PR #4811, opened 2022-04-22
fix: open local state

nliu-ms:
when user select local env:
image

when user select other env:
image


nliu-ms:
E2E TEST: N/A

HuihuiWu-Microsoft (approved):

PR #4814, opened 2022-04-22
chore(sample): add Graph Connector sample

formulahendry:
image

image


packages/fx-core/src/common/samples-config.json
```diff
@@ -162,6 +162,16 @@
...
             "suggested": false
+        },
+        {
+            "id": "graph-connector-app",
+            "title": "Graph Connector App",
+            "shortDescription": "Showcases how to build custom Graph Connector",
+            "fullDescription": "This sample app showcases how to build custom Graph Connector with Azure Functions and query data using Microsoft Graph Client and TeamsFx SDK.",
+            "tags": ["Tab", "Azure function", "TS", "SSO"],
+            "time": "10min to run",
+            "configuration": "Ready for local debug",
```
MuyangAmigo, line 173: Is it ready for local debug? I think it needs manual steps to consent the permission first?

HuihuiWu-Microsoft (approved):

packages/fx-core/src/common/samples-config.json
formulahendry: Updated:
image

HuihuiWu-Microsoft (approved):

blackchoey:
We may need to optimize the size of gif. Since it seems larger than other samples. Large gif size increases VS Code extension size, there was a bug previously for this.

HuihuiWu-Microsoft:
> We may need to optimize the size of gif. Since it seems larger than other samples. Large gif size increases VS Code extension size, there was a bug previously for this.

Suggested size is 1600*920, you can check whether your gif has correct size.

formulahendry:
> We may need to optimize the size of gif. Since it seems larger than other samples. Large gif size increases VS Code extension size, there was a bug previously for this.

@blackchoey Try to minify the size to 1MB.

blackchoey (approved):

HuihuiWu-Microsoft (approved):

PR #4822, opened 2022-04-24
fix: avoid error log when opening manifest template while provisioning

nliu-ms:
Workaround for this bug: 14225849

When executing commands like provisioning, and then open manifest template, it will has some acquire lock errors in output window. But when provisioning done, the real value will show successfully.

So change the log level to avoid this terrifying logs.
Before:
image

After:
image


nliu-ms:
E2E TEST: n/a

jayzhang (approved):

PR #4823, opened 2022-04-24
chore(vsc+core+cli): enable preview features by default

tecton:
Enable "Preview" feature by default.

- [ ] Fix E2E tests


HuihuiWu-Microsoft (dismissed):

packages/fx-core/tests/common/tools.test.ts
```diff
@@ -192,24 +192,5 @@ describe("tools", () => {
       chai.assert.isDefined(result);
       chai.assert.isTrue(result);
     });
-
```
HuihuiWu-Microsoft, line 195: Why remove this case?

tecton: The function `canAddSso()` has been changed and this test case is outdated and broken so I removed this case.

HuihuiWu-Microsoft (dismissed):

tecton:
Use 'enable-preview-features' branch instead, thus closing this PR.

PR #4825, opened 2022-04-24
perf(aad-manifest): show learn more button when deploy aad success

SLdragon:
Address this bug: [Bug 14177210](https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/14177210): Update wording/notification for deploy AAD manifest

image
image
image


packages/fx-core/src/plugins/solution/fx-solution/v2/deploy.ts
```diff
@@ -223,7 +231,18 @@ export async function deploy(
             }
           });
       } else {
-        ctx.userInteraction.showMessage("info", msg, false);
+        if (isDeployAADManifestFromVSCode) {
+          ctx.userInteraction
+            .showMessage("info", msg, false, getLocalizedString("core.deploy.aadManifestLearnMore"))
```
KennethBWSong, line 236: What will happen in cli? Will there be a button to click in cli?

KennethBWSong: Will have follow up after offline discussion.

KennethBWSong (approved):

packages/fx-core/src/plugins/solution/fx-solution/v2/deploy.ts
SLdragon: This logic is only for VSCode, for CLI part, I will create a task and discuss with Junjie how to update it in the future PR.

nliu-ms (approved):

HuihuiWu-Microsoft (approved):
LGTM

PR #4831, opened 2022-04-25
chore: align env question cicd

IvanJobs:
Skip env question in single env case for Teams Toolkit.

https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/14104788


MuyangAmigo:
We probably don't want to use an OS level pop up in this case? Provision used this because it will create resources in user's subscription, as `Add CICD` drops some files only, I think it's not that critical to ask for double confirmation?

eriolchan (approved):

PR #4841, opened 2022-04-25
chore(bot): add incoming webhook sample

XiaofuHuang:
Add imcoming webhook sample into create sample question options


swatDong:
UT fails?

XiaofuHuang:
> UT fails?

Retry the UT and passed.

PR #4852, opened 2022-04-26
perf(add sso): open readme and add telemetry

KennethBWSong:


packages/fx-core/src/plugins/solution/fx-solution/v2/executeUserTask.ts
```diff
@@ -1028,7 +1033,17 @@ export async function addSso(
     );
   }
 
-  return ok(undefined);
+  ctx.telemetryReporter.sendTelemetryEvent(SolutionTelemetryEvent.AddSso, {
```
HuihuiWu-Microsoft, line 1036: There's only AddSso-start and AddSso(success), AddSso(fail) is also needed for error case.

HuihuiWu-Microsoft (approved):

SLdragon (approved):

1yefuwang1 (approved):

PR #4853, opened 2022-04-26
fix(vsc): condition of time trigger question

tecton:
Bug: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/14246730


tecton:
E2E TEST: N/A

IvanJobs (approved):

PR #4854, opened 2022-04-26
chore: separate pkgs to release

wenytang-ms:
CD support release specific releases with inputs as following picture. 
1. if empty will release all the pkgs.
2. if inputs with space, will release the specific pkgs. 
The inputs only contain sdk, cli, vscode, cli, server and template
the following picture will only release sdk, vscode and cli.
image

### DIFF
--- a/.github/scripts/lernaDeps.json
+++ b/.github/scripts/lernaDeps.json
@@ -1,0 +1,47 @@
+{
+    "common": [
+        "packages/eslint-config"
+    ],
+    "cli": [
+        "packages/manifest",
+        "packages/api",
+        "packages/fx-core",
+        "packages/cli",
+        "templates"
+    ],
+    "vscode": [
+        "packages/manifest",
+        "packages/api",
+        "packages/fx-core",
+        "packages/vscode-extension",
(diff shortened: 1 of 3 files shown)


.github/scripts/lernaDeps.json
```diff
@@ -0,0 +1,32 @@
...
+        "templates"
+    ],
+    "server": [
+        "packages/manifest",
+        "packages/api",
+        "packages/fx-core",
+        "templates"
+    ],
+    "sdk": [
+        "packages/sdk",
```
tecton, line 26: if sdk is the input of pkg option, what's the expected behavior of sdk version in template?

LongOddCode (approved):

.github/scripts/lernaDeps.json
wenytang-ms: template will not sync up version with sdk
if you want to update the sdk version in template, you may input " sdk template"

tecton: I see, thank you!

hund030 (approved):

PR #4863, opened 2022-04-27
fix(m365): fix task not terminated after stopping F5

kuojianlu:
To fix https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/14252955.

E2E TEST: N/A

### DIFF
--- a/packages/vscode-extension/src/debug/teamsfxTaskHandler.ts
+++ b/packages/vscode-extension/src/debug/teamsfxTaskHandler.ts
@@ -364,24 +364,13 @@ async function onDidStartDebugSessionHandler(event: vscode.DebugSession): Promis
       (debugConfig.url || debugConfig.port) && // it's from launch.json
       !debugConfig.postRestartTask
     ) {
+      allRunningDebugSessions.add(event.id);
+
       // and not a restart one
       // send f5 event telemetry
-      const localAppId = (await getLocalTeamsAppId()) as string;
-      const isLocal =
-        (debugConfig.url as string) &&
-        localAppId &&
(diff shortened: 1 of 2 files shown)


packages/vscode-extension/src/debug/teamsfxTaskHandler.ts
```diff
@@ -364,37 +364,24 @@ async function onDidStartDebugSessionHandler(event: vscode.DebugSession): Promis
...
       }
 
       ExtTelemetry.sendTelemetryEvent(TelemetryEvent.DebugStart, {
         [TelemetryProperty.DebugSessionId]: event.id,
         [TelemetryProperty.DebugType]: debugConfig.type,
         [TelemetryProperty.DebugRequest]: debugConfig.request,
         [TelemetryProperty.DebugPort]: debugConfig.port + "",
-        [TelemetryProperty.DebugRemote]: isLocal ? "false" : "true",
-        [TelemetryProperty.DebugAppId]: appId,
+        [TelemetryProperty.DebugRemote]: debugConfig.teamsfxIsRemote + "",
```
a1exwang, line 381: Is this property needed? Can we determine whether it is local by checking env == sha256("local")?

kuojianlu: This property exists before multi env feature. Additionally, "local" env is introduced recently in config unify feature. So I think better to keep this property currently and may consider retire it afterwards.

a1exwang: It makes sense.

a1exwang (approved):

PR #4864, opened 2022-04-27
fix(vsc+core): addFeature bug in existing tab app

tecton:
Related bug:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/14263279
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/14263279

E2E TEST: N/A

### DIFF
--- a/packages/fx-core/src/common/tools.ts
+++ b/packages/fx-core/src/common/tools.ts
@@ -456,7 +456,10 @@ export function canAddSso(
   }
 
   const solutionSettings = projectSettings.solutionSettings as AzureSolutionSettings;
-  if (!(solutionSettings.hostType === HostTypeOptionAzure.id)) {
+  if (
+    !projectSettings.solutionSettings ||
+    !(solutionSettings.hostType === HostTypeOptionAzure.id)
+  ) {
     return returnError
       ? err(
           new SystemError(
--- a/packages/fx-core/src/common/tools.ts
+++ b/packages/fx-core/src/common/tools.ts
@@ -459,8 +459,7 @@ export function canAddSso(
   const solutionSettings = projectSettings.solutionSettings as AzureSolutionSettings;
   if (
     isExistingTabApp(projectSettings) &&
-    solutionSettings &&
-    !solutionSettings.capabilities.includes(TabSsoItem.id)
+    !(solutionSettings && solutionSettings.capabilities.includes(TabSsoItem.id))
   ) {
     return ok(Void);
   }
--- a/packages/fx-core/src/common/tools.ts
+++ b/packages/fx-core/src/common/tools.ts
@@ -455,11 +455,12 @@ export function canAddSso(
       : false;
   }
 
-  const solutionSettings = projectSettings.solutionSettings as AzureSolutionSettings;
-  if (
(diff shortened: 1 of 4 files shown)


packages/fx-core/src/plugins/solution/fx-solution/v2/executeUserTask.ts
```diff
@@ -242,6 +247,9 @@ export function canAddResource(
       sendErrorTelemetryThenReturnError(SolutionTelemetryEvent.AddResource, e, telemetryReporter)
     );
   }
+  if (!projectSetting.solutionSettings) {
```
qinezh, line 250: Would be better to use `isExistingTabApp(projectSettings)` to determine whether it's existing web app case.

packages/fx-core/src/common/tools.ts
```diff
@@ -456,7 +456,10 @@ export function canAddSso(
   }
 
   const solutionSettings = projectSettings.solutionSettings as AzureSolutionSettings;
-  if (!(solutionSettings.hostType === HostTypeOptionAzure.id)) {
+  if (
+    !projectSettings.solutionSettings ||
```
qinezh, line 460: 'Add SSO' is allowed for existing tab app scenario.

jayzhang (approved):

packages/fx-core/src/plugins/solution/fx-solution/v2/executeUserTask.ts
tecton: Thank you, fixed.

HuihuiWu-Microsoft (dismissed):

packages/fx-core/src/common/tools.ts
tecton: Thanks, updated.

packages/fx-core/src/common/tools.ts
```diff
@@ -455,11 +455,12 @@ export function canAddSso(
       : false;
   }
 
-  const solutionSettings = projectSettings.solutionSettings as AzureSolutionSettings;
-  if (
-    !projectSettings.solutionSettings ||
-    !(solutionSettings.hostType === HostTypeOptionAzure.id)
-  ) {
+  const solutionSettings = projectSettings.solutionSettings;
+  if (!solutionSettings) {
```
qinezh, line 459: How about using `isExistingTab(projectSettings)` to check if it's exisiting tab, and then check if SSO already added with `solutionSettings.capabilities.includes(TabSsoItem.id)`?

tecton: OK, so it would be like this?
```
if (isExistingTabApp(projectSettings) && !solutionSettings.capabilities.includes(TabSsoItem.id)) {
  return ok(Void);
}
```

qinezh: yeah, it looks good to me 😊

qinezh (approved):

jayzhang (approved):

HuihuiWu-Microsoft (approved):

PR #4875, opened 2022-04-27
chore(m365): open bot framework portal in browser with new profile folder

kuojianlu:
- open bot framework portal in a browser with new profile folder to avoid already signing in a wrong account. Private window cannot be used since it does not support corp account.

To fix https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/14254520.


packages/vscode-extension/src/debug/launch.ts
```diff
@@ -1,7 +1,14 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
+import { UserError } from "@microsoft/teamsfx-api";
+import open = require("open");
```
tecton, line 5: Can this package be imported using "import from"?

kuojianlu: No, it requires `esModuleInterop` to be true in `tsconfig.json`.

tecton (approved):

packages/vscode-extension/src/debug/launch.ts
```diff
@@ -28,3 +35,32 @@ export async function openHubWebClient(
...
+      ],
+    });
+  } catch {
+    const error = new UserError(
+      ExtensionSource,
+      ExtensionErrors.OpenBrowserFailed,
+      util.format(getDefaultString("teamstoolkit.localDebug.openBrowserFailed"), url),
+      util.format(localize("teamstoolkit.localDebug.openBrowserFailed"), url)
+    );
+    showError(error);
```
HuihuiWu-Microsoft, line 64: Is telemetry track needed for this error?

swatDong:
Does bot framework site support to login corp account in private window? Just recall that we cannot login to Teams in private window using corp account...

packages/vscode-extension/src/debug/launch.ts
kuojianlu: It think no need. Three reasons:
1. Opening bot framework protal is just a workaround since automatically adding Outlook channel (just like adding Teams channel) is not ready currently. When it is ready, this workaround will be retired.
2. This error is less likely to happen, only if there are no Chrome, Edge and Firefox in user's system.
3. This is not a blocking issue, even if error happens, users can open the link manually.

HuihuiWu-Microsoft (dismissed):

kuojianlu:
> Does bot framework site support to login corp account in private window? Just recall that we cannot login to Teams in private window using corp account...

Good catch. I only tested non-corp account before. Just tested my @microsoft.com account, it is indeed not supported...

swatDong (approved):

HuihuiWu-Microsoft (approved):

Alive-Fish (approved):

kimizhu (approved):

PR #4884, opened 2022-04-28
chore(sql): update e2e test

xzf0587:
As the command of creating SQL has changed, update e2e test.


blackchoey (approved):
The previous code passes process.env to the function, do we still need to pass process.env in the new code?

xzf0587:
> The previous code passes process.env to the function, do we still need to pass process.env in the new code?

The test case does not need env information. So remove it.

adashen (approved):

PR #4885, opened 2022-04-28
chore: use latest version of axios

blackchoey:
Use latest axios release which has some impactful breaking change (especially on error handling) and security fixes. The release note can be found here: https://github.com/axios/axios/releases

Using latest release will avoid introducing such breaking change in the future.


tecton (approved):

wenytang-ms (approved):

blackchoey:
E2E test passed: https://github.com/OfficeDev/TeamsFx/actions/runs/2237351935

dilin-MS (approved):

PR #4889, opened 2022-04-28
perf(aad-manifest): merge aad migration with unify config migration

SLdragon:
Address this bug: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/14134470

Merge aad migration with unify config migration:
1. Merge unify config/aad migrate if it is older version before 418 release.
2. If just aad manifest migration, migrate silently and only show message box right bottom with help link to [AAD manifest doc](https://github.com/SLdragon/wikipage/blob/aad-manifest-update/Upgrade-to-support-unify-config-and-aad-manifest.md)

Matix: 
  #### For project with sso support before 4.18 release
     - Will show modal/ not modal alert to tell user to migrate
     - Unify config with aad manifest migration
  
  #### For project without sso support before 4.18 release
     - Will show modal/ not modal alert to tell user to migrate
     - Only Unify config migration
  
  #### For project with sso support after 4.18 release
     - Migrate silently
     - only show message box right bottom with help link to AAD manifest doc
  
  #### For project without sso support after 4.18 release
     - Do not need migration


packages/fx-core/resource/package.nls.json
```diff
@@ -51,15 +51,12 @@
...
   "core.consolidateLocalRemote.Message": "Your project will be upgraded to work with the latest version of Teams Toolkit. The upgrade will only backup and change the local config and manifest template file without changing your custom code. If you want to skip auto-upgrade, please install older version of Teams Toolkit (<=3.7.0)",
-  "core.consolidateLocalRemote.DifferentManifest": "Since Teams Toolkit will use manifest.remote.template.json as a single manifest template file after the upgrade and we detected you may have customized the manifest.local.template.json, you may need to update the manifest.template.json to include your change. See Learn More for details.",
+  "core.consolidateLocalRemoteWithAadManifest.Message": "Your project will be upgraded to work with the latest version of Teams Toolkit. The upgrade will only backup and change the local config, manifest template and projectSettings file without changing your custom code. If you want to skip auto-upgrade, please install older version of Teams Toolkit (<=3.7.0)",
```
adashen, line 54: It is not specific to the file, seems not clear as previous upgrade message. Besides do we still need line 55 message in the new version? Since seems there is no scenario that only unify config will be upgraded right?

packages/fx-core/resource/package.nls.json
```diff
@@ -51,15 +51,12 @@
...
-  "core.aadManifestMigration.outputMsg": "Congratulations! You have successfully upgraded your Teams App project structure. We have added aad.template.json for AAD app configuration. And we have backed up projectSettings.json, permissions.json to .backup folder.",
-  "core.aadManifestMigration.SuccessMessage": "Upgrade succeeded. Your project is upgraded to new configuration files.",
+  "core.consolidateLocalRemoteWithAadManifest.SuccessMessage": "Congratulations! You have successfully upgraded your Teams App project structure. We have added config.local.json for local debug, manifest.template.json for both Teams app running locally and remotely and aad.template.json for AAD manifest template. And we have backed up localSettings.json, manifest.local.template.json, manifest.remote.template.json, projectSettings.json to .backup folder.",
+  "core.aadManifestMigration.outputMsg": "Congratulations! You have successfully upgraded your Teams App to support AAD manifest feature. We have added aad.template.json and updated your projectSettings.json file. And we have backed up projectSettings.json to .backup folder.",
```
adashen, line 58: Since we don't want to treat only aad manifest as "upgrade". Can we avoid this wording?

packages/fx-core/resource/package.nls.json
```diff
@@ -51,15 +51,12 @@
...
+  "core.aadManifestMigration.SuccessMessage": "Upgrade succeeded. You have successfully upgraded your Teams App to support AAD manifest feature. We have added aad.template.json and updated your projectSettings.json. And we have backed up projectSettings.json to .backup folder. For more information about AAD manifest, please refer %s",
```
adashen, line 59: same as above

packages/fx-core/resource/package.nls.json
SLdragon: > Besides do we still need line 55 message in the new version? Since seems there is no scenario that only unify config will be upgraded right?

We still need message on line 55, it is used by unify config migration when user changed teams app manifest file.  For project like SPFX without SSO support, it will only have unify config migration.

packages/fx-core/resource/package.nls.json
SLdragon: Thanks updated

packages/fx-core/resource/package.nls.json
SLdragon: updated

packages/fx-core/resource/package.nls.json
SLdragon: updated

packages/fx-core/src/core/middleware/MigrationUtils.ts
```diff
@@ -0,0 +1,82 @@
...
+      return false;
+    }
+
+    const permissionFileExist = await fs.pathExists(
+      path.join(inputs.projectPath as string, "permissions.json")
+    );
+    const projectSettingsJson = await fs.readJson(
+      path.join(inputs.projectPath as string, ".fx", "configs", "projectSettings.json")
+    );
+    const aadPluginIsActive = projectSettingsJson.solutionSettings.activeResourcePlugins.includes(
```
KennethBWSong, line 70: For existing app scenario, there won't be solutionSettings object here. Maybe will throw error here?

SLdragon: Oh, thanks, will update this block

packages/fx-core/src/core/middleware/consolidateLocalRemote.ts
```diff
@@ -307,8 +347,88 @@ async function consolidateLocalRemote(ctx: CoreHookContext): Promise<boolean> {
...
+      aadManifestJson.requiredResourceAccess = requiredResourceAccess;
+      const aadManifestPath = path.join(
+        inputs.projectPath as string,
+        "templates",
+        "appPackage",
+        "aad.template.json"
+      );
+      const projectSettingsJson = await fs.readJson(projectSettingsPath);
+
+      if (projectSettingsJson.solutionSettings.capabilities.includes("Tab")) {
```
KennethBWSong, line 370: Seems logic here are the same as in the AAD plugin, can we just call scaffold() here or maybe we can wrap it?

SLdragon: Thanks, updated

KennethBWSong (approved):

xiaolang124 (approved):

nliu-ms (approved):

1yefuwang1 (approved):

adashen (approved):

PR #4898, opened 2022-04-29
perf(api-connector): open scaffold files once user task done

wenytang-ms:
https://dev.azure.com/msazure/Microsoft%20Teams%20Extensibility/_workitems/edit/14276715

### DIFF
--- a/packages/vscode-extension/src/handlers.ts
+++ b/packages/vscode-extension/src/handlers.ts
@@ -71,6 +71,7 @@ import {
   isValidProject,
   LocalEnvManager,
   ProjectSettingsHelper,
+  UserTaskFunctionName,
 } from "@microsoft/teamsfx-core";
 import * as vscode from "vscode";
 import GraphManagerInstance from "./commonlib/graphLogin";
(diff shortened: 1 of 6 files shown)


packages/vscode-extension/src/handlers.ts
```diff
@@ -718,6 +719,14 @@ export async function addFeatureHandler(args?: any[]): Promise<Result<null, FxEr
           await commands.executeCommand("markdown.preview.toggleLock");
         });
       }
+    } else if (workspacePath && result.value.func == UserTaskFunctionName.ConnectExistingApi) {
+      const components: string[] = result.value.component;
+      for (const item of components) {
+        const apiFilePath = path.join(workspacePath, item, result.value.fileName);
```
blackchoey, line 725: can we  return the paths of generated sample code in api connector plugin? The logic in VS Code just need to open the files we return. In this way, if we want to do any change in the future (e.g. generate the files to another place), we just need to update api connector plugin.

wenytang-ms: sure

blackchoey (approved):

jayzhang (approved):

tecton (approved):

PR #4901, opened 2022-04-29
chore: update cache wording

swatDong:
Fix https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/14263253/

- Update cache wording to make the message clearer.
- Use `warning` instead of `error` since it does not break development


packages/cli/src/commonlib/cacheAccess.ts
```diff
@@ -145,15 +145,19 @@ export class CryptoCachePlugin {
           }
         }
       } catch (err) {
-        VsCodeLogInstance.error("read token fail: " + err.message);
+        VsCodeLogInstance.warning(
+          "Cannot read token from cache, clean up account cache. " + err.message
```
kuojianlu, line 149: `clean up account cache` is a user action? Does user need to do something for this warning?

swatDong: No, it's acted by toolkit. No user action needed, so make this warning.

kuojianlu (approved):

HuihuiWu-Microsoft (approved):

PR #4908, opened 2022-05-05
refactor(vsc): account treeview and SVG image folder structure

tecton:
1. Refactor account view using dynamic node similar to environment tree.
2. Reorganize SVG images and clean outdated files.

E2E TEST: N/A


packages/vscode-extension/.vscodeignore
```diff
@@ -1,22 +1,22 @@
...
 build/**
-.prettierrc.json
-.prettierignore
-webpack.config.js
-.github
-node_modules
-!out/templates/**
 coverage
-.coverconfig.json
 img/**
```
HuihuiWu-Microsoft, line 16: Resources inside img folder will not be included in the package, moving the sample resources to this folder will lost them, right?

tecton: Images used in [controls/SampleGallery.tsx](https://github.com/OfficeDev/TeamsFx/pull/4908/files#diff-10480cac513d5bc999a7a187df546398f616ce8b9381d7da18fafd200960faff) are actually bundled into client.js by Webpack. So, they are not necessary to be kept in media folder.

HuihuiWu-Microsoft: Will images used in tree view bundled into extension.js by Webpack? If so, does that mean we can also exclude media folder from .vscodeignore?

tecton: No, they are 2 concepts. Sample gallery is using image like `import xx from "xxx.png"` and renders in a web view. Media folders are used in runtime and since we are not using webview as treeview, the icons in treeview will be rendered dynamically by VS Code.

packages/vscode-extension/src/treeview/account/azureNode.ts
```diff
@@ -0,0 +1,139 @@
...
+    if (this.status === AccountItemStatus.SigningIn) {
+      return;
+    }
+    this.status = AccountItemStatus.SigningIn;
+    this.contextValue = "";
+    // refresh
+    this.eventEmitter.fire(this);
+  }
+
+  public setSignedOut() {
```
HuihuiWu-Microsoft, line 51: Will this remove the subscription node from Azure node?

tecton: Yes, since the status is set to `SignedOut` here, `getTreeItem()` triggered by `eventEmitter` will update the collapsible state to `None` which removes the subscription node.

HuihuiWu-Microsoft (approved):

xiaolang124 (approved):

Issue #4859, opened 2022-04-26
TeamsFx token generation is failing in Android


SubbaReddi:
TeamsFx [sample app](https://github.com/OfficeDev/TeamsFx-Samples/tree/ga/hello-world-tab) is deployed. Tab app is loading fine. If we click on "Authorize" button, consent popup opens and popup will close after consent. And then profile details are showing in tab. It is working fine Desktop Teams/ Web teams/ iOS as well.
But it is failing in Android mobile. Once popup closed, following error message is showing:
Unhandled Rejection(ErrorWithCode. InvalidResponse): Failed to parse response to Json

Issue is consistently reproducible in other apps as well. 

**To Reproduce**
1. Deploy following app: https://github.com/OfficeDev/TeamsFx-Samples/tree/ga/hello-world-tab
2. Once app installed for user, open app in android teams
3. Click on "Authrorize" and provide consent in popup
4. You will see error after popup closes.

**Expected behavior**
User profile details like name, profile pic should be shown in tab

**Screenshots**
[Attached recording](https://msteamspoc-my.sharepoint.com/:f:/g/personal/johndoe_msteamspoc_onmicrosoft_com/Ekg1ae69jwREkrCLFi9O3KUBklXNIt8WMLRNPddYDEIDLw?e=ZNirNd)

**Additional context**
[Attached recording](https://msteamspoc-my.sharepoint.com/:f:/g/personal/johndoe_msteamspoc_onmicrosoft_com/Ekg1ae69jwREkrCLFi9O3KUBklXNIt8WMLRNPddYDEIDLw?e=ZNirNd)

formulahendry:
Fix is merged in `ga` branch: https://github.com/OfficeDev/TeamsFx/pull/4910
Will close this issue once there is an available SDK in https://www.npmjs.com/package/@microsoft/teamsfx

formulahendry:
Hi @SubbaReddi , this issue has been fixed. And this fix is available in the beta version of the `@microsoft/teamsfx` SDK: https://www.npmjs.com/package/@microsoft/teamsfx/v/0.7.1-beta.c23501411.0

You could try this beta version in you Teams app.

PR #4910, opened 2022-05-05
fix(sdk): handle Json paring for Android

formulahendry:
Fix https://github.com/OfficeDev/TeamsFx/issues/4859

There seems an issue from Teams SDK: https://github.com/OfficeDev/microsoft-teams-library-js/issues/473

Refer to the previous code logic:  https://github.com/OfficeDev/TeamsFx/blob/44ed5ea97551d6abba1c46a891f74bae163e68b2/packages/sdk/src/util/utils.ts#L92, try to handle when `result` is an `object` instead of a `string`.

E2E TEST: N/A

### DIFF
--- a/packages/sdk/src/credential/teamsUserCredential.browser.ts
+++ b/packages/sdk/src/credential/teamsUserCredential.browser.ts
@@ -119,7 +119,7 @@ export class TeamsUserCredential implements TokenCredential {
 
             let resultJson: any = {};
             try {
-              resultJson = JSON.parse(result);
+              resultJson = typeof result == "string" ? JSON.parse(result) : result;
             } catch (error) {
               // If can not parse result as Json, will throw error.
               const failedToParseResult = "Failed to parse response to Json.";


packages/sdk/src/credential/teamsUserCredential.browser.ts
```diff
@@ -119,7 +119,7 @@ export class TeamsUserCredential implements TokenCredential {
 
             let resultJson: any = {};
             try {
-              resultJson = JSON.parse(result);
+              resultJson = typeof result == "string" ? JSON.parse(result) : result;
```
blackchoey, line 122: if android only returns a token string, we may not able to login users (since we cannot set session storage correctly)?  @KennethBWSong

KennethBWSong: I don't have time to debug the app now, but @formulahendry said it works on android after updating this line.

blackchoey: sorry I understand the code incorrectly. The string should be a json string instead a pure token string.

blackchoey (approved):

blackchoey:
@formulahendry is it possible to add a test case for it? if it's hard to add automation test, we could add a case to our manual test plan.

formulahendry:
> @formulahendry is it possible to add a UT for it? if it's hard to add automation test case, we could add a case to our manual test plan.

Seems hard to automate it. I have add test steps to include this scenario in https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/12684060

adashen (approved):

PR #4927, opened 2022-05-06
refactor(bot): implement scaffold

yukun-dong:


yukun-dong:
E2E TEST: bot refactor not complete.

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
```diff
@@ -306,19 +309,34 @@ export class TeamsBotV2Impl {
...
+        case PluginActRoles.Notification:
+          if (hostType === HostType.Function) {
+            scenarios.push(TemplateProjectsScenarios.NOTIFICATION_FUNCTION_BASE_SCENARIO_NAME);
+            triggers.map((trigger) => scenarios.push(TriggerTemplateScenarioMappings[trigger]));
+          }
+          if (hostType === HostType.AppService) {
+            scenarios.push(TemplateProjectsScenarios.NOTIFICATION_RESTIFY_SCENARIO_NAME);
+          }
+          break;
+        case PluginActRoles.Bot || PluginActRoles.MessageExtension:
```
hund030, line 330: Do we need to support notification bot + message extension in the future?

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
```diff
@@ -272,6 +273,8 @@ export class TeamsBotV2Impl {
         return HostType.AppService;
       case "azure-functions":
         return HostType.Function;
+      case undefined:
+        return HostType.AppService;
```
hund030, line 277: When will rawHostType be undefined?

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
yukun-dong: I added the logic In line 333. When bot and message extension are both selected, the legacy bot will be pushed only once/

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
yukun-dong: The lagacy bot. Not sure if there are better ways to handle it here.

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
hund030: Get it. I just misunderstood.

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
eriolchan: I guess the undefined is to handle backward compatibility, e.g., there is no host type in project settings if user created the bot by previous Teams Toolkit version, so we need to handle this case and use webapp to host it.

For this case, it is better to use [default](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch) in switch...case

packages/fx-core/src/plugins/resource/bot/v2/scaffold.ts
```diff
@@ -2,11 +2,16 @@
...
 
 import {
+  defaultActionSeq,
   ScaffoldAction,
+  ScaffoldActionName,
   ScaffoldContext,
   scaffoldFromTemplates,
 } from "../../../../common/template-utils/templatesActions";
 import { CodeTemplateInfo } from "./interface/codeTemplateInfo";
+import { CommandExecutionError, TemplateZipFallbackError, UnzipError } from "../errors";
```
hund030, line 12: Unused import `CommandExecutionError`?

hund030 (approved):
LGTM

packages/fx-core/src/plugins/resource/bot/v2/scaffold.ts
yukun-dong: Thanks for the catch. It's removed.

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
```diff
@@ -164,10 +171,13 @@ export class TeamsBotV2Impl {
...
+      PluginBot.HOST_TYPE
+    ] as HostType;
     const lang = this.resolveProgrammingLanguage(ctx);
+    const isM365 = ctx.projectSetting?.isM365;
 
-    const scenarios = this.resolveScenarios(actRoles, triggers, hostType);
+    const scenarios = this.resolveScenarios(actRoles, triggers, hostType, isM365);
 
     return scenarios.map((scenario) => {
       return {
```
eriolchan, line 183: Add comment for getBicepConfigs() in L192, since there is only one type of programming language, the line between 197 and 202 is a relation of if...else. The value can be decided by lang and it could be a map, instead of multiple if conditions.

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
```diff
@@ -306,19 +316,35 @@ export class TeamsBotV2Impl {
...
+      switch (actRole) {
+        case PluginActRoles.CommandAndResponse:
+          scenarios.push(TemplateProjectsScenarios.COMMAND_AND_RESPONSE_SCENARIO_NAME);
+          break;
+        case PluginActRoles.Notification:
+          if (hostType === HostTypes.AZURE_FUNCTIONS) {
+            scenarios.push(TemplateProjectsScenarios.NOTIFICATION_FUNCTION_BASE_SCENARIO_NAME);
+            triggers.map((trigger) => scenarios.push(TriggerTemplateScenarioMappings[trigger]));
+          }
+          if (hostType === HostTypes.APP_SERVICE) {
```
eriolchan, line 333: The logic looks a little weird. The scenario seems to be decided by user input directly, why we need rely on hostType? P.S. hostType is also decided by the same user input.

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
```diff
@@ -306,19 +316,35 @@ export class TeamsBotV2Impl {
...
+            scenarios.push(TemplateProjectsScenarios.NOTIFICATION_RESTIFY_SCENARIO_NAME);
+          }
+          break;
+        case PluginActRoles.Bot:
+        case PluginActRoles.MessageExtension:
+          if (isM365) {
```
eriolchan, line 339: It is weird to have this isM365 logic here, let's sync offline and find a better to organize them

packages/fx-core/src/plugins/resource/bot/v2/scaffold.ts
```diff
@@ -15,10 +20,24 @@ export async function scaffold(template: CodeTemplateInfo, dst: string): Promise
...
     onActionEnd: async (action: ScaffoldAction, context: ScaffoldContext) => {
-      // TODO
+      if (action.name === ScaffoldActionName.FetchTemplatesUrlWithTag) {
+        Logger.info(Messages.SuccessfullyRetrievedTemplateZip(context.zipUrl ?? ""));
+      }
     },
-    onActionError: async (action: ScaffoldAction, context: ScaffoldContext) => {
-      // TODO
+    onActionError: async (action: ScaffoldAction, context: ScaffoldContext, error: Error) => {
+      Logger.info(error.toString());
```
eriolchan, line 27: If it is an error, we'd better to use Logger.error()

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
```diff
@@ -52,13 +54,7 @@ export class TeamsBotV2Impl {
   async scaffoldSourceCode(ctx: Context, inputs: Inputs): Promise<Result<Void, FxError>> {
     const workingPath = path.join(inputs.projectPath ?? "", "bot");
```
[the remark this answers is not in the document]
JerryYangKai: @eriolchan  Sure, I will first provide some back ground. 
For VS code side, this is the right way to scaffold.
For VS side, do not need to join path. The inputs.projectPath is the place we scaffold the code.
Using platform to check if it is in VS Code or VS

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
```diff
@@ -52,13 +54,7 @@ export class TeamsBotV2Impl {
...
     const hostTypeTriggers = inputs[QuestionNames.BOT_HOST_TYPE_TRIGGER];
```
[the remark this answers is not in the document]
yukun-dong: This is a string combination of notification trigger and host, like "timer-functions", "http-restify" so it's hard to give it a better name. I think it can be improved after refactoring the question model.

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
```diff
@@ -164,10 +171,13 @@ export class TeamsBotV2Impl {
   private getTemplates(ctx: Context, inputs: Inputs): CodeTemplateInfo[] {
     const actRoles = this.resolveActRoles(ctx, inputs);
```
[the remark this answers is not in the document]
yukun-dong: The resolveActRoles is removed. The resolve trigger is to get the notifcation trigger like "timer" or "http".

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
```diff
@@ -164,10 +171,13 @@ export class TeamsBotV2Impl {
...
     const triggers = this.resolveTriggers(inputs);
-    const hostType = this.resolveHostType(ctx);
+    const hostType = ctx.projectSetting?.pluginSettings?.[PluginBot.PLUGIN_NAME]?.[
+      PluginBot.HOST_TYPE
+    ] as HostType;
     const lang = this.resolveProgrammingLanguage(ctx);
+    const isM365 = ctx.projectSetting?.isM365;
 
-    const scenarios = this.resolveScenarios(actRoles, triggers, hostType);
+    const scenarios = this.resolveScenarios(actRoles, triggers, hostType, isM365);
```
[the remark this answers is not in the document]
yukun-dong: The resolveScenarios method is refactored.

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
yukun-dong: Function bot needs to add an extra "function-base" template so we need `hostType` to make the judgement.

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
```diff
@@ -71,6 +67,17 @@ export class TeamsBotV2Impl {
     return ok(Void);
   }
 
+  private resolveHostType(hostTypeTriggers: string[]): HostType {
+    let hostType;
+    if (Array.isArray(hostTypeTriggers)) {
+      const hostTypes = hostTypeTriggers.map(
+        (item) => HostTypeTriggerOptions.find((option) => option.id === item)?.hostType
+      );
+      hostType = hostTypes ? hostTypes[0] : undefined;
```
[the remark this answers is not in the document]
yukun-dong: For function notification bot the array could have length 2. I didn't figure out how to handle it better here.

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
```diff
@@ -71,6 +67,17 @@ export class TeamsBotV2Impl {
     return ok(Void);
   }
 
+  private resolveHostType(hostTypeTriggers: string[]): HostType {
```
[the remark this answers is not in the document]
yukun-dong: After refactor this resolveHostType is referenced in resolveScenarios. To be consistent with other methods, I think it's better not to change the name.

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
yukun-dong: Change it to use map here. Please check.

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
```diff
@@ -313,39 +269,78 @@ export class TeamsBotV2Impl {
...
-            triggers.map((trigger) => scenarios.push(TriggerTemplateScenarioMappings[trigger]));
-          }
-          if (hostType === HostTypes.APP_SERVICE) {
-            scenarios.push(TemplateProjectsScenarios.NOTIFICATION_RESTIFY_SCENARIO_NAME);
-          }
+  private resolveScenarios(ctx: Context, inputs: Inputs): string[] {
+    const templateScenarios: string[] = [];
+    const solutionSettings = ctx.projectSetting.solutionSettings as AzureSolutionSettings;
+    const capabilities = solutionSettings.capabilities;
+    capabilities.map((capability: string) => {
```
hund030, line 276: Is it possible that capabilities contains both bot and message extension?
In that case do we push default scenario twice?

yukun-dong: After merge templates there will be only one left. Also it's probable that msgext template won't share same template with bot in the future.

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
```diff
@@ -24,41 +25,43 @@ import * as utils from "../utils/common";
...
   async scaffoldSourceCode(ctx: Context, inputs: Inputs): Promise<Result<Void, FxError>> {
-    const workingPath = path.join(inputs.projectPath ?? "", "bot");
-    const hostTypeTriggers = inputs[QuestionNames.BOT_HOST_TYPE_TRIGGER];
-    let hostType;
-    if (Array.isArray(hostTypeTriggers)) {
-      const hostTypes = hostTypeTriggers.map(
-        (item) => HostTypeTriggerOptions.find((option) => option.id === item)?.hostType
-      );
-      hostType = hostTypes ? hostTypes[0] : undefined;
+    let workingPath: string;
```
eriolchan, line 58: We can do a more refactor to remove duplicated logic of getting project path from inputs
```
let workingPath: string = inputs.projectPath ?? "";
if (inputs.platform !== Platform.VS) {
  workingPath = path.join(workingPath, "bot");
}
```

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
```diff
@@ -180,19 +191,15 @@ export class TeamsBotV2Impl {
   }
 
   private getBicepConfigs(ctx: Context, inputs: Inputs): BicepConfigs {
-    const lang = this.resolveProgrammingLanguage(ctx);
-
+    const languageMapping: { [key: string]: string } = {
```
eriolchan, line 194: This map could be a const outside this method, so that it won't be created again and again when this method is called. It could be somewhere bot plugin defines const.

The map name could be runtimeMap

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
```diff
@@ -235,61 +242,20 @@ export class TeamsBotV2Impl {
...
-  }
-
-  private resolveHostType(ctx: Context): HostType {
+  private resolveServiceType(ctx: Context): ServiceType {
     const rawHostType = ctx.projectSetting?.pluginSettings?.[PluginBot.PLUGIN_NAME]?.[
       PluginBot.HOST_TYPE
     ] as string;
 
     switch (rawHostType) {
       case "app-service":
```
eriolchan, line 251: This can also be a mapping from host-type to serviceType

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
```diff
@@ -303,23 +269,78 @@ export class TeamsBotV2Impl {
...
+        }
+      });
     }
-    if (hostType === HostType.AppService) {
-      // TODO: support command & respond bot and notification bot
-      scenarios.push(TemplateProjectsScenarios.DEFAULT_SCENARIO_NAME);
+    return templateScenarios;
+  }
+
+  private resolveTriggers(inputs: Inputs): BotTrigger[] {
```
eriolchan, line 332: Let's sync this logic offline

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
```diff
@@ -84,7 +99,7 @@ export class TeamsBotV2Impl {
       configs: bicepConfigs,
     };
 
-    const hostTypes = [this.resolveHostType(ctx), HostType.BotService];
+    const hostTypes = [this.resolveServiceType(ctx), ServiceType.BotService];
```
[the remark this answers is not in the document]
yukun-dong: Yes, I will change these varable names

packages/fx-core/src/plugins/resource/bot/v2/constants.ts
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { ServiceType } from "../../../../common/azure-hosting/interfaces";
+
+export const languageMapping: { [key: string]: string } = {
```
eriolchan, line 6: Good to see the refactor here. I am curious that why L7-L9, the key is not with "", but L13-L14 has "". Maybe we can build a class for this mapping and provide method easy to use. Let's sync tomorrow

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
```diff
@@ -162,37 +176,51 @@ export class TeamsBotV2Impl {
...
-    const scenarios = this.resolveScenarios(actRoles, triggers, hostType);
-
-    return scenarios.map((scenario) => {
-      return {
+    const templateInfos: CodeTemplateInfo[] = [];
+    const botGroupScenarios = new Set<string>();
+    const msgextGroupScenarios = new Set<string>();
+    const solutionSettings = ctx.projectSetting.solutionSettings as AzureSolutionSettings;
+    const capabilities = solutionSettings.capabilities;
+    capabilities.map((capability: string) => {
```
eriolchan, line 185: Let's sync more details about this logic. For VSCode, if teams app contains both ME and Bot, there is only one template is scaffolded instead of two. Let's align on the business scenario first.

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
```diff
@@ -162,37 +176,51 @@ export class TeamsBotV2Impl {
...
-
-    if (lang === "js" || lang === "ts") {
-      bicepConfigs.push("node");
-    }
-    if (lang === "csharp") {
-      bicepConfigs.push("dotnet");
-    }
-
+    const lang = this.resolveProgrammingLanguage(ctx);
+    bicepConfigs.push(languageMapping[lang]);
```
eriolchan, line 222: What if the value of lang is not defined in languageMapping?

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
```diff
@@ -235,59 +263,11 @@ export class TeamsBotV2Impl {
...
-
-    if (capabilities?.includes(PluginActRoles.MessageExtension)) {
-      actRoles.push(PluginActRoles.MessageExtension);
-    }
-
-    return actRoles;
-  }
-
-  private resolveHostType(ctx: Context): HostType {
+  private resolveServiceType(ctx: Context): ServiceType {
```
eriolchan, line 266: I'd like to rename it as decideServiceType() since it is not simply resolved by file scheme

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
```diff
@@ -235,59 +263,11 @@ export class TeamsBotV2Impl {
...
-      return [];
-    }
-    // convert HostTypeTrigger question to trigger name
-    return rawHostTypeTriggers
-      .map((hostTypeTrigger) => {
-        const option = HostTypeTriggerOptions.find((option) => option.id === hostTypeTrigger);
-        return option?.trigger;
-      })
-      .filter((item): item is BotTrigger => item !== undefined);
+    return hostServiceTypeMapping[rawHostType];
```
eriolchan, line 270: We need to add some error handling here, e.g. what if rawHostType is not defined in the mapping to make code robust.

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
```diff
@@ -303,23 +283,65 @@ export class TeamsBotV2Impl {
...
+    if (!botScenarios) {
+      botGroupScenarios.add(TemplateProjectsScenarios.DEFAULT_SCENARIO_NAME);
+    } else {
+      botScenarios.map((scenario: string) => {
+        switch (scenario) {
+          case BotScenario.CommandAndResponseBot:
+            botGroupScenarios.add(TemplateProjectsScenarios.COMMAND_AND_RESPONSE_SCENARIO_NAME);
+            break;
+          case BotScenario.NotificationBot:
+            const hostType = this.resolveHostType(inputs);
```
eriolchan, line 320: The logic should depends on user input of trigger type instead of hostType. e.g. if it is a restify HTTP trigger, the scenario is NOTIFICATION_RESTIFY_SCENARIO_NAME, otherwise it contains NOTIFICATION_FUNCTION_BASE_SCENARIO_NAME and trigger related template.

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
```diff
@@ -303,23 +283,65 @@ export class TeamsBotV2Impl {
...
+            if (hostType === HostTypes.AZURE_FUNCTIONS) {
+              botGroupScenarios.add(
+                TemplateProjectsScenarios.NOTIFICATION_FUNCTION_BASE_SCENARIO_NAME
+              );
+              const triggers = this.resolveTriggers(inputs);
+              triggers.map((trigger) =>
+                botGroupScenarios.add(TriggerTemplateScenarioMappings[trigger])
```
eriolchan, line 327: Since this is also another mapping definition, we can keep all mapping into one separate file to make us easy to config it for business.

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
```diff
@@ -303,23 +283,65 @@ export class TeamsBotV2Impl {
...
-    actRoles: PluginActRoles[],
-    triggers: BotTrigger[],
-    hostType: HostType
-  ): string[] {
-    const scenarios: string[] = [];
-    if (hostType === HostType.Function) {
-      if (actRoles.includes(PluginActRoles.Notification)) {
-        scenarios.push(TemplateProjectsScenarios.NOTIFICATION_FUNCTION_BASE_SCENARIO_NAME);
-        triggers.map((trigger) => scenarios.push(TriggerTemplateScenarioMappings[trigger]));
+  private resolveScenariosForMessageExtension(
```
eriolchan, line 286: I still don't understand why we need to have two method to resolve scenario for ME and Bot, let's sync this offline.

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
```diff
@@ -24,41 +25,42 @@ import * as utils from "../utils/common";
...
-    const workingPath = path.join(inputs.projectPath ?? "", "bot");
-    const hostTypeTriggers = inputs[QuestionNames.BOT_HOST_TYPE_TRIGGER];
-    let hostType;
-    if (Array.isArray(hostTypeTriggers)) {
-      const hostTypes = hostTypeTriggers.map(
-        (item) => HostTypeTriggerOptions.find((option) => option.id === item)?.hostType
-      );
-      hostType = hostTypes ? hostTypes[0] : undefined;
+    let workingPath = inputs.projectPath ?? "";
+    if (inputs.platform !== Platform.VS) {
```
IvanJobs, line 60: How about CLI scenario under VS?

packages/fx-core/src/common/azure-hosting/hostingFactory.ts
```diff
@@ -4,15 +4,15 @@
...
 import { AzureHosting } from "./azureHosting";
 import { BotServiceHosting } from "./botServiceHosting";
-import { HostType } from "./interfaces";
+import { ServiceType } from "./interfaces";
 
 const HostingMap: { [key: string]: () => AzureHosting } = {
-  [HostType.Function]: () => new AzureFunctionHosting(),
-  [HostType.BotService]: () => new BotServiceHosting(),
+  [ServiceType.Function]: () => new AzureFunctionHosting(),
+  [ServiceType.BotService]: () => new BotServiceHosting(),
```
IvanJobs, line 11: NIT: Although it's not fully related to the changes of the PR, how about using `{[key: string]: AzureHosting}` instead of using a function for the value? Would it be simpler?

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
```diff
@@ -71,6 +73,18 @@ export class TeamsBotV2Impl {
     return ok(Void);
   }
 
+  private resolveHostType(inputs: Inputs): HostType {
+    const hostTypeTriggers = inputs[QuestionNames.BOT_HOST_TYPE_TRIGGER];
+    let hostType;
+    if (Array.isArray(hostTypeTriggers)) {
+      const hostTypes = hostTypeTriggers.map(
```
IvanJobs, line 80: I'm wondering if it's better to use `filter`.

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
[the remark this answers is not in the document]
yukun-dong: To support CLI I use language to make the check here.

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
```diff
@@ -71,6 +74,18 @@ export class TeamsBotV2Impl {
...
+        (item) => HostTypeTriggerOptions.find((option) => option.id === item)?.hostType
+      );
+      hostType = hostTypes ? hostTypes[0] : undefined;
+    }
+    return hostType ? hostType : HostTypes.APP_SERVICE;
```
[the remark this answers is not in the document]
yukun-dong: Changing HostTypes to enum causes a lot of regression. It's better to refactor these in another PR.

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
yukun-dong: These methods are removed.

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
yukun-dong: Changed it to use language to check.

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
yukun-dong: Moved these maps to mapping.ts.

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
yukun-dong: Added.

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
yukun-dong: According to our discussion, I added the logic at the start of the `decideTemplateScenarios` function. Will check if it can be improved later.

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
yukun-dong: According to the discussion, since it's simply resolved through mapping, will still use `resolveServiceType` here.

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
```diff
@@ -21,50 +11,37 @@ import {
...
-    const hostTypeTriggers = inputs[QuestionNames.BOT_HOST_TYPE_TRIGGER];
-    let hostType;
-    if (Array.isArray(hostTypeTriggers)) {
-      const hostTypes = hostTypeTriggers.map(
-        (item) => HostTypeTriggerOptions.find((option) => option.id === item)?.hostType
-      );
-      hostType = hostTypes ? hostTypes[0] : undefined;
+    let workingPath = inputs.projectPath ?? "";
+    const lang = resolveProgrammingLanguage(ctx);
+    if (lang === "csharp") {
```
IvanJobs, line 35: 1. Better to ensure if there's some enum can be used to replace the plain string 'csharp'.
2. I guess we should use `!==` here?

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
```diff
@@ -161,48 +138,21 @@ export class TeamsBotV2Impl {
...
-
-    if (lang === "js" || lang === "ts") {
-      bicepConfigs.push("node");
-    }
-    if (lang === "csharp") {
-      bicepConfigs.push("dotnet");
-    }
-
+    const lang = resolveProgrammingLanguage(ctx);
+    bicepConfigs.push(runtimeMap[lang]);
```
IvanJobs, line 144: Nit: Add error handling to improve robustness when `lang` is not found in runtimeMap.

packages/fx-core/src/plugins/resource/bot/v2/common.ts
```diff
@@ -0,0 +1,103 @@
...
+            const options = resolveTriggerOption(inputs);
+            templateScenarios.push(
+              TemplateProjectsScenarios.NOTIFICATION_FUNCTION_BASE_SCENARIO_NAME
+            );
+            options.map((option) => templateScenarios.push(triggerScenarioMap[option.trigger!]));
+          }
+          break;
+      }
+    });
+  } else {
```
IvanJobs, line 59: 1. botScenarios should be in type of array anyway.
2. The default template case should not be associated with `! isArray`.

packages/fx-core/src/plugins/resource/bot/v2/common.ts
```diff
@@ -0,0 +1,103 @@
...
+        case BotScenario.NotificationBot:
+          const notificationTriggerType = inputs[QuestionNames.BOT_HOST_TYPE_TRIGGER] as string[];
+          if (notificationTriggerType.includes(AppServiceOptionItem.id)) {
+            templateScenarios.push(TemplateProjectsScenarios.NOTIFICATION_RESTIFY_SCENARIO_NAME);
+          } else {
+            const options = resolveTriggerOption(inputs);
+            templateScenarios.push(
+              TemplateProjectsScenarios.NOTIFICATION_FUNCTION_BASE_SCENARIO_NAME
+            );
+            options.map((option) => templateScenarios.push(triggerScenarioMap[option.trigger!]));
```
IvanJobs, line 54: Nit: Better to add error handling.

packages/fx-core/src/plugins/resource/bot/v2/common.ts
```diff
@@ -0,0 +1,103 @@
...
+  } else {
+    templateScenarios.push(TemplateProjectsScenarios.DEFAULT_SCENARIO_NAME);
+  }
+  return templateScenarios;
+}
+
+export function resolveProgrammingLanguage(ctx: Context): string {
+  const lang = ctx.projectSetting.programmingLanguage;
+  if (!lang || !(lang in langMap)) {
+    throw new Error(`notification trigger '${lang}' is invalid.`);
```
IvanJobs, line 68: Localize error string.

packages/fx-core/src/plugins/resource/bot/v2/common.ts
```diff
@@ -0,0 +1,103 @@
...
+    });
+  } else {
+    templateScenarios.push(TemplateProjectsScenarios.DEFAULT_SCENARIO_NAME);
+  }
+  return templateScenarios;
+}
+
+export function resolveProgrammingLanguage(ctx: Context): string {
+  const lang = ctx.projectSetting.programmingLanguage;
+  if (!lang || !(lang in langMap)) {
```
IvanJobs, line 67: `!(lang.toLowerCase() in langMap)`?

packages/fx-core/src/plugins/resource/bot/v2/common.ts
```diff
@@ -0,0 +1,103 @@
...
+    throw new Error(`notification trigger '${lang}' is invalid.`);
+  }
+  return langMap[lang.toLocaleLowerCase()];
```
IvanJobs, line 70: I guess using `toLowerCase()` is good enough.

packages/fx-core/src/plugins/resource/bot/v2/common.ts
```diff
@@ -0,0 +1,103 @@
...
+}
+
+export function resolveTriggerOption(inputs: Inputs): HostTypeTriggerOptionItem[] {
+  const notificationTriggerType = inputs[QuestionNames.BOT_HOST_TYPE_TRIGGER];
```
IvanJobs, line 74: Inconsistency: Check L74

packages/fx-core/src/plugins/resource/bot/v2/mapping.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export const runtimeMap: { [key: string]: string } = {
+  js: "node",
+  ts: "node",
+  csharp: "dotnet",
+};
+
+export const serviceMap: { [key: string]: ServiceType } = {
+  [HostTypes.APP_SERVICE]: ServiceType.AppService,
+  [HostTypes.AZURE_FUNCTIONS]: ServiceType.Function,
```
IvanJobs, line 17: Nit: Functions vs Function, kind of inconsistency.

packages/fx-core/src/plugins/resource/bot/v2/common.ts
```diff
@@ -0,0 +1,103 @@
+// Copyright (c) Microsoft Corporation.
```
IvanJobs, line 1: Nit: Better to add basic UT cases for functions in this file.

IvanJobs (dismissed):
lgtm

packages/fx-core/src/plugins/resource/bot/v2/common.ts
```diff
@@ -0,0 +1,113 @@
...
+    const hostTypes = notificationTriggerType.map(
+      (item) => HostTypeTriggerOptions.find((option) => option.id === item)?.hostType
+    );
+    hostType = hostTypes ? hostTypes[0] : undefined;
+  }
+  return hostType ? hostType : HostTypes.APP_SERVICE;
+}
+
+export function resolveServiceType(ctx: Context): ServiceType {
+  const rawHostType =
```
eriolchan, line 107: Same comments as resolveProgrammingLanguage()

packages/fx-core/src/plugins/resource/bot/v2/enum.ts
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+export enum ProgrammingLanguage {
```
eriolchan, line 4: Where is this enum used?

packages/fx-core/src/plugins/resource/bot/v2/mapping.ts
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { ServiceType } from "../../../../common/azure-hosting/interfaces";
+import { BotNotificationTriggers } from "../../../solution";
+import { TemplateProjectsScenarios } from "../constants";
+import { HostTypes } from "../resources/strings";
+
+export const runtimeMap: { [key: string]: string } = {
```
eriolchan, line 9: I still recommend to use normal Map instead of indexable type

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
```diff
@@ -161,48 +140,24 @@ export class TeamsBotV2Impl {
...
-
   private getBicepConfigs(ctx: Context, inputs: Inputs): BicepConfigs {
-    const lang = this.resolveProgrammingLanguage(ctx);
-
     const bicepConfigs: BicepConfigs = [];
-
-    if (lang === "js" || lang === "ts") {
-      bicepConfigs.push("node");
+    const lang = resolveProgrammingLanguage(ctx);
+    if (runtimeMap[lang] === undefined) {
```
eriolchan, line 146: Usually we don't compare with undefined, but use below. But I still recommend to use normal Map and provide a method for it. Let's sync offline
```
if (!runtimeMap[lang])
```

packages/fx-core/src/plugins/resource/bot/v2/common.ts
```diff
@@ -0,0 +1,113 @@
...
+
+export function decideTemplateScenarios(ctx: Context, inputs: Inputs): string[] {
+  const isM365 = ctx.projectSetting?.isM365;
+  const templateScenarios = [];
+  if (isM365) {
+    templateScenarios.push(TemplateProjectsScenarios.M365_SCENARIO_NAME);
+    return templateScenarios;
+  }
+  const botScenarios = inputs?.[AzureSolutionQuestionNames.Scenarios];
+  if (botScenarios) {
```
eriolchan, line 40: Has some comments on this method

packages/fx-core/src/plugins/resource/bot/v2/common.ts
```diff
@@ -0,0 +1,113 @@
...
+  } else {
+    templateScenarios.push(TemplateProjectsScenarios.DEFAULT_SCENARIO_NAME);
+  }
+  return templateScenarios;
+}
+
+export function resolveProgrammingLanguage(ctx: Context): string {
+  const lang = ctx.projectSetting.programmingLanguage;
+  if (!lang || !(lang.toLowerCase() in langMap)) {
+    throw new Error(getLocalizedString("error.bot.InvalidLanguageError", lang));
```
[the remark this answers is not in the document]
IvanJobs: Yes, it will be shown in UI.

packages/fx-core/src/plugins/resource/bot/v2/mapping.ts
```diff
@@ -0,0 +1,68 @@
...
+  [HostTypes.AZURE_FUNCTIONS, ServiceType.Functions],
+]);
+
+const langMap: Map<string, string> = new Map<string, string>([
+  ["javascript", ProgrammingLanguage.Js],
+  ["typescript", ProgrammingLanguage.Ts],
+  ["csharp", ProgrammingLanguage.Csharp],
+]);
+
+const triggerScenarioMap: Map<string, string> = new Map<string, string>([
```
eriolchan, line 27: This could be as below to include all notification scnearios
```
const triggerScenarioMap: Map<string, string[]> = new Map<string, string[]>([
[
    BotNotificationTriggers.Http,
    [TemplateProjectsScenarios.NOTIFICATION_FUNCTION_TRIGGER_HTTP_SCENARIO_NAME, TemplateProjectsScenarios.NOTIFICATION_FUNCTION_BASE_SCENARIO_NAME],
  ],
  [
    BotNotificationTriggers.Timer,
    [TemplateProjectsScenarios.NOTIFICATION_FUNCTION_TRIGGER_TIMER_SCENARIO_NAME, TemplateProjectsScenarios.NOTIFICATION_FUNCTION_BASE_SCENARIO_NAME],
  ],
  [ 
     AppServiceOptionItem.id
    [TemplateProjectsScenarios.NOTIFICATION_RESTIFY_SCENARIO_NAME],
  ],
```

packages/fx-core/src/plugins/resource/bot/v2/common.ts
```diff
@@ -0,0 +1,83 @@
...
+        const notificationTriggerType = inputs[QuestionNames.BOT_HOST_TYPE_TRIGGER] as string[];
+        if (notificationTriggerType.includes(AppServiceOptionItem.id)) {
+          templateScenarios.push(TemplateProjectsScenarios.NOTIFICATION_RESTIFY_SCENARIO_NAME);
+        } else {
+          templateScenarios.push(
+            TemplateProjectsScenarios.NOTIFICATION_FUNCTION_BASE_SCENARIO_NAME
+          );
+          const notificationTriggerType = inputs[QuestionNames.BOT_HOST_TYPE_TRIGGER] as string[];
+          notificationTriggerType
+            .map((item) => HostTypeTriggerOptions.find((option) => option.id === item)!.trigger!)
```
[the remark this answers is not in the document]
eriolchan: It seems the logic of Notification bot can be as below
```
const notificationTriggerType = inputs[QuestionNames.BOT_HOST_TYPE_TRIGGER] as string[];
notificationTriggerType.map((item) => HostTypeTriggerOptions.find((option) => option.id === item)).forEach((option) => templateScenarios.push(...getTemplateScenario(option.trigger)));
break;
```
Once we have some change in the triggerScenarioMap of file packages/fx-core/src/plugins/resource/bot/v2/mapping.ts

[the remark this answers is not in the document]
yukun-dong: This will be refactored later.

eriolchan (approved):

PR #4929, opened 2022-05-06
chore(m365): add bot outlook channel configuration in bicep template

kuojianlu:
- update bicep version to install
- add bot outlook channel configuration in bicep template

### DIFF
--- a/packages/fx-core/src/plugins/solution/fx-solution/utils/depsChecker/bicepChecker.ts
+++ b/packages/fx-core/src/plugins/solution/fx-solution/utils/depsChecker/bicepChecker.ts
@@ -36,9 +36,9 @@ import { sendErrorTelemetryThenReturnError } from "../util";
 import { isBicepEnvCheckerEnabled } from "../../../../../common/tools";
 
 export const BicepName = "Bicep";
-export const installVersion = "v0.4";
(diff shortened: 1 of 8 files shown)


meifans (approved):
LGTM

blackchoey (commented):
can we also bump the version in projectSetting.json? Currently it's 2.1.0, we may change it to 2.2.0.

We would like to use this version to track some big changes in bicep, in case we need distinguish old and new projects in the future. You could reach @wenytang-ms if you don't know how to bump this version.

Also cc @jayzhang for awareness

kuojianlu:
> can we also bump the version in projectSetting.json? Currently it's 2.1.0, we may change it to 2.2.0.
> 
> We would like to use this version to track some big changes in bicep, in case we need distinguish old and new projects in the future. You could reach @wenytang-ms if you don't know how to bump this version.
> 
> Also cc @jayzhang for awareness

As offline discussed with @jayzhang, the `version` in `projectSettings.json` just indicates the schema of this file, but not the project folder structure. If we want to track the changes in bicep, how about adding a new key such as `solutionSettings.bicepTemplateVersion`?

kuojianlu:
Closed since it is not targted GA.

PR #4946, opened 2022-05-09
perf(version-checker): add project version checker middleware

SLdragon:
Add project version checker middleware to address this bug: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/14134470

Our extension currently supports project version < 3.0.0, if in the future, there is a project that project version >= 3.0.0, then it will show a non-modal dialog to tell user to upgrade the extension.

Below is an example when extension open a project with project version 3.1.0 which is not supported:

image


jayzhang (approved):

KennethBWSong (approved):

packages/fx-core/src/core/middleware/projectVersionChecker.ts
```diff
@@ -0,0 +1,67 @@
...
+  const currentProjectVersion = projectSettings.version;
+  if (currentProjectVersion) {
+    if (!semver.satisfies(currentProjectVersion, currentSupportProjectVersion)) {
+      return true;
+    }
+  }
+  return false;
+}
+
+async function showDialog(ctx: CoreHookContext) {
```
adashen, line 48: Is it possible to put show Dialog to vscode extension, show alert to cli, and maybe there is also UI needs in VS?

adashen (approved):

PR #4952, opened 2022-05-09
fix(vsc): empty project id in rare cases

tecton:
1. upgrade telemetry package.
2. set project id when first activated in a teamsfx project.

E2E TEST: N/A


packages/vscode-extension/src/commonlib/telemetry.ts
```diff
@@ -92,8 +92,10 @@ export class VSCodeTelemetryReporter extends vscode.Disposable implements Teleme
       properties = { ...this.sharedProperties, ...properties };
     }
 
-    const projectId = getProjectId();
-    properties[TelemetryProperty.ProjectId] = projectId ? projectId : "";
+    if (properties[TelemetryProperty.ProjectId] === "unknown") {
```
HuihuiWu-Microsoft, line 95: Is there telemetry event that are sent before you set 'unknown' to project-id? In this situation the property will not be attached(because the projectId property is undefined in this line), right?

tecton: The project-id is set to sharedProperties in handlers.ts:activate() using `ExtTelemetry.addSharedProperty(TelemetryProperty.ProjectId, projectId);` It's before any telemetry events so ideally every telemetry event has this property except it's not found in a teamsfx project. This line of code is to try to get project id again that should not happen in normal situation.

HuihuiWu-Microsoft: Yes, you're right. But I'm concerned if anyone adds telemetry event before handlers.activate(). In that case, the property-id will not be attached. Can we maybe use the statement `if (properties[TelemetryProperty.ProjectId] === "unknown" || properties[TelemetryProperty.ProjectId] === undefined)`  to avoid such condition?

tecton: Thank you! That's right and I've fixed it.

HuihuiWu-Microsoft (approved):

swatDong (approved):

PR #4979, opened 2022-05-11
fix: enhance log

nliu-ms:
work item: 14290560

differentiate log string and info window string

image
image


nliu-ms:
E2E TEST: N/A

HuihuiWu-Microsoft (approved):

PR #4980, opened 2022-05-11
perf(aad-manifest): automatically create or use existing scope id

SLdragon:
For existing aad app scenario, allow user only input "objectId", "clientId" and "clientSecret" without "accessAsUserScopeId"
```json
"auth": {
        "objectId": "58ad2e68-f442-4f6c-9282-xxx",
        "clientId": "c075d7a2-29c8-4d2b-8609-xxx",
        "clientSecret": "xxx",
        "accessAsUserScopeId": "xxx" // now this property is optional
    }
```
Automatically generate `accessAsUserScopeId` will following the sequence below:

1. Try to find it from "auth" config
2. Try to find it from state.xxx.json file
3. Try to find it from remote aad app
4. Generate a new one

### DIFF
--- a/packages/fx-core/src/plugins/resource/aad/aadAppClient.ts
+++ b/packages/fx-core/src/plugins/resource/aad/aadAppClient.ts
@@ -33,6 +33,7 @@ import { getAllowedAppIds } from "../../../common/tools";
 import { TOOLS } from "../../../core/globalVars";
 import { AADManifest } from "./interfaces/AADManifest";
 import { AadAppManifestManager } from "./aadAppManifestManager";
+import { v4 as uuidv4 } from "uuid";
 
 function delay(ms: number) {
   // tslint:disable-next-line no-string-based-set-timeout
@@ -105,9 +106,13 @@ export class AadAppClient {
     skip = false
   ): Promise<void> {
     try {
-      await AadAppClient.retryHanlder(stage, () =>
-        AadAppManifestManager.updateAadApp(TokenProvider.token as string, manifest)
-      );
+      await AadAppClient.retryHanlder(stage, async () => {
+        const preAuthorizedApplications = manifest.preAuthorizedApplications;
+        manifest.preAuthorizedApplications = [];
+        await AadAppManifestManager.updateAadApp(TokenProvider.token as string, manifest);
+        manifest.preAuthorizedApplications = preAuthorizedApplications;
+        await AadAppManifestManager.updateAadApp(TokenProvider.token as string, manifest);
+      });
     } catch (error) {
       if (skip) {
         const message = Messages.StepFailedAndSkipped(
@@ -233,9 +238,9 @@ export class AadAppClient {
     stage: string,
     objectId: string,
     clientSecret: string | undefined,
+    oauth2PermissionScopeId: string | undefined,
     graphTokenProvider?: GraphTokenProvider,
-    envName?: string,
-    skip = false
+    envName?: string
   ): Promise<ProvisionConfig> {
     let manifest: AADManifest;
     try {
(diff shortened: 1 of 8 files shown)


qinezh (dismissed):
LGTM. Approved for the env schema part.

packages/fx-core/src/plugins/resource/aad/plugin.ts
```diff
@@ -342,16 +345,13 @@ export class AadAppForTeamsImpl {
     return ResultFactory.Success();
   }
 
-  public async postProvisionUsingManifest(
-    ctx: PluginContext,
-    isLocalDebug = false
-  ): Promise<AadResult> {
+  public async postProvisionUsingManifest(ctx: PluginContext): Promise<AadResult> {
     TelemetryUtils.init(ctx);
     Utils.addLogAndTelemetryWithLocalDebug(
```
KennethBWSong, line 350: By making this change, I think we can not get any post local debug telemetry?

KennethBWSong, line 350: By making this change, I think we can not get any post local debug telemetry?

SLdragon: Thanks, fixed

KennethBWSong (approved):

qinezh (approved):

packages/fx-core/src/plugins/resource/aad/aadAppClient.ts
```diff
@@ -105,9 +106,13 @@ export class AadAppClient {
...
     try {
-      await AadAppClient.retryHanlder(stage, () =>
-        AadAppManifestManager.updateAadApp(TokenProvider.token as string, manifest)
-      );
+      await AadAppClient.retryHanlder(stage, async () => {
+        const preAuthorizedApplications = manifest.preAuthorizedApplications;
+        manifest.preAuthorizedApplications = [];
+        await AadAppManifestManager.updateAadApp(TokenProvider.token as string, manifest);
+        manifest.preAuthorizedApplications = preAuthorizedApplications;
+        await AadAppManifestManager.updateAadApp(TokenProvider.token as string, manifest);
```
adashen, line 114: will it fail, if there is already pre-authed apps?

packages/fx-core/src/plugins/resource/aad/plugin.ts
```diff
@@ -191,16 +191,18 @@ export class AadAppForTeamsImpl {
...
-      if (!skip) {
+      const existingOauth2PermissionScopeId = ctx.envInfo.config.auth?.accessAsUserScopeId;
+
+      if (!skip || !existingOauth2PermissionScopeId) {
         await DialogUtils.progress?.next(ProgressDetail.GetAadApp);
         config = await AadAppClient.getAadAppUsingManifest(
           telemetryMessage,
           manifest.id,
           config.password,
+          manifest.oauth2Permissions[0]?.id,
```
adashen, line 205: can we make sure oauth2Permissions[0] is access_as_user?

packages/fx-core/src/plugins/resource/aad/aadAppClient.ts
SLdragon: We first only update oauth2Permissions and set pre-auth apps with empty value, it will not fail. 
pre-auth apps should be update after oauth2permissions

packages/fx-core/src/plugins/resource/aad/plugin.ts
SLdragon: Oh, thanks, if user modify our template, oauth2Permissions[0] may be not access_as_user, will udpate this part

SLdragon: Updated

KennethBWSong (approved):

qinezh (approved):
Approved for the env schema.

PR #4981, opened 2022-05-11
feat: add notification dotnet sdk

swatDong:
- Add notification dotnet sdk to /packages/dotnet-sdk/src/TeamsFx/Conversation
- Update code owners as well
- Code comment added
- UT will be added in later PR

E2E TEST: N/A

### DIFF
--- a/packages/dotnet-sdk/src/TeamsFx/Conversation/Channel.cs
+++ b/packages/dotnet-sdk/src/TeamsFx/Conversation/Channel.cs
@@ -1,0 +1,104 @@
+﻿namespace Microsoft.TeamsFx.Conversation
+{
+    using Microsoft.Bot.Builder;
+    using Microsoft.Bot.Schema;
+    using Microsoft.Bot.Schema.Teams;
+
+    /// <summary>
+    /// An <see cref="INotificationTarget"/> that represents a team channel.
+    /// </summary>
+    /// <remarks>
+    /// It's recommended to get channels from <see cref="TeamsBotInstallation.GetChannelsAsync"/>.
+    /// </remarks>
+    public class Channel : INotificationTarget
+    {
+        /// <summary>
+        /// Constructor.
+        /// </summary>
+        /// <param name="parent">The parent <see cref="TeamsBotInstallation"/> where this channel is created from.</param>
+        /// <param name="info">Detailed channel information.</param>
+        /// <exception cref="ArgumentNullException">Throws if provided parameter is null.</exception>
+        /// <remarks>
+        /// It's recommended to get channels from <see cref="TeamsBotInstallation.GetChannelsAsync"/>.
+        /// </remarks>
+        public Channel(TeamsBotInstallation parent, ChannelInfo info)
+        {
+            Parent = parent ?? throw new ArgumentNullException(nameof(parent));
+            Info = info ?? throw new ArgumentNullException(nameof(info));
+        }
+
+        /// <summary>
+        /// The parent <see cref="TeamsBotInstallation"/> where this channel is created from.
+        /// </summary>
+        public TeamsBotInstallation Parent { get; private set; }
+
+        /// <summary>
+        /// Detailed channel information.
+        /// </summary>
+        public ChannelInfo Info { get; private set; }
+
+        /// <summary>
+        /// The type of target. For channel it's always <see cref="NotificationTargetType.Channel"/>.
+        /// </summary>
+        public NotificationTargetType Type { get => NotificationTargetType.Channel; }
+
+        /// <inheritdoc/>
+        public Task SendMessage(string message, CancellationToken cancellationToken = default)
+        {
+            return Parent.Adapter.ContinueConversationAsync
+            (
+                Parent.BotAppId,
+                Parent.ConversationReference,
+                async (context1, ct1) => {
+                    var conversation = NewConversation(context1);
+                    await Parent.Adapter.ContinueConversationAsync
+                    (
+                        Parent.BotAppId,
+                        conversation,
+                        (context2, ct2) => context2.SendActivityAsync(message, cancellationToken: ct2),
+                        ct1
+                    ).ConfigureAwait(false);
+                },
+                cancellationToken
+            );
+        }
+
+        /// <inheritdoc/>
+        public Task SendAdaptiveCard(object card, CancellationToken cancellationToken = default)
+        {
+            return Parent.Adapter.ContinueConversationAsync
+            (
+                Parent.BotAppId,
+                Parent.ConversationReference,
+                async (context1, ct1) => {
+                    var conversation = NewConversation(context1);
+                    await Parent.Adapter.ContinueConversationAsync
+                    (
+                        Parent.BotAppId,
+                        conversation,
+                        (context2, ct2) => context2.SendActivityAsync
+                        (
+                            MessageFactory.Attachment
+                            (
+                                new Attachment {
+                                    ContentType = "application/vnd.microsoft.card.adaptive",
+                                    Content = card,
+                                }
+                            ),
+                            cancellationToken: ct2),
+                        ct1
+                    ).ConfigureAwait(false);
+                },
+                cancellationToken
+            );
+        }
+
+        private ConversationReference NewConversation(ITurnContext context)
+        {
+            var reference = context.Activity.GetConversationReference();
+            var channelConversation = reference.Clone();
+            channelConversation.Conversation.Id = Info.Id ?? string.Empty;
+            return channelConversation;
+        }
+    }
+}
--- a/packages/dotnet-sdk/src/TeamsFx/Conversation/ConversationBot.cs
+++ b/packages/dotnet-sdk/src/TeamsFx/Conversation/ConversationBot.cs
@@ -48,7 +48,7 @@
 
             Adapter = options.Adapter ?? throw new ArgumentException("Adapter is null.", nameof(options));
 
-            if (options.Notification != null && options.Notification.Enabled)
+            if (options.Notification != null)
             {
                 Notification = new NotificationBot(Adapter, options.Notification);
             }
--- a/packages/dotnet-sdk/src/TeamsFx/Conversation/ConversationBot.cs
+++ b/packages/dotnet-sdk/src/TeamsFx/Conversation/ConversationBot.cs
@@ -1,0 +1,57 @@
+﻿namespace Microsoft.TeamsFx.Conversation
+{
+    using Microsoft.Bot.Builder;
+
+    /// <summary>
+    /// Provide utilities for bot conversation, including:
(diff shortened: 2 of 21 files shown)


packages/dotnet-sdk/src/TeamsFx/Conversation/Channel.cs
```diff
@@ -0,0 +1,104 @@
...
+                        (context2, ct2) => context2.SendActivityAsync(message, cancellationToken: ct2),
+                        ct1
+                    ).ConfigureAwait(false);
+                },
+                cancellationToken
+            );
+        }
+
+        /// <inheritdoc/>
+        public Task SendAdaptiveCard(object card, CancellationToken cancellationToken = default)
```
dooriya, line 67: What if the schema of the card object is not a valid adaptive card, then how user troubleshooting with such error. Sometime Teams may have no response if the adaptive card content is incorrect

packages/dotnet-sdk/src/TeamsFx/Conversation/NotificationMiddleware.cs
```diff
@@ -0,0 +1,130 @@
...
+                if ("teamDeleted".Equals(eventType, StringComparison.OrdinalIgnoreCase))
+                {
+                    return ActivityType.TeamDeleted;
+                }
+                else if ("teamRestored".Equals(eventType, StringComparison.OrdinalIgnoreCase))
+                {
+                    return ActivityType.TeamRestored;
+                }
+            }
+            else if ("message".Equals(activityType, StringComparison.OrdinalIgnoreCase))
```
dooriya, line 76: You can use the pre-defined type from bot-builder instead of hard-coded string: e.g.: `ActivityTypes.Message`, `ActivityTypes.ConversationUpdate`

packages/dotnet-sdk/src/TeamsFx/Conversation/NotificationOptions.cs
```diff
@@ -0,0 +1,39 @@
...
+{
+    /// <summary>
+    /// Options to initialize <see cref="NotificationBot"/>.
+    /// </summary>
+    public class NotificationOptions
+    {
+        /// <summary>
+        /// Whether to enable notification or not. Default true.
+        /// </summary>
+        public bool Enabled { get; set; } = true;
```
dooriya, line 11: Why default true? In TS SDK, this is default to false.

packages/dotnet-sdk/src/TeamsFx/Conversation/Channel.cs
swatDong: I tried an invalid card object and Teams show an empty border:
image

I believe the rendering logic is at Teams client side to hard to tell schema error here.
BTW, to provide any card validation utilities, I prefer using another sdk in user's code, not conversation sdk.

packages/dotnet-sdk/src/TeamsFx/Conversation/NotificationMiddleware.cs
swatDong: Good catch. Updated.

packages/dotnet-sdk/src/TeamsFx/Conversation/NotificationOptions.cs
swatDong: TS SDK makes `Enabled` out of this `NotificationOptions`, but only in `ConversationOptions`.
Here (.net) has no combined type, so making `Enabled` inside `NotificationOptions`, then I suggest default true since "*developer who creates `NotificationOptions` probably needs notification functionality*".

dooriya: "developer who creates NotificationOptions probably needs notification functionality".
If so, then since we don't need an "Enabled" flag in the `NotificationOptions`, since we are not expect the case that user created a `NotificationOptions` but set the `Enabled` to `false`?

swatDong: Yep... this looks weird, but want to align with `CommandOptions`. Will `CommandOptions` include `Enabled`?
- If yes, `NotificationOptions` should include as well.
- If no, does that mean an empty `CommandOptions` is acceptable?

packages/dotnet-sdk/src/TeamsFx/Conversation/ConversationBot.cs
```diff
@@ -0,0 +1,57 @@
...
+    {
+        /// <summary>
+        /// The bot adapter.
+        /// </summary>
+        public BotAdapter Adapter { get; private set; }
+
+        /// <summary>
+        /// The entrypoint of notification.
+        /// </summary>
+        public NotificationBot Notification { get; private set; }
```
JerryYangKai, line 31: Just a question, why a conversation bot including a notification bot

packages/dotnet-sdk/src/TeamsFx/Conversation/ConversationBot.cs
```diff
@@ -0,0 +1,57 @@
...
+    /// <para>
+    /// For notification, set <c>Notification.Storage</c> in <see cref="ConversationOptions"/> to use your own storage implementation.
+    /// </para>
+    /// </remarks>
+    public class ConversationBot
+    {
+        /// <summary>
+        /// The bot adapter.
+        /// </summary>
+        public BotAdapter Adapter { get; private set; }
```
JerryYangKai, line 26: So the adapter is inited outside the SDK,  I mean user created it and put it as an input?

packages/dotnet-sdk/src/TeamsFx/Microsoft.TeamsFx.csproj
```diff
@@ -28,6 +28,7 @@
       <_Parameter1>Microsoft.TeamsFx.Test</_Parameter1>
     </AssemblyAttribute>
     <PackageReference Include="Microsoft.AspNetCore.Components.Web" Version="6.0.0" />
+    <PackageReference Include="Microsoft.Bot.Builder" Version="4.14.1" />
```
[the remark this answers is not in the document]
swatDong: I just copy this from current .net bot template. I'm OK with 4.16.0.

packages/dotnet-sdk/src/TeamsFx/Conversation/NotificationOptions.cs
dooriya: For me, looks like remove `Enabled` would be more straightforward:
- This make the `NotificationBot`/`CommandBot` initialization more clean. (we don't need to check the `Enabled` in the constructor)
- This also simplifies the `ConversationBot` initialization. if user doesn't need to enable notification/command feature, then they don't need to create a `NotificationOptions`/`CommandOptions`, even a empty  `NotificationOptions`, `CommandOptions`.

packages/dotnet-sdk/src/TeamsFx/Conversation/ConversationBot.cs
swatDong: We'll generate the initializing part in template, so user can still focus on business logic.

JerryYangKai: OK, got you. Now the CloudAdapter is using the BotFrameworkAuthentication, it forced the appSettings to a fixed key name -- MicrosoftAppId, a very confusing name.  
What in my mind is put the init logic into SDK, user just using BOT_ID and BOT_PASSOWRD as input to our SDK. Like what we delivered in js SDK.  
Just an idea of ​​mine, you can think about it in the future

swatDong: I've tried following code to create BotFrameworkAuthentication:
``` csharp
// Prepare Configuration for ConfigurationBotFrameworkAuthentication
builder.Configuration["MicrosoftAppType"] = "MultiTenant";
builder.Configuration["MicrosoftAppId"] = builder.Configuration.GetSection("BOT_ID")?.Value;
builder.Configuration["MicrosoftAppPassword"] = builder.Configuration.GetSection("BOT_PASSWORD")?.Value;

// Create the Bot Framework Authentication to be used with the Bot Adapter.
builder.Services.AddSingleton<BotFrameworkAuthentication, ConfigurationBotFrameworkAuthentication>();
```

Then we can leverage the built-in logic of `ConfigurationBotFrameworkAuthentication` such as client cache and credential. But I feel like these codes should be put in template, not SDK.

swatDong: The key point is, if we put the init logic in user's code, it's easy for user to integrate with ASP.Net Core's DI (Service Provider).
But if we put the init logic in SDK, we actually hide the DI part so it's a bit complex for user to integrate with service provider.

JerryYangKai: Agreed! Let the authentication part outside SDK provide more ways for user to use.
Just a question, outside this pr. Does our js SDK has same logic or this is the different between these two SDK. (because I saw js SDK just put id and password as input)

swatDong: That's another story...
In js sdk we use `BotFrameworkAdapter` whose initialization is simple, so we are able to directly use id and password.
But in .net sdk, `BotFrameworkAdapter` is **obsoleted** so we need to move to `CloudAdapter`, whose initialization is a bit complex.

I have no idea why js `BotFrameworkAdapter` is not obsoleted. But if later we need to move to `CloudAdapter` in js as well, we may change the js initialization logic as well.

packages/dotnet-sdk/src/TeamsFx/Conversation/NotificationOptions.cs
swatDong: Removed `Enabled`.

packages/dotnet-sdk/src/TeamsFx/Microsoft.TeamsFx.csproj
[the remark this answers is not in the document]
swatDong: Updated to 4.16.0.

BTW, 4.16.0 brings `Microsoft.Azure` namespaces, so I need to update existing `Azure.Core` namespace to `global::Azure.Core`. Please review that part as well.

packages/dotnet-sdk/src/TeamsFx/Conversation/Channel.cs
dooriya: Agree to provide card utility in a separate SDK, like "@microsoft/adaptiveCards-tools" for TS/JS.

dooriya (approved):

JerryYangKai (approved):

PR #4989, opened 2022-05-11
chore(m365): remove msteams-react-base-component dependency from tab

kuojianlu:


swatDong (approved):

AndrewCraswell:
Got it. Looks like that hasn't trickled through all the libraries yet. My understanding is that now the Teamsfx React SDK needs to start using their own implementation, which hasn't happened yet

kuojianlu:
> Got it. Looks like that hasn't trickled through all the libraries yet. My understanding is that now the Teamsfx React SDK needs to start using their own implementation, which hasn't happened yet

Seems you replied in the wrong place :)

AndrewCraswell:
Thanks :)

I meant to reply in another issue, but it's a relevant question for here. It looks like this PR is adding an internal implementation of the `useTeams` hook which might be able to replace the `msteams-react-base-component` dependency. Do you know if there are plans in the future for the TeamsFx React SDK to start using this? Our team would like to remove that dependency, and are excited to see similar code getting merged in :P

wictorwilen:
As the original author of the `msteams-react-base-component` and the `useTeams` hook I find this implementation very sad. 

- It does not provide near all the flexibility and features as the original `useTeams` hook
- Since you're creating a new implementation it's not a good practice to use the same name for something that's already the de-facto standard for React based Teams apps, since it would collide with apps that wants to use both the TeamsFx things and the real `useTeams` hook
- It wouldn't have been to hard to merge the projects together in a smart way to support both users of TeamsFx and those who prefer not to use TeamsFx
- Also, it's very sad that this is essentially an exact copy (with some stripped code) and not even a re-write of the code I've written. I know it's under MIT so it's free to modify...but only if the original copyright and message is left intact.

swatDong:
This is a workaround when we just started to support m365 apps running on Teams/Outlook/Office, which use teams-js 2.0.
At the PR time, teams-js and msteams-react-base-component are in beta/preview state. Then we had this workaround to avoid some compatibility issues.

Since teams-js and msteams-react-base-component now have stable releases, we are glad to add it back.
BTW, this template only affects m365 apps. For normal teams app with more functionality, we are keeping on teams-js 1.x and msteams-react-base-component 3.x.

kuojianlu:
@wictorwilen Added back `msteams-react-base-component` in https://github.com/OfficeDev/TeamsFx/pull/5734. Thanks for your detailed explanation.

MuyangAmigo:
Hey @wictorwilen, I'm reaching out to you offline to see how we can collaborate and avoid such workarounds from happening in the future.

PR #4992, opened 2022-05-12
fix(m365): fix opening outlook channel config page no response

kuojianlu:
To fix https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/14423884.

E2E TEST: N/A

### DIFF
--- a/packages/cli/src/cmds/preview/launch.ts
+++ b/packages/cli/src/cmds/preview/launch.ts
@@ -127,16 +127,16 @@ export async function openUrlWithNewProfile(url: string): Promise<boolean> {
 
     const apps = [
       {
-        name: open.apps.firefox,
-        arguments: ["-profile", profileFolderPath],
-      },
-      {
         name: open.apps.chrome,
         arguments: [`--user-data-dir=${profileFolderPath}`],
       },
       {
         name: open.apps.edge,
         arguments: [`--user-data-dir=${profileFolderPath}`],
+      },
+      {
+        name: open.apps.firefox,
+        arguments: ["-profile", profileFolderPath],
       },
     ];
     for (const app of apps) {
--- a/packages/cli/src/cmds/preview/launch.ts
+++ b/packages/cli/src/cmds/preview/launch.ts
@@ -17,7 +17,7 @@ import {
   TelemetrySuccess,
 } from "../../telemetry/cliTelemetryEvents";
 import { Colors, LogLevel } from "@microsoft/teamsfx-api";
-import { getColorizedString } from "../../utils";
+import { getColorizedString, sleep } from "../../utils";
 import { ConfigFolderName } from "@microsoft/teamsfx-api";
 import { TempFolderManager } from "./tempFolderManager";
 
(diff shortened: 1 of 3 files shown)


packages/cli/src/cmds/preview/launch.ts
```diff
@@ -99,24 +99,52 @@ export async function openUrlWithNewProfile(url: string): Promise<boolean> {
...
+      {
+        name: open.apps.edge,
+        arguments: [`--user-data-dir=${profileFolderPath}`],
+      },
+      {
+        name: open.apps.firefox,
+        arguments: ["-profile", profileFolderPath],
+      },
+    ];
+    for (const app of apps) {
```
dooriya, line 142: Is that possible to select browser when preview, like the local debug?

a1exwang (approved):

swatDong (approved):

packages/cli/src/cmds/preview/launch.ts
```diff
@@ -99,24 +99,52 @@ export async function openUrlWithNewProfile(url: string): Promise<boolean> {
...
+            resolve(false);
+          }
+          resolve(true);
+        } catch {
+          resolve(false);
+        }
+      });
+    };
+
+    const apps = [
```
dooriya, line 128: Can we move this to global const so CLI and extension can share?

packages/cli/src/cmds/preview/launch.ts
```diff
@@ -99,24 +99,52 @@ export async function openUrlWithNewProfile(url: string): Promise<boolean> {
...
+        name: open.apps.firefox,
+        arguments: ["-profile", profileFolderPath],
+      },
+    ];
+    for (const app of apps) {
+      if (await tryToOpen(url, app)) {
+        return true;
+      }
+    }
+    return false;
```
dooriya, line 147: Will we provide error message to user if none of the expected browser found?

packages/cli/src/cmds/preview/launch.ts
kuojianlu: The UX for manually configuring Outlook channel is a workaround and will be removed once App Studio API is ready. So I just copy/paste this piece of code for simplicity.

packages/cli/src/cmds/preview/launch.ts
kuojianlu: If `openUrlWithNewProfile` fails, will directly open the url in the default browser, so no error message needed. Introducing `openUrlWithNewProfile` is to enforce users to always login in to bot framework portal.

packages/cli/src/cmds/preview/launch.ts
kuojianlu: The browser selected is only for sideloading Teams/Outlook.
For extension part, the browser type cannot be known outside debug provider. So we have to try each browser. And for CLI part, to keep consistent with the extension.

dooriya (approved):

packages/cli/src/cmds/preview/launch.ts
dooriya: Make sense. Thanks for the explanation~

PR #4994, opened 2022-05-12
chore(m365): update teams-js and teamsfx version and upgrade command wording

kuojianlu:
Resolve https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/14415478 and https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/14295094.


.github/scripts/sync-version.js
```diff
@@ -45,7 +45,7 @@ function updateFileDeps(file, deps) {
             if(!(semver.prerelease(semver.minVersion(dep_[key])).includes("alpha") || semver.prerelease(semver.minVersion(dep_[key])).includes("rc") || semver.prerelease(semver.minVersion(dep_[key])).includes("beta"))){
                 continue;
             }
-            if(key === "@microsoft/teamsfx" && dep_[key] === "0.6.0-beta.0") {
+            if(key === "@microsoft/teamsfx" && (dep_[key] === "0.6.0-beta.0" || dep_[key] === "0.7.0-beta.0")) {
```
swatDong, line 48: Do we have more common patten like "*-beta.0"? Or 0.7.0 is the latest one before GA?

kuojianlu: Good point. Updated to use `x.x.x-beta.x` or `^x.x.x-beta.x` pattern.

wenytang-ms (approved):

PR #4997, opened 2022-05-12
chore(vsc): cache telemetry event during deactivation

tecton:
Telemetry events are not sent in `deactivate()` function. Cache it locally and resend next time.

### DIFF
--- a/packages/vscode-extension/src/telemetry/extTelemetry.ts
+++ b/packages/vscode-extension/src/telemetry/extTelemetry.ts
@@ -5,7 +5,7 @@
 
 import * as vscode from "vscode";
 import { FxError, Stage, UserError } from "@microsoft/teamsfx-api";
-import { globalStateGet, globalStateUpdate } from "@microsoft/teamsfx-core";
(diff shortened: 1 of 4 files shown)


chagong (approved):

packages/vscode-extension/src/telemetry/extTelemetry.ts
```diff
@@ -186,6 +193,34 @@ export namespace ExtTelemetry {
...
   }
 
+  export async function cacheTelemetryEventAsync(
+    eventName: string,
+    properties?: { [p: string]: string }
+  ) {
+    const telemetryEvents = {
+      eventName: eventName,
+      properties: {
+        [TelemetryProperty.CorrelationId]: lastCorrelationId,
```
HuihuiWu-Microsoft, line 203: Is this correlation id related to deactivation event?

tecton: Yes, it helps to connect the deactivate event with last operation. Otherwise, the last operation has no "end" event.

HuihuiWu-Microsoft (approved):

PR #5000, opened 2022-05-12
docs: update comment in teamsSsoBot

KennethBWSong:
task: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_sprints/taskboard/AuthAndData/Microsoft%20Teams%20Extensibility/CY22-5.1?workitem=14434487


packages/fx-core/templates/plugins/resource/aad/auth/bot/js/sso/showUserInfo.js
```diff
@@ -1,16 +1,23 @@
 // This file implements a function to call Graph API with TeamsFx SDK to get user profile with SSO token.
+// You can modify this file to add your business with SSO token.
+// See auth/bot/README.md to learn more about adding new command to your bot.
```
blackchoey, line 3: Should we provide an aka link here? The file path may change when user moves it to another place.

blackchoey (approved):

PR #5005, opened 2022-05-13
chore(vsc): use codicon instead of custom SVG

tecton:
Custom SVG doesn't support high contrast theme.
Related bug: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/14431565

Effect:
image image image image


HuihuiWu-Microsoft:
I think this should be a bug of vs code since they provided 'dark' and 'light' for icon path. For light high contrast, they should use the dark version. But it's always better for us to use native icons if there is. I just want to say could we also open an issue to vscode team for other customized icons?

HuihuiWu-Microsoft (approved):

tecton:
> I think this should be a bug of vs code since they provided 'dark' and 'light' for icon path. For light high contrast, they should use the dark version. But it's always better for us to use native icons if there is. I just want to say could we also open an issue to vscode team for other customized icons?

You are right. I'll find whether there's an existing issue or workaround.

HuihuiWu-Microsoft (approved):

PR #5008, opened 2022-05-13
fix(sdk): fix adaptive card builder API

dooriya:
Add type constraint to the `TData` for adaptive card builder API.

E2E TEST: covered by UI test

### DIFF
--- a/packages/sdk/src/conversation/messageBuilder.ts
+++ b/packages/sdk/src/conversation/messageBuilder.ts
@@ -57,7 +57,7 @@ export class MessageBuilder {
    * @beta
    */
   public static attachAdaptiveCard<TData extends Record<string, unknown>>(
-    cardTemplate: any,
+    cardTemplate: unknown,
     data: TData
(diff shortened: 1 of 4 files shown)


swatDong (approved):

packages/sdk/src/conversation/messageBuilder.ts
```diff
@@ -56,7 +56,10 @@ export class MessageBuilder {
    *
    * @beta
    */
-  public static attachAdaptiveCard<TData>(cardTemplate: any, data: TData): Partial<Activity> {
+  public static attachAdaptiveCard<TData extends Record<string, unknown>>(
+    cardTemplate: any,
```
a1exwang, line 60:
```suggestion
    cardTemplate: unknown,
```
Similar to `data`, we don't use any property of `template` and just pass to adaptive card tools?

qinezh (approved):

swatDong (approved):

PR #5011, opened 2022-05-13
fix(vsc): reorder logic for default option in quickpick

tecton:
Related bug: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/14414796
Fixing effect:
1. choose 'Bot'.
2. go back.
image

E2E TEST: N/A


HuihuiWu-Microsoft:
If user selected the item in second category, will the default item in the top of whole list or just the top of current category?

tecton:
> If user selected the item in second category, will the default item in the top of whole list or just the top of current category?

You can see the image in description. "Bot" is the item in second category, it is in the top of whole list and its category is also moved to top position.

HuihuiWu-Microsoft (approved):

PR #5021, opened 2022-05-13
feat(sdk): add command response dotnet sdk

dooriya:
UT will be added in another PR.

E2E TEST: N/A


packages/dotnet-sdk/src/TeamsFx/Conversation/ActivityCommandResponse.cs
```diff
@@ -0,0 +1,38 @@
...
+        /// <exception cref="ArgumentNullException"><paramref name="activity"/> is null.</exception>
+        public ActivityCommandResponse(IActivity activity)
+        {
+            Activity = activity ?? throw new ArgumentNullException(nameof(activity));
+        }
+
+        /// <inheritdoc/>
+        public async Task SendResponseAsync(ITurnContext turnContext, CancellationToken cancellationToken = default)
+        {
+            await turnContext.SendActivityAsync(Activity, cancellationToken);
```
swatDong, line 35: `ConfigureAwait(false)` for all await operation.
(Fix all warnings/suggestions provided by VS itself.)

packages/dotnet-sdk/src/TeamsFx/Conversation/RegExpTrigger.cs
```diff
@@ -0,0 +1,37 @@
...
+        /// <exception cref="ArgumentNullException"><paramref name="pattern"/> is null.</exception>
+        public RegExpTrigger(string pattern)
+        {
+            Pattern = pattern ?? throw new ArgumentNullException(nameof(pattern));
+        }
+
+        /// <inheritdoc/>
+        public bool ShouldTrigger(string input)
+        {
+            var regex = new Regex(Pattern, RegexOptions.IgnoreCase);
```
swatDong, line 33: How about directly accepting `Regex` from user? To let caller to determine options and timeout.
E.g., `ITriggerPattern` has one method `ShouldTrigger`. Then, `StringTrigger` takes `string` as input, while `RegexTrigger` takes `Regex` as input.

packages/dotnet-sdk/src/TeamsFx/Conversation/ActivityCommandResponse.cs
dooriya: Good catch and already fixed.

packages/dotnet-sdk/src/TeamsFx/Conversation/RegExpTrigger.cs
dooriya: Agree, that would be more flexible for developers.

packages/dotnet-sdk/src/TeamsFx/Conversation/CommandTriggerType.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.TeamsFx.Conversation
+{
+    /// <summary>
+    /// Type of trigger pattern used to trigger a <see cref="ITeamsCommandHandler"/> instance.
+    /// </summary>
+    public enum CommandTriggerType
```
swatDong, line 9: So, is this trigger type redundant now?

dooriya: Yeah~ Let me remove it later.

swatDong (approved):

PR #5024, opened 2022-05-15
test(e2e): fix bot e2e test to exclude aad plugin

dilin-MS:
Fix ga branch E2E test failure.

Root cause:
After adding the AAD manifest feature,
1. some bot project does not have aad plugin by default and should not check aad output properties in e2e test.
2. keyvault provision bicep file should consider m365 client secret parameter as optional.


dilin-MS:
Close this PR since Bowen has a same PR open.

PR #5025, opened 2022-05-16
fix: cannot open readme from quick start for spfx sample

HuihuiWu-Microsoft:
ADO: 
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/14431825

E2E TEST: will add ut later

### DIFF
--- a/packages/vscode-extension/src/handlers.ts
+++ b/packages/vscode-extension/src/handlers.ts
@@ -1650,10 +1650,10 @@ export async function openReadMeHandler(args: any[]) {
     const workspaceFolder = workspace.workspaceFolders[0];
     const workspacePath: string = workspaceFolder.uri.fsPath;
     let targetFolder: string | undefined;
-    if (isSPFxProject(workspacePath)) {
-      targetFolder = `${workspacePath}/SPFx`;
-    } else if (await getIsFromSample()) {
+    if (await getIsFromSample()) {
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/handlers.ts
```diff
@@ -1650,10 +1650,10 @@ export async function openReadMeHandler(args: any[]) {
     const workspaceFolder = workspace.workspaceFolders[0];
     const workspacePath: string = workspaceFolder.uri.fsPath;
     let targetFolder: string | undefined;
-    if (isSPFxProject(workspacePath)) {
-      targetFolder = `${workspacePath}/SPFx`;
-    } else if (await getIsFromSample()) {
+    if (await getIsFromSample()) {
       openSampleReadmeHandler(args);
+    } else if (isSPFxProject(workspacePath)) {
```
tecton, line 1655: I've got an idea that might help to fix this in long term: start searching from predefined folder in sample/template (e.g. workspace/SPFx) and if not found, search the parent folder until the project root. In this way, it can handle different project structures that have README in different folders.

HuihuiWu-Microsoft: For samples, all readme files exist under root folder, I think we don't need to iterate to search for readme file.

tecton (approved):

PR #5044, opened 2022-05-17
chore: update frontend plugin and change template structure

JerryYangKai:


templates/tab/csharp/default/appsettings.Development.json.tpl
```diff
@@ -6,15 +6,12 @@
       "Microsoft.Hosting.Lifetime": "Information"
     }
   },
-  "AllowedHosts": "*"{{#IS_TAB}},
```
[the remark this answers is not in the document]
JerryYangKai: Good catch!

templates/tab/csharp/default/appsettings.json.tpl
```diff
@@ -6,15 +6,12 @@
       "Microsoft.Hosting.Lifetime": "Information"
     }
   },
-  "AllowedHosts": "*"{{#IS_TAB}},
```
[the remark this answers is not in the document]
JerryYangKai: Good catch!

templates/tab/csharp/default/ProjectName.csproj.tpl
```diff
@@ -13,14 +13,8 @@
     <PackageReference Include="Microsoft.AspNetCore.Authentication.JwtBearer" Version="6.0.0" />
     <PackageReference Include="Microsoft.Graph" Version="4.10.0" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
```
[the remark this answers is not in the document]
JerryYangKai: This file need to change name, the placeholder is in the file name

templates/msgext/csharp/default/GettingStarted.txt
```diff
@@ -13,27 +13,17 @@ Quick start
 
 Before you run or Start Debugging for the first time, follow these steps:
 
-1. If your project only contains a Tab capability, skip to step #3.
```
eriolchan, line 16: Why do we move ME template from bot folder to msgext?

JerryYangKai: Because in the future msgext is a indepent capability, compare with bot.
Since message extension is no longer including bot, also bot no longer including msgext. 
I think create a new folder named msgext is make sense.

JerryYangKai: Also, we plan to move msgext (vs code version) indenpent too.

packages/fx-core/src/plugins/resource/frontend/dotnet/constants.ts
```diff
@@ -54,15 +52,19 @@ export class DotnetCommands {
     `dotnet publish --configuration Release --runtime ${runtime} --self-contained`;
 }
 
+export class ScaffoldInfo {
```
eriolchan, line 55: Since there is already an interface of CodeTemplateInfo, why don't we leverage this?

packages/fx-core/src/plugins/resource/frontend/dotnet/ops/scaffold.ts
```diff
@@ -17,19 +17,18 @@ import {
...
 export async function scaffoldFromZipPackage(
   componentPath: string,
-  templateInfo: TemplateInfo
+  variable: { [key: string]: string }
 ): Promise<void> {
   await scaffoldFromTemplates({
-    group: templateInfo.group,
-    lang: templateInfo.language,
-    scenario: templateInfo.scenario,
+    group: ScaffoldInfo.group,
```
eriolchan, line 26: Why this value of group, lang, scenario are hardcoded?

packages/fx-core/src/plugins/resource/frontend/dotnet/plugin.ts
```diff
@@ -119,14 +116,8 @@ export class DotnetPluginImpl implements PluginImpl {
     if (!ctx.projectSettings) {
       throw new NoProjectSettingError();
     }
-
-    const selectedCapabilities = (ctx.projectSettings?.solutionSettings as SolutionSettings)
```
eriolchan, line 123: We can still create an instance of CodeTemplateInfo and pass it to scaffoldFromZipPackage(), instead of hardcode the logic in scaffoldFromZipPackage(). Because according to the method name scaffoldFromZipPackage(), it is a common method.

packages/fx-core/src/plugins/resource/frontend/dotnet/plugin.ts
```diff
@@ -242,8 +233,6 @@ export class DotnetPluginImpl implements PluginImpl {
     let clientSecret = Placeholders.clientSecret;
     let tenantId = "";
     let oauthAuthority = Placeholders.oauthAuthority;
```
eriolchan, line 235: There is a Code scanning warning here, please fix it as GitHub suggested.
image

JerryYangKai: Got it

packages/fx-core/src/plugins/resource/frontend/dotnet/plugin.ts
JerryYangKai: Agreed

templates/msgext/csharp/default/GettingStarted.txt
JerryYangKai: As discussed, change the folder path

packages/fx-core/src/plugins/resource/frontend/dotnet/plugin.ts
eriolchan: I mean the CodeTemplateInfo class in the packages/fx-core/src/plugins/resource/bot/v2/interface/codeTemplateInfo.ts file, @hund030 if CodeTemplateInfo is a common interface for scaffolding, maybe we can refactor it later to avoid each plugin create a similar class for scaffold source code

eriolchan (approved):

packages/fx-core/src/plugins/resource/frontend/dotnet/plugin.ts
JerryYangKai: @eriolchan  I got you mean. 
Because the codeTemplateInfo is in bot plugin, so what I do is let the existing templateInfo in tab plugin back. 
I agreed with you about move this interface into the common pack.  I have not done this because this pr is inlcuding too much things. 
I can create a new PR to do this refactor,  also I think we should have a sync with Zhijie first.

Issue #5048, opened 2022-05-17
[cli]Provision failed for the first time with user defined config


nliu-ms:
**Describe the bug**
When there are some user-defined configs in manifest template file, and execute provision. Teamsfx doesn't load the latest content of config file, which causes validation error. And when executing provision for the second time, it works.

**To Reproduce**
1. Scaffold a project
2. Update `manifest.template.json` and `config.xx.json` with user-defined config
3. Run provision, it fails
4. Run provision again, it succeeds

**Expected behavior**

**Screenshots**
![image](https://user-images.githubusercontent.com/71362691/168734340-3f50aa2f-e9a0-4e90-a11a-86e4601a1101.png)

**VS Code Extension Information (please complete the following information):**

**CLI Information (please complete the following information):**
 - Version 0.14.0

**Additional context**

PR #5057, opened 2022-05-17
fix: disable HTML escape to avoid user-defined config render error

nliu-ms:
resolve issue: https://github.com/OfficeDev/TeamsFx/issues/5048

image


nliu-ms:
E2E TEST: N/A

HuihuiWu-Microsoft (approved):

PR #5058, opened 2022-05-17
fix(bot): suggest user run failed command

hund030:
E2E TEST: https://github.com/OfficeDev/TeamsFx/blob/dev/packages/cli/tests/e2e/bot/CommandExecutionIsUserError.ts#L35

### DIFF
--- a/packages/fx-core/src/plugins/resource/bot/languageStrategy.ts
+++ b/packages/fx-core/src/plugins/resource/bot/languageStrategy.ts
@@ -138,9 +138,13 @@ export class LanguageStrategy {
       //Typescript needs tsc build before deploy because of windows app server. other languages don"t need it.
       try {
         await utils.execute("npm install", packDir);
+      } catch (e) {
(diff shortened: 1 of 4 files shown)


packages/fx-core/src/plugins/resource/bot/languageStrategy.ts
```diff
@@ -138,9 +138,13 @@ export class LanguageStrategy {
       //Typescript needs tsc build before deploy because of windows app server. other languages don"t need it.
       try {
         await utils.execute("npm install", packDir);
+      } catch (e) {
```
eriolchan, line 141: I think it is OK to combine "npm install and npm run build" and ask user to try it again, so we don't need to separate them into two try...catch. What do you think?

hund030: Both ok for me. Let's revert the change.

eriolchan (approved):

PR #5072, opened 2022-05-18
chore: unify m365 login

xiaolang124:
1.Add m365 login api.
2.Add m365 login implements for VsCode, CLI.

Old appStudio, graph and SPFx logins will be replaced by m365 login.


packages/api/src/utils/login.ts
```diff
@@ -228,9 +236,45 @@ export type SubscriptionInfo = {
...
+export declare type TokenRequest = {
+  scopes: Array<string>;
+  showDialog?: boolean;
+};
+export type LoginStatus = {
+  status: string;
+  token?: string;
+  accountInfo?: Record<string, unknown>;
+};
+export interface M365TokenProvider {
```
swatDong, line 248: Any document comment of this new interface, similar to the old `AppStudioTokenProvider`

packages/api/src/utils/login.ts
```diff
@@ -228,9 +236,45 @@ export type SubscriptionInfo = {
...
   graphTokenProvider: GraphTokenProvider;
+  /**
+   * @deprecated The method should not be used, please update to M365TokenProvider
+   */
   appStudioToken: AppStudioTokenProvider;
+  /**
+   * @deprecated The method should not be used, please update to M365TokenProvider
+   */
   sharepointTokenProvider: SharepointTokenProvider;
+  m365TokenProvider: M365TokenProvider;
```
swatDong, line 279: For further upgrading, any side-effect if partial upgrading?
E.g., if some components call `AppStudioTokenProvider` while others call `M365TokenProvider`, will that causes multiple logins?

packages/fx-core/src/common/basicLogin.ts
```diff
@@ -0,0 +1,54 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+"use strict";
+
+import { FxError, LoginStatus, ok, Result, TokenRequest } from "@microsoft/teamsfx-api";
+
+export abstract class BasicLogin {
```
swatDong, line 8: Any concern if moving this to api package?

packages/cli/src/commonlib/m365Login.ts
```diff
@@ -0,0 +1,135 @@
...
+  },
+};
+
+export class M365Login extends BasicLogin implements M365TokenProvider {
+  private static instance: M365Login;
+  private static codeFlowInstance: CodeFlowLogin;
+
+  private constructor() {
+    super();
+    M365Login.codeFlowInstance = new CodeFlowLogin([], config, SERVER_PORT, "m365");
```
swatDong, line 55: Does this mean we deprecate the old token cache file?

packages/cli/src/commonlib/m365LoginUserPassword.ts
```diff
@@ -0,0 +1,130 @@
...
+
+    const config: AxiosRequestConfig = {
+      method: "post",
+      url: `https://login.microsoftonline.com/${
+        cfg.M365_TENANT_ID || "organizations"
+      }/oauth2/v2.0/token`,
+      headers: {
+        "Content-Type": "application/x-www-form-urlencoded",
+        Cookie:
+          "fpc=AmzaQu9yHbpLtMD2LmHazdRCGxwGAQAAAIW47NcOAAAA; x-ms-gateway-slice=estsfd; stsservicecookie=estsfd",
```
swatDong, line 64: what's this cookie for, hard-coded value?

packages/vscode-extension/src/commonlib/codeFlowLogin.ts
```diff
@@ -287,6 +294,57 @@ export class CodeFlowLogin {
     }
   }
 
+  async getTokenByScopes(scopes: Array<string>, refresh = true): Promise<Result<string, FxError>> {
+    try {
+      if (!this.account) {
+        const accessToken = await this.login();
```
a1exwang, line 300: Also need to pass `scopes` to `login()`? Otherwise, `login()` calls `this.pca!.acquireTokenByCode(this.scopes)`, the `scopes` will be wrong. (uses `this.scopes` but it should be `scopes` from argument)

packages/cli/src/commonlib/m365Login.ts
swatDong: Oh, I see, it's controlled by cache name, not account name.

packages/vscode-extension/src/commonlib/codeFlowLogin.ts
```diff
@@ -287,6 +294,57 @@ export class CodeFlowLogin {
...
+        return ok(accessToken);
+      } else {
+        return this.pca!.acquireTokenSilent({
```
a1exwang, line 303: Probably not related to this PR (maybe change in a separate PR), but I see many `this.pca!`, `this.scopes!`. Is it better to just declare `pca` as `pca: PublicClientApplication` because `pca` can never be `undefined`. It is initialized in constructor. I think it is better to avoid unsafe operations like this as much as possible.

packages/cli/src/commonlib/m365Login.ts
swatDong: But what's the side effect if changing account name from `appStudio` to `m365`?

packages/cli/src/commonlib/codeFlowLogin.ts
```diff
@@ -298,6 +319,57 @@ export class CodeFlowLogin {
...
+      if (!this.account) {
+        const accessToken = await this.login();
+        return ok(accessToken);
+      } else {
+        return this.pca!.acquireTokenSilent({
+          account: this.account,
+          scopes: scopes,
+          forceRefresh: false,
+        })
+          .then((response) => {
```
a1exwang, line 336: Is it better to unify promise code with `async`/`await` to make to code more readable? I think it is a bit hard to read when mixing `async`/`await` and `then`/`catch` together.

packages/cli/src/commonlib/codeFlowLogin.ts
```diff
@@ -298,6 +319,57 @@ export class CodeFlowLogin {
...
+          .catch(async (error) => {
+            CliCodeLogInstance.necessaryLog(
+              LogLevel.Error,
+              "[Login] silent acquire token : " + error.message
+            );
+            if (!(await checkIsOnline())) {
+              return error(CheckOnlineError());
+            }
+            await this.logout();
+            (this.msalTokenCache as any).storage.setCache({});
```
a1exwang, line 352: What is the behavior of `logout` and `setCache({})`? Will this clean up the token only for the specified scopes or all tokens in this cache?

packages/api/src/utils/login.ts
xiaolang124: Thanks for the advice, already add.

packages/api/src/utils/login.ts
xiaolang124: Normally, all these logins use the same cache logic, they should share the login status.

packages/fx-core/src/common/basicLogin.ts
xiaolang124: Thanks for the advice, move it to API package.

packages/cli/src/commonlib/m365Login.ts
xiaolang124: My bad, this should be "appStudio". Because `CodeFlowLogin` use this name to manipulate cache file. If we change from `appStudio` to `m365`, the user need to login again if he has already signed in.

packages/cli/src/commonlib/m365LoginUserPassword.ts
xiaolang124: This has been removed, this does not have any effect on username password login.

packages/vscode-extension/src/commonlib/codeFlowLogin.ts
xiaolang124: Thanks for the advice, I add scopes to login. By the way, the `scopes` will be removed in later PR.

packages/vscode-extension/src/commonlib/codeFlowLogin.ts
xiaolang124: Fixed.

packages/cli/src/commonlib/codeFlowLogin.ts
xiaolang124: Fixed.

packages/cli/src/commonlib/codeFlowLogin.ts
xiaolang124: All scopes will be removed. Because multi `msal` versions may have different cache logic, they may work in one version but not work in the other version.

swatDong (approved):

a1exwang (dismissed):

a1exwang (approved):

a1exwang (approved):

LongOddCode (approved):

packages/cli/src/commonlib/codeFlowLogin.ts
```diff
@@ -43,6 +52,12 @@ class ErrorMessage {
   static readonly loginPortConflictTitle = "LoginPortConflict";
   static readonly loginPortConflictDescription = "Timeout waiting for port. Try again.";
   static readonly loginComponent = "login";
+  static readonly checkOnlineFailTitle = "CheckOnlineFail";
+  static readonly checkOnlineFailDetail =
```
HuihuiWu-Microsoft, line 56: Why have this string while it's already in package.nls.json? As error.message, you can get default string using the one in package.nls.json.

packages/cli/src/commonlib/codeFlowLogin.ts
```diff
@@ -43,6 +52,12 @@ class ErrorMessage {
...
+    "You appear to be offline. Please check your network connection.";
+  static readonly loginUsernamePasswordFailTitle = "UsernamePasswordLoginFail";
+  static readonly loginUsernamePasswordFailDetail =
```
HuihuiWu-Microsoft, line 59: Same here

packages/vscode-extension/src/commonlib/m365Login.ts
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    if (tokenRes.isOk()) {
+      return ok(tokenRes.value);
+    } else {
+      return tokenRes;
+    }
+  }
+
+  async getJsonObject(
```
HuihuiWu-Microsoft, line 134: Does this mean we should write following logics where we need a sharepoint token?
1. get graphReadUser token
2. use the token to get spfx tenenat and obtain sharepoint token scope
3. use the scope to request sharepoint token

Alive-Fish (approved):
LGTM

packages/vscode-extension/src/commonlib/m365Login.ts
xiaolang124: The process is right. We can follow this code to create a spfx token.
```
const graphTokenRes = await M365TokenInstance.getAccessToken({scopes: GraphReadUserScopes});
  if (graphTokenRes.isOk()) {
    const spfxToken = M365TokenInstance.getAccessToken({scopes: SPFxScopes(await getSPFxTenant(graphTokenRes))});
  }
```

packages/vscode-extension/package.nls.json
```diff
@@ -40,6 +40,8 @@
     "teamstoolkit.codeFlowLogin.loginTimeoutTitle": "LoginTimeout",
     "teamstoolkit.codeFlowLogin.resultFileNotFound": "Result file not found.",
     "teamstoolkit.codeFlowLogin.silentAcquireToken": "Failed to retrieve token silently: %s",
+    "teamstoolkit.codeFlowLogin.checkOnlineFailTitle": "CheckOnlineFail",
```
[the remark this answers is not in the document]
xiaolang124: Good point, but for now there are lots of these word patterns in this file. Before localizing user information, it looks like we use these to be `constant` file.

KennethBWSong (approved):

packages/vscode-extension/package.nls.json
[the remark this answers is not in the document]
HuihuiWu-Microsoft: Sure. For the existing strings with the same pattern, I'll summarize and ask owners to cleanup later.

HuihuiWu-Microsoft (approved):
LGTM

blackchoey (approved):
looks good for sql part

swatDong (approved):

XiaofuHuang (approved):

XiaofuHuang (approved):

IvanJobs (approved):
lgtm for bot

PR #5081, opened 2022-05-18
fix(vsc): remove Azure signing in status on UI

tecton:
Azure account only shows "signed out" and "signed in" on UI to prevent forever signing in status.
Will restore until "Azure Account" extension has fixed this issue.

Related bug: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/14474430

E2E TEST: N/A

### DIFF
--- a/packages/vscode-extension/src/treeview/account/accountTreeViewProvider.ts
+++ b/packages/vscode-extension/src/treeview/account/accountTreeViewProvider.ts
@@ -110,7 +110,8 @@ async function azureAccountStatusChangeHandler(
       envTreeProviderInstance.refreshRemoteEnvWarning();
     }
   } else if (status === "SigningIn") {
-    instance.azureAccountNode.setSigningIn();
+    // "Azure Account" extension only sends SigningIn signal without SignededOut in 0.10.x, so remove this status change until it's fixed.
+    // instance.azureAccountNode.setSigningIn();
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/treeview/account/accountTreeViewProvider.ts
```diff
@@ -110,7 +110,8 @@ async function azureAccountStatusChangeHandler(
       envTreeProviderInstance.refreshRemoteEnvWarning();
     }
   } else if (status === "SigningIn") {
-    instance.azureAccountNode.setSigningIn();
+    // "Azure Account" extension only sends SigningIn signal without SignededOut in 0.10.x, so remove this status change until it's fixed.
```
HuihuiWu-Microsoft, line 113: Since we don't block user after user clicks 'Sign in to Azure' now, user may click multiple times before logged in. Have you tested related scenarios to make sure this won't introduce other issues?

HuihuiWu-Microsoft (approved):

packages/vscode-extension/src/treeview/account/accountTreeViewProvider.ts
tecton: Good point, thanks. Users will be blocked by modal prompt and if they click during signing in, it will start a new progress.

PR #5090, opened 2022-05-19
chore: update postlocaldebug in bot dotnet plugin

yukun-dong:


hund030:
Have we already removed this from frontend plugin?

JerryYangKai:
@hund030  I removed this logic in last PR

packages/fx-core/src/plugins/resource/bot/dotnet/errors.ts
```diff
@@ -0,0 +1,40 @@
...
+    suggestions: string[],
+    helpLink?: string,
+    innerError?: Error
+  ) {
+    super(errorType, code, messages, suggestions, helpLink);
+    this.innerError = innerError;
+  }
+}
+
+export class FileIOError extends DotnetPluginError {
```
eriolchan, line 28: Since it is still Bot plugin but dotnet implementation, is that possible to leverage the existing PluginError or FileIOError from bot level instead of creating new class?

packages/fx-core/src/plugins/resource/bot/dotnet/plugin.ts
```diff
@@ -51,6 +52,22 @@ export class DotnetBotImpl extends TeamsBotImpl {
     return ResultFactory.Success(result);
   }
 
+  public async postLocalDebug(context: PluginContext): Promise<FxResult> {
+    await super.postLocalDebug(context);
+    const appSettingsPath = path.join(context.root, PathInfo.appSettingDevelopment);
+    try {
+      let appSettings = await fs.readFile(appSettingsPath, "utf-8");
```
eriolchan, line 59: Not quite understand the logic, why we replace a regular expression with a placeholder? Are they the same thing? @JerryYangKai

packages/fx-core/src/plugins/resource/bot/dotnet/plugin.ts
```diff
@@ -51,6 +52,22 @@ export class DotnetBotImpl extends TeamsBotImpl {
...
+      const botId = AppSettingsPlaceholders.botId;
+      const botPassword = AppSettingsPlaceholders.botPassword;
+      appSettings = appSettings.replace(RegularExpr.botId, botId);
```
hund030, line 62: Replace $botId$ with $botId$?
I think we need to read bot id from local envInfo and replace the placeholder with actual value.

packages/fx-core/src/plugins/resource/bot/dotnet/plugin.ts
```diff
@@ -57,8 +59,23 @@ export class DotnetBotImpl extends TeamsBotImpl {
     const appSettingsPath = path.join(context.root, PathInfo.appSettingDevelopment);
     try {
       let appSettings = await fs.readFile(appSettingsPath, "utf-8");
-      const botId = AppSettingsPlaceholders.botId;
-      const botPassword = AppSettingsPlaceholders.botPassword;
+      let botId = "";
+      let botPassword = "";
+      if (isConfigUnifyEnabled()) {
```
hund030, line 64: This feature flag is always true. I don't think we still need this.

JerryYangKai:
Do we need a unit test? I think post local debug has not covered.

yukun-dong:
> Do we need a unit test? I think post local debug has not covered.

Thanks for reminding. I will see if I can add some.

packages/fx-core/src/plugins/resource/bot/dotnet/plugin.ts
```diff
@@ -51,6 +53,31 @@ export class DotnetBotImpl extends TeamsBotImpl {
     return ResultFactory.Success(result);
   }
 
+  public async postLocalDebug(context: PluginContext): Promise<FxResult> {
+    await super.postLocalDebug(context);
+    const appSettingsPath = path.join(context.root, PathInfo.appSettingDevelopment);
+    try {
+      let appSettings = await fs.readFile(appSettingsPath, "utf-8");
+      const botId = context.envInfo.state.get(PluginNames.BOT)?.get(DependentPluginInfo.botId);
```
eriolchan, line 61: I wonder how does this code know to get botId and Bot password for local settings?

packages/fx-core/src/plugins/resource/bot/dotnet/constants.ts
```diff
@@ -11,4 +11,15 @@ export class PathInfo {
     "bicep"
   );
   public static readonly ProvisionModuleTemplateFileName = "botServiceProvision.template.bicep";
+  static readonly appSettingDevelopment = "appsettings.Development.json";
+}
+
+export class DependentPluginInfo {
```
eriolchan, line 17: Since the logic exist in Frontend plugin before, so it use dependentPluginInfo. But as we move the logic to Bot plugin itself, do we still need this?

packages/fx-core/src/plugins/resource/bot/dotnet/plugin.ts
```diff
@@ -51,6 +53,31 @@ export class DotnetBotImpl extends TeamsBotImpl {
...
+      const botPassword = context.envInfo.state
+        .get(PluginNames.BOT)
+        ?.get(DependentPluginInfo.botPassword);
+      if (!botId && !botPassword) {
```
eriolchan, line 65: What return success when botId and botPassword is null?

packages/fx-core/src/plugins/resource/bot/dotnet/constants.ts
yukun-dong: This class is removed.

eriolchan (approved):

PR #5097, opened 2022-05-20
fix: fix code ql

KennethBWSong:


KennethBWSong:
E2E TEST: None

packages/fx-core/templates/plugins/resource/aad/auth/bot/ts/public/auth-end.html
```diff
@@ -53,8 +53,8 @@
       // Show error information
       function handleAuthError(errorType, errorMessage) {
         const err = JSON.stringify({
-          error: errorType,
-          message: JSON.stringify(errorMessage),
+          error: escape(errorType),
```
blackchoey, line 56: since escape() is deprecated, can we use another function to escape the value?

blackchoey (approved):

PR #5113, opened 2022-05-23
chore: close all progress bar when error occur

Siglud:


packages/fx-core/src/plugins/resource/bot/progressBars.ts
```diff
@@ -32,6 +33,18 @@ export class ProgressBarFactory {
     return handler;
   }
 
+  public static async createBarFromContext(
```
hund030, line 36: Why not use existing `newProgressBar`? I think that method is also compatible to `v2.Context`.

Siglud: OK, also changed the old behavior to avoid returning a progress bar whose status is ended.

Siglud:
fix a bug for ts source code will not trigger redeploy

eriolchan (approved):

packages/fx-core/src/plugins/resource/bot/progressBars.ts
```diff
@@ -2,29 +2,24 @@
...
 export class ProgressBarFactory {
-  // To store working progress bars.
+  // To store working in progress bars.
   private static progressBars: Map<string, IProgressHandler | undefined> = new Map();
 
   public static async newProgressBar(
     name: string,
     steps_num: number,
     context: PluginContext | v2.Context
-  ): Promise<IProgressHandler | undefined> {
```
IvanJobs, line 14: As we discussed previously, I suggest not breaking the original design. If you're confident enough, please ignore this.

Siglud: ok, done it

packages/fx-core/src/plugins/resource/bot/progressBars.ts
```diff
@@ -4,27 +4,25 @@
...
-    if (ProgressBarFactory.progressBars.has(name)) {
-      const handler = ProgressBarFactory.progressBars.get(name);
-
-      await handler?.end(true);
-
-      return handler;
+  ): Promise<IProgressHandler> {
+    const exists = ProgressBarFactory.progressBars.get(name);
```
IvanJobs, line 15: Nit: Better naming for `exists`.

IvanJobs (dismissed):
lgtm

IvanJobs (approved):

packages/fx-core/src/plugins/resource/bot/progressBars.ts
```diff
@@ -4,27 +4,25 @@
...
+      await existsHandler.end(true);
+      return existsHandler;
     }
 
     const handler =
       (context as PluginContext).ui?.createProgressBar(name, steps_num) ||
       (context as v2.Context).userInteraction?.createProgressBar(name, steps_num);
     if (!handler) {
-      context.logProvider?.warning(`Fail to create progress bar for ${name}.`);
+      await context.logProvider?.warning(`Fail to create progress bar for ${name}.`);
```
eriolchan, line 25: I see somewhere we also use Logger.warning()?

yukun-dong (approved):

PR #5114, opened 2022-05-23
chore: use new login for tunneling

a1exwang:
unit test passed locally

### DIFF
--- a/packages/vscode-extension/src/debug/microsoftTunnelingTask.ts
+++ b/packages/vscode-extension/src/debug/microsoftTunnelingTask.ts
@@ -77,7 +77,7 @@ export class MicrosoftTunnelingTaskTerminal implements vscode.Pseudoterminal {
     this.manager = new MicrosoftTunnelingManager(
       () =>
         m365Login.getAccessToken({
-          showDialog: true,
+          showDialog: false,
           scopes: [MicrosoftTunnelingAadScope],
         }),
       this.logger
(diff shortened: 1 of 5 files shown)


packages/vscode-extension/src/debug/microsoftTunnelingTask.ts
```diff
@@ -73,14 +74,14 @@ export class MicrosoftTunnelingTaskTerminal implements vscode.Pseudoterminal {
...
-      const token = await appStudioLogin.getAccessToken();
-      if (!token) {
-        throw new Error("No login");
-      }
-      return token;
-    }, this.logger);
+    this.manager = new MicrosoftTunnelingManager(
+      () =>
+        m365Login.getAccessToken({
+          showDialog: true,
```
swatDong, line 80: Interactive login on starting task may not be a good UX... any plan to move login to pre-checker?
Better to silently login here.

a1exwang: I think prereqs check already ensures M365 account is signed in. Is this correct?

If so, this will succeed in most cases and won't pop-up user interaction (whether login or consent).
Because we are first party app and preauthorized for tunneling service, [acquiring token for this scope will not require user interaction](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/resources-and-scopes.md#dynamic-scopes-and-incremental-consent). 

For the rare cases (e.g. token expired between prereqs check and tunneling), I agree changing to `showDialog: false` is better. But no need to add more checks to prereqs-check.

xiaolang124 (approved):

swatDong (approved):

PR #5120, opened 2022-05-23
fix(bot): fix some issue in bot plugin

hund030:
1. Need deploy does not work right.
2. Log message should be with component name.
3. Reuse v1 error handling and telemetry helper.

E2E TEST: no need

### DIFF
--- a/packages/fx-core/src/common/azure-hosting/azureHosting.ts
+++ b/packages/fx-core/src/common/azure-hosting/azureHosting.ts
@@ -88,11 +88,9 @@ export abstract class AzureHosting {
     module = AzureHosting.replacePluginId(module, pluginId);
 
     return {
-      Configuration: this.configurable
-        ? {
-            Modules: { [this.hostType]: module },
-          }
-        : undefined,
+      Configuration: {
+        Modules: { [this.hostType]: module },
+      },
       Reference: this.reference,
     } as ResourceTemplate;
   }
--- a/packages/fx-core/src/plugins/resource/bot/index.ts
+++ b/packages/fx-core/src/plugins/resource/bot/index.ts
@@ -2,7 +2,6 @@
 // Licensed under the MIT license.
 import {
   AzureSolutionSettings,
-  err,
   Func,
   FxError,
   ok,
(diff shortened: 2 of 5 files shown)


packages/fx-core/src/common/azure-hosting/azureHosting.ts
```diff
@@ -88,11 +88,9 @@ export abstract class AzureHosting {
     module = AzureHosting.replacePluginId(module, pluginId);
 
     return {
-      Configuration: this.configurable
-        ? {
-            Modules: { [this.hostType]: module },
-          }
-        : undefined,
+      Configuration: {
```
IvanJobs, line 91: Any explanations for :
1. What is `this.configurable` for?
2. Why removing this flag?

eriolchan (dismissed):

packages/fx-core/src/common/azure-hosting/azureHosting.ts
hund030: If the hosting is not configurable, it does not need to update bicep template.
https://github.com/OfficeDev/TeamsFx/blob/eec8ff4f80d64001d5438cb2a805ac35e0500cb8/packages/fx-core/src/common/azure-hosting/azureHosting.ts#L81

packages/fx-core/src/plugins/resource/bot/index.ts
```diff
@@ -265,90 +252,6 @@ export class TeamsBot implements Plugin {
     );
   }
 
-  private static wrapError(
```
hund030, line 268: Move this method to error.ts

packages/fx-core/src/plugins/resource/bot/index.ts
```diff
@@ -265,90 +252,6 @@ export class TeamsBot implements Plugin {
...
-              getLocalizedString("plugins.bot.UnhandledError", errorMsg),
-            ],
-            isPluginError(e) ? e.innerError : undefined
-          )
-        );
-      return ResultFactory.SystemError(UnhandledErrorCode, [errorMsg, errorMsg], innerError);
-    }
-  }
-
-  private static async runWithExceptionCatching(
```
hund030, line 335: Move this method to error.ts

eriolchan (approved):

PR #5126, opened 2022-05-23
feat(bot): support scaffold csharp template

hund030:
E2E TEST: will add e2e test after feature done.

### DIFF
--- a/packages/fx-core/src/plugins/resource/bot/index.ts
+++ b/packages/fx-core/src/plugins/resource/bot/index.ts
@@ -6,6 +6,7 @@ import {
   Func,
   FxError,
   ok,
+  Platform,
   Plugin,
   PluginContext,
   QTreeNode,
@@ -39,7 +40,11 @@ import { PluginImpl } from "./interface";
 import { isVSProject, BotHostTypes, isBotNotificationEnabled } from "../../../common";
 import { FunctionsHostedBotImpl } from "./functionsHostedBot/plugin";
 import { ScaffoldConfig } from "./configs/scaffoldConfig";
-import { createHostTypeTriggerQuestion, showNotificationTriggerCondition } from "./question";
+import {
+  createHostTypeTriggerQuestion,
+  createHostTypeTriggerQuestionForVS,
+  showNotificationTriggerCondition,
+} from "./question";
 import { getDefaultString, getLocalizedString } from "../../../common/localizeUtils";
 import { CommonHostingError } from "../../../common/azure-hosting/hostingError";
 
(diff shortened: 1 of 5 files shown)


yukun-dong (approved):

packages/fx-core/src/plugins/resource/bot/index.ts
```diff
@@ -208,7 +213,11 @@ export class TeamsBot implements Plugin {
       return await TeamsBot.runWithExceptionCatching(
         context,
         async () => {
-          if (isBotNotificationEnabled()) {
+          if (isVSProject(context.projectSettings) || context.answers?.platform === Platform.VS) {
+            const res = new QTreeNode(createHostTypeTriggerQuestionForVS());
```
eriolchan, line 217: The logic between L216 and L219 looks the same as L221 and L229, so for the long term (after VS 17.3 preview 2), createHostTypeTriggerQuestionForVS() seems to be part of createHostTypeTriggerQuestion() logic since it already takes context.answers?.platform as param. @yukun-dong

yukun-dong (commented):
Do we need to add question for command bot?

packages/fx-core/src/plugins/resource/bot/question.ts
```diff
@@ -39,12 +45,46 @@ export const AppServiceOptionItem: HostTypeTriggerOptionItem = optionWithL10n({
   // trigger of app service host is hard-coded to http, so no need to set here
 });
 
+// TODO: this option will not be shown in UI, leave messages empty.
+export const AppServiceOptionItemForVS: HostTypeTriggerOptionItem = {
+  id: "http-webapi",
+  hostType: HostTypes.APP_SERVICE,
+  label: "",
```
eriolchan, line 52: Do we need label value which will be show as project type in creating project?

packages/fx-core/src/plugins/resource/bot/v2/common.ts
```diff
@@ -47,7 +47,8 @@ export function decideTemplateScenarios(ctx: Context, inputs: Inputs): Set<strin
         const notificationTriggerType = (inputs[
           QuestionNames.BOT_HOST_TYPE_TRIGGER
         ] as string[]) ?? [AppServiceOptionItem.id];
-        notificationTriggerType.forEach((triggerType) => {
+        // notificationTriggerType may be string in VS scenario
```
eriolchan, line 50: It looks a little weird that VS and VSCode return different type because they are the same question model and VS UI need to handle it as the correct type instead of asking plugin to workaround it.
We could write here as a workaround to unblock integration test, but it is better to ask Zhiyu to see if there is any missing from UI side and if he can have a better solution for it.

eriolchan (approved):

packages/fx-core/src/plugins/resource/bot/v2/common.ts
eriolchan: So, when we support more notification trigger type in preview 3, we will revisit this logic and revert it back to previous one?

eriolchan (approved):

PR #5130, opened 2022-05-24
chore(bot): use enum for hosttype

yukun-dong:
The original `hostType` and `hostTypes` are confusing.


packages/fx-core/src/plugins/resource/bot/configs/scaffoldConfig.ts
```diff
@@ -113,7 +112,7 @@ export class ScaffoldConfig {
     const rawHostType = context.projectSettings?.pluginSettings?.[PluginBot.PLUGIN_NAME]?.[
       PluginBot.HOST_TYPE
     ] as string;
-    return utils.convertToConstValues(rawHostType, HostTypes);
+    return utils.convertToConstValues(rawHostType, HostType);
```
eriolchan, line 115: Since we change the type of HostTypes to HostType, do we need to update the method of convertToConstValues() due to the param type is different?

packages/fx-core/src/plugins/resource/bot/resources/strings.ts
```diff
@@ -123,12 +123,10 @@ export const QuestionBotScenarioToBotCapability = new Map<BotScenario, BotCapabi
...
 ]);
 
-export const HostTypes = {
-  APP_SERVICE: "app-service",
-  AZURE_FUNCTIONS: "azure-functions",
-} as const;
-
-export type HostType = typeof HostTypes[keyof typeof HostTypes];
+export enum HostType {
+  APP_SERVICE = "app-service",
```
eriolchan, line 127: It seems this is not the convention of Enum, maybe APP_SERVICE => AppService

eriolchan: Since it is an enum, do we move to mapping.ts to keep all enum together?

packages/fx-core/src/plugins/resource/bot/utils/common.ts
```diff
@@ -240,6 +240,13 @@ export function convertToConstValues<V extends string, T extends { [key in strin
   return Object.values(targetValues).find((itemValue) => value === itemValue);
 }
 
+export function convertToEnumValues<T extends string>(
```
eriolchan, line 243: I have some question about this method. what we want to convert from and to what?

packages/fx-core/src/plugins/resource/bot/resources/strings.ts
yukun-dong: Done.

packages/fx-core/src/plugins/resource/bot/utils/common.ts
yukun-dong: convert string type to enum type.

packages/fx-core/src/plugins/resource/bot/utils/common.ts
```diff
@@ -240,6 +240,13 @@ export function convertToConstValues<V extends string, T extends { [key in strin
...
+  value: unknown,
```
eriolchan, line 244: When we use Enum instead of constant string, maybe there is update for this method. The type of T and value, targetEnum looks a little weird. You can try below and check some related docs. https://www.technicalfeeder.com/2021/07/mastering-enum-in-typescript/

```
function getEnumByValue<T extends {[key:string]:string}>(enumType: T, enumValue:string): keyof T|undefined {
    return Object.keys(enumType).find(k => enumType[k] === enumValue);
}
```

eriolchan (approved):

Siglud (approved):

PR #5139, opened 2022-05-24
chore: use zip deploy need set SCM_SCRIPT_GENERATOR_ARGS

Siglud:
* use zip deploy need set SCM_SCRIPT_GENERATOR_ARGS
* insert enableAlwaysOn to bicep context

### DIFF
--- a/packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
+++ b/packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
@@ -17,7 +17,6 @@ import { AzureHostingFactory } from "../../../../common/azure-hosting/hostingFac
 import { Commands, CommonStrings, ConfigNames, PluginBot } from "../resources/strings";
 import { checkAndThrowIfMissing, checkPrecondition, CommandExecutionError } from "../errors";
 import { BicepConfigs, ServiceType } from "../../../../common/azure-hosting/interfaces";
-import { isBotNotificationEnabled } from "../../../../common";
 import {
   DEFAULT_DOTNET_FRAMEWORK,
   DeployConfigs,
@@ -80,7 +79,6 @@ export class TeamsBotV2Impl {
     const bicepContext = {
       plugins: plugins.map((obj) => obj.name),
       configs: bicepConfigs,
-      enableAlwaysOn: isBotNotificationEnabled(),
       useNode: getLanguage(ctx.projectSetting.programmingLanguage) !== ProgrammingLanguage.Csharp,
     };
 
--- a/packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
+++ b/packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
@@ -79,7 +79,6 @@ export class TeamsBotV2Impl {
     const bicepContext = {
       plugins: plugins.map((obj) => obj.name),
       configs: bicepConfigs,
-      useNode: getLanguage(ctx.projectSetting.programmingLanguage) !== ProgrammingLanguage.Csharp,
     };
 
     const serviceTypes = [resolveServiceType(ctx), ServiceType.BotService];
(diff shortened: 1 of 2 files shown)


packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
```diff
@@ -80,6 +80,8 @@ export class TeamsBotV2Impl {
     const bicepContext = {
       plugins: plugins.map((obj) => obj.name),
       configs: bicepConfigs,
+      enableAlwaysOn: isBotNotificationEnabled(),
```
hund030, line 83: I think `isBotNotificationEnabled` is always true. Can we just remove this config and always set true in bicep?

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
```diff
@@ -80,6 +80,8 @@ export class TeamsBotV2Impl {
...
+      useNode: getLanguage(ctx.projectSetting.programmingLanguage) !== ProgrammingLanguage.Csharp,
```
hund030, line 84: I think there is already a `node` config in `bicepConfigs`. Can we reuse that one?

hund030: And please also take `updateResourceTemplate` method into consideration when you update configs.

packages/fx-core/templates/plugins/resource/hosting/bicep/webapp/webappProvision.template.bicep
```diff
@@ -9,6 +9,11 @@ var webAppSKU = contains(provisionParameters, 'botWebAppSKU') ? provisionParamet
 {{else}}
 var webAppSKU = contains(provisionParameters, 'botWebAppSKU') ? provisionParameters['botWebAppSKU'] : 'F1' // Try to read SKU for Azure Web App from parameters
 {{/if}}
+{{#if useNode}}
+var scriptGenerator = '--node'
```
IvanJobs, line 13: Nit: Name with `scmScriptGeneratorArgs` would be better from my perspective.

IvanJobs (dismissed):
lgtm

packages/fx-core/templates/plugins/resource/hosting/bicep/webapp/webappProvision.template.bicep
```diff
@@ -9,6 +9,11 @@ var webAppSKU = contains(provisionParameters, 'botWebAppSKU') ? provisionParamet
...
+{{else}}
```
eriolchan, line 14: There is also a "dotnet" in BicepConfigs it is a .NET project, you can check more details about the BicepConfigs design with @hund030 
It is better we use implict if "dotnet" instead of else to avoid accidentally handle it for other languages

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
Siglud: done

packages/fx-core/templates/plugins/resource/hosting/bicep/webapp/webappProvision.template.bicep
```diff
@@ -48,7 +48,7 @@ resource webApp 'Microsoft.Web/sites@2021-02-01' = {
         }
         {
           name: 'SCM_SCRIPT_GENERATOR_ARGS'
-          value: scriptGenerator
+          value: scmScriptGeneratorArgs
```
hund030, line 51: I prefer we use if condition here and do not introduce a varaible, which may be more readable for end user.
```
{{#if (contains "dotnet" configs)}}
        value: '--aspNetCore'
{{/if}}
{{#if (contains "node" configs)}}
        value: '--node'
{{/if}}
```

eriolchan (approved):

PR #5148, opened 2022-05-24
fix(bot): support deploying dotnet project

hund030:
E2E TEST: will add e2e test after feature complete.

### DIFF
--- a/packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
+++ b/packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
@@ -268,7 +268,7 @@ export class TeamsBotV2Impl {
           `dotnet publish --configuration Release --runtime ${runtime} --self-contained`,
           packDir
         );
-        return packDir;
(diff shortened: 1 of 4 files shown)


packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
```diff
@@ -135,7 +135,12 @@ export class TeamsBotV2Impl {
     tokenProvider: TokenProvider
   ): Promise<Result<Void, FxError>> {
     const projectPath = checkPrecondition(Messages.WorkingDirIsMissing, inputs.projectPath);
-    const workingDir = path.join(projectPath, CommonStrings.BOT_WORKING_DIR_NAME);
+    const language = getLanguage(ctx.projectSetting.programmingLanguage);
+    const workingDir =
```
eriolchan, line 139: There is similar logic of getting working dir in scaffold(), we can extract it as a separate method for single source of truth later.

eriolchan (approved):

eriolchan (approved):

PR #5153, opened 2022-05-25
feat: check compatibility of Node.js and Azure Functions Core Tools

meifans:
VSC: 
image

CLI: 
image

E2E TEST: https://github.com/OfficeDev/TeamsFx/actions/runs/2382029740


packages/fx-core/src/common/deps-checker/depsError.ts
```diff
@@ -35,6 +35,14 @@ export class LinuxNotSupportedError extends DepsCheckerError {
   }
 }
 
+export class FuncNodeNotMatchedError extends DepsCheckerError {
+  constructor(message: string, helpLink: string) {
+    super(message, helpLink);
+
+    Object.setPrototypeOf(this, LinuxNotSupportedError.prototype);
```
a1exwang, line 42:
```suggestion
    Object.setPrototypeOf(this, FuncNodeNotMatchedError.prototype);
```

meifans: Fixed.

packages/fx-core/src/common/deps-checker/internal/funcToolChecker.ts
```diff
@@ -24,6 +25,26 @@ export enum FuncVersion {
   v4 = "4",
 }
 
+const FuncNodeVersionWhiteList: { [key: string]: { [key: string]: boolean } } = {
```
a1exwang, line 28: nit: You can declare like this to avoid `!` in `FuncNodeVersionWhiteList[funcVersion.toString()]![nodeVersion]`
```suggestion
const FuncNodeVersionWhiteList: { [key: FuncVersion]: { [key: string]: boolean } } = {
```

packages/fx-core/src/common/deps-checker/internal/funcToolChecker.ts
```diff
@@ -24,6 +25,26 @@ export enum FuncVersion {
...
+  // func-core-tools version
+  "1": {
```
swatDong, line 30: What's the meaning of key 1,2,3,4?

swatDong, line 30: Can we remove func v1 and v2 since we never support that...

swatDong (dismissed):

packages/fx-core/src/common/deps-checker/internal/funcToolChecker.ts
meifans: image
anyway, the object of func version may be empty that need `!`

a1exwang (approved):

packages/fx-core/src/common/deps-checker/internal/funcToolChecker.ts
meifans: Removed.

packages/fx-core/src/common/deps-checker/internal/funcToolChecker.ts
a1exwang: image

a1exwang (approved):

swatDong (approved):

PR #5178, opened 2022-05-26
chore: disable remote build and disable upload resource file

Siglud:


packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
```diff
@@ -239,6 +239,7 @@ export class TeamsBotV2Impl {
       try {
         await utils.execute("npm install", workingPath);
         await utils.execute("npm run build", workingPath);
+        await utils.execute("npm prune --production", workingPath);
```
IvanJobs, line 242: NIT: I'm wondering if it will make any impact on the local debugging lifecycle. Consider the case when customers run local debug after deployment.

packages/fx-core/templates/plugins/resource/hosting/bicep/webapp/webappProvision.template.bicep
```diff
@@ -29,15 +29,20 @@ resource webApp 'Microsoft.Web/sites@2021-02-01' = {
...
         {
           name: 'WEBSITE_NODE_DEFAULT_VERSION'
           value: '~14' // Set NodeJS version to 14.x for your site
         }
+        {
+          name: 'SCM_SCRIPT_GENERATOR_ARGS'
+          value: '--node' // Register as node server
+        }
+        {{else}}
+        {
```
IvanJobs, line 42: Why not making them consistent? I mean, node (--node) vs dotnet (remote build) are in different behaviors.

Siglud: yes, that's a good idea

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
Siglud: npm prune --production will clear all dev dependencies, when project lang is TS, that means the typescript package will be removed in this process.
I don't know if npm install will run before the local debug?

templates/tab/js/default/.webappignore
```diff
@@ -0,0 +1,17 @@
+.fx
```
hund030, line 1: tab is not hosted by Azure Web App. I think we only need to add this file in restify bot.

templates/bot/ts/default/.webappignore
```diff
@@ -0,0 +1,17 @@
+.fx
```
hund030, line 1: Do we also add this file in csharp bot template?

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
hund030: What about `npm install --production` and then `npm run build`?

hund030: I prefer not to remove anything from user's project. 
Maybe we can prune in CD environment?

templates/bot/ts/default/.webappignore
Siglud: I think in your last PR, you upload all build output but not all of the file in repository, so don't think we need this in csharp template

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
Siglud: Do you mean we do "npm prune --production" after the local build method but before we package the zip that we want to upload?
Or we can just run npm install again after the upload process?

IvanJobs: From my perspective, the goal you want to achieve is packing the project for remote in a production/release build style which may do extra optimizations. But the local debug doesn't need that. To achieve both local and remote goal, try thinking about Asp.Net project's build types of Debug and Release.

For a long term consideration, TeamsFx project may need some kind of Build Type abstractions. 
For now, I prefer not to do this kind of optimizations to avoid affecting the local debug experience.

eriolchan (approved):

PR #5180, opened 2022-05-27
chore(bot): add log message

hund030:


packages/fx-core/src/common/azure-hosting/azureFunctionHosting.ts
```diff
@@ -7,6 +7,7 @@ import { TokenProvider, Void } from "@microsoft/teamsfx-api";
 import { azureWebSiteDeploy } from "./utils";
 import { AzureOperations } from "./azureOps";
 import { getResourceGroupNameFromResourceId, getSiteNameFromResourceId } from "../tools";
+import { Messages } from "./messages";
```
eriolchan, line 10: There is a code scanning warning of "Unused import Messages."

packages/fx-core/src/common/azure-hosting/azureOps.ts
```diff
@@ -51,7 +53,8 @@ export class AzureOperations {
   public static async zipDeployPackage(
     zipDeployEndpoint: string,
     zipBuffer: Buffer,
-    config: AzureUploadConfig
+    config: AzureUploadConfig,
```
eriolchan, line 56: I am thinking if we need to add logger into some utility class? The alternative way is leverage Error to pass some information, otherwise it force unit test to take care of logger?

eriolchan (approved):

packages/fx-core/src/common/azure-hosting/azureOps.ts
hund030: The logger is an optional parameter. Does unit test need to take care of it?

packages/fx-core/src/common/azure-hosting/azureFunctionHosting.ts
hund030: Fixed.

eriolchan (approved):

PR #5198, opened 2022-05-30
fix: some success events with error properties

nliu-ms:
plugin.commonProperties is passed to sendTelemetry() method, and as a reference, the values will be altered.

So the success event may contain error properties, like:
image


nliu-ms:
E2E TEST: N/A

HuihuiWu-Microsoft (approved):

PR #5208, opened 2022-05-31
chore(bot): convert get map method to throw error

yukun-dong:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/14561201/

### DIFF
--- a/packages/fx-core/src/plugins/resource/bot/v2/mapping.ts
+++ b/packages/fx-core/src/plugins/resource/bot/v2/mapping.ts
@@ -3,7 +3,6 @@
 
 import { ServiceType } from "../../../../common/azure-hosting/interfaces";
 import { TemplateProjectsScenarios } from "../constants";
-import { checkPrecondition } from "../errors";
 import {
   FunctionsHttpTriggerOptionItem,
   FunctionsTimerTriggerOptionItem,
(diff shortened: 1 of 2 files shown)


packages/fx-core/src/plugins/resource/bot/v2/mapping.ts
```diff
@@ -61,23 +58,23 @@ export function getRuntime(lang: ProgrammingLanguage): Runtime {
...
+  throw new Error("invalid bot input");
 }
 
 export function getServiceType(hostType?: string): ServiceType {
   const serviceType = serviceMap.get(hostType ?? "");
   if (serviceType) {
     return serviceType;
   }
-  return defaultServiceType;
+  throw new Error("invalid bot input");
```
IvanJobs, line 69: Just for a reminder - consider localization for the error message.

packages/fx-core/src/plugins/resource/bot/v2/mapping.ts
```diff
@@ -61,23 +58,23 @@ export function getRuntime(lang: ProgrammingLanguage): Runtime {
   if (runtime) {
     return runtime;
   }
-  return defaultRuntime;
+  throw new Error("invalid bot input");
```
eriolchan, line 61: Maybe we could make this string "invalid bot input" as a const?

yukun-dong: Done.

PR #5214, opened 2022-05-31
chore: add creation-version telemetry property

a1exwang:
- add a common property "creation-version" in CLI and VSC
- move these common properties to VSCTelemetryReporter and CliTelemetryReporter respectively. This is to fix the issue that only events sent in vsc/cli contain these properties. (fx-core events do not have them)
- add the [missing "is-from-sample"](https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/14587942/) to cli

All event between prerequisite-check and debug-start should contain "creation-version":
VSC:
image

CLI:
image

https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/14549235/

PR #5231, opened 2022-06-06
feat(vsc): change activation workflow to show icon and welcome view by default

tecton:
1. Replace "*" activation event with "onStartupFinished" and remove other redundant events.
2. Reorder and group commands into 4 sets: activation, teamsfx only, menu and internal. These commands are registered in different activation stages.
3. Split async contents in registerTreeViews() and make it sync method.
4. Use category "Teams" instead of title prefix "Teams: ".
5. Clean unused context value and command handlers.

E2E TEST: N/A


packages/vscode-extension/package.json
```diff
@@ -40,37 +40,21 @@
...
-    "onCommand:fx-extension.deploy",
-    "onCommand:fx-extension.publish",
-    "onCommand:fx-extension.updateAad",
-    "onCommand:fx-extension.openAppManagement",
-    "onCommand:fx-extension.openBotManagement",
-    "onCommand:fx-extension.openReportIssues",
     "onCommand:fx-extension.getNewProjectPath",
-    "onCommand:fx-extension.openReadMe",
     "onCommand:fx-extension.openDeploymentTreeview",
+    "onCommand:fx-extension.openDocument",
```
HuihuiWu-Microsoft, line 47: Why remove some of commands while keeping others?

packages/vscode-extension/package.json
```diff
@@ -108,80 +92,82 @@
...
           "name": "Deployment",
-          "when": "fx-extension.sidebarWelcome.treeview && !fx-extension.customizedTreeview"
+          "when": "fx-extension.isTeamsFx"
         },
         {
           "type": "webview",
           "id": "teamsfx-deployment-webview",
           "name": "Deployment",
-          "when": "fx-extension.sidebarWelcome.treeview && fx-extension.customizedTreeview"
+          "when": "fx-extension.isTeamsFx && fx-extension.customizedTreeview"
```
HuihuiWu-Microsoft, line 136: Remove customizedTreeView here to align with other commands?

packages/vscode-extension/package.nls.json
```diff
@@ -45,57 +45,53 @@
...
-    "teamstoolkit.commands.connectExistingApi.title": "Teams: Add connection to an API",
+    "teamstoolkit.commands.addFeature.title": "Add features",
+    "teamstoolkit.commands.azureAccountSettings.title": "Azure portal",
     "teamstoolkit.commands.createAccount.azure": "Create an Azure account",
     "teamstoolkit.commands.createAccount.free": "Free",
     "teamstoolkit.commands.createAccount.m365": "Create a M365 testing tenant",
     "teamstoolkit.commands.createAccount.title": "Create an account",
-    "teamstoolkit.commands.createEnvironment.title": "Teams: Create new environment",
-    "teamstoolkit.commands.createProject.title": "Teams: Create a new Teams app",
+    "teamstoolkit.commands.createEnvironment.title": "Create new environment",
```
HuihuiWu-Microsoft, line 57: If we remove the 'Teams:' prefix, does that mean we'll need to update the svg in quick start?

packages/vscode-extension/src/extension.ts
```diff
@@ -838,6 +775,44 @@ function registerDebugConfigProviders(context: vscode.ExtensionContext) {
...
 }
 
+/**
+ * Tasks that doesn't block the user interaction so that they can be processed in background.
+ */
+async function runBackgroundAsyncTasks(
+  context: vscode.ExtensionContext,
+  isTeamsFxProject: boolean
+) {
+  ExtTelemetry.isFromSample = await handlers.getIsFromSample();
```
HuihuiWu-Microsoft, line 785: If the variable is set after handlers.activate(), will the sent telemetry attach wrong properties?

packages/vscode-extension/package.json
tecton: These commands are the ones in welcome view and walkthrough, so they are kept making sure extension will be activated to run them. The removed ones are used in teamsfx project, they will not show in command palette until extension is activated and detects that current project is teamsfx. User are not allowed to run these commands in a non-teamsfx project, thus they should not be used to activate extension.

packages/vscode-extension/package.json
tecton: It is kept for the legacy webview as treeview, can we remove the whole item?

packages/vscode-extension/package.nls.json
tecton: In command palette it will show "Teams: Create new environment" as before, so we don't need to update the SVG.

packages/vscode-extension/package.json
HuihuiWu-Microsoft: So before extension activate complete, nothing will happen if user execute these commands(like provision, deploy, ...) in command palette?

packages/vscode-extension/package.json
HuihuiWu-Microsoft: OK, I think we can keep for now.

packages/vscode-extension/src/extension.ts
tecton: `isFromSample` is meaningful to teamsfx project and it will only happen to be undefined when user runs command like "build" before this code is executed. It should be very rare case. To fix it, we'd better make this sync function, and we can see whether it happens in telemetry later.

HuihuiWu-Microsoft: There's also telemetry sent in handlers.activate for openTeamsApp

packages/vscode-extension/package.json
tecton: Yes, actually user cannot execute these teamsfx commands (provision, etc.) from command palette because they depend on  context value which is set after activation is completed.

packages/vscode-extension/src/extension.ts
tecton: Yes it's a bug and can be fixed in another PR, will check again after #5214 is merged.

HuihuiWu-Microsoft (approved):

PR #5235, opened 2022-06-06
fix: hide deploy aad manifest command when aad is not enabled

SLdragon:
Currently, even sso feature is not enabed, the command "deploy aad manifest" still exist in command palette.

E2E TEST: Will add test plan for this

### DIFF
--- a/packages/vscode-extension/src/extension.ts
+++ b/packages/vscode-extension/src/extension.ts
@@ -56,6 +56,7 @@ import {
 import { loadLocalizedStrings } from "./utils/localizeUtils";
 import { ExtensionSurvey } from "./utils/survey";
 import { ExtensionUpgrade } from "./utils/upgrade";
+import { isAADEnabled } from "@microsoft/teamsfx-core";
 
 export let VS_CODE_UI: VsCodeUI;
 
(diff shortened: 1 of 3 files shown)


packages/vscode-extension/src/extension.ts
```diff
@@ -56,6 +56,7 @@ import {
 import { loadLocalizedStrings } from "./utils/localizeUtils";
 import { ExtensionSurvey } from "./utils/survey";
 import { ExtensionUpgrade } from "./utils/upgrade";
+import { isAADEnabled } from "@microsoft/teamsfx-core";
```
tecton, line 59: This import can be merged with Line 19.

packages/vscode-extension/src/extension.ts
```diff
@@ -623,6 +624,17 @@ async function initializeContextKey() {
     workspaceUri && (await isM365Project(workspaceUri.fsPath))
   );
 
+  const workspacePath = workspaceUri?.fsPath;
+  if (workspacePath) {
```
tecton, line 628: Is this intended to check whether it's teamsfx project? If so, you could use `isTeamsFxProject`.

packages/vscode-extension/src/extension.ts
SLdragon: Thanks, will update

packages/vscode-extension/src/extension.ts
SLdragon: Yes, you're right, will updatge

HuihuiWu-Microsoft (approved):

tecton (approved):

PR #5238, opened 2022-06-07
fix: add is-from-sample telemetry property

Alive-Fish:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/14587942
image
E2E TEST:

### DIFF
--- a/packages/cli/src/commonlib/telemetry.ts
+++ b/packages/cli/src/commonlib/telemetry.ts
@@ -21,9 +21,14 @@ import { CliConfigOptions, CliConfigRunFrom, UserSettings } from "../userSetttin
 export class CliTelemetryReporter implements TelemetryReporter {
   private readonly reporter: Reporter;
   private rootFolder: string | undefined;
+  private isFromSample: boolean | undefined = undefined;
 
   constructor(key: string, cliName: string, cliVersion: string, appRoot?: string) {
     this.reporter = new Reporter(cliName, cliVersion, key, appRoot);
+  }
+
+  setIsFromSample(isFromSample?: boolean) {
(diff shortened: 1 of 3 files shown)


kuojianlu (approved):

packages/cli/src/commonlib/telemetry.ts
```diff
@@ -44,6 +49,10 @@ export class CliTelemetryReporter implements TelemetryReporter {
       properties = {};
     }
 
+    if (this.isFromSample !== undefined) {
+      properties[TelemetryProperty.IsFromSample] = this.isFromSample.toString();
```
a1exwang, line 53: Other common properties are directly retrieved using utility functions, for example:
```
 const projectId = getProjectId(this.rootFolder);
 properties[TelemetryProperty.ProjectId] = projectId ? projectId : "";
```

But `is-from-sample` is stored in a global variable in `activate()`.

Why is `is-from-sample` different from other properties?

Alive-Fish: This is to call core to get the property. For any command, running once is enough. If writing the logic this place, it will waste some time. Also, I think the other previous properties can be aligned with it.

a1exwang (approved):

packages/cli/src/commonlib/telemetry.ts
chagong: If we put "is-from-sample" property here, it will be added to all events including core and CLI levels. Please re-consider if this is aligned with vscode extension, and what's the goal of adding "is-from-sample" property.

a1exwang (approved):

chagong (approved):

PR #5239, opened 2022-06-07
chore: upgrade to v1.13

nliu-ms:


nliu-ms:
E2E TEST passed: https://github.com/OfficeDev/TeamsFx/actions/runs/2451955858

packages/fx-core/src/plugins/resource/appstudio/plugin.ts
```diff
@@ -1359,6 +1361,37 @@ export class AppStudioPluginImpl {
       appDefinition.outlineIcon = appManifest.icons.outline;
     }
 
+    appDefinition.configurableProperties = appManifest.configurableProperties;
```
HuihuiWu-Microsoft, line 1364: Does this affect project with old version manifest?

nliu-ms: it's optional, so older version without this field will just provide undefined.

HuihuiWu-Microsoft (approved):

jayzhang (approved):

PR #5242, opened 2022-06-07
refactor(bot): refactor trigger question

yukun-dong:
E2E TEST: 
https://github.com/OfficeDev/TeamsFx/blob/b8e4b822560811edf5721fcde6de2a7c616432bd/packages/cli/tests/e2e/bot/NotificationBotHappyPathCommon.ts#L35


yukun-dong:
e2e test result: https://github.com/OfficeDev/TeamsFx/actions/runs/2452263139

packages/fx-core/src/plugins/resource/bot/question.ts
```diff
@@ -63,62 +58,47 @@ export const HostTypeTriggerOptions: HostTypeTriggerOptionItem[] = [
...
+export function createHostTypeTriggerQuestion(
+  platform?: Platform,
+  cliRuntime?: string
+): MultiSelectQuestion {
   const prefix = "plugins.bot.questionHostTypeTrigger";
 
-  let staticOptions: HostTypeTriggerOptionItem[];
+  let staticOptions: HostTypeTriggerOptionItem[] = HostTypeTriggerOptions;
+  let defaultOptionItem = AppServiceOptionItem;
+  if (cliRuntime === "dotnet" || platform === Platform.VS) {
```
eriolchan, line 72: Do we leverage existing Runtime enum for "dotnet"?

packages/fx-core/src/plugins/resource/bot/index.ts
```diff
@@ -207,54 +205,25 @@ export class TeamsBot implements Plugin {
...
-              const res = new QTreeNode({
-                type: "group",
-              });
-              res.addChild(new QTreeNode(createHostTypeTriggerQuestion(context.answers?.platform)));
-              res.condition = showNotificationTriggerCondition;
-              return ok(res);
-            } else {
-              return ok(undefined);
-            }
+            const dotnetNode = new QTreeNode(createHostTypeTriggerQuestion(Platform.CLI, "dotnet"));
```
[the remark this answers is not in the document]
yukun-dong: In bot plugin code the enums for runtime are "dotnet" and "node", not "nodejs", and there is no existing enum/const for "nodejs". Adding extra enum for "nodejs" may be more confusing. Changing "nodejs" to "node" may be a big change, may need to check CLI side.

packages/fx-core/src/plugins/resource/bot/question.ts
```diff
@@ -129,11 +109,11 @@ export function createHostTypeTriggerQuestion(platform?: Platform): MultiSelectQ
       currentSelectedIds: Set<string>,
       previousSelectedIds: Set<string>
     ): Promise<Set<string>> {
-      if (currentSelectedIds.size > 1 && currentSelectedIds.has(AppServiceOptionItem.id)) {
```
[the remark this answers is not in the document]
yukun-dong: It means hoosing both appservice item and other function item is invalid.

packages/cli/tests/e2e/bot/NotificationBotHappyPathCommon.ts
```diff
@@ -34,7 +34,7 @@ export function happyPathTest(runtime: "dotnet" | "nodejs"): void {
 
     it("Provision Resource: app service hosted notification", async function () {
       const cmd =
-        runtime === "nodejs"
+        runtime === "node"
```
IvanJobs, line 37: Nit: I remember there's an enum, how about using enum here instead of a plain string `node`?

packages/cli/tests/e2e/bot/NotificationBotHappyPathCommon.ts
```diff
@@ -16,7 +16,7 @@ import { environmentManager } from "@microsoft/teamsfx-core";
 
 import { it } from "../../commonlib/it";
 
-export function happyPathTest(runtime: "dotnet" | "nodejs"): void {
+export function happyPathTest(runtime: "dotnet" | "node"): void {
```
IvanJobs, line 19: Nit: Same here, how about using Enum defined elsewhere?

packages/fx-core/src/plugins/resource/bot/index.ts
```diff
@@ -207,54 +206,19 @@ export class TeamsBot implements Plugin {
...
-                } else {
-                  return "Runtime is not .net.";
-                }
-              },
-            };
-            node.addChild(dotnetNode);
-
-            const nodejsNode = new QTreeNode(
-              createHostTypeTriggerQuestion(context.answers?.platform)
+          const res = new QTreeNode({
```
IvanJobs, line 209: Good refactor for this piece of code. I gave a comment before when yefu added this in his PR. The refactor will make the code clear and provides better readability.

packages/fx-core/src/plugins/resource/bot/index.ts
```diff
@@ -207,54 +206,19 @@ export class TeamsBot implements Plugin {
...
+            type: "group",
+          });
+          const runtimes = [Runtime.Dotnet, Runtime.Node];
```
IvanJobs, line 212: Nit: Consider using the Enum type itself to reflect this runtime list if feasible.

packages/fx-core/src/plugins/resource/bot/index.ts
```diff
@@ -207,54 +206,19 @@ export class TeamsBot implements Plugin {
...
-                type: "group",
-              });
-              res.addChild(new QTreeNode(createHostTypeTriggerQuestion(context.answers?.platform)));
-              res.condition = showNotificationTriggerCondition;
-              return ok(res);
-            } else {
-              return ok(undefined);
-            }
-          }
+            node.condition = getTriggerQuestionCondition(runtime);
```
IvanJobs, line 217: Nit: The naming seems inconsistent compared to L220. `getTriggerQuestionCondition` vs `showNotificationTriggerCondition`.

packages/cli/tests/e2e/bot/NotificationBotHappyPathCommon.ts
yukun-dong: Defined in fx-core. Is it ok to import from another package?

packages/cli/tests/e2e/bot/CommandBotHappyPathCommon.ts
```diff
@@ -24,7 +24,7 @@ import { environmentManager } from "@microsoft/teamsfx-core";
 
 import { it } from "../../commonlib/it";
 
-export function happyPathTest(runtime: "dotnet" | "nodejs"): void {
+export function happyPathTest(runtime: "dotnet" | "node"): void {
```
eriolchan, line 27: There is a Runtime enum in Bot plugin, why don't we reuse that?

packages/fx-core/src/plugins/resource/bot/index.ts
yukun-dong: Some e2e tests fail at scaffolding stage after I switched to looping using Enum and I think there exists bugs in either question model or cli after doing some experiments. Using array can be a workaround here. May need discussions offline.

packages/fx-core/src/plugins/resource/bot/question.ts
```diff
@@ -171,6 +153,29 @@ export const showNotificationTriggerCondition = {
...
+            return undefined;
+          }
+        } else if (runtime == Runtime.Node) {
+          return undefined;
+        }
+      }
+      if (inputs?.platform === Platform.VS && runtime === Runtime.Dotnet) {
+        return undefined;
+      }
+      if (inputs?.platform === Platform.VSCode && runtime == Runtime.Node) {
```
eriolchan: Same here. Maybe we can consider if we need a map to describe the decision tree, instead of a lot of if condition. The logic looks like, when isCLIDotNetEnabled() is turn off, there is a mapping (Platform -> Runtime) that
```
Platform.CLI -> Runtime.Node
Platform.VS-> Runtime.Dotnet
Platform.VSCode-> Runtime.Node
```
So, the logic is to return undefined when ```runtime === getMap(inputs?.platform)```, else return ```runtime is not ${runtime}```

packages/fx-core/src/plugins/resource/bot/question.ts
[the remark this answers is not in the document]
yukun-dong: Found a handleSelectionConflict function we can leverage. I have rewritten this function.

packages/fx-core/src/plugins/resource/bot/question.ts
```diff
@@ -171,6 +152,18 @@ export const showNotificationTriggerCondition = {
   containsAny: [NotificationOptionItem.id],
 };
 
+export function getConditionOfNotificationTriggerQuestion(runtime: Runtime) {
+  return {
+    validFunc: async (input: unknown, inputs?: Inputs) => {
+      if (inputs && inputs[CoreQuestionNames.Runtime] === runtime) {
```
IvanJobs, line 158: Nit: Maybe using `?` would make the condition simpler.

packages/fx-core/src/plugins/resource/bot/v2/mapping.ts
```diff
@@ -48,6 +49,13 @@ const triggerScenariosMap: Map<string, string[]> = new Map<string, string[]>([
   [AppServiceOptionItemForVS.id, [TemplateProjectsScenarios.NOTIFICATION_WEBAPI_SCENARIO_NAME]],
 ]);
 
+const PlatformRuntimeMap: Map<Platform, Runtime> = new Map<Platform, Runtime>([
+  [Platform.VS, Runtime.Dotnet],
+  [Platform.VSCode, Runtime.Node],
+  [Platform.CLI, Runtime.Node],
```
IvanJobs, line 55: Does it mean `Platform.CLI` does support `Runtime.Dotnet`?

yukun-dong: There is a new feature to create dotnet projects in CLI being developed and it's controlled by isCliDontnetEnabled() flag. But currently CLI only supports Node projects.

packages/cli/tests/e2e/bot/CommandBotHappyPathCommon.ts
```diff
@@ -21,10 +21,10 @@ import {
 } from "../commonUtils";
 import AppStudioLogin from "../../../src/commonlib/appStudioLogin";
 import { environmentManager } from "@microsoft/teamsfx-core";
-
+import { Runtime } from "@microsoft/teamsfx-core/src/plugins/resource/bot/v2/enum";
```
eriolchan, line 24: Can we use the similar package import instead of file import? e.g. import { environmentManager } from "@microsoft/teamsfx-core"; in L23?

packages/cli/tests/e2e/bot/NotificationBotHappyPathCommon.ts
```diff
@@ -15,8 +15,9 @@ import {
 import { environmentManager } from "@microsoft/teamsfx-core";
 
 import { it } from "../../commonlib/it";
+import { Runtime } from "@microsoft/teamsfx-core/src/plugins/resource/bot/v2/enum";
```
eriolchan, line 18: Same here

packages/cli/tests/e2e/bot/ProvisionAppServiceNotificationBot.dotnet.tests.ts
```diff
@@ -5,6 +5,7 @@
  * @author Yefu Wang <yefuwang@microsoft.com>
  */
 
+import { Runtime } from "@microsoft/teamsfx-core/src/plugins/resource/bot/v2/enum";
```
eriolchan, line 8: Same for other places

eriolchan (approved):

packages/fx-core/src/plugins/resource/bot/v2/mapping.ts
```diff
@@ -59,6 +67,14 @@ export const moduleMap: { [key: string]: string } = {
   [ServiceType.Functions]: BicepModules.Functions,
 };
 
+export function getPlatformRuntime(platform: Platform): Runtime {
+  const runtime = PlatformRuntimeMap.get(platform);
+  if (runtime) {
+    return runtime;
+  }
+  throw new Error(invalidInputMsg);
```
hund030, line 75: NIT: We better to clarify what input is invalid in the error message. User can't figure out what happened when getting such error `BT.UnhandledError: Invalid bot input`.

yukun-dong: This error message will never be shown to the users unless our code has bugs, and that's why I didn't localize this error message. But you are right at some point, I will find ways to make these error message better in the future.

jayzhang (approved):

Alive-Fish (approved):

PR #5254, opened 2022-06-09
perf(aad-manifest): align need provision message for deploying aad

SLdragon:
Currently, preview aad and deploy aad have different message when not provisioned:
align

This PR align these two messages:
aad

### DIFF
--- a/packages/fx-core/src/plugins/solution/fx-solution/v2/deploy.ts
+++ b/packages/fx-core/src/plugins/solution/fx-solution/v2/deploy.ts
@@ -100,7 +100,7 @@ export async function deploy(
         )
       );
     }
-  } else if (envInfo.envName !== "local") {
+  } else if (envInfo.envName !== "local" && inputs[Constants.DEPLOY_AAD] !== "yes") {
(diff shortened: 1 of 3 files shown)


packages/fx-core/src/plugins/solution/fx-solution/v2/deploy.ts
```diff
@@ -100,7 +100,7 @@ export async function deploy(
         )
       );
     }
-  } else if (envInfo.envName !== "local") {
+  } else if (envInfo.envName !== "local" && inputs[Constants.DEPLOY_AAD] !== "yes") {
```
1yefuwang1, line 103: This is more than a message change. Any background on this one?

SLdragon: This condition will hide select subscription if user hasn't been provisioned yet. Because aad related deploy do not need to select Azure related resource

KennethBWSong (approved):

1yefuwang1 (approved):

adashen (approved):

PR #5272, opened 2022-06-13
fix: project name restriction

yuqizhou77:
- Change the restriction of app name. It will allow almost all characters except "<>:\\?/*|\u0000-\u001F and cannot end with dot or space. And it has to start with a letter, and contain at least two alphanumeric characters, and the length should be less than 30. (Same as what we have today).
- AppName in projectSettings will be the original app name that may include special characters
- Remove non-alphanumeric characters in appName in plugins when provisioning resources.
- Update templates for C# projects to use safeProjectName for namespaces and projectName for file name. Related PR: https://devdiv.visualstudio.com/InternalTools/_git/VSProject/pullrequest/404491
- E2E test:
    - https://github.com/OfficeDev/TeamsFx/actions/runs/2506449140 original one
    - https://github.com/OfficeDev/TeamsFx/actions/runs/2507343488 projectName with dot

E2E TEST: na

### DIFF
--- a/packages/cli/tests/e2e/commonUtils.ts
+++ b/packages/cli/tests/e2e/commonUtils.ts
@@ -94,6 +94,10 @@ export function getUniqueAppName() {
   return getAppNamePrefix() + Date.now().toString() + uuidv4().slice(0, 2);
 }
 
+export function convertToAlphanumericOnly(appName: string): string {
+  return appName.replace(/[^\da-zA-Z]/g, "");
+}
+
 export function getSubscriptionId() {
   return cfg.AZURE_SUBSCRIPTION_ID || "";
 }
--- a/packages/fx-core/src/core/question.ts
+++ b/packages/fx-core/src/core/question.ts
@@ -59,9 +59,11 @@ export enum CoreQuestionNames {
   NewResourceGroupLocation = "newResourceGroupLocation",
   NewTargetEnvName = "newTargetEnvName",
   ExistingTabEndpoint = "existing-tab-endpoint",
+  SafeProjectName = "safeProjectName",
 }
 
-export const ProjectNamePattern = "^[a-zA-Z][\\da-zA-Z]+$";
+export const ProjectNamePattern =
+  '^(?=(.*[\\da-zA-Z]){2})[a-zA-Z][^"<>:\\?/*|\u0000-\u001F]*[^"\\s.<>:\\?/*|\u0000-\u001F]$';
 
 export function createAppNameQuestion(validateProjectPathExistence = true): TextInputQuestion {
   const question: TextInputQuestion = {
--- a/packages/fx-core/src/plugins/resource/aad/utils/configs.ts
+++ b/packages/fx-core/src/plugins/resource/aad/utils/configs.ts
@@ -40,6 +40,7 @@ import { BuiltInFeaturePluginNames } from "../../../solution/fx-solution/v3/cons
 import { ResultFactory } from "../results";
 import { getPermissionRequest } from "../permissions";
 import { GraphScopes, isAadManifestEnabled } from "../../../../common";
+import { convertToAlphanumericOnly } from "../../../../common/utils";
 
 export class Utils {
   public static addLogAndTelemetryWithLocalDebug(
@@ -267,15 +268,7 @@ export class ProvisionConfig {
     inputs: v2.InputsWithProjectPath,
     localSettings: v2.LocalSettings
   ): Promise<Result<any, FxError>> {
-    const displayName: string = ctx.projectSetting.appName;
-    if (displayName) {
-      this.displayName = displayName.substr(0, Constants.aadAppMaxLength) as string;
-    } else {
-      throw ResultFactory.SystemError(
-        GetConfigError.name,
-        GetConfigError.message(Errors.GetDisplayNameError[0])
-      );
-    }
+    this.setDisplayName(ctx.projectSetting.appName);
     const permissionRes = await getPermissionRequest(inputs.projectPath);
     if (permissionRes.isErr()) {
       return err(permissionRes.error);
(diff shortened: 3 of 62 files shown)


packages/fx-core/src/plugins/resource/frontend/dotnet/plugin.ts
```diff
@@ -111,11 +116,16 @@ export class DotnetPluginImpl implements PluginImpl {
     if (!ctx.projectSettings) {
       throw new NoProjectSettingError();
     }
-    const projectName = ctx.projectSettings!.appName;
-    await scaffoldFromZipPackage(ctx.root, new TemplateInfo({ ProjectName: projectName }));
+
+    const projectName = ctx.answers?.[CoreQuestionNames.AppName];
```
hund030, line 116: If we add tab by `addFeature` command, `ctx.answers` may not contain app name. I think we should get app name from project settings.

packages/fx-core/src/core/question.ts
```diff
@@ -78,7 +80,7 @@ export function createAppNameQuestion(validateProjectPathExistence = true): Text
         const validateResult = jsonschema.validate(appName, schema);
         if (validateResult.errors && validateResult.errors.length > 0) {
           if (validateResult.errors[0].name === "pattern") {
-            return getLocalizedString("core.QuestionAppName.validation.pattern");
+            return getLocalizedString("core.QuestionAppName.validation.pattern"); // TODO: update string
```
yuqizhou77, line 83: @MuyangAmigo Hi Junjie, we need to update the value of this string. The current value is "Application name must start with a letter and can only contain letters and digits."

packages/fx-core/src/plugins/resource/frontend/dotnet/plugin.ts
yuqizhou77: currently vs does not have "addFeature" command. need to sync with the team to understand the current process of that feature.

packages/fx-core/src/plugins/resource/aad/utils/configs.ts
```diff
@@ -267,7 +268,7 @@ export class ProvisionConfig {
     inputs: v2.InputsWithProjectPath,
     localSettings: v2.LocalSettings
   ): Promise<Result<any, FxError>> {
-    const displayName: string = ctx.projectSetting.appName;
+    const displayName: string = convertToAlphanumericOnly(ctx.projectSetting.appName);
```
SLdragon, line 271: Seems multiple places has these code, consider extract to a function:
```
    const displayName: string = convertToAlphanumericOnly(ctx.projectSetting.appName);
    if (displayName) {
      this.displayName = displayName.substr(0, Constants.aadAppMaxLength) as string;
    } else {
      throw ResultFactory.SystemError(
        GetConfigError.name,
        GetConfigError.message(Errors.GetDisplayNameError[0])
      );
    }
```

hund030 (approved):

packages/fx-core/src/plugins/resource/aad/utils/configs.ts
yuqizhou77: Fixed

packages/cli/tests/e2e/commonUtils.ts
```diff
@@ -94,6 +94,10 @@ export function getUniqueAppName() {
   return getAppNamePrefix() + Date.now().toString() + uuidv4().slice(0, 2);
 }
 
+export function convertToAlphanumericOnly(appName: string): string {
+  return appName.replace(/[^\da-zA-Z]/g, "");
```
nliu-ms, line 98: Just wondering will this happen: the appname contains only special characters, and after replacing it becomes an empty string.

nliu-ms: Never mind, it should start from a letter.

nliu-ms (dismissed):

packages/fx-core/src/core/question.ts
MuyangAmigo: Proposed message: `Application name must start with letters and contain at least two letters or digits. It can not contain specific ASCII characters.`

hund030 (approved):

KennethBWSong (dismissed):

yukun-dong (dismissed):

nliu-ms (approved):

SLdragon (approved):

packages/fx-core/src/plugins/resource/spfx/plugin.ts
```diff
@@ -107,7 +107,7 @@ export class SPFxPluginImpl {
 
       await progressHandler?.next(ScaffoldProgressMessage.ScaffoldProject);
       const framework = ctx.answers![SPFXQuestionNames.framework_type] as string;
-      const solutionName = ctx.projectSettings?.appName as string;
+      const solutionName = ctx.projectSettings?.appName as string; // I did not change this.
```
HuihuiWu-Microsoft, line 110: If you don't want to dismiss all the existing approvals, I'll remove the comment later in next PR.

HuihuiWu-Microsoft (approved):

packages/fx-core/src/plugins/resource/spfx/plugin.ts
yuqizhou77: Thanks for catching this! I will have another PR to remove this comment after checking this one in :)

yuqizhou77: Fixed

jayzhang (approved):

HuihuiWu-Microsoft (approved):

LongOddCode (approved):

tecton (approved):

chagong (approved):

nliu-ms (approved):

yukun-dong (approved):

Siglud (approved):

1yefuwang1 (approved):

XiaofuHuang (approved):

dooriya (approved):

hund030 (approved):

PR #5285, opened 2022-06-14
chore(bot): add support for .NET function bot scaffolding

yukun-dong:
image


packages/fx-core/src/plugins/resource/bot/question.ts
```diff
@@ -42,22 +42,22 @@ export const AppServiceOptionItem: HostTypeTriggerOptionItem = optionWithL10n({
...
+});
 
 export const HostTypeTriggerOptions: HostTypeTriggerOptionItem[] = [
   AppServiceOptionItem,
   FunctionsHttpTriggerOptionItem,
   FunctionsTimerTriggerOptionItem,
 ];
 
-export const HostTypeTriggerOptionsForVS: HostTypeTriggerOptionItem[] = [AppServiceOptionItemForVS];
+export const HostTypeTriggerOptionsForVS: HostTypeTriggerOptionItem[] = [
```
eriolchan, line 56: Since FunctionsHttpTriggerOptionItem and FunctionsTimerTriggerOptionItem are common between VS and VSCode, it looks like there could be HostTypeTriggerOptionItem[] including these 2 items. And then we add DefaultOptionItem by platform to that HostTypeTriggerOptionItem[], instead of creating 2 const array?

zyxiaoyuer: > image

For the Trigger strings/names, we can keep align with VS Code, by doing that, just need to change "webapi Server" to "Restify Server". The remainings looks good.

packages/fx-core/resource/package.nls.json
```diff
@@ -207,6 +207,10 @@
   "plugins.bot.triggers.http-restify.description": "Restify Server",
   "plugins.bot.triggers.http-restify.detail": "A running restify server hosted on Azure App Service can listen to HTTP requests.",
   "plugins.bot.triggers.http-restify.label": "HTTP Trigger",
+  "plugins.bot.triggers.http-webapi.cliName": "http-webapi",
+  "plugins.bot.triggers.http-webapi.description": "Webapi Server",
```
IvanJobs, line 211: Nit: Gave my proposal for ref: `Web API Server`

packages/fx-core/src/plugins/resource/bot/question.ts
zyxiaoyuer: thanks for clarify, then use Webapi will be good.

yukun-dong: > Since FunctionsHttpTriggerOptionItem and FunctionsTimerTriggerOptionItem are common between VS and VSCode, it looks like there could be HostTypeTriggerOptionItem[] including these 2 items. And then we add DefaultOptionItem by platform to that HostTypeTriggerOptionItem[], instead of creating 2 const array?

Done.

packages/fx-core/src/plugins/resource/bot/question.ts
```diff
@@ -68,12 +61,11 @@ export function createHostTypeTriggerQuestion(
 ): MultiSelectQuestion {
   const prefix = "plugins.bot.questionHostTypeTrigger";
 
-  let staticOptions: HostTypeTriggerOptionItem[] = HostTypeTriggerOptions;
   let defaultOptionItem = AppServiceOptionItem;
```
eriolchan, line 64: It seems to be
```
const defaultOptionItem = runtime === Rumtime.Dotnet ? AppServiceOptionItemForVS : AppServiceOptionItem;

File: packages/fx-core/src/plugins/resource/bot/configs/scaffoldConfig.ts [hunk 4]
```diff
@@ -64,7 +64,7 @@ export class ScaffoldConfig {
       // convert HostTypeTrigger question to trigger name
       this.triggers = rawHostTypeTriggers
         .map((hostTypeTrigger) => {
-          const option = HostTypeTriggerOptions.find((option) => option.id === hostTypeTrigger);
+          const option = FunctionsOptionItems.find((option) => option.id === hostTypeTrigger);
           return option?.trigger;
         })
         .filter((item): item is NotificationTrigger => item !== undefined);
```
eriolchan, line 70: Since we restrict HostTypeTriggerOptions to FunctionsOptionItems which is always a NotificationTrigger, do we still need L79 and NotificationTrigger interface anymore?

When we do the refactor, we could take a look at it further to check if there is anything out-of-date and can be removed.

packages/fx-core/src/plugins/resource/bot/configs/scaffoldConfig.ts
```diff
@@ -96,17 +96,18 @@ export class ScaffoldConfig {
...
+    const hostTypeTriggers = answers[QuestionNames.BOT_HOST_TYPE_TRIGGER];
+    let hostType: HostType = HostTypes.APP_SERVICE;
     if (Array.isArray(hostTypeTriggers)) {
-      const hostTypes = hostTypeTriggers.map(
-        (item) => HostTypeTriggerOptions.find((option) => option.id === item)?.hostType
-      );
-      return hostTypes ? hostTypes[0] : undefined;
-    } else {
-      return undefined;
+      FunctionsOptionItems.forEach((item) => {
```
eriolchan, line 104: After we use the new FunctionsOptionItems, the logic could look more clear by
```
if (Array.isArray(hostTypeTriggers)) {
 return FunctionsOptionItems.some(item => hostTypeTriggers.includes(item.id)) ? HostTypes.AZURE_FUNCTIONS : HostTypes.APP_SERVICE;
}
return HostTypes.APP_SERVICE;
```

packages/fx-core/src/plugins/resource/bot/v2/common.ts
```diff
@@ -58,14 +58,15 @@ export function decideTemplateScenarios(ctx: Context, inputs: Inputs): Set<strin
...
 export function resolveHostType(inputs: Inputs): HostType {
   const notificationTriggerType = inputs[QuestionNames.BOT_HOST_TYPE_TRIGGER];
-  let hostType;
+  let hostType: HostType = HostTypes.APP_SERVICE;
   if (Array.isArray(notificationTriggerType)) {
-    const hostTypes = notificationTriggerType.map(
-      (item) => HostTypeTriggerOptions.find((option) => option.id === item)?.hostType
-    );
-    hostType = hostTypes ? hostTypes[0] : undefined;
+    FunctionsOptionItems.forEach((item) => {
```
eriolchan, line 63: Same here

eriolchan (approved):

PR #5291, opened 2022-06-15
fix(vsc): persistence of subscription info

tecton:
Save subscription information when user specifies subscription.

Related bug: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/14716183
E2E TEST: N/A

### DIFF
--- a/packages/vscode-extension/src/commonlib/azureLogin.ts
+++ b/packages/vscode-extension/src/commonlib/azureLogin.ts
@@ -66,7 +66,6 @@ export class AzureAccountManager extends login implements AzureAccountProvider {
 
   private constructor() {
     super();
-    this.addStatusChangeEvent();
   }
 
   /**
--- a/packages/vscode-extension/src/commonlib/azureLogin.ts
+++ b/packages/vscode-extension/src/commonlib/azureLogin.ts
@@ -66,6 +66,7 @@ export class AzureAccountManager extends login implements AzureAccountProvider {
 
   private constructor() {
     super();
+    this.addStatusChangeEvent();
   }
 
   /**
@@ -78,6 +79,18 @@ export class AzureAccountManager extends login implements AzureAccountProvider {
     }
 
     return AzureAccountManager.instance;
+  }
+
+  /**
+   * Update TeamsFx project subscription information.
+   */
+  public async updateSubscriptionInfo(): Promise<void> {
+    if (AzureAccountManager.currentStatus === "LoggedIn") {
+      const subscriptioninfo = await this.readSubscription();
+      if (subscriptioninfo) {
+        this.setSubscription(subscriptioninfo.subscriptionId);
+      }
+    }
   }
 
   /**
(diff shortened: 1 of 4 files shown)


packages/vscode-extension/src/extension.ts
```diff
@@ -820,6 +821,7 @@ async function runBackgroundAsyncTasks(
   if (isTeamsFxProject) {
     await handlers.autoOpenProjectHandler();
     await TreeViewManagerInstance.updateTreeViewsByContent();
+    await AzureAccountManager.addStatusChangeEvent();
```
HuihuiWu-Microsoft, line 824: addStatusChangeEvent is more about AzureAccountManager itsself. Why do we move it outside?

HuihuiWu-Microsoft: This means statusMap was set before statusChange notify connected. If any azure events happen during this period, it will be missed. What about we doesn't export default in AzureAccountManager and use getInstance whenever needed?

packages/vscode-extension/src/handlers.ts
```diff
@@ -3131,6 +3131,7 @@ export async function selectSubscriptionCallback(args?: any[]): Promise<Result<n
     undefined
   );
   if (askSubRes.isErr()) return err(askSubRes.error);
+  await AzureAccountManager.saveSubscription(askSubRes.value);
```
xiaolang124, line 3134:
```suggestion
  await AzureAccountManager.setSubscription(askSubRes.value.subscriptionId);
```

tecton: Thanks, fixed.

packages/vscode-extension/src/extension.ts
tecton: Updated to keep event registration and split reading codes into another method.

HuihuiWu-Microsoft (approved):
LGTM

packages/vscode-extension/src/commonlib/azureLogin.ts
```diff
@@ -400,12 +412,6 @@ export class AzureAccountManager extends login implements AzureAccountProvider {
     const azureAccount: AzureAccount =
       vscode.extensions.getExtension<AzureAccount>("ms-vscode.azure-account")!.exports;
     AzureAccountManager.currentStatus = azureAccount.status;
-    if (AzureAccountManager.currentStatus === "LoggedIn") {
```
xiaolang124, line 403: If we call `updateSubscriptionInfo` both in this position and `runBackgroundAsyncTasks`, is there a problem?

tecton: I think there's no problem, in this position it probably does nothing. I'll add it back, thanks.

xiaolang124 (approved):

HuihuiWu-Microsoft (approved):

PR #5320, opened 2022-06-20
refactor(bot): use scenario to judge m365 template

yukun-dong:
E2E TEST: N/A


kuojianlu (approved):

packages/fx-core/src/plugins/solution/fx-solution/v2/utils.ts
```diff
@@ -328,6 +328,8 @@ export function fillInSolutionSettings(
     hostType = HostTypeOptionAzure.id;
   } else if (capabilities.includes(M365SearchAppOptionItem.id)) {
     capabilities = [MessageExtensionItem.id];
+    const scenarios = [M365SearchAppOptionItem.id];
+    answers[AzureSolutionQuestionNames.Scenarios] = scenarios;
```
IvanJobs, line 332: It looks like question model was used to pass information from solution to bot plugin, but it's not a real question, right?

packages/fx-core/src/plugins/resource/bot/v2/common.ts
```diff
@@ -31,17 +35,14 @@ export function getTemplateInfos(ctx: Context, inputs: Inputs): CodeTemplateInfo
 }
 
 export function decideTemplateScenarios(ctx: Context, inputs: Inputs): Set<string> {
-  const isM365 = ctx.projectSetting?.isM365;
```
IvanJobs, line 34: Any background about deprecating the use of `isM365` in projectSettings?

yukun-dong: We try to reduce the flags for deciding the template, so that the `decideTemplateScenarios` function would be simpler.

IvanJobs (approved):

eriolchan (approved):

1yefuwang1 (approved):

PR #5322, opened 2022-06-20
docs: update spfx doc to include compatibility

xu-hong:


docs/fx-core/spfx-help.md
```diff
@@ -1,14 +1,52 @@
...
+The following table lists SharePoint Framework and compatible versions of common tools and libraries, according to the [SharePoint Framework compatibility page](https://docs.microsoft.com/en-us/sharepoint/dev/spfx/compatibility#spfx-development-environment-compatibility):
+|SPFx|Node.js|NPM|TypeScript|React|
+|--------------|-----------|------------|-----------|------------|
+| 1.14 | LTS v12, LTS v14 | v5, v6 | v3.9 | v16.13.1 |
+
+### Error message
+By default,  locally installed `yo` is engaged to execute the following command with related _args_: `yo @microsoft/sharepoint`. 
+SPFx scaffolding could fail due to unsuccessful execution of above command:
+
+#### _"Failed to run command: " yo @microsoft/sharepoint ..." , 'yo' is not recognized as an internal or external command"_
```
HuihuiWu-Microsoft, line 17: There're three steps for spfx scaffolding now:
1. Check version for NPM and Node.js in create spfx question flow
2. Install yo and @microsoft/generator-sharepoint packages
3. run 'yo @microsoft/sharepoint' to scaffold
Since now we already have the version check in create question flow, I don't think user will get this message because of wrong version.
What about listing the error dialogs when version checking failed here?

HuihuiWu-Microsoft: As we already have troubleshooting part for step1 and step2.  Yeoman scaffolding can also fail in step3. I think maybe we can keep the 'check executable files' part for step3. Any ideas?

xu-hong: Listing error dialogs when version checking failed here? - agreed. Could you help provide the error message here?
Moving the "check executable files" to step3 makes sense - however, with versioning out of the way, what would contribute to this error - is it really rare? - in addition to check, what are the follow up steps developers can take?

HuihuiWu-Microsoft: Sure. I'll provide the error messages. 
Once the error happens, user can manually remove and install again in addition to check. But with version checking, actually I can't think of the reason that cause the error. Or we can remove this part and keep tracking and see what we can do according to user's reporting issues later.

HuihuiWu-Microsoft (approved):

PR #5326, opened 2022-06-20
chore: use createApiClient to call function API

blackchoey:
Use API client with BearerTokenAuthProvider to call functions, so users do not need to manually get token and set to header for every request.

### DIFF
--- a/templates/tab/js/default/src/components/sample/AzureFunctions.jsx
+++ b/templates/tab/js/default/src/components/sample/AzureFunctions.jsx
@@ -1,7 +1,7 @@
 import { useContext } from "react";
 import { Button, Loader } from "@fluentui/react-northstar";
 import { useData } from "@microsoft/teamsfx-react";
-import * as axios from "axios";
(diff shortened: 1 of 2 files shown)


hund030 (dismissed):

templates/tab/js/default/src/components/sample/AzureFunctions.jsx
```diff
@@ -11,13 +11,12 @@ async function callFunction(teamsfx) {
...
   }
   try {
-    const accessToken = await teamsfx.getCredential().getToken("");
-    const endpoint = teamsfx.getConfig("apiEndpoint");
-    const response = await axios.default.get(endpoint + "/api/" + functionName, {
-      headers: {
-        authorization: "Bearer " + accessToken.token,
-      },
-    });
+    const credential = teamsfx.getCredential();
```
MuyangAmigo, line 14: Do we want to add some comments here?

MuyangAmigo (approved):

hund030 (approved):

PR #5334, opened 2022-06-21
fix: make local debug dev cert properties more clear

a1exwang:
Add property `debug-dev-cert-status` to make cert telemetry more clear

Test (timestamp is in reverse order):
- first event is cancel in first dialog,
- second event is cancel in the second dialog
- third event is disable in vscode settings
- fourth event is confirm trust
- fourth event is second time local debug

E2E TEST: https://github.com/OfficeDev/TeamsFx-UI-Test/blob/76b1d9c8ea6d3d74fb9dbdbeddc28b3afd4259f6/src/ui-test/localdebug/localdebug-bot.test.ts

### DIFF
--- a/packages/fx-core/src/common/local/localCertificateManager.ts
+++ b/packages/fx-core/src/common/local/localCertificateManager.ts
@@ -43,6 +43,7 @@ export interface LocalCertificate {
   certPath: string;
   keyPath: string;
   isTrusted?: boolean;
+  alreadyTrusted?: boolean;
   error?: FxError;
 }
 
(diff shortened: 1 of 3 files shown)


kuojianlu (commented):
What's the behavior in CLI part?

swatDong (approved):

a1exwang:
> What's the behavior in CLI part?

CLI should be similarly handled. I will work on CLI telemetry when extension is finished altogether.

kuojianlu (approved):

tecton (approved):

PR #5335, opened 2022-06-21
chore: use createApiClient in M365 tab template

blackchoey:
Use API client with BearerTokenAuthProvider to call functions, so users do not need to manually get token and set to header for every request.

### DIFF
--- a/templates/tab/js/m365/src/components/sample/AzureFunctions.jsx
+++ b/templates/tab/js/m365/src/components/sample/AzureFunctions.jsx
@@ -1,21 +1,20 @@
 import React from "react";
 import { Button, Loader } from "@fluentui/react-northstar";
 import { useData } from "./lib/useData";
-import * as axios from "axios";
-import { TeamsFx } from "@microsoft/teamsfx";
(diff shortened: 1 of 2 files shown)


kuojianlu (approved):

templates/tab/js/m365/src/components/sample/AzureFunctions.jsx
```diff
@@ -1,21 +1,20 @@
 import React from "react";
 import { Button, Loader } from "@fluentui/react-northstar";
 import { useData } from "./lib/useData";
-import * as axios from "axios";
-import { TeamsFx } from "@microsoft/teamsfx";
+import { BearerTokenAuthProvider, createApiClient, TeamsFx } from "@microsoft/teamsfx";
```
swatDong, line 4: M365 app uses teamsfx 2.x. Are these APIs included in teamsfx 2.x?

blackchoey: Yes, have tested this template locally.

swatDong (approved):

PR #5347, opened 2022-06-22
fix: retry exclude 404

IvanJobs:
404 was treated as Exception in Axios, so extract it out as a normal response to represent nonexistence.

https://dev.azure.com/msazure/Microsoft%20Teams%20Extensibility/_workitems/edit/14727888

E2E TEST: None

### DIFF
--- a/packages/fx-core/src/plugins/resource/bot/appStudio/appStudio.ts
+++ b/packages/fx-core/src/plugins/resource/bot/appStudio/appStudio.ts
@@ -177,7 +177,7 @@ export class AppStudio {
           },
           true
         );
-        if (getBotRegistrationResponse?.status === 201) {
+        if (getBotRegistrationResponse?.status === 200) {
           Logger.info(Messages.BotResourceExist("Appstudio"));
           return;
         }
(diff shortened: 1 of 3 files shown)


a1exwang (approved):

yukun-dong (approved):

packages/fx-core/src/plugins/resource/bot/appStudio/appStudio.ts
```diff
@@ -161,11 +161,23 @@ export class AppStudio {
...
-        const getBotRegistrationResponse = await RetryHandler.Retry(
-          () => axiosInstance.get(`${AppStudio.baseUrl}/api/botframework/${registration.botId}`),
+        const getBotRegistrationResponse: AxiosResponse<any> | undefined = await RetryHandler.Retry(
+          async () => {
+            try {
+              return await axiosInstance.get(
+                `${AppStudio.baseUrl}/api/botframework/${registration.botId}`
+              );
+            } catch (e) {
+              if (e.response?.status === 404) {
```
eriolchan, line 171: Do we also need to include other 4xx user error?

Siglud (approved):

packages/fx-core/src/plugins/resource/bot/appStudio/appStudio.ts
IvanJobs: I've thought about it after previous discussion with you, but think it may be not worth the efforts to include more 4xx, because:
1. As you can see from the REST api, it's only about `Get xxx/api/{botId}` and `botId` would be a good since there's a check before. My point is there's no big chance for other 4xx except 404.
2. Think from the other way, if we include other 4xx, then the logic may be throwing some error messages for other 4XX to the end users. It may iterate all 4xx and give different error messages, kind of complex.

eriolchan (approved):

PR #5350, opened 2022-06-22
fix(bot): add bot capabilities into projectsettings.json

yukun-dong:
fix https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/14786120.
E2E TEST: N/A
In this PR the projectSettings.json looks like below, the same as 4.0.0.
image

### DIFF
--- a/packages/fx-core/src/plugins/resource/bot/v2/common.ts
+++ b/packages/fx-core/src/plugins/resource/bot/v2/common.ts
@@ -82,9 +82,7 @@ export function resolveServiceType(ctx: Context): ServiceType {
 export function resolveBotCapabilities(inputs: Inputs): BotCapability[] {
   const botScenarios = inputs?.[AzureSolutionQuestionNames.Scenarios];
   if (Array.isArray(botScenarios)) {
-    return botScenarios
-      .map((scenario) => QuestionBotScenarioToBotCapability.get(scenario))
-      .filter((item): item is BotCapability => item !== undefined);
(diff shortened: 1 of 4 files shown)


a1exwang (dismissed):

packages/fx-core/src/plugins/resource/bot/v2/common.ts
```diff
@@ -78,3 +78,14 @@ export function resolveServiceType(ctx: Context): ServiceType {
     ] as string) ?? HostType.AppService;
   return getServiceType(rawHostType);
 }
+
+export function resolveBotCapabilities(inputs: Inputs): BotCapability[] {
+  const botScenarios = inputs?.[AzureSolutionQuestionNames.Scenarios];
+  if (Array.isArray(botScenarios)) {
+    return botScenarios
+      .map((scenario) => QuestionBotScenarioToBotCapability.get(scenario))
+      .filter((item): item is BotCapability => item !== undefined);
```
Siglud, line 87: The code is hard to understand. How about changing it to 
```
return botScenarios
      .filter(it => QuestionBotScenarioToBotCapability.has(it))
      .map((scenario) => QuestionBotScenarioToBotCapability.get(scenario))
```

yukun-dong: Even if filter it first, the return type of get method is still `BotCapability|undifined` and still need a second filter. But indeed it;s hard to understand. I will change it to get()! here.

eriolchan (approved):

PR #5355, opened 2022-06-22
chore: Basic office add-in generation

millerds:
Added office add-in template questions, generation code, and VS Code configuration for add-in debugging.


packages/fx-core/src/plugins/resource/officeaddin/questions.ts
```diff
@@ -1,58 +1,76 @@
...
+  name: QuestionName.AddinNameQuestion,
+  title: "Add-in name",
+  default: "office-addin",
 };
 
-export const MultiSelectOptionTwo: OptionItem = {
-  id: "multi-option2",
-  label: "Option 2 label",
-  detail: "Option 2 detail",
+export const AddinLanguageQuestion: SingleSelectQuestion = {
```
1yefuwang1, line 37: We already have a question that asks for choosing programming language between js and ts.  Can we reuse it?

millerds: The question I made uses dynamic options based on a json file that describes the different template options we have and what supported on each of them (including which languages are available).  Is that something that can be done with the existing question?

1yefuwang1: If the json file is committed to the repo, I think the questions are actually statically determined at compile time. You can read the json file and generate the questions.

millerds: I think I get what you mean . . . but I'm not sure about the implementation.  This question creates the options available (using dynamicOptions) based on the content of the json file (yes statically added to the repo).  How would I instead conditionally create the question (using your existing question)?  I'm guessing I would use a function that returns the question instead of this type in index.ts . . . but at that point I don't know if I want to ask the question or what the options are because it depends on the answer to an earlier question.

PR #5381, opened 2022-06-27
test(bot): add ut for add capability to notification and command bot

yukun-dong:


packages/fx-core/tests/plugins/solution/solution.getQuestions.test.ts
```diff
@@ -625,4 +625,178 @@ describe("getQuestionsForScaffolding()", async () => {
       }
     }
   });
+  it("getQuestionsForUserTask - addFeature: cannot add message extension for notification bot", async () => {
+    sandbox.stub(featureFlags, "isPreviewFeaturesEnabled").returns(true);
```
eriolchan, line 629: Since some feature flag has been removed since GA, does any flag can be removed?

PR #5384, opened 2022-06-27
docs: update spfx README to simplify sharepoint site construct

HuihuiWu-Microsoft:
This is to fix this bug https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/12664738


xu-hong (approved):
LGTM

PR #5392, opened 2022-06-28
chore: remove unify config flag

xiaolang124:
1.Remove unused `unify config` feature flag.

E2E: https://github.com/OfficeDev/TeamsFx/actions/runs/2573204071 (works just as dev branch)


packages/fx-core/src/core/middleware/projectMigrator.ts
```diff
@@ -78,6 +78,7 @@ import { LocalEnvProvider } from "../../common/local/localEnvProvider";
 import { CoreHookContext } from "../types";
 import { TOOLS } from "../globalVars";
 import { getLocalizedString } from "../../common/localizeUtils";
+import { URL } from "url";
```
dooriya, line 81: Seems it is unused and can be removed.

HuihuiWu-Microsoft (approved):

PR #5393, opened 2022-06-28
chore: enhance error message

nliu-ms:
image


MuyangAmigo:
Shall we replace `app studio` with `Developer Portal`?

MuyangAmigo:
Also can we remove `Due to 404`? The next sentence seems can cover that?

packages/fx-core/src/plugins/resource/appstudio/appStudio.ts
```diff
@@ -48,6 +48,26 @@ export namespace AppStudioClient {
...
+
+    const message = getLocalizedString(
+      "error.appstudio.apiFailed",
+      e.name,
+      e.message,
+      requestPath,
+      correlationId,
+      extraData
+    );
+    const error = new Error(message);
```
HuihuiWu-Microsoft, line 64: In this case, will the display message be constructed by localized string?

HuihuiWu-Microsoft (approved):

PR #5394, opened 2022-06-28
chore: setup vsc extension unit test framework to allow test run without extension host

HuihuiWu-Microsoft:


packages/vscode-extension/test/unit/migration/data/js/others/replace-getContext-commentLine.input.js
```diff
@@ -1,10 +0,0 @@
-import * as microsoftTeams from "@microsoft/teams-js";
```
swatDong, line 1: Update CODEOWNERS since the folder path is changed.

HuihuiWu-Microsoft: Updated

swatDong (approved):

tecton (approved):

PR #5400, opened 2022-06-28
fix: open readme for walkthrough app

nliu-ms:
For bug: 14523696

image

### DIFF
--- a/packages/vscode-extension/src/handlers.ts
+++ b/packages/vscode-extension/src/handlers.ts
@@ -459,11 +459,10 @@ export async function updateAutoOpenGlobalKey(
 ): Promise<void> {
   if (isTriggerFromWalkThrough(args)) {
     await globalStateUpdate(GlobalKey.OpenWalkThrough, true);
-    await globalStateUpdate(GlobalKey.OpenReadMe, "");
   } else {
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/handlers.ts
```diff
@@ -459,11 +459,10 @@ export async function updateAutoOpenGlobalKey(
 ): Promise<void> {
   if (isTriggerFromWalkThrough(args)) {
     await globalStateUpdate(GlobalKey.OpenWalkThrough, true);
-    await globalStateUpdate(GlobalKey.OpenReadMe, "");
   } else {
     await globalStateUpdate(GlobalKey.OpenWalkThrough, false);
-    await globalStateUpdate(GlobalKey.OpenReadMe, projectUri.fsPath);
   }
+  await globalStateUpdate(GlobalKey.OpenReadMe, projectUri.fsPath);
```
HuihuiWu-Microsoft, line 465: In this case, when user create empty project from walkthrough, both readme and walkthrough will be opened. I suppose for empty project, we open walkthrough and for sample we open readme.  Correct me if I was wrong.

HuihuiWu-Microsoft (approved):

PR #5409, opened 2022-06-29
perf: add common propery in vsc, cli and core

KennethBWSong:
task: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_sprints/taskboard/AuthAndData/Microsoft%20Teams%20Extensibility/CY22-6.2?workitem=14813970

### DIFF
--- a/packages/cli/src/commonlib/telemetry.ts
+++ b/packages/cli/src/commonlib/telemetry.ts
@@ -4,7 +4,7 @@
 
 import Reporter from "../telemetry/telemetryReporter";
 import { TelemetryReporter } from "@microsoft/teamsfx-api";
-import { Correlator } from "@microsoft/teamsfx-core";
+import { Correlator, getProjectSettingsForCommonTelemetry } from "@microsoft/teamsfx-core";
 import { TelemetryProperty } from "../telemetry/cliTelemetryEvents";
 import { getAllFeatureFlags, getProjectId } from "../utils";
 import { CliConfigOptions } from "../userSetttings";
@@ -45,8 +45,7 @@ export class CliTelemetryReporter implements TelemetryReporter {
       properties = {};
     }
 
-    const projectId = getProjectId(this.rootFolder);
-    properties[TelemetryProperty.ProjectId] = projectId ? projectId : "";
+    this.addCommonProperties(this.rootFolder, properties);
     properties[TelemetryProperty.CorrelationId] = Correlator.getId();
 
     properties[CliConfigOptions.RunFrom] = tryDetectCICDPlatform();
@@ -66,8 +65,7 @@ export class CliTelemetryReporter implements TelemetryReporter {
       properties = {};
     }
 
-    const projectId = getProjectId(this.rootFolder);
-    properties[TelemetryProperty.ProjectId] = projectId ? projectId : "";
+    this.addCommonProperties(this.rootFolder, properties);
     properties[TelemetryProperty.CorrelationId] = Correlator.getId();
 
     properties[CliConfigOptions.RunFrom] = tryDetectCICDPlatform();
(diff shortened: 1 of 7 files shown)


packages/fx-core/src/common/tools.ts
```diff
@@ -948,3 +950,39 @@ export async function getSPFxToken(
...
+    return {
+      projectId: projectSettings.projectId ?? undefined,
+      version: projectSettings.version ?? undefined,
+      isFromSample: projectSettings.isFromSample ?? undefined,
+      solutionSettings: projectSettings.solutionSettings
+        ? JSON.stringify(projectSettings.solutionSettings)
+        : undefined,
+      programmingLanguage: projectSettings.programmingLanguage ?? undefined,
+      pluginSettings: projectSettings.pluginSettings
+        ? JSON.stringify(projectSettings.pluginSettings)
```
hund030, line 982: pluginSettings may contains csproj file path. Will that cause the whole pluginSettings be suppressed?

packages/fx-core/src/common/tools.ts
```diff
@@ -948,3 +950,39 @@ export async function getSPFxToken(
...
+      `.${ConfigFolderName}`,
+      InputConfigsFolderName,
+      ProjectSettingsFileName
+    );
+
+    if (!projectSettingsPath || !fs.pathExistsSync(projectSettingsPath)) {
+      return undefined;
+    }
+
+    const projectSettings = fs.readJsonSync(projectSettingsPath);
```
hund030, line 972: Is there a common method for loading project settings? Reading the raw project setting file may be not a good practice.

packages/cli/src/commonlib/telemetry.ts
```diff
@@ -102,4 +99,37 @@ export class CliTelemetryReporter implements TelemetryReporter {
...
+    if (!properties[TelemetryProperty.IsFromSample]) {
+      properties[TelemetryProperty.IsFromSample] = projectSettings?.isFromSample;
+    }
+
+    if (!properties[TelemetryProperty.ProjectSettingsVersion]) {
+      properties[TelemetryProperty.ProjectSettingsVersion] = projectSettings?.version;
+    }
+
+    if (!properties[TelemetryProperty.SolutionSettings]) {
+      properties[TelemetryProperty.SolutionSettings] = projectSettings?.solutionSettings;
```
blackchoey, line 124: do we really need to record the whole solution settings and plugin settings? The content of them may be out of control and cause problems.

packages/vscode-extension/src/commonlib/telemetry.ts
```diff
@@ -177,4 +160,38 @@ export class VSCodeTelemetryReporter extends vscode.Disposable implements Teleme
...
+    if (
+      properties[TelemetryProperty.ProjectId] === "unknown" ||
+      properties[TelemetryProperty.ProjectId] === undefined
+    ) {
+      properties[TelemetryProperty.ProjectId] = projectSettings?.projectId
+        ? projectSettings?.projectId
+        : "unknown";
+    }
+
+    if (!properties[TelemetryProperty.IsFromSample]) {
```
tecton, line 177: This is covered in `sharedProperties`. It will not be changed during the lifecycle of a project, so it's initialized once and not updated. Does it make sense to set dynamic properties in this method and move static ones to `sharedProperties`.

packages/fx-core/src/common/tools.ts
a1exwang: Should we whitelist fields for `pluginSettings`? 

If so, `host-type` and `capabilities` in bot plugin should be whitelisted.

```
    "pluginSettings": {
        "fx-resource-bot": {
            "host-type": "app-service",
            "capabilities": [
                "notification"
            ]
        }
    },
```

KennethBWSong:
Closed according to discussion. Will open new prs to add common property.

PR #5419, opened 2022-06-30
fix: differentiate manual publish

nliu-ms:
There is a question for publish, which only build the package. Update telemetry to treat it as build event.
image


nliu-ms:
E2E TEST: N/A

HuihuiWu-Microsoft (approved):

PR #5421, opened 2022-06-30
fix: able to f5 .net project without appsettings.development.json

swatDong:
Fix https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/14869344
In case `appsettings.development.json` is git ignored so let F5 to generate one. Fix both Tab and Bot projects.

E2E TEST: N/A

### DIFF
--- a/packages/fx-core/src/plugins/resource/frontend/dotnet/plugin.ts
+++ b/packages/fx-core/src/plugins/resource/frontend/dotnet/plugin.ts
@@ -215,18 +215,40 @@ export class DotnetPluginImpl implements PluginImpl {
 
   public async postLocalDebug(ctx: PluginContext): Promise<TeamsFxResult> {
     const appSettingsPath = path.join(ctx.root, PathInfo.appSettingDevelopment);
+    let appSettings: string;
     if (
       !(await runWithErrorCatchAndThrow(
         new FileIOError(appSettingsPath),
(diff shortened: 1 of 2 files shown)


packages/fx-core/src/plugins/resource/frontend/dotnet/plugin.ts
```diff
@@ -215,18 +215,40 @@ export class DotnetPluginImpl implements PluginImpl {
...
         new FileIOError(appSettingsPath),
         async () => await fs.pathExists(appSettingsPath)
       ))
     ) {
-      return ok(undefined);
+      // if appsetting file not exist, generate a new one
+      // TODO(qidon): load content from resource file or template
+      appSettings =
+        '\
+{\r\n\
```
hund030, line 229: NIT: Prefer os.EOL.

kimizhu (approved):

qinezh (approved):

packages/fx-core/src/plugins/resource/frontend/dotnet/plugin.ts
swatDong: I believe we only support VS(windows) now, so this quick fix just use `\r\n`.
For further improvement considering share the same content with template.

PR #5424, opened 2022-06-30
chore(bot): add bot app registration error telemetry

XiaofuHuang:
Add bot app registration error telemetry
Detail: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/14549235

### DIFF
--- a/packages/fx-core/src/plugins/resource/bot/appStudio/appStudio.ts
+++ b/packages/fx-core/src/plugins/resource/bot/appStudio/appStudio.ts
@@ -14,6 +14,9 @@ import { CommonStrings, ConfigNames } from "../resources/strings";
 import { RetryHandler } from "../utils/retryHandler";
 import { Messages } from "../resources/messages";
 import { Logger } from "../logger";
+import { telemetryHelper } from "../utils/telemetry-helper";
+import { TelemetryReporter } from "@microsoft/teamsfx-api";
+import { TelemetryEventNames } from "../constants";
 /**
  * Get app studio endpoint for prod/int environment, mainly for ux e2e test
  */
@@ -24,6 +27,12 @@ export function getAppStudioEndpoint(): string {
     return "https://dev.teams.microsoft.com";
   }
 }
+
+function sendAppStudioApiTelemetry(telemetryReporter: TelemetryReporter | undefined, e: any) {
+  telemetryHelper.sendAxiosApiError(telemetryReporter, TelemetryEventNames.AppStudioApi, e);
+  throw e;
+}
+
 export class AppStudio {
   private static baseUrl = getAppStudioEndpoint();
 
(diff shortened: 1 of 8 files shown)


yukun-dong (approved):

yukun-dong (dismissed):

a1exwang (approved):

packages/fx-core/src/plugins/resource/bot/constants.ts
```diff
@@ -125,6 +125,7 @@ export class ErrorNames {
   public static readonly MSG_ENDPOINT_UPDATING_ERROR = "MessageEndpointUpdatingError";
   public static readonly COMMAND_EXECUTION_ERROR = "CommandExecutionError";
   public static readonly CALL_APPSTUDIO_API_ERROR = "CallAppStudioAPIError";
+  public static readonly API_ERROR = "ApiError";
```
eriolchan, line 128: It seems that the telemetry is tracking app studio api error, can it reuse the above CALL_APPSTUDIO_API_ERROR ?

packages/fx-core/src/plugins/resource/bot/appStudio/appStudio.ts
```diff
@@ -176,7 +186,7 @@ export class AppStudio {
             }
           },
           true
-        );
+        ).catch((e) => sendAppStudioApiTelemetry(telemetryReporter, e));
```
eriolchan, line 189: Is it possible to do this catch and send telemetry outside the appStudio.ts? It looks too intrusive to the existing code to pass telemetryReport to each function call, which make code not clean. There is already a guideline in Bot plugin on how to handle exception and telemetry in a unified way, @IvanJobs can you help review this PR and suggest the right way to add telemetry for appStudio call?

IvanJobs: @XiaofuHuang I agree with @eriolchan on the intrusive part.
According to the original design, for example, https://github.com/OfficeDev/TeamsFx/blob/60d497a45bf7987c3037a0c06b131393ed644f37/packages/fx-core/src/plugins/resource/bot/index.ts#L100, all exceptions are supposed to be caught on upper layer - `index.ts` which will significantly simplify the logic, provide single source of truth and reduce a lot of scattered code like what it is in the PR.

How about moving the logic on top layer? We don't want `context` or `telemetryReporter` being passed for all lower layers functions like this.

XiaofuHuang: Thx. Moved it to telemetryReporter

packages/fx-core/src/plugins/resource/bot/constants.ts
XiaofuHuang: Removed the `ApiError` error event.

IvanJobs (approved):
lgtm

PR #5444, opened 2022-07-04
perf: add fixed common property in telemetry

KennethBWSong:
common telemetry part 1: 
1. add property that will not change in a project into telemetry
2. add shared property in cli
3. use empty string instead of unknown for empty property


packages/vscode-extension/src/commonlib/telemetry.ts
```diff
@@ -93,11 +94,30 @@ export class VSCodeTelemetryReporter extends vscode.Disposable implements Teleme
     }
 
     if (
-      properties[TelemetryProperty.ProjectId] === "unknown" ||
-      properties[TelemetryProperty.ProjectId] === undefined
+      !properties[TelemetryProperty.ProjectId] ||
```
a1exwang, line 97: It seems overriding a common property is supported, right? If so, is it better to separate these conditions because if the users of this function want to override a common property they have to specify **all three properties**?

packages/vscode-extension/src/commonlib/telemetry.ts
```diff
@@ -156,11 +195,30 @@ export class VSCodeTelemetryReporter extends vscode.Disposable implements Teleme
...
```
a1exwang, line 198: Maybe add a function `addFixedCommonProperties(properties, sharedProperties)` to prevent copy-pasting these code.

packages/vscode-extension/src/commonlib/telemetry.ts
a1exwang: Offline discussed with @KennethBWSong, it is not intended behavior to override common properties.

packages/vscode-extension/src/commonlib/telemetry.ts
KennethBWSong: Updated, thx!

a1exwang (approved):

packages/cli/src/commonlib/telemetry.ts
```diff
@@ -31,22 +32,40 @@ export class CliTelemetryReporter implements TelemetryReporter {
...
+
+      // add shared properties
+      const fixedProjectSettings = getFixedCommonProjectSettings(rootPath);
+      this.addSharedProperty(TelemetryProperty.ProjectId, fixedProjectSettings?.projectId ?? "");
+      this.addSharedProperty(
+        TelemetryProperty.IsFromSample,
+        fixedProjectSettings?.isFromSample ?? ""
+      );
+      this.addSharedProperty(
+        TelemetryProperty.ProgrammingLanguage,
```
chagong, line 44: I don't know if it's worthy to add programming language as a common property. Do we use this property anywhere for BI purpose?

chagong: One more question is: will it be only one programming language for one Teams App project?

KennethBWSong: I remember @hund030 suggested to add this in last vteam meeting.
For one project, the language will be selected at first and can not change later.

packages/cli/src/commonlib/telemetry.ts
```diff
@@ -31,22 +32,40 @@ export class CliTelemetryReporter implements TelemetryReporter {
...
+        fixedProjectSettings?.programmingLanguage ?? ""
+      );
+      this.addSharedProperty(TelemetryProperty.HostType, fixedProjectSettings?.hostType ?? "");
```
chagong, line 47: Is user able to change "HostType" after a project is created?

KennethBWSong: Host type will be decide when create a project as spfx or azure. It can not be changed.

chagong:
Did you check if it's working for VS extension?

packages/fx-core/src/common/tools.ts
```diff
@@ -945,3 +947,32 @@ export async function getSPFxToken(
...
+      `.${ConfigFolderName}`,
+      InputConfigsFolderName,
+      ProjectSettingsFileName
+    );
+
+    if (!projectSettingsPath || !fs.pathExistsSync(projectSettingsPath)) {
+      return undefined;
+    }
+
+    const projectSettings = fs.readJsonSync(projectSettingsPath);
```
chagong, line 968: Do we still need to do IO operation for read the project settings file every time? Thus we don't improve any performance?

packages/cli/src/commonlib/telemetry.ts
```diff
@@ -31,22 +32,40 @@ export class CliTelemetryReporter implements TelemetryReporter {
     if (rootPath) {
       this.rootFolder = rootPath;
       this.reporter.setAppRoot(rootPath);
+
+      // add shared properties
+      const fixedProjectSettings = getFixedCommonProjectSettings(rootPath);
+      this.addSharedProperty(TelemetryProperty.ProjectId, fixedProjectSettings?.projectId ?? "");
```
adashen, line 38: instead of add lots of check here. How about move the check string part into addSharedProperty method?

packages/fx-core/src/common/tools.ts
```diff
@@ -945,3 +947,32 @@ export async function getSPFxToken(
...
+    return {
+      projectId: projectSettings?.projectId ?? undefined,
+      isFromSample: projectSettings?.isFromSample ?? undefined,
+      programmingLanguage: projectSettings?.programmingLanguage ?? undefined,
+      hostType: projectSettings?.solutionSettings?.hostType,
```
adashen, line 973: no need fallback undefined for hostType?

packages/fx-core/src/common/tools.ts
KennethBWSong: For fixed properties, it will be read only once

packages/cli/src/commonlib/telemetry.ts
KennethBWSong: Fixed, thx!

packages/fx-core/src/common/tools.ts
KennethBWSong: Fixed, thx!

packages/vscode-extension/src/commonlib/telemetry.ts
```diff
@@ -52,7 +53,7 @@ export class VSCodeTelemetryReporter extends vscode.Disposable implements Teleme
   }
 
   addSharedProperty(name: string, value: string): void {
-    this.sharedProperties[name] = value;
+    this.sharedProperties[name] = value ?? "";
```
tecton, line 56: The parameter `value` can be changed to optional? Like `value?: string`

KennethBWSong: Fixed, thx!

tecton (dismissed):

chagong (dismissed):

XiaofuHuang (approved):

chagong (approved):

tecton (approved):

XiaofuHuang (approved):

PR #5452, opened 2022-07-05
chore: upgrade the newton version

wenytang-ms:


packages/function-extension/src/Microsoft.Azure.WebJobs.Extensions.TeamsFx.csproj
```diff
@@ -27,6 +27,7 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
     <PackageReference Include="Microsoft.Identity.Client" Version="4.22.0" />
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
```
blackchoey, line 30: Can we add a comment mentioning that this is only used to fix vulnerability?

wenytang-ms: added

blackchoey (approved):

PR #5453, opened 2022-07-05
chore(app-studio): send status code when call app studio api failed

XiaofuHuang:
send status code when call app studio api failed
Detail: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/14549235


nliu-ms (approved):

nliu-ms:
Avoid request path in telemetry data, only contains API name:
image

Show request path in display error message:
image

PR #5455, opened 2022-07-05
chore: change default publish path for vs

JerryYangKai:


hund030 (approved):
LGTM, but user still cannot specify build output path. It is expected?

JerryYangKai:
> LGTM, but user still cannot specify build output path. Is it expected?

In this way we just get the output into a specific folder.
User can change the build path in VS and it won't affect the Teams Toolkit deploy pipeline

packages/fx-core/src/plugins/resource/bot/v2/plugin.ts
```diff
@@ -291,11 +291,8 @@ export class TeamsBotV2Impl {
 
     if (lang === ProgrammingLanguage.Csharp) {
       try {
-        const framework = await TeamsBotV2Impl.getFrameworkVersion(
-          path.join(workingPath, projectFileName)
-        );
-        await utils.execute(`dotnet publish --configuration Release`, workingPath);
-        return path.join(workingPath, "bin", "Release", framework, "publish");
+        await utils.execute(`dotnet publish --output publish --configuration Release`, workingPath);
```
eriolchan, line 294: Can you help me learn more about this code change?
I think the expected result is that 
1. In scaffolding, we put our default build folder in csproj setting, so that it behaves the same as before.
2. If user want to customize the output value in csproj, toolkit will respect that value as the source of deployment.

Maybe I can offline check with you whether the fix meet above scenarios.

eriolchan (approved):

PR #5459, opened 2022-07-06
chore(vsc): fix doc link and ITP content

tecton:
Related bug: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/14000411

### DIFF
--- a/packages/vscode-extension/media/itp/itp.md
+++ b/packages/vscode-extension/media/itp/itp.md
@@ -16,7 +16,7 @@ If you don't have a personal Microsoft account, or a work or school account. You
 
 ![Create account](./createone.png)
 
->Note: You cannot Sign in with *.onmicrosoft.com account, otherwise you will meet following error in next step.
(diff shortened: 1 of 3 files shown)


packages/vscode-extension/media/itp/itp.md
```diff
@@ -59,5 +61,3 @@ If you don't have a personal Microsoft account, or a work or school account. You
 ![Instant sandbox](./turn-on-sideload.png)
 
 - Select `Save`.
-
->Note: It can take up to 24 hours for the sideloading option to be active.
```
HuihuiWu-Microsoft, line 63: Does this mean user doesn't need to wait for 24h before being able to use sideloading?

tecton: Let's check with PM.

HuihuiWu-Microsoft (approved):

PR #5462, opened 2022-07-06
chore(vsc): split the trigger-from to command palette and treeview

tecton:
Related bug: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/14893247


packages/vscode-extension/src/extension.ts
```diff
@@ -452,8 +457,7 @@ function registerMenuCommands(context: vscode.ExtensionContext) {
 
   const createAccountCmd = vscode.commands.registerCommand(
     "fx-extension.createAccount",
-    (...args) =>
-      Correlator.run(handlers.createAccountHandler, [TelemetryTriggerFrom.ViewTitleNavigation])
+    (...args) => Correlator.run(handlers.createAccountHandler, [TelemetryTriggerFrom.TreeView])
```
HuihuiWu-Microsoft, line 460: Why use TelemetryTriggerFrom.TreeView instead of TelemetryTriggerFrom.ViewTitleNavigation?

tecton: Because the command is still in tree view and have no conflict with other positions. So change it back to tree view to reduce confusion.

HuihuiWu-Microsoft: In this way, for events with 'TelemetryTriggerFrom.TreeView' trigger-from property, some are from tree items and some are from navigation icon on tree items. Also if the 'createAccount' operation is added to tree item in the further, we'll have to change this one to 'TelemetryTriggerFrom.ViewTitleNavigation' which is quite confusing to me.

tecton: Yes, we can discuss with @dooriya

tecton: Updated.

HuihuiWu-Microsoft (approved):

PR #5468, opened 2022-07-07
chore: sync before updating endpoint

IvanJobs:
Updating message endpoint for bot will flush original config on remote, for example, calling endpoint.
Fix:
    Sync bot registration before updating message endpoint.

https://dev.azure.com/msazure/Microsoft%20Teams%20Extensibility/_workitems/edit/14892895


nliu-ms (approved):

jayzhang (approved):

packages/fx-core/src/plugins/resource/bot/appStudio/appStudio.ts
```diff
@@ -152,6 +152,33 @@ export class AppStudio {
     return app;
   }
 
+  public static async getBotRegistration(
+    accessToken: string,
+    botId: string
+  ): Promise<IBotRegistration | undefined> {
+    const axiosInstance = AppStudio.newAxiosInstance(accessToken);
+
+    const getBotRegistrationResponse: AxiosResponse<any> | undefined = await RetryHandler.Retry(
```
Siglud, line 161: I think if you change 
```
AxiosResponse<any> 
```
to 
```
AxiosResponse<IBotRegistration>
```
then you can use those code
```
return getBotRegistrationResponse.data;
```
in line 176

IvanJobs: I agree with you. It may be slightly better. 
How about **updating this along with future PRs** as I don't want all PR checks to run again because of this subtle change.

IvanJobs: I agree with you - It may be slightly better. 
How about **updating this along with future PRs** as I don't want all PR checks to run again because of this subtle change.

packages/fx-core/src/plugins/resource/bot/appStudio/appStudio.ts
```diff
@@ -152,6 +152,33 @@ export class AppStudio {
     return app;
   }
 
+  public static async getBotRegistration(
+    accessToken: string,
+    botId: string
+  ): Promise<IBotRegistration | undefined> {
+    const axiosInstance = AppStudio.newAxiosInstance(accessToken);
```
eriolchan, line 159: The axiosInstance is also created in L158, is there a better way to avoid 2 axios instances?

packages/fx-core/src/plugins/resource/bot/appStudio/appStudio.ts
```diff
@@ -200,17 +212,23 @@ export class AppStudio {
   public static async updateMessageEndpoint(
     accessToken: string,
     botId: string,
-    registration: IBotRegistration
+    endpoint: string
   ): Promise<void> {
     const axiosInstance = AppStudio.newAxiosInstance(accessToken);
 
+    const botReg = await AppStudio.getBotRegistration(accessToken, botId);
```
eriolchan, line 219: Same issue here

packages/fx-core/src/plugins/resource/bot/appStudio/appStudio.ts
```diff
@@ -200,17 +212,23 @@ export class AppStudio {
...
+    if (!botReg) {
+      throw new MessageEndpointUpdatingError(endpoint);
+    }
+    botReg.messagingEndpoint = endpoint;
+
     let response = undefined;
     try {
       response = await RetryHandler.Retry(() =>
-        axiosInstance.post(`${AppStudio.baseUrl}/api/botframework/${botId}`, registration)
+        axiosInstance.post(`${AppStudio.baseUrl}/api/botframework/${botId}`, botReg)
```
eriolchan, line 228: Does app studio have some api to only update/patch for endpoint field?

packages/fx-core/src/plugins/resource/bot/appStudio/appStudio.ts
IvanJobs: Good Catch! 
We may choose to pass `axiosInstance` into `AppStudio.getBotRegistration` to avoid this kind of duplication, but I prefer not to do it because it will break the consistency of these AppStudio api - I mean, each AppStudio API here, for example, `updateMessageEndpoint` got a unify interface `(accessToken, <necessary params>)`.
I prefer to keep it as it is although it has this kind of duplication.

packages/fx-core/src/plugins/resource/bot/appStudio/appStudio.ts
IvanJobs: As far as I know, the answer is no.
1. Updating message endpoint on Teams Dev Portal also used the same REST API as it is here.
2. It seems no other API except this one to update bot registration configs.

Will double confirm with Ning and look into AppStudio's code later on, and let's assume the answer is no for now.

eriolchan (approved):

PR #5478, opened 2022-07-08
chore(vsc): add cache to telemetry reporter

tecton:
This is to reduce the possibility of losing telemetry data. It is designed to deliver at-least-once.

- Refactor `commonlib/telemetry.ts` and `telemetry/extTelemetry.ts` to single file `telemetry/telemetry.ts`.
- Add cache layer before sending telemetry events and error events.
  - Send if the cached number reaches 10 or after 3 seconds.
- Persist unsent telemetry events to disk before deactivation and send again next time.


packages/vscode-extension/src/extension.ts
```diff
@@ -156,17 +154,13 @@ function registerActivateCommands(context: vscode.ExtensionContext) {
     "createProject"
   );
   context.subscriptions.push(
-    vscode.commands.registerCommand("fx-extension.getNewProjectPath", async (...args) => {
-      if (!isSupportAutoOpenAPI()) {
```
HuihuiWu-Microsoft, line 160: Won't we support old version vscode that doesn't have autoOpen API?

tecton: I've updated the required VS Code version in `package.json`. It should only affect a very small portion of users since VS Code is auto upgraded by default.

packages/vscode-extension/src/telemetry/telemetry.ts
```diff
@@ -0,0 +1,339 @@
...
+  }
+
+  public sendTelemetryEvent(
+    eventName: string,
+    properties?: { [p: string]: string },
+    measurements?: { [p: string]: number }
+  ): void {
+    const filledProperties = this.fillProperties(properties);
+
+    if (this.testFeatureFlag) {
```
HuihuiWu-Microsoft, line 97: Keep legacy behavior when testFeatureFlag enabled?

packages/vscode-extension/src/telemetry/telemetry.ts
```diff
@@ -0,0 +1,339 @@
...
+      configure({
+        appenders: { everything: { type: "file", filename: logFile } },
+        categories: { default: { appenders: ["everything"], level: "debug" } },
+      });
+      this.logger = getLogger("TelemTest");
+    }
+    this.cache = new TelemetryCache(this.reporter);
+  }
+
+  public addSharedProperty(name: string, value: string): void {
```
HuihuiWu-Microsoft, line 86: Will this refactor affect AB testing scorecard since exp service is still using the clientTimestamp for analysis?

packages/vscode-extension/src/telemetry/telemetry.ts
```diff
@@ -0,0 +1,339 @@
...
+/**
+ *  VSCode telemetry reporter used by fx-core and extension.
+ *  Usage:
+ *    let reporter = new VSCodeTelemetryReporter(key, extensionVersion, extensionId)
+ *  Illustrate:
+ *    key = <'the application insights key'>, 'aiKey' in package.json
+ *    extensionVersion = '<extension version>', extension version will be reported as a property with each event
+ *    extensionId = '<your extension unique name>', all events will be prefixed with this event name. eg: 'extensionId/eventname'
+ */
+export class VSCodeTelemetryReporter extends vscode.Disposable implements TelemetryReporter {
```
HuihuiWu-Microsoft, line 58: Since VSCodeTelemetryReport itself is also vscode.Disposable, should we dispose it too?

packages/vscode-extension/src/telemetry/cache.ts
```diff
@@ -0,0 +1,102 @@
...
+        };
+        switch (event.type) {
+          case "normal":
+            this.reporter.sendTelemetryEvent(event.eventName, properties, event.measurements);
+            break;
+          case "error":
+            this.reporter.sendTelemetryErrorEvent(event.eventName, properties, event.measurements);
+            break;
+        }
+        event.sendTime = new Date();
```
HuihuiWu-Microsoft, line 64: Why set event.sendTime after the event is sent?

packages/vscode-extension/src/telemetry/telemetry.ts
tecton: It is disposed in `dispose()` function which is invoked in `deactivate()`

packages/vscode-extension/src/telemetry/cache.ts
tecton: This is to determine which events should be written to disk. Current assumption is to persist all event which is not sent or sent within 10 seconds.

packages/vscode-extension/src/telemetry/telemetry.ts
tecton: As discussed, it should have little impact when grouping event time. Another impact is duplicated events which will affect both control and treatment group.

packages/vscode-extension/src/telemetry/telemetry.ts
HuihuiWu-Microsoft: VSCodeTelemetryReporter.dispose() is called in extension.deactivate() which only disposes this.reporter, right?

tecton: VSCodeTelemetryReporter.dispose() does 3 things:
1. send events in cache.
2. persist events on disk.
3. dispose `this.reporter`.

PR #5492, opened 2022-07-11
chore(core): validation for create new resource group

jayzhang:


packages/fx-core/src/core/question.ts
```diff
@@ -495,23 +497,34 @@ export const QuestionSelectResourceGroup: SingleSelectQuestion = {
...
       if (!match) {
         return getLocalizedString("core.QuestionNewResourceGroupName.validation");
       }
-
+      const maybeExist = await resourceGroupHelper.checkResourceGroupExistence(name, rmClient);
+      if (maybeExist.isErr()) {
+        return maybeExist.error.message;
+      }
+      if (maybeExist.value) {
+        return `resource group already exists: ${name}`;
```
yuqizhou77, line 517: Do we need to localize this string?

jayzhang: We can add later in a unified manner. Current code logic has no localization:
image

LongOddCode (approved):

PR #5502, opened 2022-07-12
chore(vsc): improve select folder ux

tecton:
1. Let user choose default folder or browse a specified folder when creating new project.
2. Use A/B test to rollout.
3. Localize related strings.
4. Will add default folder configuration after exp is finished.

Related WIT: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/14787020


packages/vscode-extension/src/qm/vsc_ui.ts
```diff
@@ -446,7 +448,85 @@ export class VsCodeUI implements UserInteraction {
...
+                id: "browse",
+                label: `$(folder) ${localize("teamstoolkit.qm.browse")}`,
+              },
+            ];
+
+            const onDidAccept = async () => {
+              const selectedItems = quickPick.selectedItems;
+              if (selectedItems && selectedItems.length > 0) {
+                const item = selectedItems[0];
+                ExtTelemetry.sendTelemetryEvent(TelemetryEvent.SelectFolder, {
```
HuihuiWu-Microsoft, line 487: Do we need to add this event since question model already sends events with question name and answer?

tecton: The answer is user's folder path, and we cannot query it in telemetry data. This telemetry event helps us to understand which option they choose.

packages/vscode-extension/src/qm/vsc_ui.ts
```diff
@@ -446,7 +448,85 @@ export class VsCodeUI implements UserInteraction {
   }
 
   async selectFolder(config: SelectFolderConfig): Promise<Result<SelectFolderResult, FxError>> {
-    return this.selectFileInQuickPick(config, "folder", config.default);
+    if (!TreatmentVariableValue.useFolderSelection) {
+      // control group
+      return this.selectFileInQuickPick(config, "folder", config.default);
```
HuihuiWu-Microsoft, line 453: Which folder will be shown since the default folder might not exist?

packages/vscode-extension/src/telemetry/extTelemetryEvents.ts
```diff
@@ -266,6 +269,8 @@ export enum TelemetryProperty {
   DocumentationName = "documentation-name",
   // Used with OpenNewProject
   VscWindow = "vscode-window",
+  // Used with SelectFolder
+  Option = "option",
```
HuihuiWu-Microsoft, line 273: Can we make it more specific to avoid confusion?

packages/vscode-extension/src/qm/vsc_ui.ts
tecton: It depends on the behavior of VS Code. Currently it's user's home folder.

packages/vscode-extension/src/telemetry/extTelemetryEvents.ts
tecton: No problem.

HuihuiWu-Microsoft:
The result might be different if we provide user setting for default folder. User could choose a folder where he always creates teams apps as default and select 'default' option all the time after that. Without the setting, user can only select browser and choose folder if current default one(HOME/TeamsApps) is not expected.
But I guess this is all we can do since user setting is in package.json and we can't add AB test logic for that part. Just note that the actual result could be better due to this limitiation.

jayzhang (approved):

HuihuiWu-Microsoft (approved):

PR #5509, opened 2022-07-13
chore(v3): improve migration

jayzhang:


LongOddCode (approved):

HuihuiWu-Microsoft (approved):

packages/fx-core/src/component/feature/tab.ts
```diff
@@ -138,10 +138,18 @@ const configTab: Action = {
   name: "fx.configTab",
   type: "function",
   plan: (context: ContextV3, inputs: InputsWithProjectPath) => {
-    return ok(["config 'teams-tab' in projectSettings"]);
+    const tabConfig = getComponent(context.projectSetting, ComponentNames.TeamsTab);
+    if (tabConfig) {
```
hund030, line 142: I have skip adding this action if tab has been added. So I think we don't need this condition.
https://github.com/OfficeDev/TeamsFx/blob/2ddc8f81d3b25567ea2ca1febc82807613832392/packages/fx-core/src/component/feature/tab.ts#L69

hund030 (approved):

packages/fx-core/src/component/feature/tab.ts
jayzhang: It's a double ensurence. We can just keep it.

KennethBWSong (approved):

PR #5535, opened 2022-07-15
chore(deps-checker): reduce duplicate deps checker calls

XiaofuHuang:
Detail: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/14924646

The isInstalled() / getXXXVersion are called multiple times when the deps is installed. Reduce the duplicate calls.


packages/fx-core/src/common/deps-checker/depsChecker.ts
```diff
@@ -1,18 +1,27 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 import { DepsCheckerError } from "./depsError";
-import { Result } from "@microsoft/teamsfx-api";
 
 export interface DepsChecker {
-  isInstalled(): Promise<boolean>;
+  getInstallationInfo(): Promise<DependencyStatus>;
```
swatDong, line 6: Any test case to ensure `getInstallationInfo` called only once?

meifans (dismissed):

packages/fx-core/tests/common/deps-checker/cases/dotnet.it.ts
```diff
@@ -46,15 +49,15 @@ describe("DotnetChecker E2E Test - first run", async () => {
...
-    assert.isFalse(depsInfo.isLinuxSupported, "Linux should not support .NET");
+    assert.isFalse(depsInfo.isInstalled, ".NET is not installed, but isInstalled() return true");
+    assert.isFalse(depsInfo.details.isLinuxSupported, "Linux should not support .NET");
 
-    const res = await dotnetChecker.resolve();
-    assert.isTrue(res.isOk() && res.value);
+    const spyChecker = sandbox.spy(dotnetChecker);
+    const res = await spyChecker.resolve();
+    assert.isTrue(res.isInstalled);
+    assert.isTrue(spyChecker.getInstallationInfo.calledTwice);
```
swatDong, line 60: Why called twice? Is this fix to reduce to only one call?

XiaofuHuang: If it is the first time installation, will check whether the installation is success.

swatDong: make sense

swatDong (approved):

PR #5566, opened 2022-07-19
fix: codelens with v3 key and state

nliu-ms:
fix: 12615502

The state has been converted to v3, but the key is not. So cannot find corresponding value.

image


nliu-ms:
E2E TEST: n/a

HuihuiWu-Microsoft (approved):

jayzhang (approved):

PR #5637, opened 2022-07-27
feat: enable switch tenant for local debug

yuqizhou77:
- Allow to switch m365 tenant when local debugging. 
- Updated vscode output when we found users switch m365 tenant. (words may be updated later.)

E2E TEST: https://github.com/OfficeDev/TeamsFx/actions/runs/2751657861

### DIFF
--- a/packages/fx-core/src/plugins/solution/fx-solution/v2/utils.ts
+++ b/packages/fx-core/src/plugins/solution/fx-solution/v2/utils.ts
@@ -15,6 +15,8 @@ import {
   UserError,
   SystemError,
   M365TokenProvider,
+  v3,
+  EnvInfo,
 } from "@microsoft/teamsfx-api";
 import fs from "fs-extra";
 import { LocalSettingsTeamsAppKeys } from "../../../../common/localSettingsConstants";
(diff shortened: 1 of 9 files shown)


packages/fx-core/src/plugins/solution/fx-solution/v2/utils.ts
```diff
@@ -219,25 +223,34 @@ export async function checkWhetherLocalDebugM365TenantMatches(
...
       if (
         projectPath !== undefined &&
         (await fs.pathExists(`${projectPath}/bot/.notification.localstore.json`))
       ) {
-        localFiles.push("bot/.notification.localstore.json");
+        const errorMessage = getLocalizedString(
+          "core.localDebug.tenantConfirmNotice",
+          localDebugTenantId,
+          maybeM365UserAccount.value,
+          "bot/.notification.localstore.json"
```
dooriya, line 234: Here should check the local state file `.fx/states/state.local.json` rather than the local storage file for notification app

packages/fx-core/src/plugins/solution/fx-solution/v2/utils.ts
```diff
@@ -219,25 +223,34 @@ export async function checkWhetherLocalDebugM365TenantMatches(
     }
 
     if (maybeM365TenantId.value !== localDebugTenantId) {
-      const localFiles = [".fx/states/state.local.json"];
```
dooriya, line 222: Need to add back the original logic here

packages/fx-core/src/plugins/solution/fx-solution/v2/utils.ts
yuqizhou77: we will clear local envInfo here, so we won't fail here but just proceed.

jayzhang (approved):

dooriya (approved):

PR #5643, opened 2022-07-28
chore: add telemetry for code templates

IvanJobs:
1. Add central telemetry for code templates scaffolding.
2. Add telemetry property to detect fallback cases.

https://dev.azure.com/msazure/Microsoft%20Teams%20Extensibility/_workitems/edit/15066484

### DIFF
--- a/packages/fx-core/src/common/telemetry.ts
+++ b/packages/fx-core/src/common/telemetry.ts
@@ -29,6 +29,10 @@ export enum TelemetryProperty {
   IsSideloadingAllowed = "is-sideloading-allowed",
   NeedMigrateAadManifest = "need-migrate-aad-manifest",
   CheckSideloadingHttpStatus = "check-sideloading-http-status",
+  TemplateGroup = "template-group",
+  TemplateLanguage = "template-language",
+  TemplateScenario = "template-scenario",
+  TemplateFallback = "template-fallback",
 }
 
 export enum TelemetryEvent {
@@ -82,6 +86,8 @@ export enum TelemetryEvent {
   DetectPortStart = "detect-port-start",
   DetectPort = "detect-port",
   FillProjectId = "fill-project-id",
+  ScaffoldFromTemplatesStart = "scaffold-from-templates-start",
+  ScaffoldFromTemplates = "scaffold-from-templates",
 }
 
 export enum TelemetrySuccess {
--- a/packages/fx-core/src/common/template-utils/templatesActions.ts
+++ b/packages/fx-core/src/common/template-utils/templatesActions.ts
@@ -3,6 +3,7 @@ import fs from "fs-extra";
 import path from "path";
 import { getTemplatesFolder } from "../../folder";
 import { FeatureFlagName } from "../constants";
+import { Component, sendTelemetryEvent, TelemetryEvent, TelemetryProperty } from "../telemetry";
 import { fetchTemplateUrl, fetchZipFromUrl, renderTemplateContent, unzip } from "./templatesUtils";
 
 // The entire progress has following actions:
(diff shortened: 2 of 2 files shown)


packages/fx-core/src/common/template-utils/templatesActions.ts
```diff
@@ -215,6 +216,24 @@ export async function scaffoldFromTemplates(
...
+
+  if (!context.lang) {
+    throw new Error(missKeyErrorInfo("lang"));
+  }
+
+  if (!context.scenario) {
+    throw new Error(missKeyErrorInfo("scenario"));
+  }
+  // To track code templates usage.
+  sendTelemetryEvent(Component.core, TelemetryEvent.ScaffoldFromTemplatesStart, {
```
hund030, line 231: Do we really need start event? Maybe we can discuss offline.

packages/fx-core/src/common/template-utils/templatesActions.ts
```diff
@@ -215,6 +216,24 @@ export async function scaffoldFromTemplates(
   context: ScaffoldContext,
   actions: ScaffoldAction[] = defaultActionSeq
 ): Promise<void> {
+  if (!context.group) {
+    throw new Error(missKeyErrorInfo("group"));
```
hund030, line 220: This error could not be handled by context.onActionError, which may break the design of template utils.

packages/fx-core/src/common/template-utils/templatesActions.ts
```diff
@@ -227,4 +246,11 @@ export async function scaffoldFromTemplates(
       await context.onActionError(action, context, e);
```
hund030, line 246: Send error event here?

packages/fx-core/src/common/template-utils/templatesActions.ts
IvanJobs: `This error could not be handled by context.onActionError`  - An assumption of this statement is: this error should be handled by `context.onActionError`, but it may not be true because `context.onActionError` is to handle errors throwed by Actions and here the errors are out of the Actions execution loop.

One thing I'm not so sure is whether we need to precheck these three fileds here, maybe undefined is acceptable in the very beginning of this `scaffoldFromTemplates` function.

packages/fx-core/src/common/template-utils/templatesActions.ts
IvanJobs: In short, I think whether we need to sendErrorTelemetry here depends on what kind of BI insights do we want to get from it. 
1. It's kind of tricky because of the design - I mean, if there's some error throwed here, `context.onActionError` may not treat it as an Error (Fallback logic). 
2. I choose not to `sendErrorTelemetry` here - Because of the fallback logic, errors are in low possibility, and we have Error Distribution + Error Details Table for each plugin.

packages/fx-core/src/common/template-utils/templatesActions.ts
IvanJobs: I used start event to track all scaffold intentions in the very beginning of scaffold. You know, these intentions may not be all successful, so I got another telemetry in the end with fallback property.

packages/fx-core/src/common/telemetry.ts
```diff
@@ -29,6 +29,10 @@ export enum TelemetryProperty {
   IsSideloadingAllowed = "is-sideloading-allowed",
   NeedMigrateAadManifest = "need-migrate-aad-manifest",
   CheckSideloadingHttpStatus = "check-sideloading-http-status",
+  TemplateGroup = "template-group",
+  TemplateLanguage = "template-language",
+  TemplateScenario = "template-scenario",
+  TemplateFallback = "template-fallback",
```
eriolchan, line 35: In which case, we need template-fallback event?

packages/fx-core/src/common/template-utils/templatesActions.ts
```diff
@@ -156,6 +160,8 @@ export const fetchTemplateZipFromLocalAction: ScaffoldAction = {
       return;
     }
 
+    context.fallback = true;
```
eriolchan, line 163: If it is fallback to local template, maybe this action is already tracked in the error handling of download from remote?

packages/fx-core/src/common/telemetry.ts
IvanJobs: When scaffolding falls back to use local templates packed inside VS Code Extension.

packages/fx-core/src/common/template-utils/templatesActions.ts
IvanJobs: Seems not - errors will be muted (do nothing) in error handling of "download from remote", see code snippets below:
image

packages/fx-core/src/common/template-utils/templatesActions.ts
```diff
@@ -227,4 +240,11 @@ export async function scaffoldFromTemplates(
       await context.onActionError(action, context, e);
     }
   }
+
+  sendTelemetryEvent(Component.core, TelemetryEvent.ScaffoldFromTemplates, {
```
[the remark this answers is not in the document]
IvanJobs: It's not duplicated but serves for different purposes.
1. In the beginning of `scaffoldFromTemplates`, we can get all scaffolding intentions.
2. In the end of it, we can get all successful scaffolding with fallback tags.

hund030 (approved):

PR #5653, opened 2022-07-29
chore: update cli local debug message

yuqizhou77:
We will add log if we have noticed that user has switched m365 tenant in cil.

image

todo: add the doc link after updating the doc.


packages/cli/src/cmds/preview/preview.ts
```diff
@@ -1486,6 +1503,22 @@ export default class Preview extends YargsCommand {
...
+    if (tenantId) {
+      const localEnv = await localEnvManager.getLocalEnvInfo(projectPath, {
+        projectId: projectSettings.projectId,
+      });
+
+      if (
+        localEnv &&
+        localEnv.state["solution"] &&
+        localEnv.state["solution"]["teamsAppTenantId"] &&
+        localEnv.state["solution"]["teamsAppTenantId"] != tenantId
```
swatDong, line 1516: `!==`

yuqizhou77: resolved.

swatDong (approved):

PR #5660, opened 2022-08-01
perf: support add sso in core and server

KennethBWSong:
1. Allow language == csharp in add sso checks
2. Add api in server to support add sso command in VS

### DIFF
--- a/packages/fx-core/src/plugins/solution/fx-solution/v2/executeUserTask.ts
+++ b/packages/fx-core/src/plugins/solution/fx-solution/v2/executeUserTask.ts
@@ -1209,6 +1209,10 @@ export function validateAndParseLanguage(language: string): Result<string, FxErr
     return ok("js");
   }
 
+  if (language.toLowerCase() == Language.JavaScript) {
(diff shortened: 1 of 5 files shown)


packages/fx-core/src/plugins/solution/fx-solution/v2/executeUserTask.ts
```diff
@@ -1209,6 +1209,10 @@ export function validateAndParseLanguage(language: string): Result<string, FxErr
     return ok("js");
   }
 
+  if (language.toLowerCase() == Language.JavaScript) {
```
Alive-Fish, line 1212: Will this be `Language.CSharp`?

KennethBWSong: Sure, fixed, thx!

Alive-Fish (approved):
LGTM

1yefuwang1 (approved):

SLdragon (approved):

PR #5669, opened 2022-08-01
chore(vsc): add experiment for opening folder in new window

tecton:
Add A/B test experiment to open folder in new VS Code window after creation.
It helps to reduce the percentage of lost telemetry events caused by VS Code reload.

Effect:
image


packages/vscode-extension/src/controls/webviewPanel.ts
```diff
@@ -173,12 +173,7 @@ export class WebviewPanel {
     if (res.isOk()) {
       props[TelemetryProperty.Success] = TelemetrySuccess.Yes;
       ExtTelemetry.sendTelemetryEvent(TelemetryEvent.DownloadSample, props);
-      await globalStateUpdate(GlobalKey.OpenSampleReadMe, true);
```
HuihuiWu-Microsoft, line 176: Will GlobalKey.OpenSampleReadMe updated after this change?

tecton: Yes, it's updated in handlers.ts:L518

HuihuiWu-Microsoft:
Do we also consider reloadWindow after migration(Although it should be smaller number compared with create project)?

tecton:
> Do we also consider reloadWindow after migration(Although it should be smaller number compared with create project)?

No, we don't need to change it because:
1. Only 2 migration events happened in 4.x and telemetry data were not lost.
2. It makes sense to reload in current window because it's the same project.

HuihuiWu-Microsoft (dismissed):

jayzhang (approved):

HuihuiWu-Microsoft (approved):

PR #5718, opened 2022-08-05
chore(componentize): update sql bicep

xzf0587:


jayzhang (approved):

hund030:
Maybe also update the output value in constants?
```
 sqlDatabaseName: {
    key: "sqlDatabaseName",
    bicepVariable: "provisionOutputs.azureSqlOutput.value.sqlDatabaseName",
  },
```

PR #5732, opened 2022-08-08
chore: send emails when Team JS new version released

yiqing-zhao:


.github/workflows/teamsjs-released.yml
```diff
@@ -4,10 +4,58 @@ on:
     types: [teamsjs-released]
 jobs:
   notification:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-latest
```
tecton, line 7: minor: you could remove these redundant spaces

yiqing-zhao: Spaces are removed. Thx!

tecton (approved):

PR #5734, opened 2022-08-08
chore(m365): add back msteams-react-base-component to tab

kuojianlu:
Fix https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/15115429.


templates/tab/js/m365/package.json
```diff
@@ -3,14 +3,15 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "@fluentui/react-northstar": "^0.60.1",
+    "@fluentui/react-northstar": "^0.62.0",
```
swatDong, line 6: What's this change for?

kuojianlu: msteams-react-base-component depends on `msteams-react-base-component@^0.62.0`.

swatDong (approved):

PR #5737, opened 2022-08-08
chore: update subscription npm version

xiaolang124:
1.Update the subscription version to the newest.

This is a part of removing cli deprecated work.
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/15066253/

In the future pr, we will remove Azure `getAccountCredentialAsync` method.


xiaolang124:
e2e test: https://github.com/OfficeDev/TeamsFx/actions/runs/2815229319

Failed cases are same with daily pipeline.

packages/cli/package.json
```diff
@@ -42,7 +42,7 @@
   },
   "aiKey": "1c56be97-bb74-42cf-b04b-8f1aabf04592",
   "devDependencies": {
-    "@azure/arm-subscriptions": "^3.0.0",
+    "@azure/arm-subscriptions": "^5.0.0",
```
swatDong, line 45: Remove this since it's already in `dependencies`?

xiaolang124: Done.

packages/fx-core/src/plugins/solution/fx-solution/utils/util.ts
```diff
@@ -17,6 +17,16 @@ import {
 import { SubscriptionClient } from "@azure/arm-subscriptions";
 import { SolutionTelemetryComponentName, SolutionTelemetryProperty } from "../constants";
 import { TokenCredential } from "@azure/core-auth";
+import { BuiltInFeaturePluginNames } from "../v3/constants";
```
swatDong, line 20: What are these import changes for?

swatDong (approved):

PR #5740, opened 2022-08-08
chore(bot): print detailed App studio error msg from response

yukun-dong:
before:
3ab4719ef4e49930a269235a9753c99
after:
image

### DIFF
--- a/packages/fx-core/src/plugins/resource/bot/errors.ts
+++ b/packages/fx-core/src/plugins/resource/bot/errors.ts
@@ -298,35 +298,35 @@ export function wrapError(
   sendTelemetry: boolean,
   name: string
 ): FxResult {
-  let detailedErrorMsg = isErrorWithMessage(e) ? e.message : "";
+  let errorMsg = isErrorWithMessage(e) ? e.message : "";
   const innerError = isPluginError(e) ? e.innerError : undefined;
   if (innerError) {
-    detailedErrorMsg += getLocalizedString(
+    errorMsg += getLocalizedString(
       "plugins.bot.DetailedError",
       isErrorWithMessage(innerError) ? innerError.message : ""
     );
     if (isHttpError(innerError)) {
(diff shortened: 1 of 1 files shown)


packages/fx-core/src/plugins/resource/bot/errors.ts
```diff
@@ -104,18 +105,33 @@ export class PluginError extends Error {
...
+    msg += this.getAppStudioErrorDetails(this.innerError);
     if (this.suggestions.length > 0) {
       msg += getDefaultString("plugins.bot.ErrorSuggestions", this.suggestions.join(" "));
     }
     return msg;
   }
+
   genDisplayMessage(): string {
     let msg = `${this.details[1]} `;
+    msg += this.getAppStudioErrorDetails(this.innerError);
```
eriolchan, line 117: Why don't we also output inner error besides AppStudio Error?

eriolchan: We can have a better structure of the error message. E.g.
message: ..., suggestion: ..., innerError: ...

yukun-dong: The `getAppStudioErrorDetails` actually returns the inner error message, I will change it to a better name.

packages/fx-core/src/plugins/resource/bot/errors.ts
```diff
@@ -297,42 +298,50 @@ export function wrapError(
   sendTelemetry: boolean,
   name: string
 ): FxResult {
-  let errorMsg = isErrorWithMessage(e) ? e.message : "";
+  let detailedErrorMsg = isErrorWithMessage(e) ? e.message : "";
```
eriolchan, line 301: Looking through the code, it looks we can still use errorMsg variable.

eriolchan (approved):

IvanJobs (approved):
LGTM, Great Job!

PR #5746, opened 2022-08-09
chore: add telemetry for sample gallery

HuihuiWu-Microsoft:
Add following telemetry events:
1> When user click sample app card with app id attached
2> When user click 'View in GitHub' button in sample detail page with app id attached

### DIFF
--- a/packages/vscode-extension/src/controls/sampleDetailPage.tsx
+++ b/packages/vscode-extension/src/controls/sampleDetailPage.tsx
@@ -4,6 +4,11 @@ import "./sampleDetailPage.scss";
 import { VSCodeButton, VSCodeTag } from "@vscode/webview-ui-toolkit/react";
 import { Watch, Setting } from "./resources";
 import { Commands } from "./Commands";
+import {
+  TelemetryEvent,
+  TelemetryProperty,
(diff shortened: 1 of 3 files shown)


chagong (approved):

packages/vscode-extension/src/controls/sampleDetailPage.tsx
```diff
@@ -66,6 +71,16 @@ export default class SampleDetailPage extends React.Component<SampleDetailProps,
   };
 
   onViewGithub = () => {
+    vscode.postMessage({
```
tecton, line 74: Could we send telemetry in OpenExternalLink handler?

HuihuiWu-Microsoft: I prefer to send it in business code where util API is called. The openUrl API is just an util function without any context. Many of the telemetry events are already sent in places where the API is called like in openDocumentHandler(), 'TelemetryEvent.Documentation' telemetry event is sent with property TelemetryProperty.DocumentationName: "general" attached. So I suppose that's better practice.

tecton (approved):

PR #5760, opened 2022-08-10
chore(v3): fix and improve debug

jayzhang:
1. unify project settings loader entry
2. create local env when init a project
3. unify setupLocalEnvironment, confLocalEnvironment, scaffoldLocalDebugSettings

### DIFF
--- a/packages/fx-core/tests/common/local/localEnvManager.test.ts
+++ b/packages/fx-core/tests/common/local/localEnvManager.test.ts
@@ -75,6 +75,7 @@ describe("LocalEnvManager", () => {
 
       chai.assert.isDefined(projectSettings);
       chai.assert.isUndefined(projectSettings.appName);
+      chai.assert.isDefined(projectSettings.projectId);
     });
 
     it("missing file", async () => {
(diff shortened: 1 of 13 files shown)


packages/fx-core/tests/common/local/localEnvManager.test.ts
```diff
@@ -75,7 +75,6 @@ describe("LocalEnvManager", () => {
 
       chai.assert.isDefined(projectSettings);
       chai.assert.isUndefined(projectSettings.appName);
-      chai.assert.isUndefined(projectSettings.projectId);
```
swatDong, line 78: Why removing this?
If projectId has default value, validate the expected value instead of removing the assert.

swatDong (commented):
The local debug changes look good to me. Does this conflict with #5757 ?

packages/fx-core/tests/common/local/localEnvManager.test.ts
swatDong: So better to just put `chai.assert.isDefined(projectSettings.projectId);` here.

jayzhang: I will change this to isDefined

jayzhang:
> The local debug changes look good to me. Does this conflict with #5757 ?
this PR is closed

hund030 (approved):

swatDong (approved):

PR #5786, opened 2022-08-12
perf: update aad,bot plugin for sso in vs

KennethBWSong:


packages/fx-core/src/plugins/resource/aad/plugin.ts
```diff
@@ -569,7 +573,9 @@ export class AadAppForTeamsImpl {
     }
 
     if (botEndpoint) {
-      redirectUris.web?.redirectUris?.push(`${botEndpoint}/auth-end.html`);
+      redirectUris.web?.redirectUris?.push(
+        isVs ? `${botEndpoint}/bot-auth-end.html` : `${botEndpoint}/auth-end.html`
```
SLdragon, line 577: Why use different redirect url in vscode and vs?

1yefuwang1 (approved):

packages/fx-core/src/plugins/resource/aad/plugin.ts
KennethBWSong: According to Di's investigation, we need to separate tab and bot's auth pages in VS since we may host backend in the same webapp.

packages/fx-core/src/plugins/resource/bot/dotnet/plugin.ts
```diff
@@ -59,13 +59,41 @@ export class DotnetBotImpl extends TeamsBotImpl {
...
+      );
       if (await fs.pathExists(appSettingsPath)) {
         appSettings = await fs.readFile(appSettingsPath, "utf-8");
       } else {
         // if appsetting file not exist, generate a new one
         // TODO(qidon): load content from resource file or template
-        appSettings =
-          '\
+        appSettings = includeAad
+          ? '\
```
Siglud, line 71: maybe we should put these templates into a file?

packages/fx-core/src/plugins/resource/bot/dotnet/constants.ts
```diff
@@ -17,4 +17,10 @@ export class PathInfo {
 export class RegularExpr {
   static readonly botId = /\$botId\$/g;
   static readonly botPassword = /\$bot-password\$/g;
+  static readonly clientId = /\$clientId\$/g;
```
Siglud, line 20: There is no need to use regularExp here, I think simple string replacement is work fine and it's faster too

packages/fx-core/src/plugins/resource/bot/dotnet/plugin.ts
eriolchan: Agree, hard code the file content in code is not easy for maintenance, please change it with a better way.

packages/fx-core/src/plugins/resource/bot/dotnet/plugin.ts
```diff
@@ -92,6 +120,44 @@ export class DotnetBotImpl extends TeamsBotImpl {
       if (botPassword) {
         appSettings = appSettings.replace(RegularExpr.botPassword, botPassword);
       }
+      if (includeAad) {
```
eriolchan, line 123: If it is SSO related logic, can it be encapsulated as a separate library and ask Bot plugin to call it instead of putting the logic in Bot plugin? This will help to be more componentized for V3.

packages/fx-core/src/plugins/resource/bot/dotnet/plugin.ts
KennethBWSong: Sure, updated. @Siglud you may also need to update the appSettings without SSO part?

packages/fx-core/src/plugins/resource/bot/dotnet/constants.ts
KennethBWSong: Updated, thx!

packages/fx-core/src/plugins/resource/bot/dotnet/plugin.ts
KennethBWSong: @eriolchan That's a good point. Since we may have multiple components in V3 try to modify app settings file in parallel (just like bot and sso here), we expect core to support read and write for app settings file just like project settings. This part is still under discussion, and will update the logic here later if we have any updates.

Siglud (approved):

packages/fx-core/templates/plugins/resource/aad/auth/bot/csharp/SSO/TeamsSsoBot.cs
```diff
@@ -0,0 +1,57 @@
...
+        _dialog = dialog;
+        _logger = logger;
+    }
+
+    protected override async Task OnMessageActivityAsync(ITurnContext<IMessageActivity> turnContext, CancellationToken cancellationToken)
+    {
+        _logger.LogInformation("Receive message activity");
+        var text = turnContext.Activity.Text;
+        var removedMentionText = turnContext.Activity.RemoveRecipientMention();
+        if (!string.IsNullOrEmpty(removedMentionText))
```
SLdragon, line 25: Did you mean `if (!string.IsNullOrEmpty(text))`

packages/fx-core/templates/plugins/resource/aad/auth/bot/csharp/SSO/MainDialog.cs
```diff
@@ -0,0 +1,83 @@
...
+            // for more complex scenarios
+            var authProvider = new DelegateAuthenticationProvider((request) =>
+            {
+                request.Headers.Authorization =
+                    new System.Net.Http.Headers.AuthenticationHeaderValue("Bearer", tokenResponse.Token);
+                return Task.CompletedTask;
+            });
+            var graphClient = new GraphServiceClient(authProvider);
+            var profile = await graphClient.Me.Request().GetAsync();
+            await stepContext.Context.SendActivityAsync($"You're logged in as {profile.DisplayName} ({profile.UserPrincipalName}); you job title is: {profile.JobTitle}");
```
SLdragon, line 74: JobTitle maybe empty

packages/fx-core/templates/plugins/resource/aad/auth/bot/csharp/SSO/TeamsSsoBot.cs
KennethBWSong: Updated, thx!

packages/fx-core/templates/plugins/resource/aad/auth/bot/csharp/SSO/MainDialog.cs
KennethBWSong: Sure, i think it is OK if we just include user's display name so i removed the rest part.

SLdragon (approved):

Siglud (approved):

jayzhang (approved):

PR #5787, opened 2022-08-12
chore: remove BotAuthenticationOption from AddTeamsFx() method

blackchoey:
Temporary remove BotAuthenticationOption from AddTeamsFx() method.
This method requires RazorPages service so we may need further discussion on whether we should let users call AddTeamsFx() for every projects that uses TeamsFx SDK. The major concern is user might not use RazorPages and cause error during initialization.
We can add BotAuthenticationOption back if that's OK after discussion.


JerryYangKai (approved):

tecton:
I found that `AddServerSideBlazor()` is responsible for injecting `IJSRuntime` because it's only available in Blazor environment.
That is also mentioned in this [SO question](https://stackoverflow.com/questions/57054119/ijs-interop-not-available-in-asp-net-3-razor-pages).

tecton (approved):

PR #5806, opened 2022-08-16
perf: update sample code for bot sso dedup

KennethBWSong:


SLdragon (approved):

packages/fx-core/templates/plugins/resource/aad/auth/bot/csharp/SSO/SsoDialog.cs
```diff
@@ -16,11 +13,18 @@ public class DialogConstants
...
+{
+    public string eTag;
+}
+
 public class SsoDialog : ComponentDialog
 {
     ILogger<SsoDialog> _logger;
     BotAuthenticationOptions _botAuthOptions;
     Dictionary<string, string> _commandMapping = new Dictionary<string, string>();
+    IStorage _dedupStorage = new MemoryStorage();
```
blackchoey, line 26: is it possible to use dependency injection to get an IStorage instance? I saw L49 in readme.md already added MemotryStorage, not sure whether we can reuse it via DI.

blackchoey: We also need to tell users switch to another storage for production purpose

PR #5821, opened 2022-08-17
chore: update word

yuqizhou77:
- update wording when switching M365 tenant.


swatDong (approved):

HuihuiWu-Microsoft (approved):

packages/fx-core/src/plugins/solution/fx-solution/v2/provision.ts
```diff
@@ -537,7 +537,7 @@ export async function askForProvisionConsentNew(
     );
     if (confirm !== provisionText) {
       if (confirm === learnMoreText) {
-        ctx.userInteraction.openUrl("https://aka.ms/teamsfx-switch-account-or-subscription-help");
+        ctx.userInteraction.openUrl("https://aka.ms/teamsfx-switch-tenant-or-subscription-help");
```
nliu-ms, line 540: This does doesn't exist?

nliu-ms (approved):

PR #5826, opened 2022-08-17
chore: prepare sideloading command in CLI

swatDong:
This is preparation work for https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/15030845/

@wenytang-ms , I added script to be called by workflow. The step is required for dev release as well.
@Alive-Fish , I added new command to CLI, but is hidden from help since it's in preview.


.github/workflows/cd.yml
```diff
@@ -238,6 +238,17 @@ jobs:
...
+      - name: replace sideloading placeholders
+        run: bash .github/scripts/sideloading-replace.sh
+        env:
+          SIDELOADING_SERVICE_ENDPOINT: ${{ secrets.SIDELOADING_SERVICE_ENDPOINT }}
+          SIDELOADING_SERVICE_SCOPE: ${{ secrets.SIDELOADING_SERVICE_SCOPE }}
+
+      - name: commit change on local
+        run: |
+          git add .
+          git commit -m "chore: replace sideloading placeholders"
```
wenytang-ms, line 250: it's better to commit with path, ref commit with api key change.

wenytang-ms, line 250: it's better to commit in a specific path, such as packages/cli/src/..
Doing this ensures that other changes are not committed at the same time.

swatDong: good point, updated.

wenytang-ms (approved):
Approve with CD part.

Alive-Fish (approved):

xiaolang124 (approved):

PR #5847, opened 2022-08-19
feat(sdk): add card action handler in conversation SDK

dooriya:
With TeamsFx SDK support, developer can handle the card actions more easily without diving into the Bot Framework SDK.

Check here for the how to build a sample workflow bot with the SDK: 
- Sample Code: https://github.com/dooriya/WorkflowBot/tree/main
- How to use the SDK: https://github.com/dooriya/WorkflowBot

E2E TEST: N/A

TODO:
- Add scaffold template for workflow bot (a sample command with one action)
- Enable scaffold workflow bot in VSC/CLI


packages/sdk/src/conversation/interface.ts
```diff
@@ -173,6 +173,78 @@ export interface CommandOptions {
...
+}
+
+/**
+ * Interface for adaptive card action handler that can process card action invoke and return a response.
+ */
+export interface TeamsFxAdaptiveCardActionHandler {
+  /**
+   * The verb defined in adaptive card action that can trigger this handler.
+   */
+  triggerVerb: string;
```
swatDong, line 213: Is the verb case sensitive/insensitive? Better to add to doc.

packages/sdk/src/conversation/middlewares/cardActionMiddleware.ts
```diff
@@ -0,0 +1,99 @@
...
+    }
+  }
+
+  async onTurn(context: TurnContext, next: () => Promise<void>): Promise<void> {
+    if (context.activity.name === "adaptiveCard/action") {
+      const action = context.activity.value.action;
+      const actionVerb = action.verb;
+
+      for (const handler of this.actionHandlers) {
+        if (handler.triggerVerb === actionVerb) {
```
swatDong, line 32: What if two handlers define the same verb?

packages/sdk/src/conversation/interface.ts
dooriya: We'd better make it case-insensitive. Let me update it.

packages/sdk/src/conversation/middlewares/cardActionMiddleware.ts
```diff
@@ -0,0 +1,99 @@
...
+  constructor(handlers?: TeamsFxAdaptiveCardActionHandler[]) {
+    if (handlers && handlers.length > 0) {
+      this.actionHandlers.push(...handlers);
+    }
+  }
+
+  async onTurn(context: TurnContext, next: () => Promise<void>): Promise<void> {
+    if (context.activity.name === "adaptiveCard/action") {
+      const action = context.activity.value.action;
+      const actionVerb = action.verb;
```
swatDong, line 29: Is there any scenario that two cards define the same verb? E.g., a PR review card defines `approve` verb and a release pipeline card also defines `approve` verb?
Or, it's recommended to define unique verbs such as `approve-pr` and `approve-release`?

packages/sdk/src/conversation/middlewares/commandMiddleware.ts
```diff
@@ -0,0 +1,91 @@
...
+
+      const message: CommandMessage = {
+        text: commandText,
+      };
+
+      for (const handler of this.commandHandlers) {
+        const matchResult = this.shouldTrigger(handler.triggerPatterns, commandText);
+
+        // It is important to note that the command bot will stop processing handlers
+        // when the first command handler is matched.
```
swatDong, line 32: Although this is the old code but just confused... how do we ensure only one handler is triggered? There's no `break` in the for-each iteration.

dooriya: Nice catch! we should add `break` here to stop processing if one handler is matched.

packages/sdk/src/conversation/middlewares/cardActionMiddleware.ts
dooriya: For each adaptive card invoke action, the bot should send an [InvokeResponse](https://docs.microsoft.com/en-us/adaptive-cards/authoring-cards/universal-action-model#response-format) to respond to the invoke request. If multiple invoke responses were sent, only the first invoke response will be taken in the underlying bot framework.

So even we don't have a `break` here, only the invoke response sent by the first matched handler will be taken, while the response sent from other handlers will be ignored in this case. 

But since this is hidden logic from Bot Framework, I will add `break` here to make it more clear and stop processing other handlers although it will be ignored if matched handler is invoked.

packages/sdk/src/conversation/invokeResponseFactory.ts
```diff
@@ -0,0 +1,130 @@
...
+  public static textMessage(message: string): InvokeResponse {
+    if (!message) {
+      throw new Error("The text message cannot be null or empty");
+    }
+
+    return {
+      status: StatusCodes.OK,
+      body: {
+        statusCode: StatusCodes.OK,
+        type: "application/vnd.microsoft.activity.message",
```
qinezh, line 52: Extract these values into const variable since they're also used in middleware?

packages/sdk/src/conversation/middlewares/cardActionMiddleware.ts
dooriya: The latter will be recommended in this case :)

packages/sdk/src/conversation/invokeResponseFactory.ts
```diff
@@ -0,0 +1,130 @@
...
+   * the request was failed to processed.
+   *
+   * @param errorCode The status code indicates error, available values:
+   *  - 400 (BadRequest): indicate the incoming request was invalid.
+   *  - 500 (InternalServerError): indicate an unexpected error occurred.
+   * @param errorMessage The error message.
+   *
+   * @returns {InvokeResponse} An InvokeResponse object.
+   */
+  public static errorResponse(errorCode: StatusCodes, errorMessage: string): InvokeResponse {
```
qinezh, line 97: > errorCode: StatusCodes

Since only BAD_REQUEST and INTERNAL_SERVER_ERROR are supported, how about using a customized type like `InvokeResponseErrorCode` which only supports these two status code? Compile error would be better than the runtime error in this case.

packages/sdk/src/conversation/middlewares/cardActionMiddleware.ts
```diff
@@ -0,0 +1,99 @@
...
+          switch (responseType) {
+            case "application/vnd.microsoft.activity.message":
+              await this.sendInvokeResponse(context, response);
+              break;
+            case "application/vnd.microsoft.card.adaptive":
+              const card = response.body?.value;
+              if (!card) {
+                await this.sendInvokeResponse(
+                  context,
+                  InvokeResponseFactory.textMessage(this.defaultMessage)
```
qinezh, line 55: Should an error response be returned here when the error 'Adaptive card content cannot be found in the response body' occurs? Otherwise, the end user might think the action is executed successfully however it's not...

dooriya: Sounds good~

packages/sdk/src/conversation/invokeResponseFactory.ts
dooriya: Sounds good~

packages/sdk/src/conversation/middlewares/cardActionMiddleware.ts
dooriya: The invoke response here should be an error response rather than a response with default message. Will fix it.

swatDong (dismissed):

packages/sdk/src/conversation/middlewares/cardActionMiddleware.ts
```diff
@@ -29,13 +32,13 @@ export class CardActionMiddleware implements Middleware {
       const actionVerb = action.verb;
 
       for (const handler of this.actionHandlers) {
-        if (handler.triggerVerb === actionVerb) {
+        if (handler.triggerVerb.toLocaleLowerCase() === actionVerb.toLocaleLowerCase()) {
```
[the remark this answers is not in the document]
dooriya: fixed

tecton (approved):

qinezh (dismissed):

a1exwang (dismissed):

tecton (approved):

a1exwang (approved):

tecton (dismissed):

swatDong (dismissed):

qinezh (approved):

swatDong (approved):

tecton (approved):

qinezh (approved):

a1exwang (approved):

PR #5870, opened 2022-08-23
feat(vsc): open created project in new window by default

tecton:
Remove A/B test code and ship the feature to open in new window by default.
E2E TEST: N/A

Related PR: #5669


packages/vscode-extension/src/handlers.ts
```diff
@@ -460,51 +460,10 @@ export async function openFolder(
...
-      }
-    } else {
-      // timeout
-      ExtTelemetry.sendTelemetryEvent(TelemetryEvent.OpenNewProject, {
-        [TelemetryProperty.VscWindow]: VSCodeWindowChoice.NewWindowByDefault,
-      });
-      commands.executeCommand("vscode.openFolder", folderPath, true);
-    }
-  }
+  ExtTelemetry.sendTelemetryEvent(TelemetryEvent.OpenNewProject, {
```
HuihuiWu-Microsoft, line 463: Do we still need the 'TelemetryProperty.VscWindow' property while other entries are already removed?

tecton: Yes, I'm keeping this one in case we add other entries in the future.

HuihuiWu-Microsoft: Please be aware that event with this property doesn't mean the same for versions before&after this change when querying data.

HuihuiWu-Microsoft (approved):

PR #5878, opened 2022-08-23
feat(vsc): add preview command in treeview to run F5 directly in A/B test

tecton:
E2E TEST: N/A

Related feature: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/15247725


packages/vscode-extension/src/treeview/treeViewManager.ts
```diff
@@ -64,6 +65,24 @@ class TreeViewManager {
       );
       developmentTreeviewProvider.refresh();
     }
+    if (TreatmentVariableValue.previewTreeViewCommand) {
```
HuihuiWu-Microsoft, line 68: Does this command show for both new app and existing app?

HuihuiWu-Microsoft (approved):

packages/vscode-extension/src/treeview/treeViewManager.ts
tecton: Fixed comment.

PR #5879, opened 2022-08-23
chore(apiconnection): update sample file path

xzf0587:


packages/fx-core/src/plugins/resource/apiconnector/sampleHandler.ts
```diff
@@ -51,22 +52,19 @@ export class SampleHandler {
...
-      const codeFileName: string = config.APIName + "." + fileSuffix;
-      let codePath = path.join(this.projectRoot, this.component);
-      if (await fs.pathExists(path.join(codePath, "src"))) {
-        codePath = path.join(codePath, "src");
+      const sampleFileName: string = config.APIName + "." + fileSuffix;
+      const componentPath = path.join(this.projectRoot, this.component);
+      const sampleDirPath = getSampleDirPath(componentPath);
+      fs.ensureDir(sampleDirPath);
+      const sampleFilePath = path.join(sampleDirPath, sampleFileName);
+      if (await fs.pathExists(sampleFilePath)) {
```
blackchoey, line 60: should we throw error instead of delete existing file? otherwise we may delete user's file

xzf0587: We already check the duplicate files in question stage. 
Here is an inner logic to ensure the writing execution will succeed.

packages/fx-core/src/plugins/resource/apiconnector/utils.ts
```diff
@@ -22,6 +22,15 @@ export function getSampleFileName(apiName: string, languageType: string) {
   return apiName + "." + languageExt;
 }
 
+export function getSampleDirPath(componentPath: string) {
+  let basePath = componentPath;
+  if (fs.pathExistsSync(path.join(basePath, "src"))) {
```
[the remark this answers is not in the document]
xzf0587: As this function is called by some sync methods. keep the rest code stable as this is a quick fix.

wenytang-ms:
please run e2e full path. thanks

wenytang-ms (approved):

packages/fx-core/src/plugins/resource/apiconnector/sampleHandler.ts
blackchoey: can we throw error here in case the question stage didn't find duplicate files correctly?

xzf0587: Will update it in a new PR.

PR #5894, opened 2022-08-24
chore(localdebug): add base custom task terminal

XiaofuHuang:
Add base custom task terminal.
Prepare for teamsfx task.


packages/vscode-extension/src/debug/taskTerminal/baseTaskTerminal.ts
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  handleInput(data: string): void {
+    this.writeEmitter.fire(data === "\r" ? "\r\n" : data);
+    if (data.includes(ControlCodes.CtrlC)) {
+      this.stop();
+    }
+  }
+
+  // TODO: end local debug session
```
swatDong, line 36: Prefer end local debug session in `onDidEndTaskProcessHandler`, and make this pure for terminal.
Is it possible? I remember we already have such logic in `onDidEndTaskProcessHandler`.

packages/vscode-extension/src/debug/taskTerminal/baseTaskTerminal.ts
```diff
@@ -0,0 +1,47 @@
...
+  CtrlC: "\u0003",
+};
+
+export abstract class BaseTaskTerminal implements vscode.Pseudoterminal {
+  protected writeEmitter = new vscode.EventEmitter<string>();
+  onDidWrite: vscode.Event<string> = this.writeEmitter.event;
+  protected closeEmitter = new vscode.EventEmitter<number>();
+  onDidClose?: vscode.Event<number> = this.closeEmitter.event;
+
+  constructor(private taskDefinition: vscode.TaskDefinition) {}
```
swatDong, line 17: Why taking `taskDefinition` as constructor input?

packages/vscode-extension/src/debug/teamsfxTaskProvider.ts
```diff
@@ -151,7 +152,31 @@ export class TeamsfxTaskProvider implements vscode.TaskProvider {
...
+    if (task.definition.type !== TeamsfxTaskProvider.type || !task.definition.command) {
+      return undefined;
+    }
+
+    const createTerminalFuncs = {
+      "check-prerequisites": (d: vscode.TaskDefinition) => new PrerequisiteTaskTerminal(d),
+    };
+
+    const createTerminal = Object.entries(createTerminalFuncs).find(
+      ([k]) => k == task.definition.command
```
swatDong, line 164: ===

packages/vscode-extension/src/debug/teamsfxTaskProvider.ts
```diff
@@ -151,7 +152,31 @@ export class TeamsfxTaskProvider implements vscode.TaskProvider {
     task: vscode.Task,
     token?: vscode.CancellationToken | undefined
   ): Promise<vscode.Task | undefined> {
-    // Return undefined since all tasks are provided and fully resolved
+    if (task.definition.type !== TeamsfxTaskProvider.type || !task.definition.command) {
+      return undefined;
+    }
+
+    const createTerminalFuncs = {
+      "check-prerequisites": (d: vscode.TaskDefinition) => new PrerequisiteTaskTerminal(d),
```
swatDong, line 160: Add `debug` prefix for all new f5 tasks.

packages/vscode-extension/src/debug/teamsfxTaskProvider.ts
```diff
@@ -151,7 +152,31 @@ export class TeamsfxTaskProvider implements vscode.TaskProvider {
     task: vscode.Task,
     token?: vscode.CancellationToken | undefined
   ): Promise<vscode.Task | undefined> {
-    // Return undefined since all tasks are provided and fully resolved
+    if (task.definition.type !== TeamsfxTaskProvider.type || !task.definition.command) {
+      return undefined;
+    }
+
+    const createTerminalFuncs = {
```
swatDong, line 159: Make this global variable?

packages/vscode-extension/src/debug/teamsfxTaskProvider.ts
```diff
@@ -151,7 +152,31 @@ export class TeamsfxTaskProvider implements vscode.TaskProvider {
     task: vscode.Task,
     token?: vscode.CancellationToken | undefined
   ): Promise<vscode.Task | undefined> {
-    // Return undefined since all tasks are provided and fully resolved
+    if (task.definition.type !== TeamsfxTaskProvider.type || !task.definition.command) {
```
kuojianlu, line 155: Why implement in `resolveTask` rather than `provideTasks`?

packages/vscode-extension/src/debug/taskTerminal/baseTaskTerminal.ts
kuojianlu: Consider take `args` as parameter instead of `taskDefinition` since other information is not needed.

packages/vscode-extension/src/debug/taskTerminal/baseTaskTerminal.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  open(): void {
+    this.do()
+      .then(() => this.stop())
```
kuojianlu, line 21: Does this style applies to background task, i.e., starting tunnel?

packages/vscode-extension/src/debug/teamsfxTaskProvider.ts
XiaofuHuang: Changed it, thx.

packages/vscode-extension/src/debug/taskTerminal/baseTaskTerminal.ts
XiaofuHuang: Yes, this will apply to the background task. If the background task is running, the do func will not be resolved.

packages/vscode-extension/src/debug/taskTerminal/baseTaskTerminal.ts
XiaofuHuang:
```
  const additionalProperties: { [key: string]: string } = {};
  {
    // If we know this session is concurrently running with another session, send that correlationId in `debug-all-start` event.
    // Mostly, this happens when user stops debugging while preLaunchTasks are running and immediately hit F5 again.
    const session = commonUtils.getLocalDebugSession();
    if (session.id !== commonUtils.DebugNoSessionId) {
      additionalProperties[TelemetryProperty.DebugConcurrentCorrelationId] = session.id;
      // Indicates in which stage (of the first F5) the user hits F5 again.
      additionalProperties[TelemetryProperty.DebugConcurrentLastEventName] =
        localTelemetryReporter.getLastEventName();
    }
  }
  return await Correlator.runWithId(commonUtils.startLocalDebugSession(), async () => {
    if (commonUtils.checkAndSkipDebugging()) {
      // return non-zero value to let task "exit ${command:xxx}" to exit
      return "1";
    }

    await sendDebugAllStartEvent(additionalProperties);
    const result = await localPrerequisites.checkAndInstall();
    if (result.isErr()) {
      // Only local debug use validate-local-prerequisites command
      await sendDebugAllEvent(result.error);
      commonUtils.endLocalDebugSession();
      // return non-zero value to let task "exit ${command:xxx}" to exit
      return "1";
    }
  });
}
```

Now each command will call  `commonUtils.endLocalDebugSession();` when it failed. It is related to the telemetry.

packages/vscode-extension/src/debug/taskTerminal/baseTaskTerminal.ts
XiaofuHuang: The task definition contains all the custom properties in a task. (Now it is `command` and `args`). In the base class, can we keep all the custom properties?

packages/vscode-extension/src/debug/teamsfxTaskProvider.ts
XiaofuHuang: Updated the command name.

packages/vscode-extension/src/debug/teamsfxTaskProvider.ts
XiaofuHuang: Moved it to global.

packages/vscode-extension/src/debug/taskTerminal/baseTaskTerminal.ts
swatDong: So is this `TODO` already done?

kuojianlu: In `onDidEndTaskProcessHandler`, check the task result, if exit code is non-zero, then call `terminateAllRunningTeamsfxTasks`. So agree to make this class pure for the task logic only.

packages/vscode-extension/src/debug/taskTerminal/baseTaskTerminal.ts
kuojianlu: So how to stop the background task when stopping debugging?

packages/vscode-extension/src/debug/taskTerminal/baseTaskTerminal.ts
kuojianlu: Make sense.

packages/vscode-extension/src/debug/teamsfxTaskProvider.ts
XiaofuHuang: As the discussion [here](https://github.com/microsoft/vscode/issues/58836)
> As the extension author you can implement resolveTask to allow your extension to handle any tasks of your task type that appear in the user's tasks.json but aren't returned by provideTasks. I assume your provideTasks doesn't return a task with your specific args value, so the VS Code tasks system has no idea what to do with it. If you implement resolveTask, you can provide a Task for tasks.json tasks that your provideTasks doesn't return.
> Looks like the documentation doesn't outline this well. I'll update it.

packages/vscode-extension/src/debug/taskTerminal/baseTaskTerminal.ts
XiaofuHuang: > So is this `TODO` already done?

no, only vscode command will call commonUtils.endLocalDebugSession(). After we moved them to task, may be it still need to be called.

packages/vscode-extension/src/debug/teamsfxTaskProvider.ts
swatDong: I believe the key point is:
- If using `provideTasks` and user modifies `args` before F5, will the updated `args` be passed to vsc?

packages/vscode-extension/src/debug/taskTerminal/baseTaskTerminal.ts
XiaofuHuang: > So how to stop the background task managed by custom terminal when stopping debugging?

close() will be called,

packages/vscode-extension/src/debug/taskTerminal/baseTaskTerminal.ts
```diff
@@ -0,0 +1,47 @@
...
+      .catch((error) => this.stop(error));
+  }
+
+  close(): void {
+    this.stop();
+  }
+
+  handleInput(data: string): void {
+    this.writeEmitter.fire(data === "\r" ? "\r\n" : data);
```
a1exwang, line 30: Why do we need to print what user types?

packages/vscode-extension/src/debug/taskTerminal/baseTaskTerminal.ts
```diff
@@ -0,0 +1,47 @@
...
+    if (data.includes(ControlCodes.CtrlC)) {
```
a1exwang, line 31: Will some multi-byte Unicode character or control sequences contain '\x03'? If so will it be treated as Ctrl-C?

packages/vscode-extension/src/debug/taskTerminal/baseTaskTerminal.ts
a1exwang, line 30: Do we really need to support user inputs? From my understanding these tasks are output only?

packages/vscode-extension/src/debug/teamsfxTaskProvider.ts
XiaofuHuang: Yes, I can't list all args combinations in the `provideTasks`.

swatDong: No need to list all args combinations, but just `args` itself. `args` object will be handled by custom execution dynamically.

BTW, one possible reason to use `resolve` instead of `provide` is - we may need to dynamically adjust `presentation` or `runOptions` per `args`.

packages/vscode-extension/src/debug/taskTerminal/baseTaskTerminal.ts
XiaofuHuang: Since it is necessary for the user to enter `ctrl-c` (the user may want to manually end some background tasks), the input is echoed here. If the input is not echoed, the user may mistakenly believe that the window does not support any input. Echoing user input is the same as the behave of the shell terminal.

packages/vscode-extension/src/debug/teamsfxTaskProvider.ts
kuojianlu: I see. `provideTasks` cannot be customized.

packages/vscode-extension/src/debug/taskTerminal/baseTaskTerminal.ts
kuojianlu: I prefer no echo. For a non-background task, pressing `Ctrl-C` to terminate it is not supported. For a backgroud task, I think users know intuitively that he can press `Ctrl-C` to terminate it.

a1exwang: > Echoing user input is the same as the behave of the shell terminal.

I think this is not the behavior of the shell, but the actual process/command. For example, `ngrok` does not echo and `nodemon` does echo.

If you want to implement something like Ctrl-C, I think the better way is just to send anything from input to the process/command's stdin. If it wants to echo it will echo, if it wants to handle Ctrl-C, they will quit on `'\x03'`, etc. Shell should not assume what the actual process/command want to do on keyboard input, Ctrl-C, etc.

XiaofuHuang: ok, I removed the echo from handle input.

packages/vscode-extension/src/debug/taskTerminal/baseTaskTerminal.ts
XiaofuHuang: Here the value of ctrl-c is the virtual terminal sequence.
```
Implement to handle incoming keystrokes in the terminal or when an extension calls [sendText](https://code.visualstudio.com/api/references/vscode-api#Terminal.sendText). data contains the keystrokes/text serialized into their corresponding VT sequence representation.
```

packages/vscode-extension/src/debug/taskTerminal/baseTaskTerminal.ts
XiaofuHuang: May be we can redesign the local debug session in commonUtils and call `commonUtils.endLocalDebugSession()` in `onDidEndTaskProcessHandler`

swatDong: `onDidEndTaskProcessHandler` already calls `terminateAllRunningTeamsfxTasks()`. Is that enough?
I remember VSC stops debug if dependent task has error, thus we only need to stop other running tasks.

XiaofuHuang: Here the debug session means the local debug session used to generate session id and send telemetry. As we discussed offline, for teamsfx tasks in a teamsfx project, we need to add the following:
1. start the debug session in `onDidStartTaskProcessHandler` if the exsiting debug session is undefined.
2. stop the debug session in `onDidEndTaskProcessHandler` for non-background task
3. stop the debug session in `onDidStartDebugSessionHandler` for background task (exsiting)

kuojianlu (approved):

a1exwang (approved):

a1exwang (approved):

swatDong (approved):

PR #5905, opened 2022-08-25
feat(sdk): add sso command bot support

SLdragon:
Command bot in Teams Toolkit extension currently only supports non-sso command by default, although user can trigger command “add sso” to add sso support, it needs a lot of manual steps, and the code structure is not very consistent with the command bot design.

This PR update the SDK to support sso command bot.

Related design doc: 
https://microsoftapc-my.sharepoint.com/:w:/g/personal/rentu_microsoft_com/EdjwurIlkn1Mt_MCiptv-mABqNi_7y1d2t_7zBdKeWp-6w?e=czXxz1

The draft sample sso comamnd bot project to show how to use the SDK:
https://github.com/SLdragon/sso-command-bot


packages/sdk/src/conversation/command.ts
```diff
@@ -22,7 +30,22 @@ export class CommandBot {
    * @param options - initialize options
    */
   constructor(adapter: BotFrameworkAdapter, options?: CommandOptions) {
-    this.middleware = new CommandResponseMiddleware(options?.commands);
+    let ssoCommandActivityHandler: SsoExecutionActivityHandler | undefined;
+    if (options?.ssoConfig?.CustomSsoExecutionActivityHandler) {
+      ssoCommandActivityHandler = new options.ssoConfig.CustomSsoExecutionActivityHandler(
+        options?.ssoConfig
+      );
+    } else if (options?.ssoCommands?.length && options?.ssoCommands?.length > 0) {
```
tecton, line 38:
```suggestion
    } else if (options?.ssoCommands?.length > 0) {
```
Can it be simplified to this one?

packages/sdk/src/conversation/command.ts
```diff
@@ -40,11 +63,43 @@ export class CommandBot {
...
   public registerCommands(commands: TeamsFxBotCommandHandler[]): void {
     if (commands) {
       this.middleware.commandHandlers.push(...commands);
     }
   }
+
+  /**
+   * Registers a sso command into the command bot.
+   *
+   * @param command The command to registered.
```
tecton, line 77:
```suggestion
   * @param command The command to register.
```

packages/sdk/src/conversation/command.ts
```diff
@@ -40,11 +63,43 @@ export class CommandBot {
...
+      }
+      this.middleware.commandHandlers.push(ssoCommand);
+      this.middleware.activityHandler?.addCommand(ssoCommand);
+    }
+  }
+
+  /**
+   * Registers commands into the command bot.
+   *
+   * @param commands The commands to registered.
```
tecton, line 92:
```suggestion
   * @param commands The commands to register.
```

packages/sdk/src/index.ts
```diff
@@ -56,8 +56,13 @@ export {
   TriggerPatterns,
   TeamsFxAdaptiveCardActionHandler,
   TeamsFxBotCommandHandler,
+  TeamsFxBotSsoCommandHandler,
+  SsoExecutionActivityHandler,
+  SsoConfig,
 } from "./conversation/interface";
 export { ConversationBot } from "./conversation/conversation";
+export { SsoExecutionDialog } from "./conversation/sso/ssoExecutionDialog";
```
swatDong, line 64: Why exposing these two class? How will these be used in user code?

packages/sdk/src/conversation/interface.ts
```diff
@@ -163,6 +171,36 @@ export interface TeamsFxBotCommandHandler {
...
+   * @param context The bot context.
+   * @param message The command message the user types from Teams.
+   * @param ssoToken The sso token which can be used to exchange access token for the bot.
+   * @returns A `Promise` representing an activity or text to send as the command response.
+   * Or no return value if developers want to send the response activity by themselves in this method.
+   */
+  handleCommandReceived(
+    context: TurnContext,
+    message: CommandMessage,
+    ssoToken: string
```
swatDong, line 200: Does this token string contains expiration info, or, how long does it ensure before expired?

packages/sdk/src/index.ts
SLdragon: For SsoExecutionDialog is a normal bot dialog that can be used directly by advance user if he doesn't want to use our ConversionBot.


For DefaultSsoExecutionActivityHandler, it can be used as below, since we will comment the ssoConfig part for the template/sample, I expose this class to let user understand what the default value we're using:
```ts
export const commandBot = new ConversationBot({
  // The bot id and password to create BotFrameworkAdapter.
  // See https://aka.ms/about-bot-adapter to learn more about adapters.
  adapterConfig: {
    appId: process.env.BOT_ID,
    appPassword: process.env.BOT_PASSWORD,
  },

  command: {
    enabled: true,
    commands: [new HelloWorldCommandHandler() ],
    ssoCommands: [new ProfileSsoCommandHandler()],
    // ssoConfig is optional and by default using the values below
    ssoConfig: {
      CustomSsoExecutionActivityHandler: DefaultSsoExecutionActivityHandler,
      scopes: ["User.Read"],
      userState: new UserState(storage),
      conversationState: new ConversationState(storage),
      dedupStorage: storage,
      teamsFxConfig: {
        clientId: process.env.M365_CLIENT_ID,
        clientSecret: process.env.M365_CLIENT_SECRET,
        tenantId: process.env.M365_TENANT_ID,
        authorityHost: process.env.M365_AUTHORITY_HOST,
        initiateLoginEndpoint: process.env.INITIATE_LOGIN_ENDPOINT,
        applicationIdUri: process.env.M365_APPLICATION_ID_URI
      }
    }
  },
```

packages/sdk/src/conversation/interface.ts
SLdragon: Yes, expiration time can be parsed directly from ssoToken as below:
```
const ssoTokenExpiration = parseJwt(ssoToken).exp;
```

swatDong: Is it possible to provide something like `tokenProvider` instead of the one-time token itself.
I believe `tokenProvider` brings more benefits:
- Totally let user to determine whether to call `getToken`, for better performance and avoid potential expire risk.
- Totally let user to determine the token scope.
- Even can merge `TeamsFxBotCommandHandler` and `TeamsFxBotSsoCommandHandler` into one interface, and save user's learning curve.
- ...

I'm not sure if there's any technical limitation, but from auth UX a `provider` is better than `token` itself. (That's also how MSAL works.)

packages/sdk/src/conversation/command.ts
SLdragon: Cannot simplified as this, because it will show error as below in typescript:
image

packages/sdk/src/conversation/command.ts
SLdragon: Oh, thanks, will update

packages/sdk/src/conversation/command.ts
SLdragon: will update

packages/sdk/src/conversation/interface.ts
swatDong: Or, if not having `provider`, how to avoid token expired error?
E.g., the handler logic takes more than 10 minutes and the input `ssoToken` is expired when used. Is there any limitation that `ssoToken` only ensures xxx minutes before expired, or any workaround to get a new token?

SLdragon: We cannot give a tokenProvider for user to get sso token directly, because sso token is come from Teams Client, for botframework, it is backend service, and it needs to use dialog to ask user to login and then obtain sso token from teams client. User can use this sso token to exchange access token for Graph as below:

```
    const teamsfx = new TeamsFx().setSsoToken(ssoToken);
    const token =  teamsfx.getToken(["User.Read"])
```

You can compare with the teams js sdk, it just return a ssoToken, then you can use this ssoToken for OBO flow to exchange real access token (need user consent, botframwork put the consent step in dialog) :

```ts
 microsoftTeams.initialize(() => {
          microsoftTeams.authentication.getAuthToken({
            successCallback: (ssoToken: string) => {
              resolve(ssoToken);
            },
            failureCallback: (errMessage: string) => {
              const errorMsg = "Get SSO token failed with error: " + errMessage;
              internalLogger.error(errorMsg);
              reject(new ErrorWithCode(errorMsg, ErrorCode.InternalError));
            },
            resources: [],
          });
        });
```

swatDong: Make sense that bot needs to trigger dialog to get token from Teams Client, just curious is it possible to proactively trigger the dialog (with turnContext and ssoConfig as inputs).

BTW, teams js sdk is more like `provider` since user can explicitly call `token = await teamsJs.authentication.getAuthToken(...)` to get token.

Let me try to build a sample and get back to this.

packages/sdk/src/conversation/interface.ts
```diff
@@ -261,6 +309,95 @@ export interface TeamsFxAdaptiveCardActionHandler {
...
+
+  /**
+   * The list of scopes for which the token will have access, if not provided, it will use graph permission ["User.Read"] by default
+   */
+  scopes?: string[];
+
+  /**
+   * Conversation state for sso command bot, if not provided, it will use internal memory storage to create a new one.
+   */
+  conversationState?: ConversationState;
```
blackchoey, line 329: is it possible to just provide a Storage and we help user create conversationState, userState? By the way, can we reuse same storage for xxxState and dedupStorage? If yes, we may also remove dedupStorage config

packages/sdk/src/conversation/interface.ts
```diff
@@ -163,6 +171,36 @@ export interface TeamsFxBotCommandHandler {
...
+  /**
+   * Handles a bot command received activity.
+   *
+   * @param context The bot context.
+   * @param message The command message the user types from Teams.
+   * @param ssoToken The sso token which can be used to exchange access token for the bot.
+   * @returns A `Promise` representing an activity or text to send as the command response.
+   * Or no return value if developers want to send the response activity by themselves in this method.
+   */
+  handleCommandReceived(
```
blackchoey, line 197: can we change `ssoToken: string` to `token: TeamsSsoPromptTokenResponse` which contains bot sso token and access token?

packages/sdk/src/conversation/command.ts
```diff
@@ -40,11 +63,43 @@ export class CommandBot {
...
+   * @param command The command to register.
+   */
+  public registerSsoCommand(ssoCommand: TeamsFxBotSsoCommandHandler): void {
+    if (ssoCommand) {
+      if (!this.middleware.getActivityHandler()) {
+        this.middleware.setActivityHandler(new DefaultSsoExecutionActivityHandler(this.ssoConfig));
+      }
+      this.middleware.commandHandlers.push(ssoCommand);
+      this.middleware.activityHandler?.addCommand(ssoCommand);
+    }
```
blackchoey, line 86: should we throw an error if ssoCommand is empty? there may be no case to pass undefined to this function, so if we found ssoCommand is undefined, there may be some user code error

packages/sdk/src/conversation/interface.ts
SLdragon: Not sure if there are some cases that user want to use different storage for conversationState, userState, so I just expose these 3 properties for flexibility.

packages/sdk/src/conversation/sso/defaultSsoExecutionActivityHandler.ts
```diff
@@ -0,0 +1,111 @@
...
+    const userState = ssoConfig?.userState ?? new UserState(memoryStorage);
+    const conversationState = ssoConfig?.conversationState ?? new ConversationState(memoryStorage);
+    const dedupStorage = ssoConfig?.dedupStorage ?? memoryStorage;
+    const scopes = ssoConfig?.scopes ?? ["User.Read"];
+
+    const teamsfx = new TeamsFx(IdentityType.User, ssoConfig?.teamsFxConfig);
+    const settings: TeamsBotSsoPromptSettings = {
+      scopes: scopes,
+      timeout: ssoConfig?.ssoPromptConfig?.timeout,
+      endOnInvalidMessage: ssoConfig?.ssoPromptConfig?.endOnInvalidMessage === false ? false : true,
```
blackchoey, line 48: `endOnInvalidMessage` is also optional, may be able to assign `ssoConfig?.ssoPromptConfig?.endOnInvalidMessage` to it directly

packages/sdk/src/conversation/command.ts
```diff
@@ -40,11 +63,43 @@ export class CommandBot {
...
+   *
+   * @param commands The commands to register.
+   */
+  public registerSsoCommands(ssoCommands: TeamsFxBotSsoCommandHandler[]): void {
+    if (ssoCommands) {
+      if (!this.middleware.getActivityHandler()) {
+        this.middleware.setActivityHandler(new DefaultSsoExecutionActivityHandler(this.ssoConfig));
+      }
+      for (const ssoCommand of ssoCommands) {
+        this.middleware.activityHandler?.addCommand(ssoCommand);
```
wenytang-ms, line 100: seems this.middleware.activityHandler is exist, if not, will set in line 97
when will meet this.middleware.activityHandler get null?

SLdragon: Yes, that's true, however, typescript cannot detect this, I need to use ? to avoid it complains about activityHandler may be null.

packages/sdk/src/conversation/sso/defaultSsoExecutionActivityHandler.ts
SLdragon: Thanks, will update~

packages/sdk/src/conversation/command.ts
SLdragon: OK, thanks, will update

SLdragon: Seems typescript can check whether the parameter is invalid, so I remove the check in function body

packages/sdk/src/conversation/middlewares/commandMiddleware.ts
```diff
@@ -2,55 +2,101 @@
...
+              await context.sendActivity(response);
+            } else {
+              const replyActivity = response as Partial<Activity>;
+              if (replyActivity) {
+                await context.sendActivity(replyActivity);
+              }
             }
           }
-
-          break;
```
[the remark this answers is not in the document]
SLdragon: Thanks, will update

packages/sdk/src/conversation/interface.ts
swatDong: I tried provider-style and it works. I modified code based on the bot-sso sample, by just adding following code to `teamsBot.ts` L60.

``` typescript
      if (txt === "test123") {
        /** wrap this as `token provider` */
        let deferredToken: Deferred<string>;
        const tokenPromise: Promise<string> = new Promise<string>(
          (resolve, reject) => (deferredToken = { resolve, reject })
        );
        this.dialog.setSSOOperation(async (text, token) => {
          deferredToken.resolve(token);
        });
        await this.dialog.run(context, this.dialogState);
        /** *************************** */

        const token = await tokenPromise;
        await context.sendActivity(token);
      }
```

And here's the teams result (it supports consent as well):
image

Actually, what I mean for `tokenProvider` is to provide a way to **proactively trigger sso dialog**. There's also similar issue (https://stackoverflow.com/questions/56975804/how-to-send-a-proactive-dialog-with-bot-framework) with certain answers.

Anyway, it's up to you for the final design. My concern is that - if we're going to add SSO to proactively messaging (e.g., notification), `provider` is always needed.

packages/sdk/src/conversation/interface.ts
```diff
@@ -261,6 +309,95 @@ export interface TeamsFxAdaptiveCardActionHandler {
...
+     * message not related to the auth flow. Setting the flag to false ignores the user's message instead.
+     * Defaults to value `true`
+     */
+    endOnInvalidMessage?: boolean;
+  };
+
+  /**
+   * teamsfx configuration for sso
+   */
+  teamsFxConfig?: {
```
blackchoey, line 363: is it possible to set the type to authentication config? or some other type in the upcoming change from Frank?

packages/sdk/src/conversation/sso/ssoExecutionDialog.ts
```diff
@@ -0,0 +1,327 @@
...
+    this.addDialog(commandRouteDialog);
+  }
+
+  /**
+   * Add TeamsFxBotSsoCommandHandler instance
+   * @param handler TeamsFxBotSsoCommandHandler instance
+   */
+  public addCommand(handler: TeamsFxBotSsoCommandHandler): void {
+    if (!handler.commandId) {
+      handler.commandId = uuidv4();
```
swatDong, line 84: If random id always works, why expose this to user's code?
Or, what's the purpose to let user to set commandId?

packages/sdk/src/conversation/interface.ts
```diff
@@ -163,6 +171,36 @@ export interface TeamsFxBotCommandHandler {
   ): Promise<string | Partial<Activity> | void>;
 }
 
+/**
+ * Interface for a command handler that can process sso command to a TeamsFx bot and return a response.
+ */
+export interface TeamsFxBotSsoCommandHandler {
```
swatDong, line 177: You can add yourself to sdk/conversation code owner  : )

packages/sdk/src/conversation/interface.ts
SLdragon: Sure

packages/sdk/src/conversation/sso/ssoExecutionDialog.ts
SLdragon: Seems typescript cannot check whether a class is type of a specific interface, so I need a property to determine whether it is a TeamsFxBotSsoCommandHandler or TeamsFxBotCommandHandler, do you have any suggestions without a property?

```
  private isSsoExecutionHandler(
    handler: TeamsFxBotCommandHandler | TeamsFxBotSsoCommandHandler
  ): boolean {
    return "commandId" in handler;
  }
```

packages/sdk/src/conversation/interface.ts
SLdragon: Sure

packages/sdk/src/conversation/middlewares/commandMiddleware.ts
```diff
@@ -2,55 +2,101 @@
...
+
+      for (const ssoHandler of ssoHandlers) {
+        this.activityHandler?.addCommand(ssoHandler);
+      }
     }
   }
 
+  /**
+   * Set sso execution activity handler
+   * @param activityHandler SsoExecutionActivityHandler instance
```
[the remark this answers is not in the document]
SLdragon: Oh, thanks, I think I can remove these API

packages/sdk/src/conversation/middlewares/commandMiddleware.ts
```diff
@@ -2,55 +2,101 @@
...
  * @internal
  */
 export class CommandResponseMiddleware implements Middleware {
-  public readonly commandHandlers: TeamsFxBotCommandHandler[] = [];
+  public readonly commandHandlers: (TeamsFxBotCommandHandler | TeamsFxBotSsoCommandHandler)[] = [];
 
-  constructor(handlers?: TeamsFxBotCommandHandler[]) {
-    if (handlers && handlers.length > 0) {
-      this.commandHandlers.push(...handlers);
+  public activityHandler: SsoExecutionActivityHandler | undefined;
```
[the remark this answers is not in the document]
SLdragon: Sure

packages/sdk/src/conversation/interface.ts
```diff
@@ -261,6 +309,95 @@ export interface TeamsFxAdaptiveCardActionHandler {
   handleActionInvoked(context: TurnContext, actionData: any): Promise<InvokeResponse>;
 }
 
+/**
+ * Interface for SSO configuration for BotSSO
```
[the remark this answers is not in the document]
SLdragon: Will update

packages/sdk/src/conversation/middlewares/commandMiddleware.ts
```diff
@@ -2,55 +2,101 @@
...
+
+  constructor(
+    handlers?: TeamsFxBotCommandHandler[],
+    ssoHandlers?: TeamsFxBotSsoCommandHandler[],
```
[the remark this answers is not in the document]
SLdragon: By default, if user uses ssoHandler by conversation bot just like our template, we will create a default one if it doesn't exist. 

I think I can add some checks, if user just uses command bot and do not pass a sso activity handler, then we can throw error

packages/sdk/src/conversation/interface.ts
SLdragon: Sure, thanks, will use authentication config as our meeting disscussed

packages/sdk/src/conversation/middlewares/commandMiddleware.ts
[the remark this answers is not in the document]
SLdragon: Added some check

packages/sdk/src/conversation/interface.ts
SLdragon: Thanks, I will investigate this

swatDong: Offline discussed and understand the distributed limitation...
Agree that currently hard to have the `provider`-style solution. THX~

packages/sdk/src/conversation/sso/ssoExecutionDialog.ts
swatDong: The potential risk is that user can add any property to the implementation. E.g., implement the `TeamsFxBotCommandHandler` interface but add one `commandId` property...

So suggest to totally separate them in middleware. Store in two arrays, or store as tuple with type, or explicitly add type to interface, etc.

packages/sdk/src/conversation/interface.ts
```diff
@@ -261,6 +306,60 @@ export interface TeamsFxAdaptiveCardActionHandler {
...
+     * message not related to the auth flow. Setting the flag to false ignores the user's message instead.
+     * Defaults to value `true`
+     */
+    endOnInvalidMessage?: boolean;
+  };
+
+  /**
+   * teamsfx configuration for sso
+   */
+  teamsFxConfig?: Partial<AuthenticationConfiguration>;
```
wenytang-ms, line 360: seems the AuthenticationConfiguration already a partial interface?

packages/sdk/src/conversation/sso/ssoExecutionDialog.ts
SLdragon: Thanks, will update

packages/sdk/src/conversation/interface.ts
SLdragon: Oh, thanks, will update

packages/sdk/src/conversation/interface.ts
```diff
@@ -261,6 +309,95 @@ export interface TeamsFxAdaptiveCardActionHandler {
...
+ * Interface for SSO configuration for BotSSO
+ */
+export interface SsoConfig {
+  /**
+   * Custom sso execution activity handler class which should implement the interface {@link SsoExecutionActivityHandler}. If not provided, it will use {@link DefaultSsoExecutionActivityHandler} by default
+   */
+  CustomSsoExecutionActivityHandler?: new (ssoConfig: SsoConfig) => SsoExecutionActivityHandler;
+
+  /**
+   * The list of scopes for which the token will have access, if not provided, it will use graph permission ["User.Read"] by default
```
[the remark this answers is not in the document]
SLdragon: Sure

.github/CODEOWNERS.bak
```diff
@@ -0,0 +1,138 @@
+# Lines starting with '#' are comments.
```
dooriya, line 1: Is this `bak` file checked in by accident?

SLdragon: Oh, seems it checked by accident, will remove it

packages/sdk/src/conversation/interface.ts
```diff
@@ -261,6 +301,60 @@ export interface TeamsFxAdaptiveCardActionHandler {
...
+  /**
+   * The list of scopes for which the token will have access
+   */
+  scopes: string[];
+
+  /**
+   * Custom sso execution activity handler class which should implement the interface {@link SsoExecutionActivityHandler}. If not provided, it will use {@link DefaultSsoExecutionActivityHandler} by default
+   */
+  CustomSsoExecutionActivityHandler?: new (ssoConfig: SsoConfig) => SsoExecutionActivityHandler;
+
```
blackchoey, line 317: The name may be too common, since npm package does not support namespace, we may use a more spefic name to avoid confusing?

packages/sdk/src/conversation/interface.ts
```diff
@@ -313,3 +412,21 @@ export interface ConversationOptions {
...
   };
 }
+
+/**
+ * Interface for user to customize sso execution activity handler
+ */
+export interface SsoExecutionActivityHandler extends TeamsActivityHandler {
+  /**
+   * Add {@link TeamsFxBotSsoCommandHandler} instance to {@link SsoExecutionDialog}
+   * @param handler {@link SsoExecutionDialogHandler} callback function
```
blackchoey, line 422: we also implemented many other functions like `handleTeamsSigninVerifyState`. should we add all these functions to this interface?

packages/sdk/src/conversation/sso/ssoExecutionDialog.ts
```diff
@@ -0,0 +1,307 @@
...
+let COMMAND_ROUTE_DIALOG = "CommandRouteDialog";
+
+/**
+ * Sso execution dialog, use to handle sso command
+ */
+export class SsoExecutionDialog extends ComponentDialog {
+  private dedupStorage: Storage;
+  private dedupStorageKeys: string[] = [];
+  private commandMapping: Map<string, string | RegExp | (string | RegExp)[]> = new Map<
+    string,
```
blackchoey, line 41: minor: may be add some comments to explain what does the key and value mean?

packages/sdk/src/conversation/sso/ssoExecutionDialog.ts
```diff
@@ -0,0 +1,307 @@
...
+   * @param context The context object for the current turn.
+   * @param accessor The instance of StatePropertyAccessor for dialog system.
+   */
+  public async run(context: TurnContext, accessor: StatePropertyAccessor) {
+    const dialogSet = new DialogSet(accessor);
+    dialogSet.add(this);
+
+    const dialogContext = await dialogSet.createContext(context);
+    this.ensureMsTeamsChannel(dialogContext);
+    const results = await dialogContext.continueDialog();
```
blackchoey, line 124: should we share this information to developer and let them decide how to handle sso failed cases?

packages/sdk/src/conversation/sso/ssoExecutionDialog.ts
```diff
@@ -0,0 +1,307 @@
...
+    if (results && results.status === DialogTurnStatus.empty) {
+      await dialogContext.beginDialog(this.id);
+    }
+  }
```
blackchoey, line 128: also let developers decide how to handle this case

packages/sdk/src/conversation/sso/ssoExecutionDialog.ts
```diff
@@ -0,0 +1,307 @@
...
+
+  private async dedupStep(stepContext: any) {
+    try {
+      const tokenResponse = stepContext.result;
+      // Only dedup after ssoStep to make sure that all Teams client would receive the login request
+      if (tokenResponse && (await this.shouldDedup(stepContext.context))) {
+        return Dialog.EndOfTurn;
+      }
+      return await stepContext.next({
+        tokenResponse,
```
blackchoey, line 184: also let users decide how to handle command not found case?

packages/sdk/src/conversation/sso/ssoExecutionDialog.ts
```diff
@@ -0,0 +1,307 @@
...
+        message: stepContext.options.commandMessage,
+      });
+    } catch (error: unknown) {
+      const context = stepContext.context;
+      await context.sendActivity("Failed to run dedup step");
+      await context.sendActivity((error as Error).message);
+      return await stepContext.endDialog();
+    }
+  }
+
```
blackchoey, line 194: we may save the raw text (without removing mention and other things), in case developer needs related information

packages/sdk/src/conversation/sso/ssoExecutionDialog.ts
```diff
@@ -0,0 +1,307 @@
...
+        const matches = text.match(pattern as RegExp);
+        return matches ?? false;
+      }
+    }
+
+    return false;
+  }
+
+  private shouldTrigger(patterns: TriggerPatterns, text: string): RegExpMatchArray | boolean {
+    const expressions = Array.isArray(patterns) ? patterns : [patterns];
```
blackchoey, line 269: Our code use both `Error` and `ErrorWithCode`, should we only use one of them?

packages/sdk/src/conversation/sso/ssoExecutionDialog.ts
```diff
@@ -0,0 +1,307 @@
...
+            await context.sendActivity("Failed to retrieve user token from conversation context.");
+          }
+          return await stepContext.endDialog();
+        } catch (error: unknown) {
+          await context.sendActivity("Failed to retrieve user token from conversation context.");
+          await context.sendActivity((error as Error).message as string);
+          return await stepContext.endDialog();
+        }
+      },
+    ]);
```
blackchoey, line 106: may give `shouldTrigger` another name to help people better understand the code. At the beginning I thought we checks whether we need to execute the command here again. Actually we only want to get the result of regex capture group.

packages/sdk/src/conversation/sso/ssoExecutionDialog.ts
```diff
@@ -0,0 +1,307 @@
...
+      await dialogContext.beginDialog(this.id);
+    }
+  }
+
+  private getActivityText(activity: Activity): string {
+    let text = activity.text;
+    const removedMentionText = TurnContext.removeRecipientMention(activity);
+    if (removedMentionText) {
+      text = removedMentionText
+        .toLowerCase()
```
blackchoey, line 135: minor: may put this logic to the beginning?

packages/sdk/src/conversation/sso/ssoExecutionDialog.ts
```diff
@@ -0,0 +1,307 @@
...
+      const context = stepContext.context;
+      await context.sendActivity("Failed to run SSO step");
+      await context.sendActivity((error as Error).message);
+      return await stepContext.endDialog();
+    }
+  }
+
+  private async dedupStep(stepContext: any) {
+    try {
+      const tokenResponse = stepContext.result;
```
blackchoey, line 178: should we match the original text? I'm not sure whether users know we did some preprocessing before match. If they don't, they may write wrong regex (for example, 'ShowProfile .\*' instead of 'showprofile .\*')

packages/sdk/src/conversation/interface.ts
SLdragon: sure

packages/sdk/src/conversation/sso/ssoExecutionDialog.ts
SLdragon: sure

packages/sdk/src/conversation/sso/ssoExecutionDialog.ts
SLdragon: The match in our code ignore cases, we do not need to worry about the wrong regex

packages/sdk/src/conversation/sso/ssoExecutionDialog.ts
SLdragon: OK

packages/sdk/src/conversation/interface.ts
SLdragon: Will discuss with PM to finalize the name

packages/sdk/src/conversation/sso/ssoExecutionDialog.ts
SLdragon: OK

packages/sdk/src/conversation/sso/ssoExecutionDialog.ts
SLdragon: OK

packages/sdk/src/conversation/sso/ssoExecutionDialog.ts
SLdragon: As discussed, we will keep it same logic with none sso command bot

packages/sdk/src/conversation/sso/ssoExecutionDialog.ts
SLdragon: OK

packages/sdk/src/conversation/sso/ssoExecutionDialog.ts
SLdragon: already removed

packages/sdk/src/conversation/sso/ssoExecutionDialog.ts
SLdragon: Keep the logic unchanged to align with none sso command bot

SLdragon:
Open quesitons need further dicussion:

1. ssoConfig naming
1. Whether need to expose commandId
1. commandId use uuid may introduce problem, how about use hash or anything else
1. Do we need to combine sso handler with none sso handler, and add a function to determine whether it need sso token

```
export class ProfileSsoCommandHandler implements TeamsFxBotSsoCommandHandler {
  triggerPatterns: TriggerPatterns = "profile";
  requireSso?: (context, message): boolean=>{
     return true;
  }
  async handleCommandReceived(
    context: TurnContext,
    message: CommandMessage,
    tokenResponse: TeamsBotSsoPromptTokenResponse,
  ): Promise<string | Partial<Activity> | void> {
 
}

```

swatDong:
> Open quesitons need further dicussion:
> 
> 1. ssoConfig naming
> 2. Whether need to expose commandId
> 3. commandId use uuid may introduce problem, how about use hash or anything else
> 4. commandmapping in dialog is a map, when there is a handler's pattern that override other handler's patter, what's the expect behaviour
> 5. Do we need to combine sso handler with none sso handler, and add a function to determine whether it need sso token
> 
> ```
> export class ProfileSsoCommandHandler implements TeamsFxBotSsoCommandHandler {
>   triggerPatterns: TriggerPatterns = "profile";
>   requireSso?: (context, message): boolean=>{
>      return true;
>   }
>   async handleCommandReceived(
>     context: TurnContext,
>     message: CommandMessage,
>     tokenResponse: TeamsBotSsoPromptTokenResponse,
>   ): Promise<string | Partial<Activity> | void> {
>  
> }
> ```

This is to bring following benefits:
- Unified interface for both non-sso and sso command.
- Support conditional sso command.

Add code sample for the merged interface.
``` typescript
interface TeamsFxBotCommandHandler {
    triggerPatterns: TriggerPatterns;
    isSsoRequired?(context: TurnContext,message: CommandMessage): Promise<boolean>;
    handleCommandReceived(context: TurnContext, message: CommandMessage, ssoToken?: TeamsBotSsoPromptTokenResponse): Promise<string | Partial<Activity> | void>;
}
```

For non-sso command, e.g., a simple hello command:
``` typescript
class HelloCommand implements TeamsFxBotCommandHandler {
    triggerPatterns = "hello";
    handleCommandReceived(context: TurnContext, message: CommandMessage): Promise<string | void | Partial<Activity>> {
        return Promise.resolve("Hello again");
    }
}
```

For always-sso command, e.g., a simple show-token command:
``` typescript
class TokenCommand implements TeamsFxBotCommandHandler {
    triggerPatterns = "show-token";
    isSsoRequired(context: TurnContext, message: CommandMessage): Promise<boolean> {
        return Promise.resolve(true);
    }
    handleCommandReceived(context: TurnContext, message: CommandMessage, ssoToken?: TeamsBotSsoPromptTokenResponse | undefined): Promise<string | void | Partial<Activity>> {
        return Promise.resolve(ssoToken?.token);
    }
}
```

For conditional-sso command, e.g., show private/public repo information command:
``` typescript
class RepoCommand implements TeamsFxBotCommandHandler {
    triggerPatterns = /show-repo (https:\/\/github\.com\/.*)/;
    isSsoRequired(context: TurnContext, message: CommandMessage): Promise<boolean> {
        const repoUrl = message.matches![0];
        return isPrivateRepo(repoUrl);
    }
    handleCommandReceived(context: TurnContext, message: CommandMessage, ssoToken?: TeamsBotSsoPromptTokenResponse | undefined): Promise<string | void | Partial<Activity>> {
        let repoInfo;
        if (ssoToken) {
            repoInfo = ... // exchange token to access private repo
        } else {
            repoIndo = ... // directly access public repo
        }
        return repoInfo;
    }
}
```

packages/sdk/src/conversation/middlewares/commandMiddleware.ts
```diff
@@ -2,55 +2,123 @@
...
-  constructor(handlers?: TeamsFxBotCommandHandler[]) {
-    if (handlers && handlers.length > 0) {
-      this.commandHandlers.push(...handlers);
+    this.commandHandlers.push(...handlers);
+    for (const ssoHandler of ssoHandlers) {
+      this.addSsoCommand(ssoHandler);
     }
   }
 
+  public addSsoCommand(ssoHandler: TeamsFxBotSsoCommandHandler) {
```
dooriya, line 50: explicitly return `void` here?

packages/sdk/src/conversation/interface.ts
```diff
@@ -313,3 +416,31 @@ export interface ConversationOptions {
     enabled?: boolean;
   };
 }
+
+/**
+ * Interface for user to customize sso execution activity handler
+ */
+export interface BotSsoExecutionActivityHandler {
```
dooriya, line 423: Can we add API document for all interface APIs to provide some guidance on implementing a custom SSO activity handler?

dooriya, line 423: Can we add API document for each APIs to provide some guidance on implementing a customized SSO activity handler？

packages/sdk/src/conversation/middlewares/commandMiddleware.ts
SLdragon: OK， thanks

dooriya (approved):

swatDong (approved):

dooriya (approved):

.github/CODEOWNERS
```diff
@@ -44,7 +44,7 @@
 /packages/api/src/schemas @dooriya @qinezh @a1exwang @kimizhu
 
 /packages/sdk @tecton @blackchoey @SLdragon @dilin-MS @wenytang-ms @yiqing-zhao
-/packages/sdk/src/conversation @kimizhu @swatDong @a1exwang @XiaofuHuang @dooriya
+/packages/sdk/src/conversation @kimizhu @swatDong @a1exwang @XiaofuHuang @dooriya @SLdragon
```
blackchoey, line 47: I saw GitHub complains the codeowners file has error, please take a look

SLdragon: Seems dilin-MS doesn't have write access to this repo any more, I will remove her.

packages/sdk/src/conversation/interface.ts
```diff
@@ -163,6 +173,31 @@ export interface TeamsFxBotCommandHandler {
...
+   * @param context The bot context.
+   * @param message The command message the user types from Teams.
+   * @param ssoToken The sso token which can be used to exchange access token for the bot.
+   * @returns A `Promise` representing an activity or text to send as the command response.
+   * Or no return value if developers want to send the response activity by themselves in this method.
+   */
+  handleCommandReceived(
+    context: TurnContext,
+    message: CommandMessage,
+    ssoToken: TeamsBotSsoPromptTokenResponse
```
blackchoey, line 197: do we want to rename the parameter name since it's not only a sso token?

packages/sdk/src/conversation/interface.ts
```diff
@@ -261,6 +301,64 @@ export interface TeamsFxAdaptiveCardActionHandler {
...
+     * Custom sso execution activity handler class which should implement the interface {@link BotSsoExecutionActivityHandler}. If not provided, it will use {@link DefaultBotSsoExecutionActivityHandler} by default
+     */
+    CustomBotSsoExecutionActivityHandler?: new (
+      ssoConfig: BotSsoConfig
+    ) => BotSsoExecutionActivityHandler;
+
+    /**
+     * Conversation state for sso command bot, if not provided, it will use internal memory storage to create a new one.
+     */
+    conversationState?: ConversationState;
```
blackchoey, line 329: Per my understanding, developers need to use non memory storage in production environment. How do we tell users about this?

packages/sdk/src/index.ts
```diff
@@ -56,8 +56,14 @@ export {
   TriggerPatterns,
   TeamsFxAdaptiveCardActionHandler,
   TeamsFxBotCommandHandler,
+  TeamsFxBotSsoCommandHandler,
+  BotSsoExecutionActivityHandler,
+  BotSsoConfig,
+  BotSsoExecutionDialogHandler,
 } from "./conversation/interface";
 export { ConversationBot } from "./conversation/conversation";
+export { BotSsoExecutionDialog } from "./conversation/sso/botSsoExecutionDialog";
```
blackchoey, line 65: I'm not sure whether we want to export BotSsoExecutionDialog and DefaultBotSsoExecutionActivityHandler. Can you share how they serves for the feature? My major concern is they're our internal implementation, if we exported them, we need to consider more when we want to ajust our internal implementations.

packages/sdk/src/conversation/interface.ts
SLdragon: Oh, thanks, will update

blackchoey (approved):
Overall looks good to me. For my feature related comments, you can have additional PR if necessary.

packages/sdk/src/index.ts
SLdragon: Yes, we will export BotSsoExecutionDialog, since it is a general dialog and user can use it process custom command as below:
```
    this.ssoExecutionDialog?.addCommand(
      async (
        context: TurnContext,
        tokenResponse: TeamsBotSsoPromptTokenResponse,
        message: CommandMessage
      ) => {
        ...
        user's logic
        ...
      },
      triggerPattern
    );
```

For DefaultBotSsoExecutionActivityHandler, we can hide and do not export it

packages/sdk/src/conversation/interface.ts
SLdragon: I think we can add comments in our sample or template

SLdragon:
Will merge this PR, and I will create another PR to update according to Chaoyi's comments

PR #5935, opened 2022-08-29
chore(componentize): update sql for debug

xzf0587:


packages/fx-core/src/component/utils.ts
```diff
@@ -485,9 +485,8 @@ export function getHostingComponent(
   return undefined;
 }
 
-// TODO:implement after V3 project setting update
-export function isHostedByAzure(context: ContextV3): boolean {
-  return true;
+export function isDebug(context: ContextV3): boolean {
```
hund030, line 488: `isDebug` sounds like checking development mode or production mode. Maybe `isLocalEnv` would be better.

xzf0587: Updated it. Thanks.

hund030 (approved):

jayzhang (approved):

PR #5944, opened 2022-08-30
chore(debug): add sso debug handler

kuojianlu:
- move writeManifestFileToBuildFolder from AadAppForTeamsImpl to
  AadAppManifestManager
- add sso debug handler in component, using AadAppManifestManager and
  AadAppClient to provision aad


swatDong (commented):
Better practice is to separate `plugins/resource/aad` changes into another PR, to be merged earlier...

Code change looks good to me, but test coverage blocks.

kuojianlu:
> Better practice is to separate `plugins/resource/aad` changes into another PR, to be merged earlier...
> 
> Code change looks good to me, but test coverage blocks.

Good idea! The SSO setup logic also needs Bowen's review, so I put aad plugin changes into this PR together for simplicity. Anyway, will follow the best practice in the future :).
Will fix the test coverage after Bowen saying yes on the SSO setup logic.

XiaofuHuang (approved):

packages/fx-core/src/component/debugHandler/sso.ts
```diff
@@ -0,0 +1,230 @@
...
+        // set objectId, clientId, clientSecret to state
+        envInfoV3.state[ComponentNames.AadApp].objectId = config.objectId;
+        envInfoV3.state[ComponentNames.AadApp].clientId = config.clientId;
+        envInfoV3.state[ComponentNames.AadApp].clientSecret = config.password;
+      }
+
+      // set applicationIdUris to state
+      let applicationIdUri = "api://";
+      if (ProjectSettingsHelper.includeFrontend(projectSettingsV3)) {
+        applicationIdUri += `localhost/${envInfoV3.state[ComponentNames.AadApp].clientId}`;
```
KennethBWSong, line 137: For tab + bot scenario, app id uri should be "api://localhost/botid-xxx", client id should not be included?

kuojianlu: Good catch! Fixed, thanks.

KennethBWSong (approved):

jayzhang (approved):

PR #5959, opened 2022-08-31
chore: update workflow ts template and add it to release pipeline

dooriya:


templates/bot/ts/workflow/src/cardActions/helloworldActionHandler.ts
```diff
@@ -5,10 +5,10 @@ import responseCard from "../adaptiveCards/doStuffActionResponse.json";
 import { CardData } from "../cardModels";
 
 /**
- * The `HelloWorldCardActionHandler` registers an action with the `TeamsFxBotActionHandler` and responds
+ * The `DoStuffCardActionHandler` registers an action with the `TeamsFxBotActionHandler` and responds
```
yukun-dong, line 8: File name should also be "doStuffCardActionHandler.ts".

dooriya: Good catch, will update

yukun-dong (approved):

PR #5963, opened 2022-08-31
chore(localdebug): show task error in output and notification

XiaofuHuang:
show task error in output and notification
image
image

Updated: 
image

### DIFF
--- a/packages/vscode-extension/src/debug/constants.ts
+++ b/packages/vscode-extension/src/debug/constants.ts
@@ -89,3 +89,5 @@ export class LaunchUrl {
 export const teamsAppIdPlaceholder = "${teamsAppId}";
 export const teamsAppInternalIdPlaceholder = "${teamsAppInternalId}";
 export const accountHintPlaceholder = "${account-hint}";
+
+export const outputPanelLink = "[output panel](command:fx-extension.showOutputChannel)";
--- a/packages/vscode-extension/src/debug/constants.ts
+++ b/packages/vscode-extension/src/debug/constants.ts
@@ -90,4 +90,4 @@ export const teamsAppIdPlaceholder = "${teamsAppId}";
 export const teamsAppInternalIdPlaceholder = "${teamsAppInternalId}";
 export const accountHintPlaceholder = "${account-hint}";
 
-export const outputPanelLink = "[output panel](command:fx-extension.showOutputChannel)";
+export const outputPanelCommand = "command:fx-extension.showOutputChannel";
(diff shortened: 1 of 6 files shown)


swatDong (dismissed):
Code change looks good but questions on UI:
- Is the task terminal shown by default? Or it's `reveal: never`?
- Any quick way to avoid showing `[output panel](command:...)`?

XiaofuHuang:
> Code change looks good but questions on UI:
> 
> * Is the task terminal shown by default? Or it's `reveal: never`?
> * Any quick way to avoid showing `[output panel](command:...)`?

When the task contains `reveal: never`, the terminal will not be shown, but user can see it by manually opening this terminal.

swatDong:
> > Code change looks good but questions on UI:
> > 
> > * Is the task terminal shown by default? Or it's `reveal: never`?
> > * Any quick way to avoid showing `[output panel](command:...)`?
> 
> When the task contains `reveal: never`, the terminal will not be shown, but user can see it by manually opening this terminal.

Just to ensure the default value is `never` so this error message is only shown on few cases. Keep consistency of all non-background debug tasks that output panel is the main window.

kuojianlu (approved):

XiaofuHuang:
> > > Code change looks good but questions on UI:
> > > 
> > > * Is the task terminal shown by default? Or it's `reveal: never`?
> > > * Any quick way to avoid showing `[output panel](command:...)`?
> > 
> > 
> > When the task contains `reveal: never`, the terminal will not be shown, but user can see it by manually opening this terminal.
> 
> Just to ensure the default value is `never` so this error message is only shown on few cases. Keep consistency of all non-background debug tasks that output panel is the main window.

Yes, `reveal` is not `never`  only in ngrok task .

swatDong (approved):

packages/vscode-extension/src/debug/constants.ts
```diff
@@ -89,3 +89,5 @@ export class LaunchUrl {
 export const teamsAppIdPlaceholder = "${teamsAppId}";
 export const teamsAppInternalIdPlaceholder = "${teamsAppInternalId}";
 export const accountHintPlaceholder = "${account-hint}";
+
+export const outputPanelLink = "[output panel](command:fx-extension.showOutputChannel)";
```
tecton, line 93: Does `[output panel]` need to be localized?

kuojianlu:
I am thinking do we need to show terminal when the task fails? Seems output channel is great enough.

XiaofuHuang:
> I am thinking do we ready need to show error messages in terminal when the task fails? Seems output channel is great enough.

If the user manually set  `reveal` / manually open the terminal, the user may be confused about why the task is failed. The message tell the user go to output panel to see the detail.

kuojianlu (approved):

swatDong (approved):

kuojianlu (approved):

packages/vscode-extension/src/debug/constants.ts
XiaofuHuang: Updated to localize output panel

tecton (approved):

PR #5976, opened 2022-08-31
chore: cleanup cicd

yuqizhou77:
- cleanup v2 code for cicd
- fix unit test and add more UT
- fix e2e test and remove v2
E2E TEST: https://github.com/OfficeDev/TeamsFx/actions/runs/2976740822


jayzhang (approved):

packages/fx-core/src/component/feature/cicd/cicd.ts
```diff
@@ -56,7 +54,7 @@ export class CICD {
   ): Promise<Result<undefined, FxError>> {
     const cicdImpl: CICDImpl = new CICDImpl();
     const envName = inputs.env || inputs[questionNames.Environment];
-    const res = await cicdImpl.addCICDWorkflows(context, inputs, envName);
+    const res = await cicdImpl.addCICDWorkflows(context, inputs, envName, context.envInfo);
```
IvanJobs, line 57: Do we still need to add `context.envInfo` as parameter when `context` is already passed as the first one?

IvanJobs (approved):

packages/fx-core/src/component/feature/cicd/cicd.ts
yuqizhou77: Fixed.

jayzhang (approved):

Alive-Fish (approved):

PR #5977, opened 2022-08-31
chore: deploy v3 (WIP)

Siglud:


packages/fx-core/src/component/deploy/deploy.ts
```diff
@@ -0,0 +1,43 @@
...
+    "scriptDriver",
+  ]);
+
+  private static DRIVER_INSTANCE = new Map<string, DeployDriver>([
+    ["azureAppServiceDriver", new AzureAppServiceDeployDriver()],
+    ["scriptDriver", new ScriptDeployDriver()],
+  ]);
+
+  static async runSteps(steps: DeployStep[], context: DeployContext) {
+    const noneExistsSteps = DeployStage.checkDriver(steps);
```
eriolchan, line 26: Maybe we could use invalidDriver for the name?

packages/fx-core/src/component/utils/fileOperation.ts
```diff
@@ -0,0 +1,115 @@
...
+
+      const entry = zip.getEntry(relativePath);
+      if (entry) {
+        // The header is an object, the ts declare of adm-zip is wrong.
+        const header = entry.header as any;
+        const mtime = header && header.time;
+        // Some files' mtime in node_modules are too old, which may be invalid,
+        // so we arbitrarily add a limitation to update this kind of files.
+        // If mtime is valid and the two mtime is same in two-seconds, we think the two are same file.
+        if (
```
eriolchan, line 55: I am not sure if we still need this logic for incremental zip optimization?

packages/fx-core/src/component/utils/common.ts
```diff
@@ -0,0 +1,17 @@
...
+import { exec } from "child_process";
+
+/**
+ * check parameter, throw error if value is null or undefined
+ * @param name parameter name
+ * @param value parameter value
+ */
+export function checkMissingDeployArgs<T>(name: string, value: T | null | undefined): T {
+  if (!value) {
+    throw DeployPrerequisiteError.somethingMissing(name);
```
eriolchan, line 14: Since it throws deploy specific error, is it better to be moved to component/deploy folder?

packages/fx-core/src/component/error/deployError.ts
```diff
@@ -0,0 +1,195 @@
...
+      ? messageParams
+        ? getLocalizedString(messageKey, ...messageParams)
+        : getLocalizedString(messageKey)
+      : undefined;
+  }
+
+  toFxError(): UserError {
+    if (this.errorType == "UserError") {
+      return new UserError({
+        source: "Deploy",
```
eriolchan, line 53: It seems "Deploy" can be a constant as DEPLOY_COMPONENT

packages/fx-core/src/component/error/deployError.ts
```diff
@@ -0,0 +1,195 @@
...
+  }
+
+  /**
+   * display messages will add 'Suggestions:' by default
+   * can be overwritten by subclasses
+   * @protected
+   */
+  protected toDisplayMessage(): string {
+    return this.displayMessage
+      ? getLocalizedString("plugins.bot.ErrorSuggestions", this.displayMessage)
```
eriolchan, line 79: I guess this is still WIP so it include "plugins.bot" string

packages/fx-core/src/component/error/deployError.ts
```diff
@@ -0,0 +1,195 @@
...
+    this.name = name;
+    this.helpLink = helpLink;
+    this.displayMessage = messageKey
+      ? messageParams
+        ? getLocalizedString(messageKey, ...messageParams)
+        : getLocalizedString(messageKey)
+      : undefined;
+  }
+
+  toFxError(): UserError {
```
eriolchan, line 50: Since the return type will be either UserError or SystemError, I guess you will update it later

packages/fx-core/src/component/error/deployError.ts
```diff
@@ -0,0 +1,195 @@
...
+    message: string,
+    statusCode: number,
+    messageParams?: string[],
+    helpLink?: string
+  ) {
+    super("UserError", name, message, messageParams, helpLink);
+    this.statusCode = statusCode;
+  }
+
+  static fromHttpError(name: string, message: string, error: unknown): DeployExternalApiCallError {
```
eriolchan, line 98: I guess the rest error is still WIP, will review it later when you finish

packages/fx-core/src/component/deploy/deployTypes.ts
```diff
@@ -0,0 +1,84 @@
...
+  scriptWorkingDir: string;
+  scriptPath: string;
+};
+
+export type BuildArgs = {
+  src: string;
+  buildCommand: string;
+};
+
+export type DeployContext = {
```
eriolchan, line 43: Since it is a common context about FxCore, I am not sure whether "FxContext" is a better name

packages/fx-core/src/component/deploy/azureDeployDriver.ts
```diff
@@ -0,0 +1,231 @@
...
+import { TokenCredentialsBase } from "@azure/ms-rest-nodeauth";
+import { BaseDeployDriver } from "./deployDriver";
+import { Base64 } from "js-base64";
+import * as appService from "@azure/arm-appservice";
+import { DeployConstant } from "../constant/deployConstant";
+import { default as axios } from "axios";
+import { AxiosZipDeployResult } from "./deployTypes";
+import { waitSeconds } from "../../common";
+
+export abstract class AzureDeployDriver extends BaseDeployDriver {
```
eriolchan, line 28: Overall the code looks clear and easy to understand, and I really appreciate your effort on the code/design quality.

packages/fx-core/src/component/deploy/azureDeployDriver.ts
```diff
@@ -0,0 +1,231 @@
...
+  ): Promise<void>;
+
+  /**
+   * check if resource id is legal and parse it
+   * @param resourceId deploy target
+   * @protected
+   */
+  protected parseResourceId(resourceId: string): AzureResourceInfo {
+    const result = resourceId.trim().match(this.pattern);
+    if (!result || result.length != 3) {
```
eriolchan, line 63: It is better to make 3 as a constant of meaningful variable name instead of a magic number

packages/fx-core/src/component/constant/deployConstant.ts
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+export class DeployConstant {
+  public static readonly DEPLOYMENT_TMP_FOLDER = ".deployment";
```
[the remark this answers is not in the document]
Siglud: This folder named .deployment is just compatible with the old version. If we change the folder name, the migration process will need to remove the old folder if it exists.

packages/fx-core/src/component/utils/fileOperation.ts
Siglud: Maybe we can remove the last trust time logic  and keep the last two-seconds logic

packages/fx-core/src/component/utils/common.ts
Siglud: Yes, maybe we should throw a more common error here.

packages/fx-core/src/component/deploy/deployTypes.ts
Siglud: I don't know if other components need to use the content of other Contexts, I've only picked some of the content needed for Deploy here. Because I want to try to control the number of parameters to avoid confusion

packages/fx-core/src/component/deploy/azureDeployDriver.ts
Siglud: Although this 3 looks like a magic number, it actually corresponds to three backreferences in the regular expression

jayzhang (dismissed):

packages/fx-core/src/component/deploy/azureDeployDriver.ts
```diff
@@ -0,0 +1,250 @@
...
+   * create azure deploy config for Azure Function and Azure App service
+   * @param azureResource azure resource info
+   * @param azureCredential user azure credential
+   * @protected
+   */
+  protected async createAzureDeployConfig(
+    azureResource: AzureResourceInfo,
+    azureCredential: TokenCredentialsBase
+  ): Promise<AzureUploadConfig> {
+    this.managementClient = new appService.WebSiteManagementClient(
```
1yefuwang1, line 193: It seems a little bit confusing to initialize `managementClient` here. 
Since it is a data member, I guess it should be reused. But every call to `createAzureDeployConfig` creates a new instance.

Siglud: Yes, it's required by Azure Function and Azure APP Service, but Azure Storage don't need this client. So, it can't be initialized in constructor

1yefuwang1: So maybe we can revisit the inheritance hierarchy and see whether we can make `managementClient` only known to Components that need it.
Or can we just make it a local variable, and construct it every time we need it. Drives won't be long running. Reusing an Azure Client doesn't bring much benefit.

1yefuwang1 (approved):

packages/fx-core/src/component/coordinator/lifecycleStage.ts
```diff
@@ -0,0 +1,24 @@
...
+
+import { DriverContext, Step } from "../interface/buildAndDeployArgs";
+import { PrerequisiteError } from "../error/componentError";
+import { StepDriver } from "../interface/stepDriver";
+import { Container } from "typedi";
+
+/**
+ * entrance of the all lifecycle stage
+ */
+class LifecycleStage {
```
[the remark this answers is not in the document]
hund030: Unused class?

packages/fx-core/src/component/deploy/azureAppServiceDeployDriver.ts
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { AzureResourceInfo, DeployStepArgs, DriverContext } from "../interface/buildAndDeployArgs";
+import { AzureDeployDriver } from "./azureDeployDriver";
+import { TokenCredentialsBase } from "@azure/ms-rest-nodeauth";
+import { StepDriver } from "../interface/stepDriver";
+import { Service } from "typedi";
+
+@Service("deploy/azureAppService")
```
hund030, line 10: We may need a constant for service name.

hund030 (approved):
packages/fx-core/src/component/error/errorConstant.ts
Empty file?

packages/fx-core/src/component/deploy/azureDeployDriver.ts
```diff
@@ -0,0 +1,255 @@
...
+  protected parseResourceId(resourceId: string): AzureResourceInfo {
+    const result = resourceId.trim().match(this.pattern);
+    if (!result || result.length != 4) {
+      throw PrerequisiteError.somethingIllegal("resourceId", "plugins.bot.InvalidValue", [
+        "resourceId",
+        resourceId,
+      ]);
+    }
+    return {
+      subscriptionId: resourceId[1].trim(),
```
IvanJobs, line 76: It seems you wanted to use `result` here but used `resourceId` instead. Right?

packages/fx-core/src/component/coordinator/lifecycleStage.ts
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { DriverContext, Step } from "../interface/buildAndDeployArgs";
+import { PrerequisiteError } from "../error/componentError";
+import { StepDriver } from "../interface/stepDriver";
+import { Container } from "typedi";
+
+/**
+ * entrance of the all lifecycle stage
```
IvanJobs, line 10: Nit: all lifecycle stage + s.

packages/fx-core/src/component/deploy/azureDeployDriver.ts
```diff
@@ -0,0 +1,255 @@
...
+    logger?: LogProvider
+  ): Promise<string> {
+    let res: AxiosZipDeployResult;
+    try {
+      res = await AzureDeployDriver.AXIOS_INSTANCE.post(zipDeployEndpoint, zipBuffer, config);
+    } catch (e) {
+      throw DeployExternalApiCallError.zipDeployError(e);
+    }
+
+    if (res?.status !== 200) {
```
IvanJobs, line 127: Nit: 200 can be replaced by some constant defined previously.

packages/fx-core/src/component/deploy/azureStorageDeployDriver.ts
```diff
@@ -0,0 +1,183 @@
...
+    );
+    const responses = await Promise.all(tasks);
+    const errorResponse = responses.find((res) => res.errorCode !== undefined);
+    if (errorResponse) {
+      throw DeployExternalApiCallError.uploadToStorageError(sourceFolder, errorResponse);
+    }
+    return;
+  }
+
+  private static async createContainerClient(
```
IvanJobs, line 91: Nit: Factory pattern to generate clients.

packages/fx-core/src/component/constant/commonConstant.ts
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+export class HttpStatusCode {
+  public static readonly HTTP_OK_RESPONSE_CODE = 200;
```
eriolchan, line 5: I think it is enough to just use keyword for the http status code, e.g. "HTTP_OK_RESPONSE_CODE" -> "OK", so that in code we can just use HttpStatusCode.OK for that value. You can find the variable name from this https://developer.mozilla.org/en-US/docs/Web/HTTP/Status, such as "OK", "Created", "Accepted".

Another question is what about using enum instead of constant?

packages/fx-core/src/component/coordinator/lifecycleStage.ts
```diff
@@ -0,0 +1,24 @@
...
+ */
+class LifecycleStage {
+  static async runSteps(steps: Step[], context: DriverContext) {
+    const drivers = steps.map((it) => Container.get(it.driver) as StepDriver);
+    if (drivers.find((it) => !it)) {
```
eriolchan, line 15: I don't understand the logic of this part of code? Can you explain more?

eriolchan: What the error handling logic when there is some driver name invalid? Will it break the whole execution?

packages/fx-core/src/component/deploy/azureDeployDriver.ts
eriolchan: Maybe you can use the constant you defined in HttpStatusCode here?

packages/fx-core/src/component/deploy/azureDeployDriver.ts
```diff
@@ -0,0 +1,255 @@
...
+    try {
+      listResponse = await this.managementClient.webApps.listPublishingCredentials(
+        azureResource.resourceGroupName,
+        azureResource.instanceId
+      );
+    } catch (e) {
+      throw DeployExternalApiCallError.listPublishingCredentialsError(e);
+    }
+
+    if (listResponse._response.status !== 200) {
```
eriolchan, line 211: Same here for using http status code constant/enum

packages/fx-core/src/component/deploy/azureFunctionDeployDriver.ts
```diff
@@ -0,0 +1,51 @@
...
+
+import { AzureResourceInfo, DeployStepArgs, DriverContext } from "../interface/buildAndDeployArgs";
+import { TokenCredentialsBase } from "@azure/ms-rest-nodeauth";
+import { AzureDeployDriver } from "./azureDeployDriver";
+import { DeployExternalApiCallError } from "../error/deployError";
+import { AxiosResponseWithStatusResult } from "../../common/azure-hosting/interfaces";
+import { Service } from "typedi";
+import { StepDriver } from "../interface/stepDriver";
+
+@Service("deploy/azureFunction")
```
eriolchan, line 12: The official name is azureFunctions, please add "s" for that name

packages/fx-core/src/component/deploy/azureStorageDeployDriver.ts
```diff
@@ -0,0 +1,183 @@
...
+        };
+        const client = containerClient.getBlockBlobClient(destFilePath);
+        tasks.push(client.uploadFile(filePath, options));
+      },
+      (itemPath: string) => {
+        return ig.test(path.relative(sourceFolder, itemPath)).unignored;
+      }
+    );
+    const responses = await Promise.all(tasks);
+    const errorResponse = responses.find((res) => res.errorCode !== undefined);
```
eriolchan, line 84: In TypeScript, it is recommend to use !res.errorCode instead of directly comparing with undefined, since there are other "empty value" besides undefined.

packages/fx-core/src/component/deploy/azureStorageDeployDriver.ts
```diff
@@ -0,0 +1,183 @@
...
+
+    const deleteJobs: Promise<BlobDeleteResponse>[] = [];
+    for await (const blob of client.listBlobsFlat()) {
+      if (AzureStorageDeployDriverImpl.isBlobFile(blob)) {
+        deleteJobs.push(client.deleteBlob(blob.name));
+      }
+    }
+
+    const responses = await Promise.all(deleteJobs);
+    const errorResponse = responses.find((res) => res.errorCode !== undefined);
```
eriolchan, line 170: Same here for comparing with undefined

packages/fx-core/src/component/deploy/npmBuildDriver.ts
```diff
@@ -0,0 +1,36 @@
...
+    const impl = new NpmBuildDriverImpl(args, context);
+    return await impl.run();
+  }
+}
+
+export class NpmBuildDriverImpl extends BaseBuildDriver {
+  private static readonly emptyMap = new Map<string, string>();
+  private static readonly NPM_BUILD_COMMAND_PREFIX = "npm";
+
+  async run(): Promise<Map<string, string>> {
```
eriolchan, line 24: It looks the run() looks similar between DotnetBuildDriver and NpmBuildDriver, is it possible to abstract more?

packages/fx-core/src/component/coordinator/lifecycleStage.ts
Siglud: Yea, It's just some sample codes without error handling.

packages/fx-core/src/component/deploy/azureDeployDriver.ts
Siglud: Ok ,changed

packages/fx-core/src/component/deploy/azureFunctionDeployDriver.ts
Siglud: Ok, but those driver names are still in draft. They will be modified after the project setting yaml file is finalized

packages/fx-core/src/component/deploy/azureStorageDeployDriver.ts
Siglud: Ok, changed to !res.errorCode

packages/fx-core/src/component/deploy/npmBuildDriver.ts
Siglud: Yes, or maybe it will both extend from the script driver. I'll make some changes when other people's code is ready

packages/fx-core/src/component/error/componentError.ts
```diff
@@ -0,0 +1,181 @@
...
+        ? getLocalizedString(messageKey, ...messageParams)
+        : getLocalizedString(messageKey)
+      : undefined;
+    this.detail = detail;
+  }
+
+  toFxError(): FxError {
+    if (this.errorType === "UserError") {
+      return new UserError({
+        source: "Deploy",
```
eriolchan, line 63: It seems that L72 use constant, but here is not, can we keep them consistent to use constant?

packages/fx-core/src/component/utils/common.ts
```diff
@@ -0,0 +1,57 @@
...
+export function checkMissingArgs<T>(name: string, value: T | null | undefined): T {
+  if (!value) {
+    throw PrerequisiteError.somethingMissing(name);
+  }
+  return value;
+}
+
+export function asOptional<T>(as: (s: unknown, key: string) => T) {
+  return function (s: unknown, key: string): T | undefined {
+    if (s === undefined) {
```
eriolchan, line 20: what about just using "if (!s)"?

packages/fx-core/src/component/utils/common.ts
```diff
@@ -0,0 +1,57 @@
...
+}
+
+type KeyValidators<T> = {
+  [P in keyof T]-?: (s: unknown, key: string) => T[P];
+};
+
+export function asFactory<T>(keyValidators: KeyValidators<T>) {
+  return function (data: unknown): T {
+    console.log(data);
+    if (typeof data === "object" && data !== null) {
```
eriolchan, line 48: Is it really "data !== null" since null is a different type from undefined

PR #5986, opened 2022-09-01
chore: add more m365 commands

swatDong:
- Add commands `m365 unacquire`, `m365 launchinfo`
- Move service call to *packageService.ts*
- Add unit tests


packages/cli/src/cmds/m365/m365.ts
```diff
@@ -59,96 +57,138 @@ async function getTokenAndUpn(): Promise<[string, string]> {
...
-    const content = new FormData();
-    content.append("package", data);
-    CLILogProvider.necessaryLog(LogLevel.Info, "Uploading package ...");
-    const uploadResponse = await instance.post(
-      "/dev/v1/users/packages",
-      content.getBuffer(),
-      content.getHeaders()
-    );
+class M365Sideloading extends YargsCommand {
+  public readonly commandHead = "sideloading";
```
kuojianlu, line 61: Have the command names already confirmed by PM?

packages/cli/src/cmds/m365/packageService.ts
```diff
@@ -0,0 +1,186 @@
...
+  private readonly axiosInstance;
+  public constructor(endpoint: string) {
+    this.axiosInstance = axios.create({
+      baseURL: endpoint,
+      timeout: 30000,
+    });
+  }
+
+  public async sideLoading(token: string, manifestPath: string): Promise<void> {
+    const data = await fs.readFile(manifestPath);
```
kuojianlu, line 25: What if `readFile` fails?

packages/cli/src/cmds/m365/m365.ts
swatDong: The `m365` command is for office-addins integration only. It has *preview* warning and is hidden from `--help`.
The names are confirmed by office-addins team. Actually they already used a private build with these names, and this PR is make these commands public for easier integration.

packages/cli/src/cmds/m365/packageService.ts
swatDong: Good catch... move `readFile` into `try...catch...`

kuojianlu (approved):

XiaofuHuang (approved):

PR #5995, opened 2022-09-01
feat(sdk): add partial authconfig to customConfig

frankqianms:
Add an optional Partial type to customConfig in constructor of TeamsFx, so that user can get some property prompts of configurations while creating an instance of TeamsFx class.


SLdragon:
Do not need to commit package-lock.json file

PR #6000, opened 2022-09-02
chore(debug): set envs in each debug handler

kuojianlu:


packages/fx-core/src/component/debugHandler/localEnvProvider.ts
```diff
@@ -0,0 +1,209 @@
...
+      SslKeyFile: "SSL_KEY_FILE",
+    },
+    teamsfx: {
+      ClientId: "REACT_APP_CLIENT_ID",
+      LoginUrl: "REACT_APP_START_LOGIN_PAGE_URL",
+      FuncEndpoint: "REACT_APP_FUNC_ENDPOINT",
+      FuncName: "REACT_APP_FUNC_NAME",
+    },
+  },
+  backend: {
```
swatDong, line 34: How to handle those place-holder envs like `SQL_xxx`?

kuojianlu: Should we put empty SQL envs in the file, or totally let user to manage the SQL envs?

swatDong: Better to have those when first time generating the file. We do let user to manage the values, but need place holders for keys.

kuojianlu: Added SQL env place holders when the project has SQL.

swatDong (approved):

PR #6015, opened 2022-09-05
fix(vsc): unnecessary work in non-teamsfx project

tecton:
E2E TEST: N/A

Related bug: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/15221284


packages/vscode-extension/src/extension.ts
```diff
@@ -814,13 +820,14 @@ async function runBackgroundAsyncTasks(
...
 
   await exp.initialize(context);
-  TreatmentVariableValue.previewTreeViewCommand = (await exp
-    .getExpService()
-    .getTreatmentVariableAsync(
-      TreatmentVariables.VSCodeConfig,
-      TreatmentVariables.PreviewTreeViewCommand,
-      true
-    )) as boolean | undefined;
+  TreatmentVariableValue.previewTreeViewCommand = true;
```
HuihuiWu-Microsoft, line 823: Any reason to make it always true to have preview Tree view command?

tecton: Thanks, fixed.

HuihuiWu-Microsoft (dismissed):

HuihuiWu-Microsoft (approved):

Issue #5978, opened 2022-08-31
Can't acquire SSO token for full trust apps


AndrewCraswell:
**Describe the bug**
The `TeamsUserCredential` is failing to return a valid SSO token when calling the `getSSOToken()` function. Our app is a Full Trust app, and when running in Ring 0, we've only been able to retrieve empty tokens (""). With further investigation, we believe the issue is caused here:
https://github.com/OfficeDev/TeamsFx/blob/dev/packages/sdk/src/credential/teamsUserCredential.browser.ts#L336

When `getSSOToken()` calls the `getAuthToken()` method from teams-js, an empty array of resources are being passed. Our thought was that this is sufficient, since our app manifest defines the resource needed (`api://{domain}/{clientId}`). However, it appears that Full Trust apps will not attempt to locate the resource from the manifest, and instead checks what resources are passed to the `getAuthToken()` function. We were able to find this line in the docs:

![image](https://user-images.githubusercontent.com/1908292/187781634-1db23fac-2a39-4c54-86c8-dcdd2809b5e9.png)

We only encountered this behavior after deploying the Full Trust app to Ring 0. Sideloaded apps still appear to be able to find the resource defined in the manifest. Unfortunately, since the `resources` argument is not exposed through the `getSSOToken()` function, we had to clone the `teamsfx` SDK library and make the change and bundle it directly into our app code to verify that passing the proper resource would resolve the issue.

A deeper investigation into the Teams Client code determined that Full Trust apps also require the resource to be expressed as `https://{domain}/{clientId}` rather than using the `api://` scheme. I haven't been able to find any documentation that describes this, so it was a complete surprise.

**Expected behavior**

1. Provide a mechanism for developers to pass the `resources` parameter to the `getSSOToken()` function to unblock Full Trust apps, and enable us to properly fetch SSO tokens.

2. Some documentation should probably be updated to identify that Full Trust apps need to define the resource using the `https://` schema rather than `api://`. I'm not really sure which team should do that, but it probably fits better here:
https://docs.microsoft.com/en-us/microsoftteams/platform/tabs/how-to/authentication/tab-sso-manifest

**VS Code Extension Information (please complete the following information):**
 - OS: Windows
 - Version: 4.0.5

**CLI Information (please complete the following information):**
 - OS: Windows
 - Version: ^1.0.0

**Additional context**

This issue is related to this bug I filed some time back:
https://github.com/OfficeDev/TeamsFx/issues/5672

[...]

PR #6019, opened 2022-09-05
feat: dastewa/full trust apps support pass 0

dastewa:
Add support for full trust apps which requires that a resource (app registration) uri be passed to the getAuthToken call of the teams js library. The aim of this pr was to make minimal changes to the interfaces, and so this list of resources are considered optional. 

Additional context: In the case of full trust apps, the app registration uri placed in the Teams app manifest is not used by the Teams client for token procurement. Instead, full trust apps need the app registration uri to be passed to the teams client when the getAuthToken method is called. This is currently the only way SSO will work for full trust apps - without this change, the Teams client will **not** return a token to the TeamsFx library.

Related issue: https://github.com/OfficeDev/TeamsFx/issues/5978


SLdragon (changes requested):
Looks good to me~
Tip: To pass Lint PR / check-format (pull_request), you should run `npm run setup` in the root folder of teamsfx, so that it will generate docs files for this change.

SLdragon:
to pass the Lint PR / pr-title check, the PR title should be semantic, such as: feat: Dastewa/full trust apps support pass 0.

dastewa:
@SLdragon - thank you so much for the feedback. I've made the updates to add the updated documentation as well as to update the title of the pr :).

SLdragon:
E2E TEST: N/A

SLdragon:
Lint PR / check-format (pull_request) still failed, seems missing this file: packages/sdk/review/teamsfx.api.md

SLdragon:
Also try to pull latest code from dev to make codecov happy

SLdragon (approved):

packages/sdk/src/credential/teamsUserCredential.browser.ts
```diff
@@ -185,7 +187,8 @@ export class TeamsUserCredential implements TokenCredential {
     options?: GetTokenOptions
   ): Promise<AccessToken | null> {
     validateScopesType(scopes);
-    const ssoToken = await this.getSSOToken();
+    const resources = (options as GetTeamsUserTokenOptions)?.resources;
```
tecton, line 190: I guess the signature of `getToken` is unchanged in order to align with `TokenCredential` interface.
Do we need to add comment and example for this API feature? It looks like developers would need to pass the `resources` value in `options` parameter if they want to use full trust Teams apps.

dastewa: @tecton that's a great call, I've updated the comments to include more information about this for full trust apps along with an example of how to use this. Is there anything else I need update for the "codecov/project" task to run?

tecton: Maybe you can merge 'dev' branch again to trigger the task.

SLdragon (approved):

SLdragon:
Seems codecov hangs, I will close this pr and reopen it to trigger it again.

PR #6025, opened 2022-09-06
chore(debug): add set up tab output

kuojianlu:
- refactor tab debug handler, splitting setUp into several steps for
  better output transparency

image

PR #6026, opened 2022-09-06
chore(localdebug): add local tunnel task

XiaofuHuang:
Detail: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/15310408
Success:
image
image

Failed:
image
image

TODO: use the handleDebugActions in pr https://github.com/OfficeDev/TeamsFx/pull/6025 to show summary.


packages/vscode-extension/package.json
```diff
@@ -927,6 +931,23 @@
                   }
                 },
                 "additionalProperties": false
+              },
+              {
+                "type": "object",
+                "required": [
+                  "configFile"
```
kuojianlu, line 938: Do we need to make the interface extensible for other tunnelling tools in the future?

XiaofuHuang: We can add `type`:  `basis` / `ngrok` in the future. As we discussed in the feature design, to avoid user confusion, there is no type attribute.

packages/vscode-extension/package.json
```diff
@@ -927,6 +931,23 @@
...
+                ],
+                "properties": {
+                  "configFile": {
+                    "type": "string",
+                    "default": "${teamsfx:workspaceFolder}/.fx/configs/ngrok.yml"
+                  },
+                  "binFolder": {
+                    "type": "string",
+                    "default": "${teamsfx:ngrokBinFolder}"
```
kuojianlu, line 947: What `default` means? If `binFolder` is empty in `task.json`, we will resolve it as `${teamsfx:ngrokBinFolder}`?

packages/vscode-extension/package.json
kuojianlu: Make sense. Then I wonder if `configFile` is also required for future tunnelling tools?

packages/fx-core/src/plugins/solution/fx-solution/debug/util/ngrok.ts
```diff
@@ -23,7 +23,7 @@ function delay(ms: number) {
   return new Promise((resolve) => setTimeout(resolve, ms));
 }
 
-export async function getNgrokHttpUrl(port: string | number): Promise<string | undefined> {
+export async function getNgrokHttpUrl(addr: string | number): Promise<string | undefined> {
```
kuojianlu, line 26: In "Set up bot" task, we do not know the bot port, how to get the ngrok url?

packages/vscode-extension/src/debug/taskTerminal/localTunnelTaskTerminal.ts
```diff
@@ -0,0 +1,264 @@
...
+    return { configFile: configFile, binFolder: binFolder };
+  }
+
+  private startNgrokChildProcess(
+    configFile: string,
+    binFolder?: string
+  ): Promise<Result<Void, FxError>> {
+    return new Promise<Result<Void, FxError>>((resolve, reject) => {
+      VsCodeLogInstance.info("Starting local tunnel task.");
+      const command = `ngrok start ${ngrokTunnelName} --config=${configFile} --log=stdout --log-format=logfmt`;
```
kuojianlu, line 94: How about also printing the command in terminal for better transparency?

packages/vscode-extension/package.json
XiaofuHuang: No, the default is only used in the auto prompt of vscode.
If the binFolder is empty in the `task.json`, the global ngrok will be used.

packages/fx-core/src/plugins/solution/fx-solution/debug/util/ngrok.ts
XiaofuHuang: Please use the static method LocalTunnelTaskTerminal.getNgrokEndpoint()

packages/vscode-extension/src/debug/taskTerminal/localTunnelTaskTerminal.ts
XiaofuHuang: Good idea!

packages/vscode-extension/package.json
kuojianlu: Then I think we should remove this `default` since the default means the actual behavior when the property is not provided in `tasks.json`.

Are the following mappings what we expect?
- `binFolder: "${teamsfx:ngrokBinFolder}"` -> resolve ngrok in ~/.fx
- `binFolder: "xxx"` -> xxx
- `binFolder: ""` -> resolve global ngrok
- not provided -> resolve global ngrok

packages/vscode-extension/package.json
XiaofuHuang: Now `configFile` is required, in the furture, it will according to the valiue of type.

packages/vscode-extension/package.json
XiaofuHuang: > Then I think we should remove this `default` since the default means the actual behavior when the property is not provided in `tasks.json`.
> 
> Are the following mappings what we expect?
> 
> * `binFolder: "${teamsfx:ngrokBinFolder}"` -> resolve ngrok in ~/.fx
> * `binFolder: "xxx"` -> xxx
> * `binFolder: ""` -> resolve global ngrok
> * not provided -> resolve global ngrok

Yes, the mapping is the expect. I will use another way to show the auto prompt .

packages/vscode-extension/package.json
kuojianlu: Got. Will there be any compatibility issue?

packages/vscode-extension/package.json
XiaofuHuang: Removed the default value.

packages/fx-core/src/plugins/solution/fx-solution/debug/util/ngrok.ts
kuojianlu: Sure. Just curious how this way will work for CLI or even VS in the future. Do you think it is necessary to have an implementation in fx-core to manage the tunnel, something like what @a1exwang did before.

packages/vscode-extension/package.json
XiaofuHuang: The tasks.json currently generated by this scheme will work on the new version of the Teams Toolkit. (The default value of `type` will be `ngrok`.) 

And the project generated in the newer version of  Teams Toolkit will not work in the older version of  Teams Toolkit.

XiaofuHuang: The schema may be contains
```json
                "if": {
                  "properties": {
                    "type": {
                      "contains": {
                        "const": "basis"
                      }
                    }
                  }
                },
                "then": {
                  "not": {
                    "required": [
                      "configFile"
                    ]
                  }
                },
                "else": {
                  "required": [
                    "configFile"
                  ]
                }
```

packages/fx-core/src/plugins/solution/fx-solution/debug/util/ngrok.ts
XiaofuHuang: Do we need to maintain ngrok command in cli? We don't need to maintain lifecycle for ngrok of cli. User can directly run the native ngrok command.

swatDong: I believe for CLI and VS we will need to maintain ngrok lifecycle in these cases:
- VS (teamsfx project) - one click F5
- CLI (teamsfx project) - one command preview
- CLI (existing app) - something like `teamsfx setup bot --tunnel` to automatically start tunneling and register bot

Anyway, these are coming features, so for now just ensure there will be no breaking change.

packages/vscode-extension/src/debug/taskTerminal/localTunnelTaskTerminal.ts
```diff
@@ -0,0 +1,265 @@
...
+                error?.message ?? "",
+                openTerminalCommand
+              )
+            )
+          )
+        );
+      });
+
+      this.childProc.on("close", (code: number) => {
+        if (code == 0) {
```
swatDong, line 146: ===

packages/vscode-extension/src/debug/taskTerminal/localTunnelTaskTerminal.ts
```diff
@@ -0,0 +1,265 @@
...
+  }
+
+  private startNgrokChildProcess(
+    configFile: string,
+    binFolder?: string
+  ): Promise<Result<Void, FxError>> {
+    return new Promise<Result<Void, FxError>>((resolve, reject) => {
+      VsCodeLogInstance.info("Starting local tunnel task.");
+      const command = `ngrok start ${ngrokTunnelName} --config=${configFile} --log=stdout --log-format=logfmt`;
+      this.writeEmitter.fire(`${command}\r\n\r\n`);
```
swatDong, line 95: Does `\r\n` work on all OS platforms? Or use `os.EOL` instead?

XiaofuHuang: It is according to the vscode API document https://code.visualstudio.com/api/references/vscode-api#Pseudoterminal
image

swatDong: Good to know!

swatDong (dismissed):

swatDong (approved):

kuojianlu (approved):

tecton (approved):

1yefuwang1 (approved):

PR #6040, opened 2022-09-07
chore: update manifest template for spfx and handle with solution nam…

HuihuiWu-Microsoft:
…e when not persisted

### DIFF
--- a/packages/fx-core/src/component/code/spfxTabCode.ts
+++ b/packages/fx-core/src/component/code/spfxTabCode.ts
@@ -86,10 +86,23 @@ export async function scaffoldSPFx(
     const webpartName = inputs[SPFXQuestionNames.webpart_name] as string;
     const framework = (inputs[SPFXQuestionNames.framework_type] as string) ?? undefined;
     let solutionName: string | undefined = undefined;
+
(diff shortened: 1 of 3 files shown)


packages/fx-core/src/component/code/spfxTabCode.ts
```diff
@@ -86,10 +86,23 @@ export async function scaffoldSPFx(
     const webpartName = inputs[SPFXQuestionNames.webpart_name] as string;
     const framework = (inputs[SPFXQuestionNames.framework_type] as string) ?? undefined;
     let solutionName: string | undefined = undefined;
+
     if (!isAddSpfx) {
       solutionName =
         ((context as ContextV3).projectSetting?.appName as string) ||
         ((context as PluginContext).projectSettings?.appName as string);
+    } else {
+      const yorcPath = path.join(inputs.projectPath, "SPFx", ".yo-rc.json");
```
HuihuiWu-Microsoft, line 95: Persist 'solutionName' in .yo-rc.json if not exist so we can add a webpart to current SPFx project.

nliu-ms (approved):

jayzhang (approved):

Issue #5815, opened 2022-08-16
Error, Request failed with status code 409


jasonjgeiger:
#Issue
Have been previously successful in running the Teams toolkit to locally develop our team's application. Late last week I attempted to run the app locally and was presented with the error noted below.

## Error
`[Error] - code:AppStudioPlugin.LocalAppIdUpdateFailed, message: Failed to update local debug Teams app in Developer Portal, due to API call to Developer Portal failed: Error, Request failed with status code 409, API name: create-app, X-Correlation-ID: undefined. This may come from some intermittent service error. Please wait for a few minutes and retry the current step.`

## Data returned with error
`data: {"teamsAppId":"REDACTED","tenantId":null,"ownerAadId":null,"userList":[{"tenantId":"REDACTED,"aadId":"REDACTED","displayName":"REDACTED","userPrincipalName":"REDACTED","isAdministrator":true}],"userAppSettings":null,"environments":null,"globalEnvironmentProperties":null,"createdAt":"0001-01-01T00:00:00+00:00","updatedAt":"0001-01-01T00:00:00+00:00","appId":"REDACTED","appName":"","appStudioVersion":"","version":"1.0.0","manifestVersion":"1.13",`

## Additional notes
- There are no notable changes to the app to indicating the app is not the culprit of the error. 
- Also a fellow teammate with a similar MacOS development setup is also encountering the same error.
- Waiting a few minutes and retrying results in the same error.

## System
| Software | Version |
| -- | -- |
| MacOS | 12.5 |
| VSCode | 1.70.1 |
|Teams Toolkit extension | 4.0.4 |

nliu-ms:
Thanks @jasonjgeiger for reporting this error. Usually the `409 status code` can be caused by duplicate Teams app id.

I'm not sure how the duplicate ids happen, but you may try this way:
1) Alter the Teams app id in `state.local.json`, e.g. here update 656ab44d-7255-4408-a662-1bc4e061654`d` to 656ab44d-7255-4408-a662-1bc4e061654`e`
![image](https://user-images.githubusercontent.com/71362691/185052358-bb6a1c7b-9c11-406d-a074-0d5c2edb89f3.png)
2) Do local debug again

This will create a new Teams app with your provided app id. 

Please let me know if it works for you, thanks!

nliu-ms:
[...]

PR #6044, opened 2022-09-07
fix: switch account under same tenant, will cause 409 conflict for creating Teams app

nliu-ms:
For local debug:
We will detect conflict app id and generate new app id

For provision:
We will guide user to visit help link
image


nliu-ms:
E2E TEST: https://github.com/OfficeDev/TeamsFx/actions/runs/3005742989

Many failure cases but not related.

jayzhang (approved):

nliu-ms:
fix https://github.com/OfficeDev/TeamsFx/issues/5815

PR #6053, opened 2022-09-08
chore: expose add feature in command palette

HuihuiWu-Microsoft:


packages/vscode-extension/src/extension.ts
```diff
@@ -659,6 +664,12 @@ async function initializeContextKey(isTeamsFxProject: boolean) {
     "fx-extension.canUpgradeToArmAndMultiEnv",
     await canUpgradeToArmAndMultiEnv(workspaceUri?.fsPath)
   );
+
+  await vscode.commands.executeCommand(
+    "setContext",
+    "fx-extension.canAddFeature",
+    !isSPFxProject || isSPFxMultiTabEnabled()
```
tecton, line 671: What features can users add if SPFx multi-tab is enabled?

HuihuiWu-Microsoft: For support multi-tab feature, 'SPFx tab' will be released to allow users to add. When this is enabled, there'll be two options:  
1. SPFx  tab
2. CI/CD

HuihuiWu-Microsoft: Updated, thanks!

tecton (approved):

PR #6056, opened 2022-09-08
chore: generator v3 

yukun-dong:


packages/fx-core/src/component/generator/scaffoldContext.ts
```diff
@@ -0,0 +1,23 @@
...
+import AdmZip from "adm-zip";
+import { ScaffoldAction } from "./scaffoldAction";
+
+export interface ScaffoldContext {
+  scenario: string;
+  destination: string;
+  logProvider: LogProvider;
+  zipUrl?: string;
+  zip?: AdmZip;
+  fallback?: boolean;
```
JerryYangKai, line 14: Do we need this parameter or just check the fallback url is none

packages/fx-core/src/component/generator/utils.ts
```diff
@@ -0,0 +1,116 @@
...
+import {
+  FetchSampleUrlWithTagError,
+  FetchZipFromUrlError,
+  TemplateZipFallbackError,
+  UnzipError,
+} from "./error";
+import { ScaffoldAction, ScaffoldActionName } from "./scaffoldAction";
+import { ScaffoldContext } from "./scaffoldContext";
+
+export async function fetchTemplateUrl(
```
JerryYangKai, line 24: This one and the following one looks the same, could we merge into one?

packages/fx-core/src/component/generator/constant.ts
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+export const templateTagListUrl =
+  "https://github.com/OfficeDev/TeamsFx/releases/download/template-tag-list/template-tags.txt";
+export const templateDownloadBaseUrl = "https://github.com/OfficeDev/TeamsFx/releases/download";
+
+export const sampleTagListUrl =
+  "https://github.com/OfficeDev/TeamsFx/releases/download/template-tag-list/template-tags.txt";
```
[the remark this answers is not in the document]
yukun-dong: Since the sample release plan is not finalized, I will leave a comment here.

packages/fx-core/src/component/generator/generator.ts
```diff
@@ -0,0 +1,78 @@
...
+
+export class Generator {
+  public static async generateFromTemplates(
+    templateName: string,
+    language: string,
+    destinationPath: string,
+    ctx: ContextV3
+  ): Promise<void> {
+    const appName = ctx.projectSetting?.appName;
+    const projectId = ctx.projectSetting?.projectId;
```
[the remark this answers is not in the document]
yukun-dong: > Here we miss a logic. Sample scenario we will just unzip a folder with a fix folder name ---- sample name. Create scenario we need to create a folder named 'project name', then place all the files in this folder

Good catch! This is fixed.

packages/fx-core/src/component/generator/scaffoldContext.ts
yukun-dong: You are right. Only fallbackUrl will suffice.

packages/fx-core/src/component/generator/utils.ts
yukun-dong: done.

packages/fx-core/src/component/generator/generator.ts
```diff
@@ -0,0 +1,78 @@
...
+      scenario: `${templateName}_${language}`,
+      destination: destinationPath,
+      logProvider: ctx.logProvider,
+      fileDataReplaceFn: genFileDataRenderReplaceFn({
+        appName: appName,
+        projectId: projectId,
+      }),
+      fileNameReplaceFn: genFileNameRenderReplaceFn({
+        appName: appName,
+        projectId: projectId,
```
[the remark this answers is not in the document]
yukun-dong: Indeed. I will remove it.

packages/fx-core/src/component/generator/generator.ts
```diff
@@ -0,0 +1,107 @@
...
+      },
+      ctx.templateReplace
+    );
+    const nameReplace = mergeReplaceMap(
+      {
+        appName: appName,
+      },
+      ctx.templateReplace
+    );
+    const destination = ctx.isAddFeature ? destinationPath : path.join(destinationPath, appName);
```
JerryYangKai, line 46: Do we actually create the folder called 'appName'?

yukun-dong: Yes. The create folder logic is at https://github.com/OfficeDev/TeamsFx/blob/5798adc3ec75b79d8672d503a02f5d680aed85cb/packages/fx-core/src/component/generator/utils.ts#L104

hund030, line 46: I think there is already a stage property in Inputs. Usually, we check `inputs.stage === Stage.create`.
I prefer split this to two interfaces instead of control its behavior by an inner flag.

packages/fx-core/src/component/generator/generateAction.ts
```diff
@@ -0,0 +1,105 @@
...
+} from "./constant";
+import { getTemplatesFolder } from "../../folder";
+import { MissKeyError } from "./error";
+
+export interface GenerateAction {
+  name: string;
+  run: (context: GenerateContext) => Promise<void>;
+}
+
+export enum GenerateActionName {
```
hund030, line 24: I don't get why do we rewrite actions? Can we just re-use the template utils?

packages/fx-core/src/component/generator/generator.ts
hund030: Imagine which do you prefer:
```
createProject() {
  context.isAddFeature = false;
  Generator.generateTemplates(..., context);
}
addxxFeature() {
  context.isAddFeature = true;
  Generator.generateTemplates(..., context);
}
```
or 
```
createProject() {
  Generator.generateTemplates(...);
}
addxxFeature() {
  Generator.addBuildingBlocks(...);
}
```

packages/fx-core/src/component/generator/generateAction.ts
yukun-dong: There are some small differences between template utils actions and new actions, so I think it's better to rewrite them.

packages/fx-core/src/component/generator/generator.ts
yukun-dong: Will sync with you about this offline.

packages/fx-core/src/component/generator/sample.ts
```diff
@@ -0,0 +1,49 @@
...
+  fullDescription: string;
+  tags: string[];
+  time: string;
+  configuration: string;
+  link?: string;
+  suggested: boolean;
+  relativePath?: string;
+}
+
+export interface SampleCollection {
```
eriolchan, line 16: Just curious why we need the interface of SampleCollection, it looks a little duplicated with SampleInfo[]

packages/fx-core/src/component/generator/utils.ts
```diff
@@ -0,0 +1,181 @@
...
+
+export function renderTemplateFileName(
+  fileName: string,
+  fileData: Buffer,
+  variables?: { [key: string]: string }
+): string {
+  return Mustache.render(fileName, variables);
+}
+
+export function genFileDataRenderReplaceFn(variables: { [key: string]: string }) {
```
eriolchan, line 63: In this class, why do we need to have some unnecessary wrapper like this. I'm a little confused on so many utils class, can you show me more high level about the relationships between the core Generator logic and those kinds of utils?

packages/fx-core/src/component/generator/utils.ts
```diff
@@ -0,0 +1,181 @@
...
+    }
+  }
+  return sampleDestination;
+}
+
+export function getSampleInfoFromName(sampleName: string): SampleInfo {
+  const samples = sampleProvider.SampleCollection.samples.filter(
+    (sample) => sample.id.toLowerCase() === sampleName.toLowerCase()
+  );
+  if (samples.length == 0) {
```
eriolchan, line 129: if (!samples.length)

packages/fx-core/src/component/generator/utils.ts
```diff
@@ -0,0 +1,181 @@
...
+      : rawEntryData;
+
+    const filePath: string = path.join(dstPath, entryName);
+    const dirPath: string = path.dirname(filePath);
+    await fs.ensureDir(dirPath);
+    await fs.writeFile(filePath, entryData);
+  }
+}
+
+export async function getValidSampleDestination(
```
eriolchan, line 108: although we may copy this code part from sample gallery, since it will be code of Generator, we could write it in a better way with high code quality.

packages/fx-core/src/component/generator/utils.ts
```diff
@@ -0,0 +1,181 @@
...
+    throw Error(`invalid sample name: '${sampleName}'`);
+  }
+  return samples[0];
+}
+
+export function mergeReplaceMap(
+  obj: { [key: string]: string },
+  obj2?: { [key: string]: string }
+): { [key: string]: string } {
+  const result = { ...obj };
```
eriolchan, line 139: why don't just as this https://bobbyhadz.com/blog/typescript-merge-objects
```
{...obj, ...obj2}
```

packages/fx-core/src/component/generator/generator.ts
```diff
@@ -0,0 +1,141 @@
...
+export class Generator {
+  public static async generateTemplate(
+    templateName: string,
+    language: string,
+    destinationPath: string,
+    ctx: ContextV3
+  ): Promise<void> {
+    const appName = ctx.projectSetting?.appName;
+    const projectId = ctx.projectSetting?.projectId;
+    const dataReplace = mergeReplaceMap(
```
eriolchan, line 33: the only difference between dataReplace and nameReplace is projectId, why can't they share the same map?

jayzhang (dismissed):

JerryYangKai (approved):

eriolchan (dismissed):

packages/fx-core/src/component/generator/utils.ts
```diff
@@ -98,83 +193,79 @@ export async function unzip(
...
+export function renderTemplateFileName(
+  fileName: string,
+  fileData: Buffer,
+  variables?: { [key: string]: string }
+): string {
+  return Mustache.render(fileName, variables).replace(templateFileExt, "");
 }
 
 export function getSampleInfoFromName(sampleName: string): SampleInfo {
   const samples = sampleProvider.SampleCollection.samples.filter(
```
hund030, line 231: NIT: maybe `find` is better than `filter`?

packages/fx-core/src/component/generator/utils.ts
```diff
@@ -98,83 +193,79 @@ export async function unzip(
...
-  action: GenerateAction,
-  context: GenerateContext,
+  action: GeneratorAction,
+  context: GeneratorContext,
   error: Error
 ) {
   switch (action.name) {
-    case GenerateActionName.FetchSampleUrlWithTag:
+    case GeneratorActionName.FetchSampleUrlWithTag:
       throw new FetchSampleUrlWithTagError();
```
hund030, line 265: It's better to log inner error.

packages/fx-core/tests/component/generator/generator.test.ts
```diff
@@ -45,39 +55,110 @@ describe("Generator utils", () => {
...
+  it("template", async () => {
+    const templateName = "bot";
+    const language = "ts";
+    sandbox
+      .stub(fetchTemplateUrlWithTagAction, "run")
+      .callsFake(async (context: GeneratorContext) => {
+        context.zipUrl =
+          "https://github.com/hund030/TemplatePackerDemo/releases/download/templates%400.1.0/bot_notification_ts_function_http.zip";
+      });
+    await Generator.generateTemplate(templateName, language, tmpDir, context);
```
hund030, line 160: Seems this test downloads the template and unzip it? It's better to mock the api call.

hund030 (approved):

hund030:
Overll looks good to me. You can merge this first and fix some comments later.

jayzhang (approved):

PR #6078, opened 2022-09-13
perf(sso-command): not export DefaultBotSsoExecutionActivityHandler, interface, code owner update

SLdragon:
- Update logic inside CommandMiddleware to make it more clear
- Update code owner to make sure it not show errors
- Not export DefaultBotSsoExecutionActivityHandler based on Chaoyi's suggestion
- Update interface parameter name


packages/sdk/src/conversation/interface.ts
```diff
@@ -436,7 +436,7 @@ export interface ConversationOptions {
  *    ...
  * });
  * ```
- * For details information about how to implement a BotSsoExecutionActivityHandler, please refer {@link DefaultBotSsoExecutionActivityHandler} class source code.
+ * For details information about how to implement a BotSsoExecutionActivityHandler, please refer DefaultBotSsoExecutionActivityHandler class source code.
```
dooriya, line 439: Maybe a link here will be more useful to discover the code.

SLdragon: Ah, thanks for your suggestion

packages/sdk/src/conversation/middlewares/commandMiddleware.ts
```diff
@@ -62,36 +62,39 @@ export class CommandResponseMiddleware implements Middleware {
...
-              message
-            );
-
-            await this.processResponse(context, response);
+            break;
           }
-          break;
         }
       }
     } else {
```
dooriya, line 100: Do we need the `else` code block here if the command is responded in a non-sso command handler?

SLdragon: This `else` condition is for none message activity type, such as `signin/tokenExchange` which will trigger login process

SLdragon: I think we do not need to process the command twice, if non-sso command has already triggered, then it will break and stop trigger

dooriya: Oh ~ I see, I mistook the line indents, sorry for that.

dooriya (dismissed):

packages/sdk/src/conversation/middlewares/commandMiddleware.ts
```diff
@@ -62,36 +62,39 @@ export class CommandResponseMiddleware implements Middleware {
...
-    this.commandHandlers.push(ssoHandler);
     this.hasSsoCommand = true;
   }
 
   public async onTurn(context: TurnContext, next: () => Promise<void>): Promise<void> {
     if (context.activity.type === ActivityTypes.Message) {
       // Invoke corresponding command handler for the command response
       const commandText = this.getActivityText(context.activity);
-
+      let alreadyProcessed = false;
```
blackchoey, line 72: Just realized one thing: there may be cases that a command text can match both a commandHandler and a ssoCommandHandler.
What's our attitude towards this scenario? An easier way is that we assume there shouldn't be intersection and document this for users. But I'm not sure whether we can make this assumption.

SLdragon: From original design command bot will stop processing handlers when the first command handler is matched. So we can give some comments to tell user about this

packages/sdk/src/conversation/interface.ts
SLdragon: Updated

packages/sdk/src/conversation/middlewares/commandMiddleware.ts
SLdragon: @MuyangAmigo how about your suggestions?

swatDong (dismissed):

packages/sdk/src/conversation/middlewares/commandMiddleware.ts
MuyangAmigo: For now we can take the assumption that developers want to build SSO/NON-SSO commands separately just not to over design at the beginning. We can always iterate this over when there's customer signal or specific asks come along.

dooriya (approved):

swatDong (approved):

dooriya (approved):

packages/sdk/src/conversation/interface.ts
```diff
@@ -436,7 +436,7 @@ export interface ConversationOptions {
  *    ...
  * });
  * ```
- * For details information about how to implement a BotSsoExecutionActivityHandler, please refer DefaultBotSsoExecutionActivityHandler class source code.
+ * For details information about how to implement a BotSsoExecutionActivityHandler, please refer DefaultBotSsoExecutionActivityHandler class source code: https://github.com/OfficeDev/TeamsFx/blob/dev/packages/sdk/src/conversation/sso/defaultBotSsoExecutionActivityHandler.ts
```
blackchoey, line 439: Use an aka link here in case the url is not valid any more?

SLdragon: Sure, will update in another PR, and to make sure this PR can catch up today's release~

wenytang-ms (approved):

blackchoey (approved):

packages/sdk/src/conversation/middlewares/commandMiddleware.ts
blackchoey: Thanks, resolving this comment.

PR #6079, opened 2022-09-13
fix(localdebug): support relative path in npm install task

XiaofuHuang:
Support relative path in `cwd` of npm install task

E2E TEST: none

### DIFF
--- a/packages/vscode-extension/src/debug/taskTerminal/npmInstallTaskTerminal.ts
+++ b/packages/vscode-extension/src/debug/taskTerminal/npmInstallTaskTerminal.ts
@@ -37,13 +37,11 @@ export class NpmInstallTaskTerminal extends BaseTaskTerminal {
         throw BaseTaskTerminal.taskDefinitionError("cwd");
       }
 
-      let resolvedCwd = BaseTaskTerminal.resolveTeamsFxVariables(projectOption.cwd);
-      if (!path.isAbsolute(resolvedCwd)) {
-        resolvedCwd = path.join(globalVariables.workspaceUri?.fsPath ?? "", resolvedCwd);
-      }
-
       return {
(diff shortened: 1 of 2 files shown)


packages/vscode-extension/src/debug/taskTerminal/npmInstallTaskTerminal.ts
```diff
@@ -36,8 +37,13 @@ export class NpmInstallTaskTerminal extends BaseTaskTerminal {
...
       }
 
+      let resolvedCwd = BaseTaskTerminal.resolveTeamsFxVariables(projectOption.cwd);
+      if (!path.isAbsolute(resolvedCwd)) {
+        resolvedCwd = path.join(globalVariables.workspaceUri?.fsPath ?? "", resolvedCwd);
+      }
+
       return {
-        cwd: path.normalize(BaseTaskTerminal.resolveTeamsFxVariables(projectOption.cwd)),
+        cwd: path.normalize(resolvedCwd),
```
swatDong, line 46: Why not just using `path.resolve()`?

XiaofuHuang: Good idea, use `path.resolve` instead.

swatDong (approved):

kuojianlu:
I wonder if it is a good practice to allow users to use relative path. I have this doubt because the `cwd` in shell command is absolute path.

swatDong:
> I wonder if it is a good practice to allow users to use relative path. I have this doubt because the `cwd` in shell command is absolute path.

I tried relative path and it works:
```
        {
            "label": "try",
            "type": "shell",
            "command": "pwd",
            "options": {
                "cwd": "tabs"
            }
        }
```

kuojianlu (approved):

PR #6092, opened 2022-09-14
chore: show failed checkers in precheck error display message

swatDong:
Improve precheck error message for https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/15525708/

Show the failed checkers without switching to output.
image


kuojianlu (approved):

a1exwang (approved):

XiaofuHuang (approved):

XiaofuHuang:
Do we need to make message same to the display message?
image

swatDong:
> Do we need to make message same to the display message? image

This is to fix the pop-up error dialog that user can have a quick sense what's going wrong. But in output window there are already enough logs so no need to change. (Just find displayMessage is shown on error dialog only... so use it)

HuihuiWu-Microsoft (approved):

PR #6095, opened 2022-09-14
fix: app id conflict with published app

nliu-ms:
Have a clear error message, rather than raw error.
image


nliu-ms:
E2E TEST: N/A

jayzhang (approved):

PR #6101, opened 2022-09-15
chore: teamsfx add spfx-tab

Alive-Fish:
image

### DIFF
--- a/packages/cli/src/cmds/add.ts
+++ b/packages/cli/src/cmds/add.ts
@@ -29,16 +29,16 @@ import {
   CapabilityAddCommandAndResponse,
   CapabilityAddMessageExtension,
   CapabilityAddNotification,
+  CapabilityAddSPFxTab,
   CapabilityAddSSOTab,
   CapabilityAddTab,
   CapabilityAddWorkflow,
 } from "./capability";
 import {
-  isBotNotificationEnabled,
   isAadManifestEnabled,
   isApiConnectEnabled,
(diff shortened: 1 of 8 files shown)


packages/cli/src/cmds/add.ts
```diff
@@ -224,25 +224,21 @@ export default class Add extends YargsCommand {
...
-          new CapabilityAddTab(),
-          new CapabilityAddBot(),
-          new CapabilityAddMessageExtension(),
+    // Category 1: Add Teams Capability
+    new CapabilityAddNotification(),
+    new CapabilityAddCommandAndResponse(),
+    ...(isWorkflowBotEnabled() ? [new CapabilityAddWorkflow()] : []),
+    new CapabilityAddSSOTab(),
+    new CapabilityAddTab(),
+    ...(isSPFxMultiTabEnabled() ? [new CapabilityAddSPFxTab()] : []),
```
HuihuiWu-Microsoft, line 233: What will happen if user execute add spfx under azure project?

HuihuiWu-Microsoft (approved):

packages/cli/src/cmds/add.ts
Alive-Fish: Open a bug because the behaviors to add features are also abnormal. https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/15554342

jayzhang (approved):

chagong (approved):

PR #6121, opened 2022-09-16
chore(localdebug): add tunnel name to local tunnel task and get endpoint from api after timeout

XiaofuHuang:
1. Add optional arg `tunnelName`, the default value is `bot`.
2. Get the ngrok endpoint from api when the local tunnel task cannot find the endpoint from output after 1min.


packages/vscode-extension/src/debug/taskTerminal/localTunnelTaskTerminal.ts
```diff
@@ -195,19 +202,80 @@ export class LocalTunnelTaskTerminal extends BaseTaskTerminal {
...
     if (matches && matches?.length > 2) {
-      return { src: matches[1], dist: matches[2] };
+      const ngrokTunnelInfo = { src: matches[1], dist: matches[2] };
+      this.isOutputSummary = true;
+      this.status.endpoint = ngrokTunnelInfo;
+      this.outputSuccessSummary(ngrokTunnelInfo);
+    }
+  }
+
+  private async saveNgrokEndpointFromApi(
```
swatDong, line 232: Which exact case that log fails but api works?

XiaofuHuang: We use regex `obj=tunnels name=${tunnelName} addr=(?<src>.*) url=(?<endpoint>.*)` to find the endpoint. 

The log may be changed in some version of ngrok, (Use global ngrok .......)

swatDong (approved):

swatDong (approved):

tecton (approved):

PR #6126, opened 2022-09-17
fix: bot runtime error caused by latest adaptivecards npm pacckage

dooriya:
Scaffold and run all bot project will got the following runtime error:
image

This is caused by the latest `adaptivecards` npm package which is just released on **9/17**.

To fix this issue, update the `adaptivecards-tools` sdk to keep using `adaptivecards@2.10.0`

E2E TEST: N/A


dooriya:
will fix in main

PR #6133, opened 2022-09-19
docs: add template builder section in template contribution guide [REVIEW ONLY]

hund030:


templates/CONTRIBUTING.md
```diff
@@ -1,3 +1,57 @@
...
+1. Add your changes to existing asset.
+1. You have no need to figure out which template should be updated. Re-generate all templates:
+    ```
+    > cd src
+    > npm run generate:all
+    ```
+1. Commit both your asset changes and template changes.
+
+
 ## How to scaffold from pre-release templates?
```
eriolchan, line 55: Is that better for "How to test template for pre-release version"?

eriolchan: Do we need case of "How to test template in local debug?"

templates/CONTRIBUTING.md
```diff
@@ -1,3 +1,57 @@
+# Contributing
+Welcome and thank you for your interest in contributing to TeamsFx templates! Before contributing to this project, please review this document for policies and procedures which will ease the contribution and review process for everyone. If you have questions, please raise your issue on github.
+
+## Built the project
```
yukun-dong, line 4: Maybe "Build the project" ?

templates/CONTRIBUTING.md
```diff
@@ -1,3 +1,57 @@
...
+
+1. Clone this repo locally. (git clone https://github.com/OfficeDev/TeamsFx.git)
+1. Open a terminal and move into your local copy. (cd TeamsFx)
```
yukun-dong, line 7: The numbers are all 1.

templates/CONTRIBUTING.md
```diff
@@ -1,3 +1,57 @@
...
+    > npm run generate ../definitions/<your-definition-file>
+    ```
+1. Commit both your definition file and your template source code.
+
+## How to synchronize template asset changes to templates?
+
+Assuming every template assets are referred by at least one template. When updating a template asset, you must synchronize the changes to those templates that refer to this asset.
+
+1. Add your changes to existing asset.
+1. You have no need to figure out which template should be updated. Re-generate all templates:
```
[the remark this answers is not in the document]
hund030: In my mind Template Builder does not **figure out** related templates. The explanation is a little different from the fact. But it's a good point. Maybe template builder should only update those templates who need to be updated.

templates/CONTRIBUTING.md
hund030: We have "How to debug template?" section.

templates/CONTRIBUTING.md
hund030: Resolved.

templates/CONTRIBUTING.md
hund030: It's ok and has better maintainability. Markdown will automatically order it.

Issue #6127, opened 2022-09-17
Failed to start bot due to error caused by the `adaptivecards` dependency


dooriya:
**Describe the bug**
Scaffold bot project with Teams toolkit, run the bot will got the following error:
![image](https://user-images.githubusercontent.com/10163840/190853901-89cddb23-e965-4b64-95b5-a8afb21d544e.png)

**To Reproduce**
1. create a bot (notification bot /command bot/ bot)
2. local debug the bot

**How to fix**
For users already have a bot project created by toolkit:
1. Open terminal, go to the `bot` folder
2. Re-install `@microsoft/teamsfx` npm package by: `npm install @microsoft/teamsfx --force`
3. F5 your bot again

For Teams Toolkit:
- **Option 1**: wait for AdaptiveCard to fix this issue in the latest version, already tracked in this issue: https://github.com/microsoft/AdaptiveCards/issues/7904.
- **Option 2**: update `bot/package.json` in all bot project templates to use `"adaptivecards": "~2.10.0"`, including command bot, notification bot, default bot.
- **Option 3**: update `package.json` in the`teamsfx` and `adaptivecards-tools` sdk to use `"adaptivecards": "~2.10.0"`.

PR #6137, opened 2022-09-19
fix(template): hotfix bot runtime error due to adaptivecards dependency

dooriya:
Fix issue: [Failed to start bot due to error caused by the adaptivecards dependency](https://github.com/OfficeDev/TeamsFx/issues/6127)

E2E TEST: N/A

### DIFF
--- a/templates/bot/js/command-and-response/package.json
+++ b/templates/bot/js/command-and-response/package.json
@@ -17,6 +17,7 @@
         "url": "https://github.com"
     },
     "dependencies": {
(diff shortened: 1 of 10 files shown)


templates/bot/js/command-and-response/package.json
```diff
@@ -17,6 +17,7 @@
         "url": "https://github.com"
     },
     "dependencies": {
+        "adaptivecards": "~2.10.0",
```
swatDong, line 20: Package order - move `adaptivecards` after `@xxx`

PR #6150, opened 2022-09-20
feat: support both spfx 1.15 and 1.16 beta 

HuihuiWu-Microsoft:
Since SPFx v1.16 will go public by Ignite, we'd like to support latest version of SPFx. However, they won't GA by December. 

So according to our discussion, we decide to support both 1.15 and 1.16 beta version in our TTK. We'll use a VSC setting to allow user to switch the version he wants.

E2E TEST: will add later


HuihuiWu-Microsoft:
@xu-hong Please help review the documentation part. Thanks!

jayzhang (approved):

tecton (approved):

PR #6168, opened 2022-09-21
chore: add sample for message extension SSO

wenytang-ms:
image


packages/fx-core/src/common/samples-config.json
```diff
@@ -205,6 +205,18 @@
...
+        {
+            "id": "query-org-user-with-message-extension-sso",
+            "title": "Query with Message Extension SSO",
+            "shortDescription": "Query Org User with Message Extension SSO",
+            "fullDescription": "This is a simple search-based message extension app demonstrating how to integrate with Bot Framework SDK, TeamsFx SDK and Microsoft Graph API to implement a feature that queries organization's user with single sign on (SSO) capability.",
+            "tags": ["Message extension", "SSO"],
+            "time": "5min to run",
+            "configuration": "Ready for local debug",
+            "suggested": false,
+            "url": "https://github.com/OfficeDev/TeamsFx-samples/tree/dev/",
```
wenytang-ms, line 218: will remove the url and packageLink property after v1.0.0 tag contains this sample

HuihuiWu-Microsoft:
Suggested image size is 1600*920 or you should use the same ratio at least. Also, can you provide a screenshot/gif of sample gallery to make sure the sample app doesn't break responsive layout?

packages/fx-core/src/common/samples-config.json
```diff
@@ -205,6 +205,18 @@
             "time": "5min to run",
             "configuration": "Ready for local debug",
             "suggested": false
+        },
+        {
+            "id": "query-org-user-with-message-extension-sso",
+            "title": "Query with Message Extension SSO",
```
MuyangAmigo, line 211: I suggest the title to be: `Org User Search Connector` similar to `NPM Search Connector` we had today.

wenytang-ms: thanks
updated!

HuihuiWu-Microsoft (approved):

MuyangAmigo (approved):

PR #6198, opened 2022-09-23
chore(debug): improve local tunnel task

kuojianlu:
- rename reuse to keepAlive
- do not scaffold useGlobalNgrok
- remove resolving bin folder logic
- fix beginsPattern

image


packages/vscode-extension/package.json
```diff
@@ -1141,7 +1141,7 @@
           }
         ],
         "background": {
-          "beginsPattern": "Local tunnel service is started successfully",
+          "beginsPattern": "Forwarding URL https?://.* to https?://.*",
```
kuojianlu, line 1144: May bad in previous PR. `beginsPattern` and `endsPattern` cannot be the same, otherwise the debugging will be blocked on this background task.

swatDong: Could it be just .*?

swatDong (approved):

packages/vscode-extension/package.json
kuojianlu: No, have to be a pattern after the ngrok starts successfully, otherwise the next task will still be run if starting ngrok fails.

HuihuiWu-Microsoft (approved):

PR #6209, opened 2022-09-26
chore(localdebug): add spfx v1.16 node checker

XiaofuHuang:
1. add spfx v1.16 node checker
2. remove func node checker
3. Scaffold transparency tasks for SPFx project
 
image

### DIFF
--- a/packages/fx-core/src/common/local/localEnvManager.ts
+++ b/packages/fx-core/src/common/local/localEnvManager.ts
@@ -59,9 +59,8 @@ export class LocalEnvManager {
 
     // NodeJS
     if (isSPFx) {
+      // TODO: add spfx 1.16.0 node
       depsTypes.push(DepsType.SpfxNode);
-    } else if (includeBackend || includeFuncHostedBot) {
-      depsTypes.push(DepsType.FunctionNode);
     } else {
       depsTypes.push(DepsType.AzureNode);
     }
--- a/packages/fx-core/src/common/local/localEnvManager.ts
+++ b/packages/fx-core/src/common/local/localEnvManager.ts
@@ -22,7 +22,7 @@ import * as localStateHelper from "./localStateHelper";
 import { LocalSettingsProvider } from "../localSettingsProvider";
 import { getNpmInstallLogInfo, NpmInstallLogInfo } from "./npmLogHelper";
 import { getPortsInUse, getPortsFromProject } from "./portChecker";
-import { waitSeconds } from "../tools";
+import { getAppSPFxVersion, waitSeconds } from "../tools";
 import { LocalCrypto } from "../../core/crypto";
 import { CoreSource, ReadFileError, NgrokConfigError } from "../../core/error";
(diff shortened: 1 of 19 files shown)


packages/vscode-extension/src/handlers.ts
```diff
@@ -1644,22 +1644,8 @@ export async function openReadMeHandler(args: any[]) {
 
 export async function promptSPFxUpgrade() {
   if (globalVariables.isSPFxProject) {
-    let projectSPFxVersion = null;
     // eslint-disable-next-line @typescript-eslint/no-non-null-assertion, @typescript-eslint/no-non-null-asserted-optional-chain
-    const yoInfoPath = path.join(globalVariables.workspaceUri?.fsPath!, "SPFx", ".yo-rc.json");
```
swatDong, line 1649: Any compatibility issue if removing this? E.g., opening old project?

HuihuiWu-Microsoft: The logic doesn't change, I just moved it to util function so prerequisite checker can reuse.

swatDong (dismissed):

packages/fx-core/src/common/local/localEnvManager.ts
```diff
@@ -59,9 +62,11 @@ export class LocalEnvManager {
 
     // NodeJS
     if (isSPFx) {
-      depsTypes.push(DepsType.SpfxNode);
-    } else if (includeBackend || includeFuncHostedBot) {
-      depsTypes.push(DepsType.FunctionNode);
+      if ((await getAppSPFxVersion(projectPath))?.startsWith("1.16.0")) {
```
tecton, line 65: Does it mean the supported version in future is 1.16.0-x instead of 1.16.x?

tecton (commented):
In the screenshot, it seems that the error message should be used in VS Code prompt is printed because there's a markdown link. If not related to this PR, we can fix it later.

packages/fx-core/src/common/local/localEnvManager.ts
XiaofuHuang: @HuihuiWu-Microsoft  As we discussed offline, after SPFx 1.16.0 GA, Teams Toolkit will only support node.js 16. The old SPFx node checker will be retired.

HuihuiWu-Microsoft: Yes, for this release, we'll support SPFx v1.16.0-beta.1

tecton (approved):

packages/fx-core/src/common/local/localEnvManager.ts
HuihuiWu-Microsoft: @XiaofuHuang Right, once SPFx v1.16.0 released, we'll only support SPFx v1.16.0(Node.js v16 is supported by SPFx v1.16.0) and there'll be only one node checker for SPFx

swatDong (approved):

jayzhang (approved):

PR #6211, opened 2022-09-26
chore(debug): reorder tasks to make output/terminal switching smooth

kuojianlu:
- move installing azure functions binding extensions task to the dependency of starting backend task
- do not reveal npm install task terminals
- do not reveal local tunnel task terminal
- hide task echo and reusing message


packages/fx-core/src/plugins/solution/fx-solution/debug/util/tasksTransparency.ts
```diff
@@ -82,6 +81,7 @@ export function generateTasks(
    *   - Start services
    *   - Start frontend
    *   - Start backend
+   *   - Install Azure Functions binding extensions
```
swatDong, line 84: Does `watch` need to depend on `install extensions`? Or, will `install extensions` CRUDs watched files?

packages/vscode-extension/src/debug/prerequisitesHandler.ts
```diff
@@ -1042,20 +1042,20 @@ function checkNpmInstall(
...
+                command: taskName,
+              },
+              vscode.workspace.workspaceFolders![0],
+              taskName,
+              ProductName,
+              new vscode.ShellExecution([baseNpmInstallCommand, ...args].join(" "), {
+                cwd: folder,
+              })
             );
+            task.presentationOptions.reveal = vscode.TaskRevealKind.Never;
```
swatDong, line 1057: Adjust the progress bar as well:
- wording - something like "Check [terminal window]() for detailed logs..."
- Is it possible to add "Open Terminal Window" button to progress bar?
- Is it possible to expand all progress bar text by default?

packages/fx-core/src/plugins/solution/fx-solution/debug/util/tasksTransparency.ts
kuojianlu: No and no. The two tasks can run in parallel.

packages/vscode-extension/src/debug/prerequisitesHandler.ts
kuojianlu: - Currently we use the progress handler from the common utils for all extension codes. It hard codes "Check **output window** for details.". Do we need to modify it or implement our own progress handler?
- vscode do not provide the API to add a button to a progress bar.
- Seems not. Do not find a way to expand the text by default.

swatDong: At least we need to tell user the detailed logs are in terminal window, so:
- If hard to replace "Check output window...", just add "Open [terminal window]()..." to the message
- Add "Open terminal window..." to output log as well
- Popup another notification (with open terminal button) together with the progress bar
- ...

kuojianlu: Sure. Let me update it in another PR.

swatDong (dismissed):

swatDong (approved):

PR #6217, opened 2022-09-27
chore(v3): fix migrator

jayzhang:


packages/fx-core/src/core/middleware/projectMigrator.ts
```diff
@@ -999,45 +1023,187 @@ async function updateConfig(ctx: CoreHookContext) {
...
+  }
+
+  // identity
+  {
+    const identity = getComponent(projectSettings, ComponentNames.Identity);
+    if (identity) {
+      const clonedInputs = cloneDeep(inputs);
+      assign(clonedInputs, {
+        componentId: "",
+        scenario: "",
```
yuqizhou77, line 1111: just for my learning. How do we determine whether we need these two properties and when to set them to be empty?

packages/fx-core/src/core/middleware/projectMigrator.ts
```diff
@@ -999,45 +1023,187 @@ async function updateConfig(ctx: CoreHookContext) {
...
+    const res = await bicepComponent.init(inputs.projectPath!);
+    if (res.isErr()) return err(res.error);
+  }
+  const biceps: Bicep[] = [];
+  const context = createContextV3(projectSettings);
+  // teams-tab
+  {
+    const config = getComponent(projectSettings, ComponentNames.TeamsTab);
+    if (config) {
+      const hosting = config.hosting! || ComponentNames.AzureStorage;
```
yuqizhou77, line 1042: also just for learning. Why we do "|| ComponentNames.AzureStorage" here?

jayzhang: For case when there is no hosting, actually, it may not happen if there is no bug for project settings migration.

packages/fx-core/src/core/middleware/projectMigrator.ts
jayzhang: It depends on implementations of generateBicep

packages/fx-core/src/core/middleware/projectMigrator.ts
```diff
@@ -71,14 +70,31 @@ import {
...
-import { getProjectTemplatesFolderPath } from "../../common/utils";
+import { convertToAlphanumericOnly, getProjectTemplatesFolderPath } from "../../common/utils";
+import Container from "typedi";
+import { BicepComponent } from "../../component/bicep";
+import { ComponentNames, Scenarios } from "../../component/constants";
+import { getComponent } from "../../component/workflow";
+import { bicepUtils, createContextV3, generateConfigBiceps } from "../../component/utils";
+import { assign, cloneDeep } from "lodash";
+import { IdentityResource } from "../../component/resource/identity";
+import { BotService } from "../../component/resource/botService";
```
hund030, line 85: Just update botService path. Please rebase latest dev branch.

jayzhang: ok

xzf0587 (approved):

hund030 (approved):

wenytang-ms (approved):
approval for the generate bicep part

PR #6222, opened 2022-09-27
chore: azure storage static website config driver

Siglud:


packages/fx-core/src/component/error/componentError.ts
```diff
@@ -114,6 +114,33 @@ export class ExternalApiCallError extends BaseComponentInnerError {
     super("UserError", name, message, messageParams, suggestionKey, detail, helpLink);
     this.statusCode = statusCode;
   }
+
+  static getAzureCredentialError(error?: unknown): ExternalApiCallError {
+    error = error ?? "";
+    return new ExternalApiCallError(
+      "GetAzureCredentialError",
+      "plugins.bot.FailRetrieveAzureCredentials",
+      -1,
```
eriolchan, line 123: Just curious that what does it  mean when value is -1?

packages/fx-core/src/component/provision/azureStorageStaticWebsiteConfigDriver.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  async isStorageStatusWebsiteEnabled(
+    azureInfo: AzureResourceInfo,
+    azureBlobClient: BlobServiceClient,
+    context: DriverContext
+  ): Promise<boolean> {
+    await context.logProvider.debug(
+      `Checking if static website feature is enabled in Azure Storage account '${azureInfo.instanceId}'.`
+    );
+    return (await azureBlobClient.getProperties()).staticWebsite?.enabled === true;
```
eriolchan, line 75: I am not sure if it can just return "return (await azureBlobClient.getProperties()).staticWebsite?.enabled" since it is equal to "enabled === true"

Siglud: We can't just return that value, it might be undefined

packages/fx-core/src/component/utils/azureResourceOperation.ts
```diff
@@ -0,0 +1,115 @@
...
+  try {
+    credential = await tokenProvider.getAccountCredentialAsync();
+  } catch (e) {
+    throw ExternalApiCallError.getAzureCredentialError(e);
+  }
+
+  if (!credential) {
+    throw PrerequisiteError.somethingIllegal(
+      "azureCredential",
+      "plugin.hosting.FailRetrieveAzureCredentials",
```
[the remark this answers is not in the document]
Siglud: yes, we will change it before the v3 version release

packages/fx-core/src/component/error/componentError.ts
Siglud: Nothing, I'll ignore the status code value if the value is -1 when I writing error log

eriolchan (approved):

eriolchan (approved):

hund030 (approved):

PR #6225, opened 2022-09-27
chore(debug): add comments for prerequisites and setup tasks

kuojianlu:


packages/fx-core/src/plugins/solution/fx-solution/debug/util/tasksTransparency.ts
```diff
@@ -388,38 +406,48 @@ function startLocalTunnel(): Record<string, unknown> {
...
     args: {
       baseUrl: "https://localhost:53000",
     },
   };
+  return commentJson.assign(commentJson.parse(comment), task);
 }
 
-function setUpBot(): Record<string, unknown> {
-  const comment = `
+function setUpBot(): CommentJSONValue {
```
XiaofuHuang, line 424: What about add an aka link here?

kuojianlu: Good idea.

swatDong (approved):

PR #6228, opened 2022-09-27
fix: dont ask user to provision when local debug failed

HuihuiWu-Microsoft:
After: 
image

ADO:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/15577951

E2E TEST: NA

### DIFF
--- a/packages/fx-core/resource/package.nls.json
+++ b/packages/fx-core/resource/package.nls.json
@@ -434,6 +434,7 @@
   "error.appstudio.invalidManifest": "Failed to parse manifest string, dut to error: %s. This might be caused by invalid configurations. %s",
   "error.appstudio.loadManifest": "Failed to read manifest file. Error: %s.",
   "error.appstudio.getLocalConfigFailed": "Missing configuration data for manifest. You may need to run 'Local debug' first. %s",
+  "error.appstudio.incorrectLocalConfig": "Missing configuration data for manifest.  %s Please remove the state.local.json file and try again.",
   "error.appstudio.getRemoteConfigFailed": "Missing configuration data for manifest. %s",
   "error.appstudio.publishFailed": "Failed to publish Teams app with ID %s.",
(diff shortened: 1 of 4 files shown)


packages/fx-core/resource/package.nls.json
```diff
@@ -434,6 +434,7 @@
   "error.appstudio.invalidManifest": "Failed to parse manifest string, dut to error: %s. This might be caused by invalid configurations. %s",
   "error.appstudio.loadManifest": "Failed to read manifest file. Error: %s.",
   "error.appstudio.getLocalConfigFailed": "Missing configuration data for manifest. You may need to run 'Local debug' first. %s",
+  "error.appstudio.incorrectLocalConfig": "Missing configuration data for manifest.  %s Please remove the state.local.json file and try again.",
```
dooriya, line 437: Remove previous local debug file will discard the previous resources (e.g. AAD, Teams app registration). Is that possible to just add back the missing configuration if user know the config value?

HuihuiWu-Microsoft: For other operations like when user click 'Zip Teams App package' button, we'll ask user to do local debug or provision if the states doesn't exist(https://github.com/OfficeDev/TeamsFx/blob/18f046bf45fdccd939b1d76dbb3e51d0687e2a41/packages/fx-core/src/component/resource/appManifest/utils/ManifestUtils.ts#L462);
For this issue, it happens when user do local debug and failed because of our Toolkit not compatible or user manually modified the state file.  So I don't think the missing config can be auto filled after local debug again without any other operations. If we don't want other resources discarded, maybe a better way is to ask user to only remove the key-value pairs under "fx-resource-bot" plugin so bot related resource will be provisioned again while other resources keep. The only concern is whether user can modify correctly. 
I think it should be OK to provision all the resources again since this is just a corner case and shouldn't happen after Zhijie's fix. Any ideas?

dooriya (approved):

jayzhang (approved):

PR #6230, opened 2022-09-27
Clear needConsent flag after successful login

xelossan:
When UIRequiredError occurs, needConsent is set to true so it calls teamsfx.login() at the next reload to ask for user consent. So far so good. The problem is: after the user gives the necessary consent, needConsent flag is not cleared, so subsequent reloads will keep calling teamsfx.login and showing popup window that will automatically disappear.

This change fixes the issue - when consent is granted, login function won't be called.


yiqing-zhao (approved):

yiqing-zhao:
Thanks for your advice! Since PR should be merged into Dev instead of Main, I changed this manually.

PR #6234, opened 2022-09-28
chore: add aadApp/create action

blackchoey:
Reimplemented some existing classes for refactor / code cleanup purpose and avoid updating too much existing code. In the feature we can cleanup unused code, which should out of V3 scope.

UT will be added later.


packages/fx-core/src/component/drivers/aad/constants.ts
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Microsoft Corporation.
```
xzf0587, line 1: The path name of file uses "drivers" instead of "driver". 
The names under component use singular style. 
Only a suggestion.

packages/fx-core/src/component/drivers/aad/create.ts
```diff
@@ -0,0 +1,118 @@
...
+    if (args.genegerateClientSecret === undefined) {
+      invalidParameters.push("genegerateClientSecret");
+    }
+
+    if (invalidParameters.length > 0) {
+      const invalidParametersString = invalidParameters.join(",");
+      throw new UserError({
+        source: actionName,
+        name: DriverConstants.invalidParameterError.errorCode,
+        message: getDefaultString(
```
Siglud, line 78: maybe you can throw PrerequisiteError here, the code is more simple than construct a UserError

blackchoey: Thanks. I found I need some non-trivial changes to PrerequisiteError in order to adopt it, so followed your idea to create my own error class. We can see how to share the common parts between each action (driver) later.

packages/fx-core/src/component/drivers/aad/constants.ts
blackchoey: Changed folder name to `driver`. For renaming `driver` to `action`, since other code in fx-core (e.g. deploy actions) already uses `driver` in multiple places, I will driver discussions offline and have additional PR to rename the folder, class names, etc.

packages/fx-core/src/component/driver/aad/create.ts
```diff
@@ -0,0 +1,114 @@
...
+        // Create new AAD app if no client id exists
+        const aadApp = await aadAppClient.createAadApp(args.name);
+        aadAppState.AAD_APP_CLIENT_ID = aadApp.appId!;
+        aadAppState.AAD_APP_OBJECT_ID = aadApp.id!;
+        await this.setAadEndpointInfo(context.m365TokenProvider, aadAppState);
+      }
+
+      if (args.generateClientSecret && !aadAppState.SECRET_AAD_APP_CLIENT_SECRET) {
+        // Create new client secret if no client secret exists
+        if (!aadAppState.AAD_APP_OBJECT_ID) {
```
adashen, line 41: So if it is the existing aad app scenario and would like to generate secret, object id must be provided?

blackchoey: Yes, there may be cases that user does not have permission to query object id by client id. So this action follows existing AAD plugin design to ask user provide object id for existing AAD app.

packages/fx-core/src/component/driver/aad/interface/createAadAppOutput.ts
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+export interface CreateAadAppOutput {
+  // Non secrets
+  AAD_APP_CLIENT_ID?: string;
```
1yefuwang1, line 6: Are these fields optional by design?

blackchoey: Yes, though it was called output, it also serves as input because we're doing create if not exist things which relies on previous output.
You can think what the action does is: if one property in output is empty, it will set the property's value and output to .env. There's one exception that when client id exists, the action will not fill missing values.

1yefuwang1 (approved):

jayzhang (approved):

KennethBWSong (approved):

PR #6248, opened 2022-09-28
fix: state value may be empty string

hund030:
E2E TEST: N/A

### DIFF
--- a/packages/fx-core/src/component/resource/appManifest/utils.ts
+++ b/packages/fx-core/src/component/resource/appManifest/utils.ts
@@ -556,7 +556,7 @@ export function resolveManifestTemplate(
         if (array.length === 3 && array[0] === "state") {
           const component = array[1];
           const configKey = array[2];
-          if (!view.state[component] || !view.state[component][configKey]) {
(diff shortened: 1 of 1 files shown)


packages/fx-core/src/component/resource/appManifest/utils.ts
```diff
@@ -556,7 +556,7 @@ export function resolveManifestTemplate(
         if (array.length === 3 && array[0] === "state") {
           const component = array[1];
           const configKey = array[2];
-          if (!view.state[component] || !view.state[component][configKey]) {
+          if (view.state[component]?.[configKey] == undefined) {
```
eriolchan, line 559: What's the reason to change the equation? It seems that there is some special logic that we don't use old one? Can I learn more about that reason, and weather to use "==" or "==="?

hund030: `==` matches both null and undefined, while `===` only matches undefined.
Empty string is a valid value for some states, while `!view.state[component][configKey]` regards it as invalid value.

eriolchan (approved):

eriolchan (approved):

PR #6249, opened 2022-09-28
chore(app-setting): app settings for ts sso-tab

formulahendry:
ADO task: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_git/teamsfx-specs/pullrequest/6505698?path=/specs/app-setting-improvement.md&_a=files

- Add config.ts to include all the configs
- Add .env.teamsfx.local to include all the configs

This PR is part of `app settings improvement` feature. Will update more templates after this one is merged.


templates/tab/ts/default/.env.teamsfx.local
```diff
@@ -0,0 +1,7 @@
+# Following variables will be overwritten by TeamsFx
```
hund030, line 1: Why do we need this file in source control?

blackchoey: Let users understand where the environment variable values come from when running the code locally. This will be more straightforward than asking users to view the package.json file. And many samples (e.g. the botframework samples) provides an .env to help user get started easily.

hund030: This file will be re-generated by debug component when user try local debug. The updated .env.teamsfx.local will be:
```
# Following variables are generated by TeamsFx
BROWSER=none
HTTPS=true
SSL_CRT_FILE=xxxxxx/.fx/certificate/localhost.crt
SSL_KEY_FILE=xxxxxx/.fx/certificate/localhost.key
REACT_APP_START_LOGIN_PAGE_URL=https://localhost:53000/auth-start.html
REACT_APP_CLIENT_ID=xxxxxxxxxxx
REACT_APP_FUNC_ENDPOINT=
REACT_APP_FUNC_NAME=
PORT=53000

# Following variables can be customized or you can add your owns
# FOO=BAR
```
Is it expected?

blackchoey: Thanks. I think we need to use your version (and remove the values) to ensure we have all the keys. This also reminds me that @swatDong is updating the wording in this file. @formulahendry can you help followup with Qianhao on the final content?

swatDong: FYI. The wordings can be found in both source code and [doc](https://github.com/OfficeDev/TeamsFx/blob/dev/docs/fx-core/localdebug-help.md#what-to-do-if-i-want-to-add-my-own-environment-variables-for-local-debug).

formulahendry, line 1: @jidddddd Please help advice the wording part. Whether we should use `Following variables are generated by TeamsFx` to align with local-debug, or we should use something like `Following variables will be overwritten by TeamsFx` as the values are all empty after scaffold.

jidddddd: @formulahendry can we change it to `TeamsFx will overwrite the following variable values when running debug`
@swatDong can we update the source code and doc for this comment as well?

swatDong: Updated in https://github.com/OfficeDev/TeamsFx/pull/6283

formulahendry: Thanks all for the valuable comments~

templates/tab/ts/default/.env.teamsfx.local
```diff
@@ -0,0 +1,13 @@
+# TeamsFx will overwrite the following variable values when running debug
+SSL_CRT_FILE=
+SSL_KEY_FILE=
+PORT=
+BROWSER=none
+HTTPS=true
```
formulahendry, line 6: @swatDong Shall we keep the values (`none`, `true`) like the original `.env` file does, or just remove the values as well?

swatDong: I'd like to keep these, since teams tab app requires https.

blackchoey (approved):

swatDong (approved):

hund030 (approved):

PR #6259, opened 2022-09-29
chore: update service principal get json method

xiaolang124:
1.Fix service principal get user data problem.


xiaolang124:
image

Service principal use server-side permission and do not support to get specific permission method. Original scope is */user_impersonation.

a1exwang (approved):

PR #6279, opened 2022-10-08
feat(vsc): add support for untrusted workspace

tecton:
Related feature: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/14775898

Effect Preview:
image

E2E TEST: N/A

### DIFF
--- a/packages/vscode-extension/src/extension.ts
+++ b/packages/vscode-extension/src/extension.ts
@@ -91,7 +91,9 @@ export async function activate(context: vscode.ExtensionContext) {
       m365TokenProvider: M365TokenInstance,
     });
 
-    registerCodelensAndHoverProviders(context);
+    if (vscode.workspace.isTrusted) {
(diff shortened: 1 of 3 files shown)


packages/vscode-extension/src/extension.ts
```diff
@@ -91,7 +91,9 @@ export async function activate(context: vscode.ExtensionContext) {
       m365TokenProvider: M365TokenInstance,
     });
 
-    registerCodelensAndHoverProviders(context);
+    if (vscode.workspace.isTrusted) {
+      registerCodelensAndHoverProviders(context);
```
HuihuiWu-Microsoft, line 95: How about other operations like debugging and tasks?

tecton: They are all blocked by VS Code.

HuihuiWu-Microsoft (approved):

PR #6284, opened 2022-10-09
fix(vsc): select workspace folder canceled in WSL

tecton:
Related bug: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/15772677

E2E TEST: N/A

### DIFF
--- a/packages/vscode-extension/src/qm/vsc_ui.ts
+++ b/packages/vscode-extension/src/qm/vsc_ui.ts
@@ -475,6 +475,7 @@ export class VsCodeUI implements UserInteraction {
             },
           ];
 
+          let hideByDialog = false;
           const onDidAccept = async () => {
             const selectedItems = quickPick.selectedItems;
             if (selectedItems && selectedItems.length > 0) {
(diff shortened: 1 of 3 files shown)


packages/vscode-extension/src/qm/vsc_ui.ts
```diff
@@ -504,9 +504,6 @@ export class VsCodeUI implements UserInteraction {
 
           disposables.push(
             quickPick.onDidAccept(onDidAccept),
-            quickPick.onDidHide(() => {
```
HuihuiWu-Microsoft, line 507: Is there any side effects when user press Esc in the quick pick?

tecton: ESC can still cancel the quick pick. It invokes `onDidHide` as well.

HuihuiWu-Microsoft: When it invokes `onDidHide`, will it trigger UserCancelError and send correct telemetry?

tecton: Yes, I'll update the code. Otherwise, the telemetry will be lost.

HuihuiWu-Microsoft (approved):

PR #6288, opened 2022-10-09
chore(app-setting): app settings for ts bot

formulahendry:
ADO task: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_git/teamsfx-specs/pullrequest/6505698?path=/specs/app-setting-improvement.md&_a=files

- Add config.ts to include all the configs
- Add .env.teamsfx.local to include all the configs

This PR is part of `app settings improvement` feature. Will update more templates.

Previous PR for ts sso-tab: https://github.com/OfficeDev/TeamsFx/pull/6249

### DIFF
--- a/templates/bot/ts/default/.env.teamsfx.local
+++ b/templates/bot/ts/default/.env.teamsfx.local
@@ -1,0 +1,17 @@
+# TeamsFx will overwrite the following variable values when running debug
+BOT_ID=
+BOT_PASSWORD=
+M365_CLIENT_ID=
+M365_CLIENT_SECRET=
+M365_TENANT_ID=
+M365_AUTHORITY_HOST=
+INITIATE_LOGIN_ENDPOINT=
+API_ENDPOINT=
+M365_APPLICATION_ID_URI=
+
+# Following variables can be customized or you can add your owns
+# FOO=BAR
+SQL_ENDPOINT=
+SQL_DATABASE_NAME=
+SQL_USER_NAME=
+SQL_PASSWORD=
--- a/templates/bot/ts/default/.env.teamsfx.local
+++ b/templates/bot/ts/default/.env.teamsfx.local
@@ -1,17 +1,6 @@
 # TeamsFx will overwrite the following variable values when running debug
 BOT_ID=
 BOT_PASSWORD=
-M365_CLIENT_ID=
-M365_CLIENT_SECRET=
-M365_TENANT_ID=
-M365_AUTHORITY_HOST=
-INITIATE_LOGIN_ENDPOINT=
-API_ENDPOINT=
-M365_APPLICATION_ID_URI=
 
 # Following variables can be customized or you can add your owns
 # FOO=BAR
-SQL_ENDPOINT=
-SQL_DATABASE_NAME=
-SQL_USER_NAME=
-SQL_PASSWORD=
--- a/templates/bot/ts/default/config.ts
+++ b/templates/bot/ts/default/config.ts
@@ -1,13 +1,6 @@
 const config = {
   botId: process.env.BOT_ID,
   botPassword: process.env.BOT_PASSWORD,
-  clientId: process.env.M365_CLIENT_ID,
-  clientSecret: process.env.M365_CLIENT_SECRET,
-  tenantId: process.env.M365_TENANT_ID,
-  authorityHost: process.env.M365_AUTHORITY_HOST,
-  initiateLoginEndpoint: process.env.INITIATE_LOGIN_ENDPOINT,
-  apiEndpoint: process.env.API_ENDPOINT,
-  applicationIdUri: process.env.M365_APPLICATION_ID_URI,
 };
 
 export default config;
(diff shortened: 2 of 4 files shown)


templates/bot/ts/default/config.ts
```diff
@@ -0,0 +1,13 @@
+const config = {
+  botId: process.env.BOT_ID,
+  botPassword: process.env.BOT_PASSWORD,
+  clientId: process.env.M365_CLIENT_ID,
+  clientSecret: process.env.M365_CLIENT_SECRET,
+  tenantId: process.env.M365_TENANT_ID,
+  authorityHost: process.env.M365_AUTHORITY_HOST,
+  initiateLoginEndpoint: process.env.INITIATE_LOGIN_ENDPOINT,
+  apiEndpoint: process.env.API_ENDPOINT,
```
SLdragon, line 9: Where is the config `apiEndpoint` used?

formulahendry: It would be used if developer wants to call Azure Functions in Bot, so list all possible environment variables here, and it is aligned with the `.env.teamsfx.local` generated by local-debug

Siglud (approved):

swatDong (dismissed):

templates/bot/ts/default/config.ts
adashen: Do we need to list all the settings or we focus on what is used by the template?

templates/bot/ts/default/package.json
```diff
@@ -24,6 +24,7 @@
     "restify": "^8.5.1"
   },
   "devDependencies": {
+    "@types/node": "^12.0.0",
```
adashen, line 27: why this dev dependencies is introduced?

formulahendry: Before `npm install`, I meet with below lint warning:
image
But after `npm install`, I find `@types/node` is already in descendant dev dependencies, so no need to explicitly include it now. Thanks for pointing out. I have removed it.

templates/bot/ts/default/config.ts
formulahendry: For `.env.teamsfx.local`, I think we need to list all the settings, and align with local-debug.
For `config.ts`, it is worthy to discuss. Add @jidddddd for PM's thought on this.
cc @blackchoey

blackchoey: If the code doesn't use apiEndpoint, I would prefer to remove it in config.ts to avoid confusing.

For .env.teamsfx.local, my opinion is the keys in it should align with `config.ts`, which means it should also not have `apiEndpoint`. Otherwise user maybe confused where does the code consumes `apiEndpoint`.

How do you think not having `apiEndpoint` in scaffolded `.env.teamsfx.local` but an empty `apiEndpoint` is added to `.env.teamsfx.local` during local debug?

Also cc @swatDong to see what's the effort to not generate apiEndpoint if the source code does not need it. Though the priority is low from my perspective.

templates/bot/ts/default/.env.teamsfx.local
```diff
@@ -0,0 +1,17 @@
...
+M365_CLIENT_SECRET=
+M365_TENANT_ID=
+M365_AUTHORITY_HOST=
+INITIATE_LOGIN_ENDPOINT=
+API_ENDPOINT=
+M365_APPLICATION_ID_URI=
+
+# Following variables can be customized or you can add your owns
+# FOO=BAR
+SQL_ENDPOINT=
```
blackchoey, line 14: can we remove SQL related items in this file?

templates/bot/ts/default/config.ts
formulahendry: In that case, for the alignment, we need to remove other keys which are not used as well, then only keep `BOT_ID` and `BOT_PASSWORD` in `.env.teamsfx.local` and  `config.ts` file?

blackchoey: In my opinion, yes. We just need to keep bot id & password.

jidddddd: I think it's better to only list variables as needed if possible. If the user wants to add extra features, they would follow a how-to guide on what and where to add variables.

cc: @MuyangAmigo, @swatDong, @blackchoey

adashen: And another thing is, can we merge the template change now? Is there any dependency needs ready first?

templates/bot/ts/default/.env.teamsfx.local
formulahendry: Updated

templates/bot/ts/default/config.ts
formulahendry: @blackchoey @jidddddd Updated.

formulahendry: > And another thing is, can we merge the template change now? Is there any dependency needs ready first?

@adashen De you mean merge this PR into `dev` branch? I think there is no dependency.

swatDong (approved):

blackchoey (approved):

SLdragon (approved):

Siglud (approved):

PR #6290, opened 2022-10-09
chore: add utility to expand environment variables

blackchoey:
Add utility to expand environment variables for V3's new placeholder syntax.
It doesn't support escaping for now. We can enrich the functionality in the future if there's explicit requirement.

### DIFF
--- a/packages/fx-core/src/component/utils/common.ts
+++ b/packages/fx-core/src/component/utils/common.ts
@@ -55,3 +55,21 @@ export function asFactory<T>(keyValidators: KeyValidators<T>) {
     throw PrerequisiteError.somethingIllegal("Deploy", "data", "plugins.bot.InvalidData");
   };
 }
+
+// Expand environment variables in content. The format of referencing environment variable is: ${{ENV_NAME}}
+export function expandEnvironmentVariable(content: string): string {
+  const placeholderRegex = /\${{[a-zA-Z_][a-zA-Z0-9_]*}}/g;
+  const placeholders = content.match(placeholderRegex);
+
+  if (placeholders) {
+    for (const placeholder of placeholders) {
+      const envName = placeholder.slice(3, -2); // removes `${{` and `}}`
+      const envValue = process.env[envName];
+      if (envValue) {
+        content = content.replace(placeholder, envValue);
+      }
+    }
+  }
+
+  return content;
+}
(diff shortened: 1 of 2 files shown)


packages/fx-core/src/component/utils/common.ts
```diff
@@ -55,3 +55,21 @@ export function asFactory<T>(keyValidators: KeyValidators<T>) {
...
 }
+
+// Expand environment variables in content. The format of referencing environment variable is: ${{ENV_NAME}}
+export function expandEnvironmentVariable(content: string): string {
+  const placeholderRegex = /\${{[a-zA-Z_][a-zA-Z0-9_]*}}/g;
+  const placeholders = content.match(placeholderRegex);
+
+  if (placeholders) {
+    for (const placeholder of placeholders) {
+      const envName = placeholder.slice(3, -2); // removes `${{` and `}}`
```
Siglud, line 66: maybe we should add .trim() to avoid the user using the environment variables like `${{  xxx }}`?

packages/fx-core/src/component/utils/common.ts
```diff
@@ -55,3 +55,21 @@ export function asFactory<T>(keyValidators: KeyValidators<T>) {
     throw PrerequisiteError.somethingIllegal("Deploy", "data", "plugins.bot.InvalidData");
   };
 }
+
+// Expand environment variables in content. The format of referencing environment variable is: ${{ENV_NAME}}
+export function expandEnvironmentVariable(content: string): string {
+  const placeholderRegex = /\${{[a-zA-Z_][a-zA-Z0-9_]*}}/g;
```
Siglud, line 61: I think environment variable parameters should not be allowed to start with an underscore

packages/fx-core/src/component/utils/common.ts
1yefuwang1: Perhaps we can use `toUpperCase()` here to normalize env var names.

blackchoey: @Siglud Thanks for the suggestion. Added trim to allow white spaces around env name. 
@1yefuwang1 though env name should be upper case, users can indeed set env name with lower case characters. So my preference is respecting the name provided by user.

packages/fx-core/src/component/utils/common.ts
blackchoey: Based on following answer, env name can start with underscore (only digit is not allowed as first character):
https://stackoverflow.com/questions/2821043/allowed-characters-in-linux-environment-variable-names

1yefuwang1 (approved):

xzf0587 (approved):

PR #6296, opened 2022-10-10
chore(v3): coordinator

jayzhang:


packages/fx-core/src/component/coordinator/index.ts
```diff
@@ -0,0 +1,172 @@
+import { hooks } from "@feathersjs/hooks/lib";
```
xzf0587, line 1: add a file header?

hund030 (approved):

PR #6306, opened 2022-10-10
chore: add aadApp/update action

blackchoey:


packages/fx-core/src/component/driver/aad/update.ts
```diff
@@ -0,0 +1,137 @@
...
+import axios from "axios";
+import { SystemError, UserError } from "@microsoft/teamsfx-api";
+import { UnhandledSystemError, UnhandledUserError } from "./error/unhandledError";
+import { getUuid } from "../../../common/tools";
+import { expandEnvironmentVariable } from "../../utils/common";
+import { AadManifestHelper } from "../../resource/aadApp/utils/aadManifestHelper";
+import { AADManifest } from "../../resource/aadApp/interfaces/AADManifest";
+import { MissingFieldInManifestUserError } from "./error/invalidFieldInManifestError";
+
+const actionName = "aadApp/update"; // DO NOT MODIFY the name
```
SLdragon, line 21: Why "DO NOT MODIFY the name", if in the future we want to refactor this name, how to do it?

packages/fx-core/src/component/driver/aad/update.ts
```diff
@@ -0,0 +1,137 @@
...
+const helpLink = "https://aka.ms/teamsfx-actions/aadapp-update";
+
+// logic from src\component\resource\aadApp\aadAppManifestManager.ts
+@Service(actionName) // DO NOT MODIFY the service name
+export class UpdateAadAppDriver implements StepDriver {
+  private guidRegex =
```
SLdragon, line 27: How about using existing library:
```ts
 import isUUID from "validator/lib/isUUID".

if(isUUID("xxxxxxxx")){
   ...
}
```

packages/fx-core/src/component/driver/aad/interface/updateAadAppArgs.ts
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+export interface UpdateAadAppArgs {
+  manifestPath: string; // Relative path to project root
+  manifestOutputPath: string; // Relative path to project root
```
wenytang-ms, line 6: if the manifestOutputPath is required?

packages/fx-core/src/component/driver/aad/update.ts
blackchoey: user will reference this name in teamsfx.yml, if we changed the name, it will be a breaking change.
``` yaml
provision:
  - uses: "aadApp/update"
    with:
        manifestPath: "xxx"
```

packages/fx-core/src/component/driver/aad/update.ts
blackchoey: thanks, will use that lib

packages/fx-core/src/component/driver/aad/interface/updateAadAppArgs.ts
blackchoey: I think we can make it optional. If user provides output path (e.g. build/aad.manifest.dev.json), we will write the actual manifest to that path. If not, we do nothing.
@SLdragon @KennethBWSong @jidddddd @adashen how do you think about above proposal? The current behavior in V2 is we always write actual manifest to `build/aad.manifest.{env}.json`.

blackchoey: also cc @MuyangAmigo for your thoughts of above proposal

adashen: I would like to have the output file which is consistent with common build behaviour.  And since the path is relative, where did we get the project root? Do we have a common utility in fxcore for project root?

adashen: Besides, when user select a custom manifest from external folder. Is it still good to use relative path here?

blackchoey: Will create additional PR to compute the actual absolute path because current context does not have project root path yet. Don't want to make this PR pending too much time.

packages/fx-core/src/component/driver/aad/update.ts
```diff
@@ -0,0 +1,142 @@
...
+        const preAuthorizedApplications = manifest.preAuthorizedApplications;
+        manifest.preAuthorizedApplications = [];
+        await aadAppClient.updateAadApp(manifest);
+        manifest.preAuthorizedApplications = preAuthorizedApplications;
+      }
+      // 2. Update AAD app again with full manifest to set preAuthorizedApplications
+      await aadAppClient.updateAadApp(manifest);
+
+      // Output actual manifest to project folder
+      await fs.ensureDir(path.dirname(args.manifestOutputPath));
```
wenytang-ms, line 62: maybe manifestOutputFile is easy to understand. 
it's not block

packages/fx-core/src/component/driver/aad/update.ts
```diff
@@ -0,0 +1,142 @@
...
+      await fs.writeFile(args.manifestOutputPath, JSON.stringify(manifest, null, 4), "utf8");
```
wenytang-ms, line 63: manifestOutputPath actually is a file, so maybe manifestOutputFile is easy to understand?

wenytang-ms (approved):

packages/fx-core/src/component/driver/aad/update.ts
blackchoey: updated to outputFilePath after synced with @jidddddd

wenytang-ms (approved):

KennethBWSong (approved):

SLdragon (approved):

xzf0587 (approved):

PR #6333, opened 2022-10-12
fix: deploy aad manifest condition update

wenytang-ms:
E2E TEST: https://github.com/OfficeDev/TeamsFx/actions/runs/3231453639


xzf0587 (dismissed):

jayzhang (dismissed):

xzf0587 (dismissed):

xzf0587 (approved):

jayzhang (approved):

packages/vscode-extension/src/handlers.ts
```diff
@@ -424,6 +425,16 @@ function registerCoreEvents() {
   }
 }
 
+export async function getAzureProjectConfigV3(): Promise<ProjectConfigV3 | undefined> {
```
HuihuiWu-Microsoft, line 428: Is the legacy getAzureSolutionSettings() function still used in other places? If not, why not just replace with v3 logic?

HuihuiWu-Microsoft (approved):

packages/vscode-extension/src/handlers.ts
wenytang-ms: this is a workaround, once v3 release, the projectSettings.json file may not be exist, so later will udpate the logic.

PR #6344, opened 2022-10-12
chore: change folder structure and interface return type add parameter to context

Siglud:
change folder structure & change return type to Result<T> & change v3 context


hund030 (approved):

packages/fx-core/tests/component/drivers/aad/create.test.ts
```diff
@@ -61,14 +61,14 @@ describe("aadAppCreate", async () => {
     args = {
       generateClientSecret: true,
     };
-    await expect(createAadAppDriver.run(args, mockedDriverContext))
+    await expect(createAadAppDriver.handler(args, mockedDriverContext))
```
blackchoey, line 64: minor: in my opinion we should test the run() function instead of the handler function because run() is our interface. You can keep current code and I will update the test cases later.

blackchoey (approved):

xzf0587 (approved):

eriolchan (approved):

PR #6375, opened 2022-10-17
chore(vsc): add experiment welcome views

tecton:
Related feature: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/15461918


packages/vscode-extension/src/extension.ts
```diff
@@ -824,6 +829,27 @@ async function runBackgroundAsyncTasks(
...
+      true
+    )) as string | undefined;
+  await vscode.commands.executeCommand(
+    "setContext",
+    "fx-extension.isExistingUser",
+    isExistingUser !== "no"
+  );
+  if (TreatmentVariableValue.welcomeViewStyle === "A") {
+    await vscode.commands.executeCommand("setContext", "fx-extension.welcomeViewTreatment", true);
+    await vscode.commands.executeCommand("setContext", "fx-extension.welcomeViewA", true);
```
HuihuiWu-Microsoft, line 847: If the group is re-randomized, it may happen that both "fx-extension.welcomeViewA" and "fx-extension.welcomeViewB" contexts are true. Can we also set another context to avoid this situation?

HuihuiWu-Microsoft (approved):

packages/vscode-extension/src/extension.ts
tecton: The context will be cleared when VS Code starts.

PR #6378, opened 2022-10-17
chore: add progress bar logic; change build arguments name

Siglud:


hund030 (dismissed):

packages/fx-core/src/component/driver/script/baseBuildDriver.ts
```diff
@@ -23,18 +41,29 @@ export abstract class BaseBuildDriver extends BaseDeployStepDriver {
...
     }
     return BaseBuildDriver.emptyMap;
   }
+
+  /**
+   * call when error happens
+   * do some resource clean up
+   */
+  async cleanup(): Promise<void> {
+    await this.progressBar?.end(false);
```
eriolchan, line 67: Just curious why here is end with false, but in some files e.g. packages/fx-core/src/component/driver/deploy/azure/azureAppServiceDeployDriver.ts is true?

Siglud: This function is only called when an error occurs.

eriolchan (approved):

hund030 (approved):

PR #6384, opened 2022-10-17
fix(debug): allow debug after m365 account swtiched

kuojianlu:
@microsoft/teamsfx-cli	v1.1.3-rc.2
@microsoft/teamsfx-core	v1.17.0-rc.2
ms-teams-vscode-extension	v4.1.0-rc.2

CLI BREAKING CHANGE:
  

CLI feat commits:
  

### DIFF
--- a/packages/fx-core/src/component/debugHandler/utils.ts
+++ b/packages/fx-core/src/component/debugHandler/utils.ts
@@ -1,0 +1,100 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+"use strict";
+
+import {
+  CryptoProvider,
+  LogProvider,
+  M365TokenProvider,
+  ProjectSettingsV3,
+  ResourceContextV3,
+  TelemetryReporter,
+  UserInteraction,
+  v3,
+  err,
+  ok,
+  InputsWithProjectPath,
+  Platform,
+  Result,
+  Void,
+  FxError,
+} from "@microsoft/teamsfx-api";
+
+import { ComponentNames } from "../constants";
+import { DefaultManifestProvider } from "../resource/appManifest/manifestProvider";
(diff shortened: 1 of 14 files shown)


kuojianlu:
Fix https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/15872897.

E2E TEST: N/A

XiaofuHuang (approved):

packages/fx-core/src/component/debugHandler/utils.ts
```diff
@@ -0,0 +1,100 @@
...
+    logger,
+    telemetry,
+    ui,
+    cryptoProvider
+  );
+  const inputs = constructInputsWithProjectPath(projectPath);
+  const tenantId =
+    envInfoV3.state[ComponentNames.AadApp]?.tenantId ||
+    envInfoV3.state[ComponentNames.AppManifest]?.tenantId;
+  const checkResult = await checkWhetherLocalDebugM365TenantMatches(
```
swatDong, line 52: Is there the case - user does not login, and run the task(s) directly. Will this ask user to login again and again?

kuojianlu: `checkWhetherLocalDebugM365TenantMatches` does ask user to login if user has not loged in.  The tasks are running in sequence, so only the first time login is required, and the following steps will run as normal.

swatDong: I mean, what if use cancels login? Or that's out of scope?

kuojianlu: If user cancels login, there will be an error to terminate the running task.
image

swatDong (approved):

PR #6403, opened 2022-10-18
chore(template v3): add sso tab templates

hund030:


templates/scenarios/sso-tab-js/.fx/configs/config.dev.json.tpl
```diff
@@ -0,0 +1,18 @@
+{
```
huimiu, line 1: Just curious, is the tpl file a new design in V3? I don't seem to see it in the current template

hund030: No. We add .tpl suffix to identify which file includes placeholders. You can find such files in v2 templates, too.
https://github.com/OfficeDev/TeamsFx/blob/dev/templates/tab/csharp/default/Interop/InteropModuleBase.cs.tpl

huimiu: OK, I got it. It looks like most of it is used in csharp's template

templates/scenarios/sso-tab-js/templates/azure/main.bicep
```diff
@@ -0,0 +1,18 @@
+@secure()
+param provisionParameters object
```
huimiu, line 2: Is this variable defined repeatedly? I guess you want to put all the variable definitions in the` config.bicep` file, I don't know if I understand it correctly.

hund030: I just copy the biceps from current teamsfx project. Here the config.bicep is useless, but I don't want to make the change in this PR.

huimiu (approved):

PR #6419, opened 2022-10-20
chore(v3template): notification-restify-js

xzf0587:


templates/scenarios/notification-restify-js/.fx/projectSettings.json
```diff
@@ -0,0 +1,5 @@
+{
+    "projectId": "92f9e2d5-2d2e-482c-9ac3-7ed261b3c4bf",
```
hund030, line 2: Maybe a placeholder? Or just remove it?

xzf0587: It is removed in template. Thanks.
@jayzhang, it seems the projectId should be generated in create project stage.

PR #6428, opened 2022-10-20
chore: v3 config manager

1yefuwang1:
closes https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_sprints/taskboard/Teams%20Extensibility%20E2E%20Team/Microsoft%20Teams%20Extensibility/CY22-10.2?workitem=15675483
and 
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_sprints/taskboard/Teams%20Extensibility%20E2E%20Team/Microsoft%20Teams%20Extensibility/CY22-10.2?workitem=15835412


packages/fx-core/src/component/configManager/lifecycle.ts
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    for (const driver of drivers) {
+      const result = await driver.instance.run(driver.with, ctx);
+      if (result.isErr()) {
+        return err(result.error);
+      }
+      for (const [envVar, value] of result.value) {
+        envOutput.set(envVar, value);
+        process.env[envVar] = value;
```
xzf0587, line 70: Here the secret env var will be written in env directly. Is it as design?

1yefuwang1: Yes, encryption/decryption only happens when values are written to/read from files, which is handled by core. In memory, they are just plain text.

jayzhang (approved):

PR #6444, opened 2022-10-21
chore: add extensions content to manifest generation

millerds:
Add code to put in content into the extensions section of the manifest template based on the office-addin template (currently only json case one available).

### DIFF
--- a/packages/manifest/src/manifest.ts
+++ b/packages/manifest/src/manifest.ts
@@ -545,7 +545,7 @@ export class DevPreviewManifest implements DevPreviewSchema {
   localizationInfo?: ILocalizationInfo;
 
   developer: IDeveloper = {
-    name: "Teams App, Inc.",
+    name: "Contoso",
     mpnId: "",
     websiteUrl: "https://localhost:3000",
     privacyUrl: "https://localhost:3000/privacy",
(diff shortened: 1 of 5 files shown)


packages/manifest/src/manifest.ts
```diff
@@ -545,7 +545,7 @@ export class DevPreviewManifest implements DevPreviewSchema {
   localizationInfo?: ILocalizationInfo;
 
   developer: IDeveloper = {
-    name: "Teams App, Inc.",
+    name: "Contoso",
```
1yefuwang1, line 548: I think it's better to change the name in devPreviewManifest.ts, otherwise teams app's manifest is also changed.

millerds: Isn't devPreviewManifest.ts just the definition of the interfaces representing the dev preview manifest structure?  This is where those interfaces become a class with values.  The DevPreviewManifest object gets this change which isn't specific to the type of app getting created while the TeamsAppManifest class keeps the Teams specific name.  

When we eventually unify these two classes (or deprecate one) we're going to have to have a generic value the represents the whole rather than being teams specific.

1yefuwang1: Yeah, you are right.

1yefuwang1 (approved):

PR #6462, opened 2022-10-25
fix: create new teams app if account in same tenant

yuqizhou77:
- We used to generate a new Teams app id if local debug with a different account from same tenant. Add this logic back to the new local debug workflow.

E2E TEST: https://github.com/OfficeDev/TeamsFx/actions/runs/3317743455

### DIFF
--- a/packages/fx-core/src/component/debugHandler/appManifest.ts
+++ b/packages/fx-core/src/component/debugHandler/appManifest.ts
@@ -28,7 +28,10 @@ import { environmentManager } from "../../core/environment";
 import { loadProjectSettingsByProjectPath } from "../../core/middleware/projectSettingsLoader";
 import { AppStudioClient } from "../resource/appManifest/appStudioClient";
 import { ComponentNames } from "../constants";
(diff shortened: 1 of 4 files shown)


packages/fx-core/src/component/debugHandler/appManifest.ts
```diff
@@ -174,6 +178,21 @@ export class AppManifestDebugHandler {
       // For SPFx manifest
       this.envInfoV3.config.isLocalDebug = true;
 
+      // Local debug if switching to a different account in same tenant
+      if (
+        this.envInfoV3.envName === environmentManager.getLocalEnvName() &&
```
kuojianlu, line 183: No need to check the envName because only local debug will enter here.

kuojianlu:
How about bot?

packages/fx-core/src/component/debugHandler/appManifest.ts
yuqizhou77: updated.

kuojianlu (approved):

nliu-ms (approved):

PR #6483, opened 2022-10-25
fix(localdebug): make local debug task args can be empty

XiaofuHuang:
Update `args` schema to make it can be empty
image
E2E TEST: none

### DIFF
--- a/packages/vscode-extension/package.json
+++ b/packages/vscode-extension/package.json
@@ -778,7 +778,7 @@
             ]
           },
           "args": {
-            "oneOf": [
+            "anyOf": [
               {
                 "type": "object",
                 "required": [


packages/vscode-extension/package.json
```diff
@@ -778,7 +778,7 @@
             ]
           },
           "args": {
-            "oneOf": [
+            "anyOf": [
```
kuojianlu, line 781: What's the effect after changed to `anyOf`? Will the warning still exist?

XiaofuHuang: The warning will disappear. It allow args to match multiple schema in the array.

kuojianlu (approved):

tecton (approved):

PR #6484, opened 2022-10-25
chore(generator): remove fetch sample tag

yukun-dong:


packages/fx-core/src/component/generator/generator.ts
```diff
@@ -82,7 +84,7 @@ export class Generator {
       destination: destinationPath,
       logProvider: ctx.logProvider,
       zipUrl: sample.link,
```
hund030, line 86: `sample.link ?? sampleConfig.defaultPackageLink`?

packages/fx-core/src/component/generator/utils.ts
```diff
@@ -135,19 +121,6 @@ export async function fetchTemplateZipUrl(
   return `${templateConfig.templateDownloadBaseURL}/${selectTemplateTag}/${name}.zip`;
```
hund030, line 121: Seems there's a bug. It should be `${templateConfig.templateDownloadBaseURL}/${selectedTag}/${name}.zip`

packages/fx-core/src/component/generator/generator.ts
yukun-dong: This logic is in samples.ts. https://github.com/OfficeDev/TeamsFx/blob/7f7c9560f1f5a13bb57eeb7a61feb29da8998d30/packages/fx-core/src/common/samples.ts#L35

packages/fx-core/src/component/generator/utils.ts
yukun-dong: Thanks. Fixed.

packages/fx-core/src/common/samples-config.json
```diff
@@ -1,6 +1,6 @@
 {
-    "baseUrl": "https://github.com/OfficeDev/TeamsFx-Samples/tree/v1.0.0/",
-    "defaultPackageLink": "https://github.com/OfficeDev/TeamsFx-Samples/archive/refs/tags/v1.0.0.zip",
+    "baseUrl": "https://github.com/OfficeDev/TeamsFx-Samples/tree/v1.1.0/",
+    "defaultPackageLink": "https://github.com/OfficeDev/TeamsFx-Samples/archive/refs/tags/v1.1.0.zip",
```
hund030, line 3: Why do we upgrade the sample version?

yukun-dong: The sample version is updated by spring release, not by me.

hund030: Then you may need to clean up your commits. It's better if your PR only includes your changes.

yukun-dong: Thanks. Updated.

hund030 (approved):

PR #6488, opened 2022-10-26
chore: apply v3 feature flag

IvanJobs:
Apply v3 feature flag for cicd.

### DIFF
--- a/packages/fx-core/templates/plugins/resource/cicd_v2/azdo/README.md
+++ b/packages/fx-core/templates/plugins/resource/cicd_v2/azdo/README.md
@@ -1,0 +1,48 @@
+# How to use pre-cooked pipelines on Azure DevOps
+
+## Prerequisites
+- Teams app projects are version controlled by Azure DevOps.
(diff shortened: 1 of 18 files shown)


packages/fx-core/templates/plugins/resource/cicd_v2/azdo/README.md
```diff
@@ -0,0 +1,48 @@
+# How to use pre-cooked pipelines on Azure DevOps
```
hund030, line 1: It's ok but I will suggest finding another location for these files. Maybe `fx-core/templates/components/cicd`?

hund030: I found there are similia files in `fx-core/templates/plugins/resource/cicd/`. Is there any change?

IvanJobs: Good suggestion. Will update it in another PR since this PR is to be merged before pre-release.

IvanJobs: > I found there are similia files in `fx-core/templates/plugins/resource/cicd/`. Is there any change?

Yes - package.json doesn't exist in root, so the very first step is not `npm ci` any more.

jayzhang (approved):

PR #6503, opened 2022-10-26
chore: add bot get failed error message

yuqizhou77:
- Add error message when get bot failed but the bot id is from env
- Also created a new aka.ms link which pointed to the same help doc while keeping the old aka.ms link valid also.
TODO: update doc


kuojianlu (approved):
Is it necessary to apply this change to `https://github.com/OfficeDev/TeamsFx/blob/dev/packages/fx-core/src/component/resource/botService/botService.ts#L72`, which is the old local debug flow?

nliu-ms (approved):

yuqizhou77:
> Is it necessary to apply this change to `https://github.com/OfficeDev/TeamsFx/blob/dev/packages/fx-core/src/component/resource/botService/botService.ts#L72`, which is the old local debug flow?

Will have another PR later today for that. Thanks for catching this!

PR #6504, opened 2022-10-27
chore: align name length

IvanJobs:
1. Length of local bot name has to be <= 35.
2. Length of remote bot handle has to be <= 42.
So min both -> 35.
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/16005991

### DIFF
--- a/packages/fx-core/src/component/resource/botService/common.ts
+++ b/packages/fx-core/src/component/resource/botService/common.ts
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 import { Uuid } from "node-ts-uuid";
+import { MaxLengths } from "./constants";
 
 export function genUUID(): string {
   return Uuid.generate();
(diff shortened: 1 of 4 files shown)


IvanJobs:
close to create another one.

packages/fx-core/src/component/resource/botService/common.ts
```diff
@@ -9,3 +10,7 @@ export function genUUID(): string {
 export function isHttpCodeOkOrCreated(code: number): boolean {
   return [200, 201].includes(code);
 }
+
+export function makeBotName(raw: string): string {
+  return raw.length > MaxLengths.BOT_NAME ? raw.substr(raw.length - MaxLengths.BOT_NAME) : raw;
```
eriolchan, line 15: According to this page https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr. This function is deprecated and it seem we also call it in wrong way.

PR #6507, opened 2022-10-27
chore: align bot name with v2

IvanJobs:
Align bot name with v2.

https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/16005991

### DIFF
--- a/packages/fx-core/src/component/driver/botAadApp/create.ts
+++ b/packages/fx-core/src/component/driver/botAadApp/create.ts
@@ -44,6 +44,7 @@ export class CreateBotAadAppDriver implements StepDriver {
       const createRes = await botRegistration.createBotRegistration(
         context.m365TokenProvider,
         args.name,
+        args.name,
         botConfig
       );
       if (createRes.isErr()) {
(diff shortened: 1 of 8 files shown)


packages/fx-core/src/component/driver/botAadApp/create.ts
```diff
@@ -44,6 +44,7 @@ export class CreateBotAadAppDriver implements StepDriver {
       const createRes = await botRegistration.createBotRegistration(
         context.m365TokenProvider,
         args.name,
+        args.name,
```
IvanJobs, line 47: Passing anything here doesn't matter too much because in remote bot registration name is controlled by bicep. The name passed here will not be used if you dive deep into the code.

yukun-dong (approved):

packages/fx-core/src/component/driver/botAadApp/create.ts
hund030: I think it's better to refine the parameter list rather than passing a useless parameter.

hund030 (approved):

packages/fx-core/src/component/driver/botAadApp/create.ts
IvanJobs: Agree - note it in future improvements.

eriolchan: I have the same feeling, let's talk about it to define a better interface for this case.

PR #6519, opened 2022-10-27
chore(debug): always show comment for customized env

kuojianlu:


swatDong (approved):
cc @formulahendry for the wordings to align in template (seems already aligned)

PR #6535, opened 2022-10-31
chore: call new envMW in V3

blackchoey:


packages/fx-core/src/core/FxCore.ts
```diff
@@ -526,10 +526,10 @@ export class FxCore implements v3.ICore {
           ui: context.userInteraction,
           logProvider: context.logProvider,
           telemetryReporter: context.telemetryReporter,
-          projectPath: context.projectPath!,
+          projectPath: inputs.projectPath!,
           platform: inputs.platform,
         };
-        await envUtil.readEnv(context.projectPath!, func.params.env);
+        // await envUtil.readEnv(context.projectPath!, func.params.env);
```
blackchoey, line 532: No longer need to read env manually. In the meanwhile, `context` does not contain `projectPath` property per my understanding, I help updated to `inputs.projectPath!` in driver context. Feel free to call out if previous code is expected and I will revert my changes.

1yefuwang1 (approved):
LGTM. But still need huajie's review.

jayzhang (approved):

nliu-ms (approved):

PR #6538, opened 2022-10-31
fix: resolve peer deps

yiqing-zhao:
@microsoft/teamsfx-cli	v1.1.4-rc.1
@microsoft/teamsfx-core	v1.18.0-rc.1
ms-teams-vscode-extension	v4.1.1-rc.1

CLI BREAKING CHANGE:
  

CLI feat commits:
  


packages/fx-core/src/common/local/constants.ts
```diff
@@ -145,7 +145,7 @@ export const TaskDefaultValue = Object.freeze({
     },
   },
   npmInstall: {
-    npmInstallArgs: ["--no-audit"],
+    npmInstallArgs: ["--no-audit", "--legacy-peer-deps"],
```
swatDong, line 148: Will this cause another issue: `teamsfx` requires `teams-js 2.0`, but user adds `teams-js 1.0`, then tab (SSO) does not work.

kuojianlu, line 148: Will `--legacy-peer-deps` has a requirement for `npm` version?

templates/scenarios/m365-tab-ts/tabs/package.json
```diff
@@ -29,7 +29,7 @@
     "scripts": {
         "dev:teamsfx": "env-cmd --silent -f .env.teamsfx.local npm run start",
         "start": "react-scripts start",
-        "install:teamsfx": "npm install",
+        "install:teamsfx": "npm install --legacy-peer-deps",
```
hund030, line 32: I wonder why do we add this? What's the issue?

wenytang-ms:
The related issue:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/16016793

packages/fx-core/src/common/local/constants.ts
yiqing-zhao: No. After testing out, it works properly, since teams-js 2.0 still provides an API translation layer (mapping v.1 SDK to v.2 SDK API calls).

yiqing-zhao: Both npm 6 and 8 can perform `--legacy-peer-deps`. Npm 6 would have warnings for peer dependency not installed. For instance, if users use `teamsfx-react@2.0.0` but `teamsfx@1.0.0`, it would have warnings.

yiqing-zhao:
Will close this PR. The bug should be resolved after 2.0.0 is released and will not impact user. Will tell testers to add the flag during testing.

PR #6553, opened 2022-10-31
chore(generator): add safe project name to replace map

yukun-dong:


packages/fx-core/src/component/generator/generator.ts
```diff
@@ -44,17 +45,18 @@ export class Generator {
     actionContext?: ActionContext
   ): Promise<Result<undefined, FxError>> {
     const appName = ctx.projectSetting?.appName;
-    const projectId = ctx.projectSetting?.projectId;
-    const nameReplaceMap = { ...{ appName: appName }, ...ctx.templateVariables };
-    const dataReplaceMap = { ...{ projectId: projectId }, ...nameReplaceMap };
+    const replaceMap = {
+      ...{ appName: appName, SafeProjectName: convertToAlphanumericOnly(appName) },
```
hund030, line 49: Maybe also add ProjectName as it is used for filename replacement.
And I wonder if we can just remove appName from the map since there isn't such a placeholder in v3 templates.

yukun-dong: https://github.com/OfficeDev/TeamsFx/blob/58a540929b3707b28925270131cae4dd2c436615/templates/scenarios/command-and-response-ts/package.json.tpl#L2 Will this be removed in the future?

hund030 (approved):

packages/fx-core/src/component/generator/generator.ts
hund030: I think so. Let me follow up this.

PR #6563, opened 2022-11-01
chore(debug): resolve env:teamsAppId placeholder in launch.json

kuojianlu:


XiaofuHuang (approved):

packages/vscode-extension/src/debug/commonUtils.ts
```diff
@@ -394,3 +395,21 @@ export class Step {
...
+  if (result.isErr()) {
+    throw result.error;
+  }
+
+  const teamsAppId = result.value.TEAMS_APP_ID;
+  if (teamsAppId === undefined) {
+    throw new UserError(
+      ExtensionSource,
+      ExtensionErrors.TeamsAppIdNotFoundError,
+      `TEAMS_APP_ID is missing in ${env} environment.`
```
kuojianlu, line 410: Will extract the strings to package.nls.json when updating wordings afterwards.

swatDong (approved):

packages/vscode-extension/src/debug/commonUtils.ts
```diff
@@ -394,3 +395,21 @@ export class Step {
...
   }
 }
+
+export async function getV3TeamsAppId(projectPath: string, env: string): Promise<string> {
+  const result = await envUtil.readEnv(projectPath, env);
+  if (result.isErr()) {
+    throw result.error;
+  }
+
+  const teamsAppId = result.value.TEAMS_APP_ID;
```
blackchoey, line 405: In use existing scenario, user may use other environment variable to set teams app id. What's our attitude to this case? @kuojianlu @swatDong  @jidddddd

kuojianlu: For existing scenario, user can directly use the raw teams app id instead of the placeholder in launch url.
@swatDong Any other ideas?

swatDong: Use `env:TEAMS_APP_ID` as placeholder?
E.g.,
- `local:TEAMS_APP_ID` means `TEAMS_APP_ID` in *teamsfx/.env.local*
- `dev:TEAMS_APP_ID` means `TEAMS_APP_ID` in *teamsfx/.env.dev*
- `TEAMS_APP_ID` means ask for env first, then from *teamsfx/.env.<env>*
- `my-dev-1:MY_TEAMS_APP_ID` means `MY_TEAMS_APP_ID` in *teamsfx/.env.my-dev-1*

Anyway, consider improve after happy path.

blackchoey: BTW, do we have env selection when local debug? If yes maybe just use `TEAMS_APP_ID` as placeholder (no need to add `dev:`)?

swatDong: Currently we have (4.1.0), but only for remote. For local better to not ask, that's why we need `local:`.
As if we support `local:`, user can use any `<env>`, e.g., `dev:`.

swatDong: That's a tradeoff:
- if we use `dev:TEAMS_APP_ID` by default, the first remote launch does not require env selection, but user needs to change that for another env.
- if we use `TEAMS_APP_ID` by default, all remote launches require env selection.

I'll discuss with PM to determine the final UX, but the underlying implementation is the same.

PR #6564, opened 2022-11-01
chore: change teamsfx.yml to app.yml

blackchoey:


packages/fx-core/resource/package.nls.json
```diff
@@ -1039,5 +1039,5 @@
   "debug.warningMessage": "To debug applications in Teams, your localhost server must be on HTTPS.\nFor Teams to trust the self-signed SSL certificate used by the toolkit, a self-signed certificate must be added to your certificate store.\n You may skip this step, but you'll have to manually trust the secure connection in a new browser window when debugging your apps in Teams.\nFor more information \"https://aka.ms/teamsfx-ca-certificate\".",
   "debug.warningMessage2": " You may be asked for your account credentials when installing the certificate.",
   "debug.install": "Install",
-  "driver.spfx.error.noValidAppCatelog": "There is no valid app catalog in your tenant. You can update the property 'createAppCatalogIfNotExist' in teamsfx.yml to true if you want us to create it for you or you can create it by yourself."
+  "driver.spfx.error.noValidAppCatelog": "There is no valid app catalog in your tenant. You can update the property 'createAppCatalogIfNotExist' in app.yml to true if you want us to create it for you or you can create it by yourself."
```
blackchoey, line 1042: @HuihuiWu-Microsoft FYI. Since driver does not know the yaml file name, not sure whether we want to change the wordings here to not mention yaml file name.

xzf0587 (approved):

packages/fx-core/resource/package.nls.json
HuihuiWu-Microsoft: Yes, I think we should update the wording to remove yml file name and provide the link to let user know how to modify the property. Maybe like:
You can update the property 'createAppCatalogIfNotExist' to true if you want us to create it for you or you can create it by yourself, see <link> for more details. cc @xu-hong 
Do we already have a public link? Or I can update it later after we have one.

blackchoey: I will merge first. You can update the wording later as it's not a blocking thing.

PR #6565, opened 2022-11-01
chore: add spfx generator and spfx-tab scenario template

HuihuiWu-Microsoft:


JerryYangKai (commented):
Hi @HuihuiWu-Microsoft , after reviewing the code I am curious about the isOfficialSPFx part, in the logic is using this check to generator different template.
So could you please give a brief description about the background and why we using two template spfx-tab and spfx-pre-tab. Thanks

HuihuiWu-Microsoft:
> 

Hi @JerryYangKai , the reason why we use different templates is that both SPFx v1.15.0 and v1.16.0-beta.1 is supported in our Teams Toolkit now. There's slight difference between the templates. So I have 'spfx-tab' for v1.15.0 and 'spfx-tab-pre' for v1.16.0-beta.1. 
But this won't be long since we'll upgrade to support SPFx v1.16.0 once the official release shipped.

JerryYangKai (approved):

jayzhang (approved):

PR #6571, opened 2022-11-01
chore(core): adjust yml engine output

jayzhang:


packages/fx-core/src/component/coordinator/index.ts
```diff
@@ -327,14 +330,49 @@ export class Coordinator {
           }
         }
         if (unresolvedPlaceHolders.length === 0) {
```
1yefuwang1, line 332: We should continue to call `execute` even if there are unresolvedPlaceholders. Because there could be cross driver references. e.g. Driver2's input depends on Driver1's output. So the placeholder can't be resolve until execution happens.

1yefuwang1 (approved):

1yefuwang1 (approved):

xzf0587 (approved):

PR #6583, opened 2022-11-02
chore(debug): add env/generate driver

kuojianlu:


packages/fx-core/resource/package.nls.json
```diff
@@ -1042,6 +1042,7 @@
   "debug.warningMessage2": " You may be asked for your account credentials when installing the certificate.",
   "debug.install": "Install",
   "driver.spfx.error.noValidAppCatelog": "There is no valid app catalog in your tenant. You can update the property 'createAppCatalogIfNotExist' in app.yml to true if you want us to create it for you or you can create it by yourself.",
-  "driver.tools.error.invalidParameter": "Following parameter is missing or invalid for %s action: %s."
-
+  "driver.tools.error.invalidParameter": "Following parameter is missing or invalid for %s action: %s.",
+  "driver.env.error.invalidParameter": "Following parameter is missing or invalid for %s action: %s.",
```
swatDong, line 1046: Any idea to share one `invalidParameter` message? Or it's expected to create for each driver?

swatDong (approved):

packages/fx-core/resource/package.nls.json
kuojianlu: Currently each driver has its own error handling, so it is expected to create its own error type as well as the error message. However, we are happy if the driver framework provides some shared error hanlding logic afterwards.

XiaofuHuang (approved):

packages/fx-core/src/component/driver/env/generate.ts
```diff
@@ -0,0 +1,93 @@
...
+    try {
+      this.validateArgs(args);
+
+      if (args.target) {
+        await fs.ensureFile(args.target);
+        const envs = dotenv.parse(await fs.readFile(args.target));
+        const content = Object.entries({ ...envs, ...args.envs })
+          .map(([key, value]) => `${key}=${value}`)
+          .join(os.EOL);
+        await fs.writeFile(args.target, content);
```
XiaofuHuang, line 48: If `.env` file contains comments, will they be still existing?

kuojianlu: No. To improve it later if needed.

jayzhang (approved):

PR #6578, opened 2022-11-01
chore(localdebug): add lifecycle task for v3

XiaofuHuang:
Configuration in `task.json`
```
        {
            "label": "Create resources",
            "type": "teamsfx",
            "command": "provision",
            "args": {
                "configFile": "${workspaceFolder}/.fx/teamsfx.local.yml",
                "env": "local"
            }
        }
```

TODO: update log and telemetry

PR #6584, opened 2022-11-02
chore(localdebug): add debug config file into tab sso v3 template

XiaofuHuang:
Add debug into tab sso v3 template

### DIFF
--- a/templates/scenarios/sso-tab-js/.vscode/tasks.json
+++ b/templates/scenarios/sso-tab-js/.vscode/tasks.json
@@ -55,9 +55,6 @@
             "type": "shell",
             "command": "npm run dev:teamsfx",
             "isBackground": true,
-            "options": {
-                "cwd": "${workspaceFolder}/tabs"
-            },
             "problemMatcher": {
                 "pattern": {
                     "regexp": "^.*$",
--- a/templates/scenarios/sso-tab-js/.vscode/tasks.json
+++ b/templates/scenarios/sso-tab-js/.vscode/tasks.json
@@ -2,44 +2,74 @@
     "version": "2.0.0",
     "tasks": [
         {
-            "label": "Pre Debug Check & Start All",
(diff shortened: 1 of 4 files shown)


swatDong:
Why adding an empty `.env.local`?
- Could just remove it since no value there?
- Or, add some placeholders and comments (similar to .env.dev)?

swatDong (dismissed):

XiaofuHuang:
> Why adding an empty `.env.local`?
> 
> * Could just remove it since no value there?
> * Or, add some placeholders and comments (similar to .env.dev)?

Now if `apply` api will throw error if `.env` file does not exist. So the empty .env.local file make the f5 work.

I added some placeholders and comments in .env.local.

kuojianlu:
Should we also update the placeholder for launch remote, e.g. `dev:teamsAppId`?

XiaofuHuang:
> Should we also update the placeholder for launch remote, e.g. `dev:teamsAppId`?

Now v3 will have a question to choose env.
image

Will remote show the same question?

swatDong (approved):

templates/scenarios/sso-tab-js/.vscode/tasks.json
```diff
@@ -2,45 +2,72 @@
...
-            "label": "prepare local environment",
-            "type": "shell",
-            "command": "exit ${command:fx-extension.pre-debug-check}",
-            "presentation": {
-                "reveal": "never"
+            "label": "Create resources",
+            "type": "teamsfx",
+            "command": "provision",
+            "args": {
+                "configFile": "${workspaceFolder}/teamsfx/app.local.yml",
```
kuojianlu, line 34: -> `template`

XiaofuHuang: Will updated in PR: https://github.com/OfficeDev/TeamsFx/pull/6578

PR #6597, opened 2022-11-02
feat(debug): add teamsfx-run-utils package

kuojianlu:
E2E TEST: N/A


packages/run-utils/src/index.ts
```diff
@@ -0,0 +1,62 @@
...
+    throw new Error(`${projectPath} does not exist.`);
+  }
+
+  if (!(await fs.pathExists(envPath))) {
+    throw new Error(`${envPath} does not exist.`);
+  }
+
+  const settings = await readSettings(projectPath);
+  if (!settings.trackingId) {
+    throw new Error("trackingId is missing in settings.json");
```
swatDong, line 49: Consider later:
How to handle decrypt error, throw or ignore?
Or, add one more crypt option to let user to choose whether to decrypt or not...

swatDong (approved):

packages/run-utils/package.json
```diff
@@ -0,0 +1,48 @@
+{
+  "name": "@microsoft/teamsfx-run-utils",
+  "version": "1.0.0",
+  "description": "Provides the utilities to run TeamsFx project locally.",
+  "main": "./lib/index.js",
+  "scripts": {
+    "clean": "rimraf lib",
+    "build": "rimraf lib && npx tsc -p ./",
```
swatDong, line 8: Consider later - use webpack to bundle all src and node_modules into one single file.

packages/run-utils/src/index.ts
kuojianlu: Sure.

packages/run-utils/package.json
kuojianlu: Sure.

PR #6610, opened 2022-11-03
chore(coordinator): pre provision check

jayzhang:


packages/fx-core/src/component/coordinator/index.ts
```diff
@@ -104,6 +105,17 @@ export const Feature2TemplateName: any = {
...
+const M365Actions = [
+  "botAadApp/create",
+  "teamsApp/create",
+  "teamsApp/update",
+  "aadApp/create",
+  "aadApp/update",
+  "m365Bot/create",
+  "m365Bot/update",
+];
+const AzureActions = ["arm/deploy"];
```
xzf0587, line 117: Could we add optional properties in action interface? Coordinator can use the properties to get the action type.
If so, we will not modify the coordinator code when the action changes

jayzhang: Can do it, but not now.

hund030 (approved):

PR #6619, opened 2022-11-03
chore(generator): update log and telemetry

yukun-dong:


packages/fx-core/src/component/generator/generator.ts
```diff
@@ -64,6 +63,10 @@ export class Generator {
     };
     await actionContext?.progressBar?.next(ProgressMessages.generateTemplate);
     await this.generate(generatorContext, TemplateActionSeq);
+    merge(actionContext?.telemetryProps, {
+      [TelemetryProperty.TemplateName]: generatorContext.name,
```
hund030, line 67: Is it possible to add the template name at the beginning of the method? So that error event also contains this property.

yukun-dong: How about adding template name to error event? Since begin event is sent before generateTemplate() is called, adding something like input[name] to the telemetry may make actionMW non-generic. You can check the latest commit.

hund030 (approved):

hund030 (approved):

PR #6653, opened 2022-11-07
chore: update hover provider

nliu-ms:
1) update hover provider
2) validation log
3) fix client error
4) special case for spfx
hover:
image
codelens:
image


nliu-ms:
E2E TEST: n/a

packages/vscode-extension/src/extension.ts
```diff
@@ -746,7 +746,9 @@ function registerCodelensAndHoverProviders(context: vscode.ExtensionContext) {
   const manifestTemplateSelector = {
     language: "json",
     scheme: "file",
-    pattern: `**/${TemplateFolderName}/${AppPackageFolderName}/manifest.template.json`,
+    pattern: isV3Enabled()
```
HuihuiWu-Microsoft, line 749: There might be problems if we only support manifest.template.json:
1. For SPFx, there's a manifest.template.local.json that also need codelens
2. For other templates that's customized by users, is it designed to not support?

HuihuiWu-Microsoft (approved):

jayzhang (approved):

PR #6661, opened 2022-11-07
chore: restore process.env for V3 lifecycles and actions

1yefuwang1:
fixes https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_sprints/taskboard/Teams%20Extensibility%20E2E%20Team/Microsoft%20Teams%20Extensibility/CY22-10.3?workitem=16161691


packages/fx-core/src/component/middleware/processEnvRestorerMW.ts
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+import { Middleware, NextFunction } from "@feathersjs/hooks";
+import { CoreHookContext } from "../../core/types";
+import _ from "lodash";
+
+export const ProcessEnvRestorerMW: Middleware = async (
```
blackchoey, line 7: can it be part of Huajie's envLoaderMW? so we don't need to add another middleware?

1yefuwang1: Adding it to existing middleware will add a big `try finally` block in existing middleware code. which is kind of messy and harder test. But the advantage is we don't need to modify the decorator list for each lifecycle.

blackchoey: One thing to consider is whether `envLoaderMW` is responsible to restore env at the end. My answer is yes (I'm not aware of a case that don't need to restore the env). Also, if we make 2 separate middleware, we need a way to ensure everyone (including new members) knows he/she needs to use `ProcessEnvRestorerMW` along with `envLoaderMW`. Failing to do so may introduce bugs. You can discuss with Huajie to make the final decision.

By the way, can you share more about your concern of `messy and harder test`?

1yefuwang1: "By the way, can you share more about your concern of messy and harder test?" There are no test for envloaderMW, I'll have to write from scratch.

jayzhang (approved):

blackchoey (approved):

PR #6666, opened 2022-11-08
chore: error message when tenant not match

yuqizhou77:


yuqizhou77:
- fixed a bug I noticed about asking for consent but the yml does not contain any m365 actions
- added error message when tenant does match

PR #6667, opened 2022-11-08
feat(debug): add m365 bot driver

kuojianlu:
- add create and update actions

E2E TEST: N/A


packages/fx-core/src/component/driver/m365Bot/create.ts
```diff
@@ -0,0 +1,90 @@
...
+
+      const botRegistration: IBotRegistration = {
+        botId: args.botId,
+        name: args.name,
+        description: "",
+        iconUrl: "",
+        messagingEndpoint: "",
+        callingEndpoint: "",
+      };
+      await AppStudioClient.createBotRegistration(tokenResult.value, botRegistration, false);
```
swatDong, line 57: What if reusing [localBotRegistration](https://github.com/OfficeDev/TeamsFx/blob/dev/packages/fx-core/src/component/resource/botService/botRegistration/localBotRegistration.ts)?

packages/fx-core/src/component/driver/m365Bot/interface/updateM365BotArgs.ts
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+export interface UpdateM365BotArgs {
+  botId: string;
+  messagingEndpoint: string;
+  name?: string;
```
swatDong, line 7: Just for m365bot actions, prefer to keep the same args for both `create` and `update`.
Or, reconsider to only one action `m365Bot/createOrUpdate` ?

packages/fx-core/src/component/driver/m365Bot/create.ts
kuojianlu: `LocalBotRegistration.createBotRegistration` needs `aadDisplayName` as input, which is not ready for use to use. Ivan will refine the interface. Will use it then.

packages/fx-core/src/component/driver/m365Bot/interface/updateM365BotArgs.ts
swatDong: I just realized that user may get confused if they can customize `name` in both `create` and `update`. So one single action seems clearer to user.

kuojianlu: Agree to use only one action `m365Bot/createOrUpdate`.

swatDong (dismissed):
There are already some references for `m365bot/create` and `m365bot/update` in core sources, remind to update those as well : )

kuojianlu:
> There are already some references for `m365bot/create` and `m365bot/update` in core sources, remind to update those as well : )

Nice reminder. Let me update them in another PR.

swatDong (approved):

jayzhang (approved):

PR #6686, opened 2022-11-08
chore(generator): error handling

yukun-dong:


packages/fx-core/src/component/generator/generator.ts
```diff
@@ -153,12 +156,25 @@ async function sampleDefaultOnActionError(
...
       throw new Error(error.message);
   }
 }
+
+async function wrapError(run: () => Promise<void>) {
+  try {
+    await run();
+  } catch (error) {
+    if (error instanceof BaseComponentInnerError) {
+      throw error.toFxError();
```
hund030, line 175: I think there are too many re-throws. Is it possible to transform the error in context.onActionError?
Like: 
```
throw (new FetchZipFromUrlError(context.zipUrl!)).toFxError();
```

PR #6703, opened 2022-11-09
feat(vsc): add in-product guide

tecton:
Covered by A/B test. Will show an in-product guide for workflow bot.

E2E TEST: N/A


LongOddCode (approved):

packages/vscode-extension/src/controls/webviewPanel.ts
```diff
@@ -87,6 +91,23 @@ export class WebviewPanel {
     // This happens when the user closes the panel or when the panel is closed programatically
     this.panel.onDidDispose(() => this.dispose(), null, this.disposables);
 
+    if (TreatmentVariableValue.inProductDoc) {
+      this.panel.onDidChangeViewState(
+        (e) => {
+          const panel = e.webviewPanel;
+          ExtTelemetry.sendTelemetryEvent(TelemetryEvent.InteractWithInProductDoc, {
```
HuihuiWu-Microsoft, line 98: Will this be triggered for all types of panel?

tecton: Yes, sample and in product doc panel. Do you think it's better to filter the in-product panel?

HuihuiWu-Microsoft: Yes, because the event name sounds like only related to in-product doc. You can filter the in-product panel or change the event name. Both are OK to me.

HuihuiWu-Microsoft (approved):

packages/vscode-extension/src/controls/webviewPanel.ts
tecton: Updated.

jayzhang (approved):

PR #6705, opened 2022-11-09
chore(core): pre provision check for vs

jayzhang:


packages/fx-core/src/component/coordinator/index.ts
```diff
@@ -290,6 +291,69 @@ export class Coordinator {
...
+    const projectModel = maybeProjectModel.value;
+    const cycles: ILifecycle[] = [
+      projectModel.registerApp,
+      projectModel.provision,
+      projectModel.configureApp,
+    ].filter((c) => c !== undefined) as ILifecycle[];
+
+    // 2. check each cycle
+    for (const cycle of cycles) {
+      cycle.resolvePlaceholders();
```
1yefuwang1, line 336: Maybe it's better to check the return value of `resolvePlaceholders()` to get unresolved placeholders,

jayzhang: There is no need to check whether placeholders are resolved or not, we only need the final value

packages/fx-core/src/component/coordinator/index.ts
```diff
@@ -290,6 +291,69 @@ export class Coordinator {
...
+      for (const driver of cycle.driverDefs) {
+        if (AzureActions.includes(driver.uses)) {
+          res.needAzureLogin = true;
+        }
+        if (M365Actions.includes(driver.uses)) {
+          res.needM365Login = true;
+        }
+      }
+      const firstDriver = cycle.driverDefs[0] || undefined;
```
1yefuwang1, line 345: It's possible that driverDefs is an empty array, in which case `cycle.driverDefs[0]` yields `undefined`

xzf0587 (dismissed):

packages/fx-core/src/component/coordinator/index.ts
jayzhang: cycle.driverDefs[0] === undefined is ok for the following code to run correctly

jayzhang: image

LongOddCode (approved):

1yefuwang1 (approved):

PR #6730, opened 2022-11-10
chore: support vs tab sso

xiaolang124:
1.Add vs tab sso local yml and related codes.


packages/fx-core/src/component/middleware/envMW.ts
```diff
@@ -37,7 +37,9 @@ export const envLoaderMWImpl: Middleware = async (ctx: CoreHookContext, next: Ne
       ctx.result = err(envListRes.error);
       return;
     }
-    question.staticOptions = envListRes.value;
+    question.staticOptions = envListRes.value.filter(
```
swatDong, line 40: What's this change for?

packages/server/src/serverConnection.ts
```diff
@@ -116,6 +116,9 @@ export default class ServerConnection implements IServerConnection {
     token: CancellationToken
   ): Promise<Result<Void, FxError>> {
     const corrId = inputs.correlationId ? inputs.correlationId : "";
+    if (inputs.env && inputs.env === environmentManager.getLocalEnvName()) {
```
swatDong, line 119: Could this change be in VS extension - set `inputs.workflowFilePath` before calling `server.provisionResourcesRequest`.

packages/fx-core/src/component/middleware/envMW.ts
swatDong: Is it possible to ensure `inputs.env` is always set for F5, then no need to have this change?

xiaolang124: For now, core support multi env. So, we need to exclude 'local' environment.
image

swatDong: So this is not for F5?
I prefer not adding local-specific logic. `local` is just the default name and user can customize any env name (`local1`, `my-desktop`, `my-laptop`, ...)
And, what if user really want to provision `local` environment?

swatDong: BTW, for F5 (prepare teams app deps), let's just hard code `local` as the default env for the 1st release.
I believe in later release we need some config/setting to let user to choose which env to be prepared.

swatDong: In addition, if there should be local-specific logic, add to the very top level (extension, template, user settings) instead of core.

xiaolang124: Make sense, I will change it back.

swatDong (approved):

jayzhang (approved):

PR #6755, opened 2022-11-11
feat(vsc): remove subscription node and adapt to new API in v3

tecton:
1. Remove subscription node.
2. refactor getAzureSolutionSettings()
3. refactor getSettingsVersion()

Related bug / task:
1. https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/16112014
2. https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/16112898
3. https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/16139513

E2E TEST: N/A


HuihuiWu-Microsoft (approved):

packages/vscode-extension/src/commonlib/azureLogin.ts
```diff
@@ -356,6 +356,9 @@ export class AzureAccountManager extends login implements AzureAccountProvider {
    * set tenantId and subscriptionId
    */
   async setSubscription(subscriptionId: string): Promise<void> {
+    if (isV3Enabled()) {
+      return;
```
swatDong, line 360: Is this a workaround to make current v3 template work, or a final solution that V3 totally does not need `setSubscription`?

tecton: Since there will be no "subscriptionInfo.json" in V3, our toolkit will not store the sub info anymore. So, I think `setSubscription` will be deprecated in V3. Users will only need to choose subscription before provision.

swatDong (approved):

PR #6771, opened 2022-11-14
chore(vsc): remove isM365 in V3

tecton:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/16111911

### DIFF
--- a/packages/vscode-extension/src/handlers.ts
+++ b/packages/vscode-extension/src/handlers.ts
@@ -292,7 +292,7 @@ export async function getIsFromSample(projectPath?: string) {
 }
 
 export async function getIsM365(): Promise<boolean | undefined> {
-  if (core) {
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/handlers.ts
```diff
@@ -292,7 +292,7 @@ export async function getIsFromSample(projectPath?: string) {
 }
 
 export async function getIsM365(): Promise<boolean | undefined> {
-  if (core) {
+  if (core && isV3Enabled()) {
```
HuihuiWu-Microsoft, line 295: Should be !isV3Enabled()?

tecton: Thanks, updated.

HuihuiWu-Microsoft (approved):

PR #6778, opened 2022-11-14
chore: render template with app name

nliu-ms:
1) Render manifest/aad template and yml with app name:
image

2) Avoid environment variables with ${{env_name}} pattern being rendered to $

### DIFF
--- a/templates/scenarios/js/sso-tab/aad.manifest.template.json.tpl
+++ b/templates/scenarios/js/sso-tab/aad.manifest.template.json.tpl
@@ -1,7 +1,10 @@
+{{=<% %>=}}
 {
     "id": "${{AAD_APP_OBJECT_ID}}",
     "appId": "${{AAD_APP_CLIENT_ID}}",
-    "name": "sso-tab-aad",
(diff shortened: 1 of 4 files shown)


templates/scenarios/js/sso-tab/aad.manifest.template.json.tpl
```diff
@@ -1,7 +1,10 @@
+{{=<% %>=}}
 {
     "id": "${{AAD_APP_OBJECT_ID}}",
     "appId": "${{AAD_APP_CLIENT_ID}}",
-    "name": "sso-tab-aad",
+<%={{ }}=%>
+    "name": "{{appName}}-aad",
```
hund030, line 6: What about using `<%appName%>`? You don't need to surround the placeholder with `<%={{ }}=%>` and `{{=<% %>=}}`.

nliu-ms: ok, as discussed with @hund030 , he will update the generator to use other pattern as the placeholder.

PR #6780, opened 2022-11-14
chore: change relative path logic and use longer time

yukun-dong:


packages/fx-core/src/component/generator/generatorAction.ts
```diff
@@ -96,7 +101,7 @@ export const fetchZipFromUrlAction: GeneratorAction = {
     if (!context.zipUrl) {
       throw new MissKeyError("zipUrl");
     }
-    context.zip = await fetchZipFromUrl(context.zipUrl, defaultTryLimits, defaultTimeoutInMs);
+    context.zip = await fetchZipFromUrl(context.zipUrl, context.tryLimits, context.timeoutInMs);
```
hund030, line 104: I will expect there are null checks.
For example:
```
await fetchZipFromUrl(context.zipUrl, context.tryLimits ?? defaultTryLimits, context.timeoutInMs ?? defaultTimeoutInMs);
```

packages/fx-core/src/component/generator/generator.ts
```diff
@@ -53,9 +61,11 @@ export class Generator {
     const replaceMap = ctx.templateVariables;
     const generatorContext: GeneratorContext = {
       name: language ?? commonTemplateName,
-      relativePath: scenario,
+      relativePath: `${scenario}/`,
       destination: destinationPath,
       logProvider: ctx.logProvider,
+      tryLimits: templateDefaultTryLimits,
+      timeoutInMs: templateDefaultTimeoutInMs,
```
hund030, line 68: What about making default timeout 20s? so that we don't need to explicitly specify timeout for templates. So as try limits.

hund030 (approved):

hund030 (approved):

PR #6791, opened 2022-11-14
chore: separate sample id with sample app folder

HuihuiWu-Microsoft:


yukun-dong (approved):

dooriya (approved):

packages/fx-core/src/common/samples.ts
```diff
@@ -34,7 +34,7 @@ class SampleProvider {
           configuration: sample.configuration,
           link: sample.packageLink ?? sampleConfig.defaultPackageLink,
           suggested: sample.suggested,
-          url: sample.url ?? sampleConfig.baseUrl,
+          url: sample.relativePath ? sample.url : sample.url ?? sampleConfig.baseUrl,
```
tecton, line 37: Looks like the logic is same with old one. `relativePath` is not used?

jayzhang (approved):

packages/fx-core/src/common/samples.ts
tecton: relativePath is used to determine whether it's external sample.

tecton (approved):

packages/fx-core/src/common/samples.ts
HuihuiWu-Microsoft: `relativePath` is used for external sample which will use its own url instead of base version.

PR #6804, opened 2022-11-15
chore: support local debug in bot registration

IvanJobs:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/16270707

Add code to support driver `botFramework/createOrUpdateBot` in bot registration.


packages/fx-core/src/component/driver/botFramework/createOrUpdateBot.ts
```diff
@@ -38,32 +39,20 @@ export class CreateOrUpdateBotFrameworkBotDriver implements StepDriver {
...
+        botId: args.botId,
+        name: args.name,
+        description: args.description ?? "",
+        iconUrl: args.iconUrl ?? "",
+        messagingEndpoint: args.messagingEndpoint,
+        callingEndpoint: "",
+      };
+
+      const botRegistration: BotRegistration = new LocalBotRegistration();
+      await botRegistration.createOrUpdateBotRegistration(
```
kuojianlu, line 52: Check if the result is error.

packages/fx-core/src/component/resource/botService/botRegistration/botRegistration.ts
```diff
@@ -30,6 +30,7 @@ export class Constants {
   public static readonly BOT_REGISTRATION: string = "BotRegistration";
   public static readonly CREATE_BOT_REGISTRATION: string = "createBotRegistration";
   public static readonly UPDATE_MESSAGE_ENDPOINT: string = "updateMessageEndpoint";
+  public static readonly UPDATE_BOT_REGISTRATION: string = "updateBotRegistration";
```
eriolchan, line 33: Given we will provide updateBotRegistration() for updating more than message endpoint, do we still need to keep the original updateMessageEndpoint()?

packages/fx-core/src/component/resource/botService/botRegistration/utils.ts
```diff
@@ -0,0 +1,22 @@
...
+import { IBotRegistration } from "../appStudio/interfaces/IBotRegistration";
+
+export class Utils {
+  public static mergeIBotRegistration(
+    local: IBotRegistration,
+    remote: IBotRegistration
+  ): IBotRegistration {
+    return {
+      botId: local.botId ?? remote.botId,
+      name: local.name ? local.name : remote.name,
```
eriolchan, line 13: It seems all logic looks like ```local.prop ?? remote.prop```. Why only L12 use this pattern, but the rest use another style?

IvanJobs: This is because in IBotRegistration, botId is the only one with type `string | undefined` and the rest are `string`.
 `??` wasn't used for the rest, for example, `iconUrl: local.iconUrl ? local.iconUrl : remote.iconUrl`,  because `??` doesn't work for empty strings. 
If `local.iconUrl` is empty, we should respect the value from remote.

packages/fx-core/src/component/resource/botService/botRegistration/botRegistration.ts
IvanJobs: It is still referenced by legacy code in some `configure` lifecycle of `botService.ts`. We may delete this interface if we're ready to remove all those legacy code.

packages/fx-core/src/component/driver/botFramework/createOrUpdateBot.ts
IvanJobs: Good catch! Will do.

packages/fx-core/src/component/resource/botService/botRegistration/utils.ts
kuojianlu: Can use `local.iconUrl || remote.iconUrl` for simplicity?

kuojianlu (approved):

packages/fx-core/src/component/driver/botFramework/createOrUpdateBot.ts
```diff
@@ -38,30 +39,23 @@ export class CreateOrUpdateBotFrameworkBotDriver implements StepDriver {
...
+      const botRegistrationData: IBotRegistration = {
+        botId: args.botId,
+        name: args.name,
+        description: args.description ?? "",
+        iconUrl: args.iconUrl ?? "",
+        messagingEndpoint: args.messagingEndpoint,
+        callingEndpoint: "",
+      };
+
+      const botRegistration: BotRegistration = new LocalBotRegistration();
```
swatDong, line 51: Is it possible to make `LocalBotResitration` a class property or static variable?
Creating new instance every time seems not a good practice.

yukun-dong (approved):

packages/fx-core/src/component/driver/botFramework/createOrUpdateBot.ts
IvanJobs: I agree. There's a `factory` which may return a static instance everytime, I may use it in another PR.

swatDong (approved):

xzf0587 (approved):

jayzhang (approved):

eriolchan (approved):

PR #6824, opened 2022-11-15
chore: publish teams app in developer portal

yuqizhou77:
- after TDP integration feature is turned on:
image
We will have "publish in developer poral" to replace the current link to developer portal which will ask user to select env and manifest file, and then redirect user to the "publish to org" page in developer portal. The link will be something like https://dev.teams.microsoft.com/apps/{appid}/distributions/app-catalog?login_hint={account}&referrer=teamstoolkit_vsc

image

- update url handler as updating url in dev portal


nliu-ms (dismissed):

packages/fx-core/resource/package.nls.json
```diff
@@ -294,7 +294,7 @@
   "plugins.appstudio.overwriteAndUpdate": "Overwrite and update",
   "plugins.appstudio.emptyAppPackage": "Could not find any files in the package of app %s.",
   "plugins.appstudio.unprocessedFile": "%s was not processed by Teams Toolkit.",
-  "plugins.appstudio.viewDevPortal": "View in Developer Portal",
```
yuqizhou77, line 297: this has not been localized yet. So I feel it is safe to do the rename here.

packages/vscode-extension/package.nls.json
```diff
@@ -171,6 +171,8 @@
     "teamstoolkit.commandsTreeViewProvider.publish.blockTooltip": "Cannot run command when publishing package. Try again when publish is finished.",
     "teamstoolkit.commandsTreeViewProvider.publish.running": "Publishing...",
     "teamstoolkit.commandsTreeViewProvider.publishDescription": "Publish to Teams",
+    "teamstoolkit.commandsTreeViewProvider.publishInDevPortalTitle": "Publish in Developer Portal",
+    "teamstoolkit.commandsTreeViewProvider.publishInDevPortalDescription": "Publish in Developer Portal",
```
tecton, line 175: Is description same with title?

packages/vscode-extension/src/handlers.ts
```diff
@@ -965,6 +965,13 @@ export async function publishHandler(args?: any[]): Promise<Result<null, FxError
   return await runCommand(Stage.publish);
 }
 
+export async function publishInDeveloperPortalHandler(
+  args?: any[]
+): Promise<Result<null, FxError>> {
+  ExtTelemetry.sendTelemetryEvent(TelemetryEvent.PublishStart, getTriggerFromProperty(args));
```
tecton, line 971: Do we need to use a new event? `Publish Start` is also used by publish action. We cannot tell the difference between them.

packages/vscode-extension/src/treeview/webViewProvider/deploymentView.tsx
```diff
@@ -3,6 +3,7 @@ import "./tree.scss";
 import { getCurrentTheme, Theme } from "./theme";
```
tecton, line 3: This file is not used actually.

packages/vscode-extension/package.nls.json
yuqizhou77: > Is description same with title?

I would keep it same for now as this is the same as what we do for "publish to teams" item. Also double confirming with PM. Will update in another PR if we decide to change.

tecton (approved):

tecton (approved):

nliu-ms (approved):

jayzhang (approved):

PR #6835, opened 2022-11-16
chore(generator): change select version logic

yukun-dong:


packages/fx-core/src/component/generator/constant.ts
```diff
@@ -10,7 +10,8 @@ export const defaultTimeoutInMs = 20000;
 export const sampleDefaultTimeoutInMs = 40000;
 export const templateFileExt = ".tpl";
 export const templateAlphaVersion = "templates-0.0.0-alpha";
```
JerryYangKai, line 12: I am not very sure about this, just double confirm, this is templates-0.0.0-alpha  not   templates@0.0.0-alpha

packages/fx-core/src/component/generator/constant.ts
```diff
@@ -10,7 +10,8 @@ export const defaultTimeoutInMs = 20000;
...
-export const templateBetaVersion = "templates-0.0.0-beta";
+export const templateRcVersion = "templates@0.0.0-rc";
+export const templateV3Version = "^2.0.0";
```
JerryYangKai, line 14: This constant is used in the logic to compare the max version, so this is just a default version, right? 
So this name will cause some misunderstand when first see it.

packages/fx-core/src/component/generator/constant.ts
yukun-dong: Yes, it's templates-0.0.0-alpha.

packages/fx-core/src/component/generator/constant.ts
hund030, line 14: I think if we do not update template-config.json, how to download fallback zip files?

packages/fx-core/src/component/generator/constant.ts
```diff
@@ -10,7 +10,7 @@ export const defaultTimeoutInMs = 20000;
 export const sampleDefaultTimeoutInMs = 40000;
 export const templateFileExt = ".tpl";
 export const templateAlphaVersion = "templates-0.0.0-alpha";
-export const templateBetaVersion = "templates-0.0.0-beta";
+export const templateRcVersion = "templates@0.0.0-rc";
```
hund030, line 13: I think we don't need to hard code this version since we decide to reuse config.version for both v2 templates and v3 templates.

hund030 (dismissed):

hund030 (approved):

PR #6845, opened 2022-11-16
chore: validate aad manifest before update

blackchoey:
Tell user a environment variable required by the manifest is missing to help better understand the error:
image


xzf0587 (approved):

blackchoey:
already confirmed with @chagong and @jidddddd on the wording offline, will merge the PR

PR #6852, opened 2022-11-16
chore: use RUN FROM PACKAGE by default. use production mode npm install

Siglud:


xzf0587:
It seems m365-tab-ts is not in the update.

templates/scenarios/ts/command-and-response/teamsfx/app.yml
```diff
@@ -22,10 +22,10 @@ provision:
 deploy:
   - uses: npm/command
     with:
-      args: install
+      args: install --production
```
hund030, line 25: I wonder without devDependency, can it be built successfully? For example, `typescript` is not installed.

Siglud: Good question. Now replaced with npm prune --production.

hund030 (approved):

PR #6863, opened 2022-11-17
chore: add vs v3 templates

xiaolang124:
1. Update vs v3 templates.


kuojianlu (approved):

templates/scenarios/csharp/notification-timer-trigger/teamsfx/app.local.yml.tpl
```diff
@@ -0,0 +1,50 @@
+version: 1.0.0
```
kuojianlu, line 1: There is already `app.local.yml`. Rename it instead of adding a new one.

xiaolang124: Yep, this is a rename pr. Git mistake this for a add operation.

kuojianlu (approved):

PR #6876, opened 2022-11-17
chore: fix provision consent

yuqizhou77:


yuqizhou77:
see https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/16299252 for the context and discussion

jayzhang (approved):

PR #6881, opened 2022-11-17
chore(sample): add video filter sample app

chaiyiping:


packages/fx-core/src/common/samples-config.json
```diff
@@ -209,6 +209,16 @@
             "time": "5min to run",
             "configuration": "Ready for debug",
             "suggested": false
+        },
+        {
+            "id": "video-filters-app-sample",
+            "title": "Video Filters App Sample",
```
qinezh, line 215: Could you please confirm if it's `Video Filters` rather than `Video Filter`? @chaiyiping

qinezh: As we discussed, let's update the wording to `Video Filter` to keep consistent?

PR #6882, opened 2022-11-18
chore: change tunnel domain to bot domain in local yml

xiaolang124:
1.change tunnel domain, endpoint to bot domain, endpoint in local yml.


templates/scenarios/csharp/command-and-response/teamsfx/app.local.yml.tpl
```diff
@@ -26,18 +26,18 @@ provision:
...
       name: {%appName%}
-      messagingEndpoint: ${{TUNNEL_ENDPOINT}}/api/messages
+      messagingEndpoint: ${{BOT_ENDPOINT}}/api/messages
       description: ""
 
 configureApp:
   - uses: teamsApp/validate
     env: 
-      BOT_DOMAIN: ${{TUNNEL_DOMAIN}}
+      BOT_DOMAIN: ${{BOT_DOMAIN}}
```
swatDong, line 35: I believe this can be removed since `BOT_DOMAIN` is already loaded from *.env.local*.

xiaolang124: Thanks for the diea, done.

swatDong (approved):

kuojianlu:
Seems no `BOT_DOMAIN` in `message-extension/teamsfx/.env.local`

PR #6900, opened 2022-11-21
chore(core): init debug|infra templates

jayzhang:


templates/scenarios/common/init-debug-vs-bot/teamsfx/app.local.yml
```diff
@@ -0,0 +1,46 @@
+version: 1.0.0
+
+registerApp:
+  - uses: teamsApp/create # Creates a Teams app
+    with:
+      name: {%appName%} # Teams app name
```
hund030, line 6: Use `.tpl` file extension when using placeholder.

jayzhang: Yes, this file is just a placeholder, you can update it if necessary

xzf0587 (approved):

PR #6907, opened 2022-11-21
perf: add update aad manifest command

wenytang-ms:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/16302316


packages/cli/src/cmds/updateAad.ts
```diff
@@ -0,0 +1,68 @@
...
+import {
+  TelemetryEvent,
+  TelemetryProperty,
+  TelemetrySuccess,
+} from "../telemetry/cliTelemetryEvents";
+import { getSystemInputs } from "../utils";
+import { YargsCommand } from "../yargsCommand";
+
+export default class UpdateAadManifest extends YargsCommand {
+  public readonly commandHead = "update aad-manifest";
```
blackchoey, line 18: should the command be `update aad-app`? @jidddddd

wenytang-ms: Will update this command to be update,
and "update aad-manifest" as the sub-command
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/16271402
this command to "update teams-app-manifest" as another sub-command

jidddddd: I think `update aad-manifest` is fine as that's exactly what we're doing.

blackchoey: offline synced with @jidddddd let's use `update aad-app` because AAD app is an entity in AAD, not AAD manifest. Azure CLI also uses `az ad app update`.
We also need to use `update teams-app`, @nliu-ms FYI.

packages/cli/src/cmds/update.ts
```diff
@@ -0,0 +1,91 @@
...
+  TelemetryProperty,
+  TelemetrySuccess,
+} from "../telemetry/cliTelemetryEvents";
+import { getSystemInputs } from "../utils";
+import { YargsCommand } from "../yargsCommand";
+
+export class UpdateAadApp extends YargsCommand {
+  public readonly commandHead = "aad-app";
+  public readonly command = this.commandHead;
+  public readonly description = "Update the Aad App Manifest in the current application.";
```
wenytang-ms, line 20: @jidddddd please help review this Description

packages/cli/src/cmds/update.ts
```diff
@@ -0,0 +1,91 @@
...
+      ...makeEnvRelatedProperty(rootFolder, inputs),
+    });
+    return ok(null);
+  }
+}
+
+export default class Update extends YargsCommand {
+  public readonly commandHead = "update";
+  public readonly command = `${this.commandHead} <application-manifest>`;
+  public readonly description = "Update the specific application manifest file.";
```
wenytang-ms, line 73: @jidddddd please help review this Descrption also, thanks

packages/cli/src/cmds/update.ts
```diff
@@ -0,0 +1,89 @@
...
+  public readonly description = "Update the Aad App Manifest in the current application.";
+
+  public builder(yargs: Argv): Argv<any> {
+    this.params = {
+      folder: {
+        describe: "Select root folder of the project",
+        string: true,
+      },
+      env: {
+        describe: "Select an existing environment for the AAD manifest",
```
frankqianms, line 29: Should the cases of 'AAD manifest' be aligned with those in line 20?

packages/cli/src/cmds/update.ts
jidddddd: Is this for Teams app? If so, should be "Update the Teams application manifest file."

packages/cli/src/cmds/update.ts
jidddddd: LGTM

packages/cli/src/cmds/update.ts
wenytang-ms: Not really, update is the main command, both "teams app" and "aad-app" are sub-command.
For example, the command add feature in the v2 version, add is the main command, and add sso is the sub-command.

jidddddd (approved):

blackchoey (approved):

packages/cli/src/cmds/update.ts
```diff
@@ -0,0 +1,89 @@
...
+        describe: "Select root folder of the project",
+        string: true,
+      },
+      env: {
+        describe: "Select an existing environment to udpate AAD App",
+        requiresArg: true,
+        string: true,
+      },
+    };
+    return yargs.options(this.params).version(false);
```
Alive-Fish, line 34: Please don't add these parameters into the `this.params` because in this way `folder` and `env` will be auto ignored in the interactive mode.

```typescript
# this may be enough
yargs.options(EnvOptions).options(RootFolderOptions);
```

PR #6920, opened 2022-11-21
chore: update vs launchSettings templates

xiaolang124:
Improve VS 'launchSettings' templates.


templates/scenarios/csharp/non-sso-tab/Properties/launchSettings.json.tpl
```diff
@@ -11,15 +11,16 @@
       },
       "hotReloadProfile": "aspnetcore"
     },
```
swatDong, line 13: Does the `,` break the json format in VS?

xiaolang124: It will show warning information, thanks for the idea.

swatDong (approved):

PR #6930, opened 2022-11-22
chore: generate development appsettings file when needed

xiaolang124:
1.Copy from 'appsettings.json' to generate development appsettings file when needed.


packages/fx-core/src/component/driver/env/appsettingsGenerate.ts
```diff
@@ -37,6 +38,15 @@ export class GenerateAppsettingsDriver implements StepDriver {
     try {
       this.validateArgs(args);
       const appsettingsPath = getAbsolutePath(args.target, context.projectPath);
+      if (!fs.existsSync(appsettingsPath)) {
+        // try to copy appsettings.json
+        const appsettingsTemplatePath = getAbsolutePath("appsettings.json", context.projectPath);
+        if (!fs.existsSync(appsettingsTemplatePath)) {
+          throw new NoAppsettingsFileUserError(actionName, args.target);
```
swatDong, line 45: If appsettings.json not exist, just create an empty json file `{}`

swatDong (approved):

packages/fx-core/src/component/driver/env/appsettingsGenerate.ts
xiaolang124: Done.

PR #6936, opened 2022-11-22
chore: block unsupport commands for video filter

a1exwang:
The list of blocked commands:
* Provision in the cloud
* Zip Teams metadata package
* Deploy to the cloud
* Publish to Teams
* Add features
* Validate manifest file

image
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/16037816/

### DIFF
--- a/packages/fx-core/src/core/middleware/videoFilterAppBlocker.ts
+++ b/packages/fx-core/src/core/middleware/videoFilterAppBlocker.ts
@@ -1,0 +1,72 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+"use strict";
+
+import { NextFunction } from "@feathersjs/hooks";
+import { Inputs, err, Func } from "@microsoft/teamsfx-api";
+import { isVideoFilterProject } from "../../common/tools";
+import { VideoFilterAppRemoteNotSupportedError } from "../error";
+import { CoreHookContext } from "../types";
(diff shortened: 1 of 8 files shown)


packages/fx-core/src/core/middleware/videoFilterAppBlocker.ts
```diff
@@ -0,0 +1,72 @@
...
+  },
+];
+
+async function shouldBlockExecution(ctx: CoreHookContext): Promise<boolean> {
+  const inputs = ctx.arguments[ctx.arguments.length - 1] as Inputs;
+  if (!inputs.projectPath) {
+    return false;
+  }
+
+  if (ctx.method === "executeUserTask") {
```
qinezh, line 35: Looks it only blocks the defined executed user tasks, where's the logic to block operation like provision, deploy?

a1exwang: Normal lifecycle methods that have this middleware are blocked by default. But some operations in `executeUserTask` should be blocked and some don't. So we need this list only for `executeUserTask`.

image

qinezh (approved):

jidddddd (approved):

jayzhang (approved):

tecton (approved):

PR #6946, opened 2022-11-22
fix: bot only project developer

nliu-ms:
1) Only set default developer when placeholder exists and without endpoint.

2) V3 do not need this logic, as samples/templates don't have placeholders for developer.


nliu-ms:
E2E TEST: N/A

jayzhang (approved):

HuihuiWu-Microsoft (approved):

PR #6962, opened 2022-11-24
chore: prepare v3 preview cmd

swatDong:
V3 preview has different behavior so prepare entrypoint for v3 preview.

### DIFF
--- a/packages/cli/src/cmds/preview/previewEnv.ts
+++ b/packages/cli/src/cmds/preview/previewEnv.ts
@@ -1,0 +1,73 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+"use strict";
+
+import * as path from "path";
+import { Argv } from "yargs";
+import { FxError, ok, Result } from "@microsoft/teamsfx-api";
+import { environmentManager } from "@microsoft/teamsfx-core/build/core/environment";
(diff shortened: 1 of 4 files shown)


kuojianlu (approved):

packages/cli/src/cmds/preview/previewEnv.ts
```diff
@@ -0,0 +1,73 @@
...
+import { YargsCommand } from "../../yargsCommand";
+
+// The new preview cmd `teamsfx preview --env ...`
+export default class PreviewEnv extends YargsCommand {
+  public readonly commandHead = `preview`;
+  public readonly command = `${this.commandHead}`;
+  public readonly description = "Preview the current application.";
+
+  public builder(yargs: Argv): Argv<any> {
+    yargs.option("folder", {
```
Alive-Fish, line 20: Maybe you can use `yargs.options(xxx).options(xxx)`? There is also `RootFolderOptions` defined in the `constants.ts` file and you can use it.

Alive-Fish (approved):

packages/cli/src/cmds/preview/previewEnv.ts
swatDong: Thanks! Will update in next PR.

PR #6973, opened 2022-11-24
chore: add param to update teams-app

nliu-ms:
non interactive mode, will continue update:
image

### DIFF
--- a/packages/cli/src/constants.ts
+++ b/packages/cli/src/constants.ts
@@ -87,6 +87,16 @@ export const AadManifestOptions: OptionsMap = {
   },
 };
 
+export const TeamsAppManifestFilePathName = "manifest-file-path";
+export const TeamsAppManifestOptions: OptionsMap = {
+  [AadManifestFilePathName]: {
+    type: "string",
+    global: false,
+    description:
+      "Enter the Teams app manifest template file path, default to './appPackage/manifest.template.json'",
+  },
+};
+
 export const templates = sampleProvider.SampleCollection.samples.map((sample) => {
   return {
     tags: sample.tags,
--- a/packages/cli/src/constants.ts
+++ b/packages/cli/src/constants.ts
@@ -93,7 +93,7 @@ export const TeamsAppManifestOptions: OptionsMap = {
     type: "string",
     global: false,
     description:
-      "Enter the Teams app manifest template file path, default to './appPackage/manifest.template.json'",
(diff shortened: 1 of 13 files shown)


packages/fx-core/src/component/resource/appManifest/appStudio.ts
```diff
@@ -747,31 +749,7 @@ export async function updateManifestV3(
...
-  const appStudioTokenRes = await ctx.tokenProvider.m365TokenProvider.getAccessToken({
-    scopes: AppStudioScopes,
-  });
-  if (appStudioTokenRes.isErr()) {
-    return err(appStudioTokenRes.error);
-  }
-  const appStudioToken = appStudioTokenRes.value;
-
   try {
-    const localUpdateTime = (await fs.stat(manifestFileName)).mtime.getTime();
```
nliu-ms, line 759: Remove this logic as file state timestamp cannot be used to distinguish local update time.

E.g.
User does not modify template, the local time will always be the same and less than dev portal update time.

HuihuiWu-Microsoft: In my understanding, localUpdateTime is always updated when building Teams app package. Correct me if I was wrong

HuihuiWu-Microsoft (dismissed):

packages/cli/src/constants.ts
```diff
@@ -87,6 +87,16 @@ export const AadManifestOptions: OptionsMap = {
   },
 };
 
+export const TeamsAppManifestFilePathName = "manifest-file-path";
+export const TeamsAppManifestOptions: OptionsMap = {
+  [AadManifestFilePathName]: {
```
Alive-Fish, line 92: Should it be `TeamsAppManifestFilePathName`?

packages/cli/src/utils.ts
```diff
@@ -173,6 +173,20 @@ export async function askManifestFilePath(): Promise<Result<string, FxError>> {
   }
 }
 
+export async function askTeamsManifestFilePath(): Promise<Result<string, FxError>> {
+  const config: InputTextConfig = {
+    name: "manifest-file-path",
```
Alive-Fish, line 178: This can be `TeamsAppManifestFilePathName`.

Alive-Fish (approved):

Alive-Fish (approved):

HuihuiWu-Microsoft (approved):

1yefuwang1 (approved):

xzf0587 (approved):

PR #6977, opened 2022-11-24
chore(migration): add naming converter for v3 migration

KennethBWSong:


packages/fx-core/src/core/middleware/MigrationUtils.ts
```diff
@@ -111,3 +112,146 @@ export async function needMigrateToAadManifest(ctx: CoreHookContext): Promise<bo
...
+  "state.fx-resource-function.functionAppResourceId",
+  "state.fx-resource-function.functionEndpoint",
+  "state.fx-resource-key-vault.m365ClientSecretReference",
+  "state.fx-resource-key-vault.botClientSecretReference",
+];
+const secretPrefix = "SECRET_";
+const configPrefix = "CONFIG__";
+const provisionOutputPrefix = "PROVISIONOUTPUT__";
+
+function generateOutputNameRegexForPlugin(pluginId: string) {
```
blackchoey, line 171: I don't know when but current bicep template in V2 changed plugin id. For example, `fx-resource-frontend-hosting` is changed to `teams-tab`, but the state file still used the old plugin id `fx-resource-frontend-hosting`. We need to support this case.

KennethBWSong: Updated, thx!

xzf0587: It is changed in component update. 
Refer to the link for the mapping
https://github.com/OfficeDev/TeamsFx/blob/569d2b01b6a77856cf63f1c32b29e7e449d57e8b/packages/fx-core/src/component/migrate.ts#L95

xzf0587 (approved):

packages/fx-core/src/core/middleware/MigrationUtils.ts
```diff
@@ -111,3 +112,162 @@ export async function needMigrateToAadManifest(ctx: CoreHookContext): Promise<bo
...
+// convert x-xx.xxx.xxx to x_xx__xxx__xxx
+function commonNamingConverterV3(name: string): string {
+  const names = name.split(".");
+  return names.join("__").replace(/\-/g, "_").toUpperCase();
+}
+
+function provisionOutputNamingConverterV3(name: string, bicepContent: string): string {
+  const names = name.split(".");
+  const pluginNames = [names[1], pluginIdMappingV3[names[1]]];
+  const keyName = names[2];
```
frankqianms, line 231: Just comfirm that all the env names used in V2 will be no more than 2 layers nested(2 "." at most). Right?

blackchoey (approved):

packages/fx-core/src/core/middleware/MigrationUtils.ts
KennethBWSong: For these special cases, we have a list above and all the items in the list are in the same format.

PR #6996, opened 2022-11-28
chore: rename bot id to microsoft app id

IvanJobs:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/15906688

Rename BOT_ID to MICROSOFT_APP_ID, BOT_PASSWORD to MICROSOFT_APP_PASSWORD.

image
image
image


packages/fx-core/src/component/driver/botAadApp/create.ts
```diff
@@ -117,8 +117,8 @@ export class CreateBotAadAppDriver implements StepDriver {
 
   private loadCurrentState(): CreateBotAadAppOutput {
     return {
-      BOT_ID: process.env.BOT_ID,
-      SECRET_BOT_PASSWORD: process.env.SECRET_BOT_PASSWORD,
+      MICROSOFT_APP_ID: process.env.MICROSOFT_APP_ID,
```
blackchoey, line 120: I'm not sure whether we need to change the output here: users will see both `AAD_APP_CLIENT_ID` and `MICROSOFT_APP_ID` in .env file. Without certain context, this may be confusing to users.

swatDong:
What will migration do?
- Migrate V2's `BOT_ID` to V3's `MICROSOFT_APP_ID`? This may require user code change.

or
- Keep `BOT_ID` and make sure all actions support both `BOT_ID` and `MICROSOFT_APP_ID`?

packages/fx-core/src/component/driver/botAadApp/create.ts
IvanJobs: I'm considering to add comments here.  @MuyangAmigo may take a look.

IvanJobs:
> What will migration do?
> 
> * Migrate V2's `BOT_ID` to V3's `MICROSOFT_APP_ID`? This may require user code change.
> 
> or
> 
> * Keep `BOT_ID` and make sure all actions support both `BOT_ID` and `MICROSOFT_APP_ID`?

Personally, I prefer the first one if anyway developers have to do some code/config changes during the migration.  If not, the second one would be better.

PR #7004, opened 2022-11-29
chore: state.{env}.json migrator V3

frankqianms:


packages/fx-core/src/core/middleware/utils/migrationContext.ts
```diff
@@ -6,7 +6,7 @@ import fs, { CopyOptions, EnsureOptions, PathLike, WriteFileOptions } from "fs-e
 import path from "path";
 import { CoreHookContext } from "../../types";
 
-const teamsfxFolder = "teamsfx";
+export const teamsfxFolder = "teamsfx";
```
blackchoey, line 9: there's `SettingsFolderName` in `@microsoft/teamsfx-api` package, you can import this in projectMigratorV3.ts. No need to export here.

packages/fx-core/src/core/middleware/utils/migrationContext.ts
```diff
@@ -106,4 +106,28 @@ export class MigrationContext {
   async cleanTeamsfx(): Promise<void> {
     await fs.remove(path.join(this.projectPath, teamsfxFolder));
   }
+
+  async readStateFile(filePath: string): Promise<any> {
```
blackchoey, line 110: My personal understanding is we don't need to put these functions to MigrationContext because it's not designed as a utility class. You can create an utils.ts for these functions or add them to projectMigratorV3.ts directly.

You can talk with Zhaofeng about his idea.

packages/fx-core/src/core/middleware/projectMigratorV3.ts
```diff
@@ -67,6 +68,47 @@ async function preMigration(context: MigrationContext): Promise<void> {
...
+  // general
+  if (await context.fsPathExists(path.join(".fx", "states"))) {
+    // if ./fx/states/ exists
+    const fileNames = await context.fsReadDirSync(path.join(".fx", "states")); // search all files, get file names
+    const fileRegex = new RegExp("state\\.[a-zA-Z0-9_]*\\.json", "g"); // state.*.json
+    for (const fileName in fileNames) {
+      const fileNamesArray = fileRegex.exec(fileName);
+      if (fileNamesArray != null && fileNamesArray.length == 1) {
+        // get envName
+        const envName = fileName.substring(
```
blackchoey, line 81: may use regex capture group to get the env name `state\.(a-zA-Z0-9_])*\.json`

packages/fx-core/src/core/middleware/utils/migrationContext.ts
```diff
@@ -106,4 +106,28 @@ export class MigrationContext {
...
+    const filepath = path.join(this.projectPath, filePath);
+    if (await fs.pathExists(filepath)) {
```
blackchoey, line 112: as internal utility function, you're able remove some checks and let consumer to ensure the path exists. Because you list files first then read the files, so the file will always exist.

packages/fx-core/src/core/middleware/utils/migrationContext.ts
```diff
@@ -106,4 +106,28 @@ export class MigrationContext {
...
+  readBicepContent(): any {
+    return fs.readFileSync(
+      path.join(this.projectPath, "templates", "azure", "provision.bicep"),
+      "utf8"
+    );
+  }
+
+  async fsReadDirSync(_path: string): Promise<string[]> {
+    const dirPath = path.join(this.projectPath, _path);
+    return fs.readdirSync(dirPath);
```
blackchoey, line 127: may use async version of readdir?

PR #7014, opened 2022-11-30
perf: remove message extension from bot template

yukun-dong:


hund030 (commented):
Do we need to update manifest?

yukun-dong:
> Do we need to update manifest?

Don't need to. The bot template manifest doesn't contains `composeExtensions` section.

hund030 (approved):

PR #7017, opened 2022-11-30
chore(migration): update

xzf0587:


jayzhang (approved):

packages/fx-core/src/core/middleware/projectMigratorV3.ts
```diff
@@ -117,3 +165,26 @@ async function loadProjectSettings(projectPath: string): Promise<ProjectSettings
     throw oldProjectSettings.error;
   }
 }
+
+export async function askUserConfirm(ctx: CoreHookContext): Promise<boolean> {
+  sendTelemetryEvent(Component.core, TelemetryEvent.ProjectMigratorNotificationStart);
+  const res = await TOOLS?.ui.showMessage(
```
wenytang-ms, line 171: is there any link for developers to read about the migrate change ?

xzf0587: We will present a summary after migration. 
There is no doc before migration currently. But user can rollback the migration change from the backup folder.

wenytang-ms (approved):

PR #7024, opened 2022-11-30
chore: support release v3 preview TTK

wenytang-ms:
change contains:
1. release preview TTK with version such as 3.2.2022112008
2. The daily release is brought forward to 8 am
3. webpack define the preview/other build

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -7,6 +7,10 @@ on:
         description: "preid name(alpha, beta, rc, stable)"
         required: true
         default: "rc"
+      isPreview:
+        description: "ensure release ttk for preview(yes or no)"
+        required: false
+        default: 'no'
       skip-version-rc:
         description: "release RC version just with lerna publish, skip run lerna version(yes or no)"
         required: false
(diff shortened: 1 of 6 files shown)


LongOddCode (approved):

tecton (approved):

.github/workflows/cd.yml
```diff
@@ -369,6 +379,49 @@ jobs:
           artifactErrorsFailBuild: true
           bodyFile: ./CHANGELOG.md
 
+      - name: Update Preview TTK ai Key
+        if: ${{ github.event_name == 'schedule' || (contains(steps.version-change.outputs.CHANGED, 'ms-teams-vscode-extension@') && github.event.inputs.isPreview != 'no')}}
```
IvanJobs, line 383: What about the case of manual run? Will the replacement of preview ai key be missed in that case?

.github/workflows/cd.yml
```diff
@@ -7,6 +7,10 @@ on:
         description: "preid name(alpha, beta, rc, stable)"
         required: true
         default: "rc"
+      isPreview:
+        description: "ensure release ttk for preview(yes or no)"
+        required: false
```
IvanJobs, line 12: Nit: required = true?

IvanJobs (approved):
LGTM, please consider my comments.

.github/workflows/cd.yml
wenytang-ms: I don't think it's required, since there may rc, stable release. 
they may not care about this input

.github/workflows/cd.yml
wenytang-ms: no, once manual run, the inputs.isPreview should not be 'no', then will trigger the preview workflow

PR #7081, opened 2022-12-05
perf(migrationV3): add name mapping for migration

KennethBWSong:


xzf0587 (approved):

packages/fx-core/src/core/middleware/MigrationUtils.ts
```diff
@@ -191,9 +196,14 @@ function generateOutputNameRegexForPlugin(pluginId: string) {
 export function namingConverterV3(
   name: string,
   type: FileType,
-  bicepContent: string
+  bicepContent: string,
+  needsRename = false
```
blackchoey, line 200: when do we need to set needsRename to `false`? if no case maybe we can remove this optional parameter until we have explicit requirement

blackchoey (approved):

PR #7093, opened 2022-12-05
perf(migrateV3): migrate app id uri

KennethBWSong:


packages/fx-core/src/core/middleware/utils/v3MigrationUtils.ts
```diff
@@ -80,3 +85,50 @@ export async function getProjectVersion(ctx: CoreHookContext): Promise<string> {
...
+  if ((projectSettings as ProjectSettingsV3).components) {
+    tabSso = (projectSettings as ProjectSettingsV3).components.some((component, index, obj) => {
+      return component.name === "teams-tab" && component.sso == true;
+    });
+    botSso = (projectSettings as ProjectSettingsV3).components.some((component, index, obj) => {
+      return component.name === "teams-bot" && component.sso == true;
+    });
+  } else {
+    // For projects that does not componentize.
+    // Should not be called.
```
blackchoey, line 103: what does this mean?

packages/fx-core/src/core/middleware/utils/v3MigrationUtils.ts
```diff
@@ -80,3 +85,50 @@ export async function getProjectVersion(ctx: CoreHookContext): Promise<string> {
   }
   return "0.0.0";
 }
+
+export function parseCapabilities(projectSettings: ProjectSettings): {
```
blackchoey, line 89: can we give it a better name, like `getCapabilitySsoStatus` or something else?

packages/fx-core/src/core/middleware/utils/v3MigrationUtils.ts
KennethBWSong: This part is for scenario that projects migrate from v2 - not componentized. I am not quite sure whether this scenario is valid, but from previous discussion, we don't expect this scenario.

blackchoey (dismissed):

xzf0587 (dismissed):

frankqianms (approved):

xzf0587 (approved):

PR #7094, opened 2022-12-05
chore: userdata migration for V3 with some fixes

frankqianms:


packages/fx-core/src/core/middleware/utils/v3MigrationUtils.ts
```diff
@@ -80,3 +88,24 @@ export async function getProjectVersion(ctx: CoreHookContext): Promise<string> {
   }
   return "0.0.0";
 }
+
+export async function fsReadAndConvertUserdata(
```
blackchoey, line 92: Can we just call it `readAndConvertUserdata`?

frankqianms: sure

packages/fx-core/src/core/middleware/projectMigratorV3.ts
```diff
@@ -366,11 +336,80 @@ export async function configsMigration(context: MigrationContext): Promise<void>
               FileType.CONFIG,
               bicepContent
             );
-            await context.fsWriteFile(path.join(SettingsFolderName, ".env." + envName), envData);
+            await context.fsWriteFile(path.join(SettingsFolderName, ".env." + envName), envData, {
+              encoding: "utf8",
+              flag: "a+",
```
blackchoey, line 341: add comments for the options?

frankqianms: good suggestion

frankqianms: and maybe mode property should be default, we just need to use flag option {encoding: "utf8", flag: "a+"}

packages/fx-core/src/core/middleware/MigrationUtils.ts
```diff
@@ -210,7 +210,9 @@ export function namingConverterV3(
         case FileType.CONFIG:
           return ok(`${configPrefix}${res}`);
         case FileType.USERDATA:
-          return ok(`${secretPrefix}${res}`);
+          if (res.includes("STATE__"))
```
frankqianms, line 213: it might be better to use:
  _if (res.startswith("STATE__"))_ 
in case that the env name includes "STATE__" in the middle of it
@blackchoey what do u think

blackchoey (approved):

xzf0587 (approved):

PR #7099, opened 2022-12-05
fix: remove dashboard entry for v3

huimiu:
- Remove the following entry

image

E2E TEST: N/A


packages/fx-core/src/core/question.ts
```diff
@@ -373,6 +378,10 @@ export function createCapabilityQuestionPreview(inputs?: Inputs): SingleSelectQu
     staticOptions.splice(newBots.length, 0, ExistingTabOptionItem);
   }
 
+  if (isV3Enabled()) {
+    staticOptions.splice(newBots.length, 1);
```
hund030, line 382: Looks tricky. I will prefer something like:
```
const Dashboard = isV3Enabled() ? [] : [ DashboardOptionItem ]
const staticOptions: StaticOptions = [
    ...newBots,
    ...Dashboard,
    ...
  ];
```

huimiu: Nice comment. I've adjusted it.

hund030 (approved):

frankqianms (approved):

PR #7101, opened 2022-12-05
Siglud/v3 helplink

Siglud:


packages/fx-core/src/component/error/deployError.ts
```diff
@@ -21,42 +22,67 @@ export class DeployExternalApiCallError extends ExternalApiCallError {
...
+  ): DeployExternalApiCallError {
     return new DeployExternalApiCallError(
       DeployConstant.DEPLOY_ERROR_TYPE,
       "ZipDeployError",
       "plugins.bot.FailedDeployZipFile",
-      statusCode ?? -1
+      statusCode ?? -1,
+      undefined,
+      undefined,
+      undefined,
```
hund030, line 42: There are too many undefined....

hund030 (approved):

packages/fx-core/src/component/error/deployError.ts
Siglud: It seems to be no good way to avoid this

PR #7112, opened 2022-12-06
chore(V3): add content to .env.{env} if apim plugin exists

frankqianms:


packages/fx-core/src/core/middleware/projectMigratorV3.ts
```diff
@@ -337,13 +338,16 @@ export async function configsMigration(context: MigrationContext): Promise<void>
...
           if (obj["manifest"]) {
             const bicepContent = readBicepContent(context);
-            // convert every name
-            const envData = jsonObjectNamesConvertV3(
-              obj["manifest"],
-              "manifest.",
-              FileType.CONFIG,
-              bicepContent
-            );
+            const teamsfx_env = fs
```
blackchoey, line 341: can we add a step to create new .env files with TEAMSFX_ENV variable? so we don't need to duplicate this logic multiple times

packages/fx-core/src/core/middleware/projectMigratorV3.ts
```diff
@@ -421,3 +440,51 @@ export async function userdataMigration(context: MigrationContext): Promise<void
...
+    // judge if apim plugin exists
+    let flag_apimPlugin = false;
+    for (const obj of projectSettingsContent["components"])
+      if (obj["name"] === "apim") {
+        flag_apimPlugin = true;
+        break;
+      }
+    // must do this step after configs migration
+    if (flag_apimPlugin) {
+      await context.fsEnsureDir("configs");
```
blackchoey, line 459: why we need to have a `configs` folder in the project?

packages/fx-core/src/core/middleware/projectMigratorV3.ts
```diff
@@ -421,3 +440,51 @@ export async function userdataMigration(context: MigrationContext): Promise<void
...
+      path.join(context.projectPath, ".fx", "configs", "projectSettings.json")
+    );
+    // judge if apim plugin exists
+    let flag_apimPlugin = false;
+    for (const obj of projectSettingsContent["components"])
+      if (obj["name"] === "apim") {
+        flag_apimPlugin = true;
+        break;
+      }
+    // must do this step after configs migration
```
blackchoey, line 457: why we must do this step after configs migration?

packages/fx-core/src/core/middleware/projectMigratorV3.ts
frankqianms: actually should be pathExists, rather than fsEnsureDir here

packages/fx-core/src/core/middleware/projectMigratorV3.ts
frankqianms: I think it is better to add apim plugin information into .env after .env files already exist. Actually, there is no need to have to do this after configs migration

packages/fx-core/src/core/middleware/projectMigratorV3.ts
frankqianms: sure

frankqianms: I want to add this logic once at the beginning of config migration, as config migration will be the first migration. what do u think? @blackchoey

packages/fx-core/src/core/middleware/projectMigratorV3.ts
frankqianms: and as the pathExists judgement is already used at the beginning of this function, so we just need to delete this line.

xzf0587 (approved):

PR #7126, opened 2022-12-07
chore: migrate VS projects

blackchoey:


packages/fx-core/templates/core/v3Migration/csharp.app.yml
```diff
@@ -0,0 +1,93 @@
+version: 1.0.0
```
blackchoey, line 1: @IvanJobs @Siglud please help review the content in this app yml template, the syntax is handlebars syntax (similar to mustache)

Siglud (approved):

packages/fx-core/src/core/middleware/projectMigratorV3.ts
```diff
@@ -192,12 +195,21 @@ export async function generateSettingsJson(context: MigrationContext): Promise<v
...
 
 export async function generateAppYml(context: MigrationContext): Promise<void> {
+  const inputs: Inputs = context.arguments[context.arguments.length - 1];
+  const bicepFilePath =
+    inputs.platform === Platform.VS
+      ? Constants.vsProvisionBicepPath
+      : Constants.vscodeProvisionBicepPath;
   const bicepContent: string = await fs.readFile(
-    path.join(context.projectPath, Constants.provisionBicepPath),
+    path.join(context.projectPath, bicepFilePath),
```
frankqianms, line 204: so this function: [readBicepContent](https://github.com/OfficeDev/TeamsFx/blob/dev/packages/fx-core/src/core/middleware/utils/v3MigrationUtils.ts#:~:text=export%20function%20readBicepContent(context%3A%20MigrationContext)%3A%20any%20%7B) should also be changed according to platform, right?

frankqianms: or could you please change the code of readBicepContent() in utils, as this is a common function that will be often refered?

packages/fx-core/templates/core/v3Migration/csharp.app.yml
```diff
@@ -0,0 +1,93 @@
...
+
+provision:
+{{#if activePlugins.fx-resource-bot}}
+  - uses: botAadApp/create # Creates a new AAD app for Bot Registration.
+    with:
+      name: {{appName}}bt$\{{RESOURCE_SUFFIX}}
```
IvanJobs, line 7: Nit: Just FYI, I saw multiple `\` used across the file content. Another way to make it clean is to use a different separator like this file: https://github.com/OfficeDev/TeamsFx/blob/dev/packages/fx-core/templates/plugins/resource/cicd_v2/github/cd.yml

blackchoey: thanks, I will have a try later

IvanJobs (dismissed):
LGTM for the part of `botAadApp/create` in `csharp.app.yml`.

packages/fx-core/src/core/middleware/projectMigratorV3.ts
blackchoey: thanks, updated

chagong (dismissed):

frankqianms (dismissed):

xzf0587 (approved):

PR #7127, opened 2022-12-07
perf(migrationV3): add migration for azure parameter files

KennethBWSong:


xzf0587 (approved):

packages/fx-core/src/core/middleware/projectMigratorV3.ts
```diff
@@ -319,6 +321,43 @@ export async function replacePlaceholderForManifests(context: MigrationContext):
   }
 }
 
+export async function replacePlaceholderForAzureParameter(
```
blackchoey, line 324: maybe just call it `migrateAzureParameter`? because it's not just replace placeholders

blackchoey (approved):

packages/fx-core/src/core/middleware/projectMigratorV3.ts
KennethBWSong: Will update the name in another pr to make some overall changes. Thanks!

PR #7130, opened 2022-12-07
chore: get region when user sign in

nliu-ms:


xiaolang124 (approved):

packages/fx-core/src/common/tools.ts
```diff
@@ -995,6 +998,15 @@ export async function getSPFxToken(
   return spoToken;
 }
 
+/**
+ * Get and set regin for App Studio client
+ * @param m365TokenProvider
+ */
+export async function setRegion(authSvcToken: string) {
+  const region = await AuthSvcClient.getRegion(authSvcToken);
+  AppStudioClient.SetRegion(region);
```
swatDong, line 1007: Does regional endpoint only work for `AppStudioClient`?
Do we need to apply this to other app studio calls (botframework reg, sideloading permission check, ...)?

nliu-ms: It should be applied to other calls as well, but I will put them in next PRs.

swatDong (approved):

PR #7144, opened 2022-12-07
Update Workflow Bot README.md

therealjohn:
Fixes a typo that causes a bug when copying and pasting the sample code.


templates/bot/ts/workflow/README.md
```diff
@@ -129,7 +129,7 @@ export class DoSomethingActionHandler {
     triggerVerb = "doSomething";
 
     async handleActionInvoked(context, message) {
```
dooriya, line 131: How about change the parameter name to `actionData` here (to keep consistency with the action handler API) ?

therealjohn: Done!

PR #7171, opened 2022-12-09
chore: add office addin generator

1yefuwang1:


packages/fx-core/src/component/generator/officeAddin/generator.ts
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  public static async doScaffolding(
+    context: ContextV3,
+    inputs: Inputs,
+    destinationPath: string
+  ): Promise<Result<undefined, FxError>> {
+    const template = getTemplate(inputs);
+    const name = inputs[AddinNameQuestion.name];
+    const addinRoot = destinationPath;
```
millerds, line 77: This is incorrect . . . the addinRoot should be destinationPath + name (make sure this dir also get's created).  The subfolder houses the code for the add-in as this seems to be the pattern for other teams constructions and also allows for future side by side add-ins.

packages/fx-core/src/component/generator/officeAddin/generator.ts
```diff
@@ -0,0 +1,149 @@
...
+    const result = await OfficeAddinGenerator.doScaffolding(context, inputs, destinationPath);
+    if (result.isErr()) {
+      return err(result.error);
+    }
+
+    const templateRes = await Generator.generateTemplate(
+      context,
+      destinationPath,
+      templateName,
+      "ts"
```
millerds, line 63: There is a language question that should be used to set this value.  Currently it is only Typescript (based on the settings json) and so the question is skipped in the UI but setting it up now allow for future changes to be easier.

millerds (changes requested):
When I try to run the branch I get an error comming from ensureTrackingId(projectPath, inputs.projectId) as it can't seem to find the settings file to update with the project id.  At that point in the code it seems like only that office add-in content has been created and on the rest of the project structure.  If I recall from before the project should have been generated and the office add-in added as a folder in the scaffolded project . . . but that doesn't appear to be happening hear.

HuihuiWu-Microsoft (dismissed):

packages/fx-core/src/component/generator/officeAddin/generator.ts
1yefuwang1: The path is changed by design. Because office add-in's taskpane template is self contained and already have a .vscode folder with vscode tasks to run F5. If scaffolded in a subfolder, we need to generate another .vscode in the project root.

millerds: The project is self-contained because it also supports the legacy workflow that doesn't use the toolkit and is designed to be just an add-in.  My understanding is that the toolkit scenario is supposed to support the combined project structure so that we can have office add-ins and things like launch pages and teams tab in the same project.  If we leave the self-contained structure doesn't that complicate the scenario where a new feature/component get's added to an existing project?

Note that in the old branch I already had code to generate stuff in the .vscode launch.json file in the project directory as well as updating the manifest template.  I still used the manifest in the subfolder for the add-in debugging, but the plan was to update our tooling packages and the launch.json to work with the shared manifest once the toolkit supported deploying it to the new service.

millerds: OK . . . looks like the other project types are being flattened as well so it looks like that part makes sense (I still get the error).  I guess the how to guides have the responsibility of explaining how to combine different features into a singular project.

1yefuwang1: TEAMSFX_DEBUG_TEMPLATE needs to be turned on to dismiss the error.

tecton (approved):
VSC looks good.

HuihuiWu-Microsoft (approved):

PR #7177, opened 2022-12-09
chore(vsc): change string

tecton:


packages/vscode-extension/package.nls.json
```diff
@@ -86,7 +86,7 @@
     "teamstoolkit.commands.getstarted.title": "Get Started",
     "teamstoolkit.commands.refresh.title": "Refresh",
     "teamstoolkit.commands.reportIssue.title": "Report issues on GitHub",
-    "teamstoolkit.commands.selectTutorials.title": "View Guided Tutorials",
+    "teamstoolkit.commands.selectTutorials.title": "View How-to Guides",
```
HuihuiWu-Microsoft, line 89: Casing pattern seems not aligned between all the command names. Maybe we can confirm with PM and update them later.

tecton: ok, thanks!

PR #7181, opened 2022-12-11
perf(migrationV3): update logic to read bicep content

KennethBWSong:


packages/fx-core/src/core/middleware/utils/v3MigrationUtils.ts
```diff
@@ -27,13 +27,22 @@ export async function readJsonFile(context: MigrationContext, filePath: string):
 }
 
 // read bicep file content
-export function readBicepContent(context: MigrationContext): any {
+export async function readBicepContent(context: MigrationContext): Promise<any> {
   const inputs: Inputs = context.arguments[context.arguments.length - 1];
   const bicepFilePath =
     inputs.platform === Platform.VS
       ? "Templates/azure/provision.bicep"
       : "templates/azure/provision.bicep";
```
frankqianms, line 35: since we already created new function getTemplateFolderPath() here, we might change codes here as:
`const bicepFilePath = path.join(getTemplateFolderPath(context), "azure", "provision.bicep");`

KennethBWSong: Good catch. Forget to update here. Thx!

xzf0587 (approved):

blackchoey (approved):

xzf0587 (approved):

PR #7189, opened 2022-12-12
chore: add A/B test for dashboard scenario

huimiu:
Add a A/B test for the dashboard option to see how the data works for both types of text.

- Option 1:
 image

- Option 2:
image

E2E TEST: N/A


packages/fx-core/src/core/question.ts
```diff
@@ -345,6 +345,10 @@ export function createCapabilityQuestionPreview(inputs?: Inputs): SingleSelectQu
     );
     WorkflowOptionItem.label = `$(hubot) ${getLocalizedString("core.WorkflowOption.label.abTest")}`;
     WorkflowOptionItem.detail = getLocalizedString("core.WorkflowOption.detail.abTest");
+    DashboardOptionItem.label = `$(browser) ${getLocalizedString(
+      "core.DashboardOption.label.abTest"
+    )}`;
+    DashboardOptionItem.detail = getLocalizedString("core.DashboardOption.detail.abTest");
```
hund030, line 351: Also update the comment?
```
// AB test for notification/command/workflow bot template naming
```

eriolchan (dismissed):

packages/fx-core/src/core/question.ts
huimiu: Yes, I missed this. I've updated the comment.

hund030 (approved):

PR #7205, opened 2022-12-13
fix: .env.teamsfx.local path for video filter app for cert

a1exwang:
Previously after F5, `.env.teamsfx.local` will be generated in `tabs` folder which should be in `app` folder for video filter. This will cause ssl cert not loaded in Vite which cause video filter failing to load.

img

Tested on Windows
img


swatDong (approved):

a1exwang:
E2E TESTS: covered by unit test

qinezh (approved):

swatDong (approved):

PR #7224, opened 2022-12-14
chore: migrate SPFx project

blackchoey:
1. Migrate SPFx project
2. Update to latest action names


xzf0587 (approved):

packages/fx-core/templates/core/v3Migration/spfx.app.yml
```diff
@@ -0,0 +1,60 @@
+version: 1.0.0
+
+deploy:
+  - uses: cli/runNpmCommand
+    with:
+      args: install
+      workingDirectory: ./SPFx
```
HuihuiWu-Microsoft, line 7: Currently SPFx solution is under `src` folder

PR #7239, opened 2022-12-14
fix: judge 'components' exists and iterable

frankqianms:
E2E TEST: N/A

### DIFF
--- a/packages/fx-core/src/core/middleware/projectMigratorV3.ts
+++ b/packages/fx-core/src/core/middleware/projectMigratorV3.ts
@@ -590,7 +590,7 @@ export async function generateApimPluginEnvContent(context: MigrationContext): P
     let flag_apimPlugin = false;
     if (
       Object.keys(projectSettingsContent).includes("components") &&
-      projectSettingsContent["components"] &&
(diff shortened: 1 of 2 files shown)


packages/fx-core/src/core/middleware/projectMigratorV3.ts
```diff
@@ -588,11 +588,17 @@ export async function generateApimPluginEnvContent(context: MigrationContext): P
     );
     // judge if apim plugin exists
     let flag_apimPlugin = false;
-    for (const obj of projectSettingsContent["components"])
-      if (obj["name"] === "apim") {
-        flag_apimPlugin = true;
-        break;
-      }
+    if (
+      Object.keys(projectSettingsContent).includes("components") &&
```
kuojianlu, line 592: Why not check it use `activeResourcePlugins`, like https://github.com/OfficeDev/TeamsFx/blob/dev/packages/fx-core/src/core/middleware/utils/appYmlGenerator.ts#L117.

frankqianms: good suggestion

xzf0587 (approved):

PR #7240, opened 2022-12-14
chore(debug): support configuring channels when creating or updating bot

kuojianlu:
```yml
  - uses: botFramework/create # Create or update the bot registration on dev.botframework.com
    with:
      botId: ${{BOT_ID}}
      name: bot
      messagingEndpoint: ${{BOT_ENDPOINT}}/api/messages
      description: ""
      channels:
        - name: msteams
        - name: outlook
```

### DIFF
--- a/packages/fx-core/src/component/driver/botFramework/createOrUpdateBot.ts
+++ b/packages/fx-core/src/component/driver/botFramework/createOrUpdateBot.ts
@@ -14,10 +14,16 @@ import { ExecutionResult, StepDriver } from "../interface/stepDriver";
 import { addStartAndEndTelemetry } from "../middleware/addStartAndEndTelemetry";
 import { InvalidParameterUserError } from "./error/invalidParameterUserError";
 import { UnhandledSystemError } from "./error/unhandledError";
-import { CreateOrUpdateBotFrameworkBotArgs } from "./interface/createOrUpdateBotFrameworkBotArgs";
+import {
+  CreateOrUpdateBotFrameworkBotArgs,
+  MicrosoftTeamsChannelSettings,
+} from "./interface/createOrUpdateBotFrameworkBotArgs";
 import { BotRegistration } from "../../resource/botService/botRegistration/botRegistration";
 import { LocalBotRegistration } from "../../resource/botService/botRegistration/localBotRegistration";
-import { IBotRegistration } from "../../resource/botService/appStudio/interfaces/IBotRegistration";
+import {
+  BotChannelType,
+  IBotRegistration,
+} from "../../resource/botService/appStudio/interfaces/IBotRegistration";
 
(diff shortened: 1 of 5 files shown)


packages/fx-core/src/component/driver/botFramework/createOrUpdateBot.ts
```diff
@@ -75,13 +81,29 @@ export class CreateOrUpdateBotFrameworkBotDriver implements StepDriver {
         getLocalizedString("driver.botFramework.progressBar.createOrUpdateBot")
       );
 
+      let callingEndpoint: string | undefined = undefined;
+      let configuredChannels: BotChannelType[] | undefined = undefined;
+      if (args.channels) {
```
swatDong, line 86: What if `channels` is undefined?
And what's the case to set `enable` to false for certain channel?

kuojianlu: If `channels` is undefined, will not update the configured channels, see https://github.com/OfficeDev/TeamsFx/blob/kuojianlu/botFramework-channel/packages/fx-core/src/component/resource/botService/botRegistration/utils.ts#L18.
If setting `enable` to false, will disable the channel.

swatDong: How about removing `enable`? To disable the channel, just to remove the whole object?

kuojianlu: Good idea, `enable` seems redundant. Let me update it.

swatDong (approved):

packages/fx-core/src/component/driver/botFramework/createOrUpdateBot.ts
```diff
@@ -75,13 +81,27 @@ export class CreateOrUpdateBotFrameworkBotDriver implements StepDriver {
...
+        configuredChannels = [];
+        for (const channel of args.channels) {
+          if (channel.name === BotChannelType.MicrosoftTeams) {
```
IvanJobs, line 89: Just FYI: My preference would be to use `switch` here.

IvanJobs:
Do we want to support this for V2? It looks like it only supported V3 for drivers.

IvanJobs (approved):

kuojianlu:
> Do we want to support this for V2 or the remote experience? It looks like it only supported V3 for drivers.

Confirmed with Junjie, he said V2 is no need to be supported. For remote, I've already updated the bicep to configure Outlook channel, see https://github.com/OfficeDev/TeamsFx/blob/dev/templates/scenarios/js/m365-message-extension/infra/botRegistration/azurebot.bicep.

PR #7248, opened 2022-12-15
chore: send telemetry for deploy result. change run from package help link to wiki page

Siglud:


packages/fx-core/src/component/driver/interface/buildAndDeployArgs.ts
```diff
@@ -58,4 +58,17 @@ export type AxiosHeaderWithLocation = {
...
 
+export type DeployResult = {
+  id: string;
+  status: number;
+  start_time: string;
+  end_time: string;
+  complete: boolean;
+  active: boolean;
+  is_readonly: boolean;
+  url: string;
```
hund030, line 69: Are these properties all necessary for BI? And I wonder `url` may contains PII. For example, the app service name.

Siglud: I want the URL to be a unique ID for the deployment, and if it is PII data, I can do a hash on it first.

hund030 (approved):
Looks good to me. Remember to classify the telemetry properties you add.

PR #7250, opened 2022-12-15
chore(debug): remove configuring outlook channel from prompt 

kuojianlu:
- remove configuring outlook channel from install app in teams prompt
- update m365-message-extension README

### DIFF
--- a/packages/vscode-extension/src/debug/teamsAppInstallation.ts
+++ b/packages/vscode-extension/src/debug/teamsAppInstallation.ts
@@ -17,27 +17,17 @@ import { openUrlWithNewProfile } from "./launch";
 
 export async function showInstallAppInTeamsMessage(env: string, appId: string): Promise<boolean> {
   const isLocal = env === environmentManager.getLocalEnvName();
-  const botId = await commonUtils.getBotId(env);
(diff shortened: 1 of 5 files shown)


packages/vscode-extension/src/debug/teamsAppInstallation.ts
```diff
@@ -17,27 +17,17 @@ import { openUrlWithNewProfile } from "./launch";
...
-  const botId = await commonUtils.getBotId(env);
+  const botId = isV3Enabled() ? undefined : await commonUtils.getBotId(env);
 
   let messages: string[] = [];
   const items = [localize("teamstoolkit.localDebug.installApp.installInTeams")];
   if (isV3Enabled()) {
     if (isLocal) {
-      if (botId) {
-        messages = [
-          localize("teamstoolkit.localDebug.installApp.bot.description"),
```
swatDong, line 28: Could these strings be removed from package.nls.json?

kuojianlu: They are still used for v2 projects. We can remove them at least after v3 is public and migration is done.

swatDong (approved):

PR #7256, opened 2022-12-15
chore(debug-migration): handle customized args in bot setup task

kuojianlu:


packages/fx-core/src/core/middleware/utils/debug/debugV3MigrationUtils.ts
```diff
@@ -81,6 +84,29 @@ export class OldProjectSettingsHelper {
   }
 }
 
+export async function updateLocalEnv(
```
XiaofuHuang, line 87: Can we keep the local env same to old project? It will be updated after first time F5.

kuojianlu: In yesterday's meeting, there is a conclusion: respect the inputs when writing `.env`. For debugging, `tasks.json` rather than `state.local.json` is the input source. This is why I handle the customized args.

swatDong (approved):

jayzhang (approved):

PR #7257, opened 2022-12-15
Simplify the TS command bot template

therealjohn:
A simplification of the command bot template and make it easier to get started and understand. This also cleans up some language to try and make things clearer. 

Some shortcuts were taken to reduce the number of files and concepts introduced so that it's simpler for developers to understand where to start/explore first. i.e. `index.ts` contains code for the restify server and the command bot configuration instead of using separate files and naming like "internal" - which weren't very clear on their intent.

The result in the scaffold:

| Before | After |
| ------------- | ------------- |
| image  | image  |


templates/bot/ts/command-and-response/src/helloworldCommandHandler.ts
```diff
@@ -17,13 +16,36 @@ export class HelloWorldCommandHandler implements TeamsFxBotCommandHandler {
   ): Promise<string | Partial<Activity> | void> {
     console.log(`Bot received message: ${message.text}`);
 
-    // Render your adaptive card for reply message
-    const cardData: CardData = {
-      title: "Your Hello World Bot is Running",
-      body: "Congratulations! Your hello world bot is running. Click the documentation below to learn more about Bots and the Teams Toolkit.",
+    // An example Adaptive Card that defines the response message of this helloWorld command.
+    const cardJson = {
```
garrytrinder, line 20: I would consider keeping this as an external file rather than bringing it inline.

You lose the context that there is a card definition from briefly looking at the project structure and also lose the ability to view the card in the Adaptive Cards Designer VSCode extension.

garrytrinder (commented):
I like this new simplified structure.

Removing the folders makes it less opinionated leaving the structure up to the developer to decide what is best for them and without having to unpick an existing folder structure.

It would be nice to implement an empty TeamsActivityHandler in `index.ts` as well to make it easier for developers to implement handlers for Teams events where there are no TeamsFx abstractions.

```ts
// teams activity handler setup, override methods to handle incoming activities from Teams
class ConversationBotActitivityHandler extends TeamsActivityHandler {
  constructor() {
    super();
  }
}

// create an instance of your activity handler
const conversationBotActitivityHandler = new ConversationBotActitivityHandler();

// bot framework endpoint to receive incoming activities
server.post("/api/messages", async (req, res) => {
  await conversationBot.adapter.process(req, res, (context) => conversationBotActitivityHandler.run(context))
});
```

PR #7258, opened 2022-12-16
chore: update multi env e2e test for v3

xiaolang124:
Skip these 3 cases because 'add tab', 'add sql' are no longer support in v3.


xiaolang124:
e2e: https://github.com/OfficeDev/TeamsFx/actions/runs/3709576499/jobs/6288472521

qinezh (approved):

PR #7268, opened 2022-12-16
chore: add telemetry for sdk version

yiqing-zhao:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/16605049
E2E TEST: N/A

### DIFF
--- a/packages/vscode-extension/src/debug/prerequisitesHandler.ts
+++ b/packages/vscode-extension/src/debug/prerequisitesHandler.ts
@@ -1187,14 +1187,18 @@ function checkNpmInstall(
         error = err;
       }
 
-      const object = await fs.readJson(path.join(folder, "package-lock.json"));
+      if (!isV3Enabled()) {
+        const packageLockFile = await fs
+          .readJson(path.join(folder, "package-lock.json"))
+          .catch(() => {});
 
-      ctx.properties[TelemetryProperty.BotbuilderVersion] =
-        object?.dependencies["botbuilder"]?.version;
-      ctx.properties[TelemetryProperty.TeamsFxVersion] =
-        object?.dependencies["@microsoft/teamsfx"]?.version;
-      ctx.properties[TelemetryProperty.TeamsJSVersion] =
-        object?.dependencies["@microsoft/teams-js"]?.version;
+        ctx.properties[TelemetryProperty.BotbuilderVersion] =
+          packageLockFile?.dependencies["botbuilder"]?.version;
+        ctx.properties[TelemetryProperty.TeamsFxVersion] =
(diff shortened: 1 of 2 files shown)


HuihuiWu-Microsoft (dismissed):

packages/vscode-extension/src/debug/prerequisitesHandler.ts
```diff
@@ -1186,6 +1186,16 @@ function checkNpmInstall(
         // treat unexpected error as installed
         error = err;
       }
+
+      const object = await fs.readJson(path.join(folder, "package-lock.json"));
```
kuojianlu, line 1190: Use better nam instead of `object`.

swatDong, line 1190: In v3 `npm install` is moved to `teamsfx:deploy` as an action, then this won't be hit.
BTW, does `fs.readJson` throws if *package-lock.json* not exist?

packages/vscode-extension/src/debug/prerequisitesHandler.ts
```diff
@@ -1186,6 +1186,16 @@ function checkNpmInstall(
...
+
+      ctx.properties[TelemetryProperty.BotbuilderVersion] =
+        object?.dependencies["botbuilder"]?.version;
+      ctx.properties[TelemetryProperty.TeamsFxVersion] =
+        object?.dependencies["@microsoft/teamsfx"]?.version;
+      ctx.properties[TelemetryProperty.TeamsJSVersion] =
+        object?.dependencies["@microsoft/teams-js"]?.version;
```
kuojianlu, line 1197: Are these telemetries also targeted for v3? Please note that npm install cheking is removed in v3.

packages/vscode-extension/src/debug/prerequisitesHandler.ts
yiqing-zhao: Thanks for the comments. Updated for not-existed `package-lock.json`, `fs.readJson` will now return `undefined`, and will not affect the telemetry assignment.

yiqing-zhao: Updated, thanks!

packages/vscode-extension/src/debug/prerequisitesHandler.ts
yiqing-zhao: Thanks for the comments! Adding check for V3, and the telemetry is currently only for v2. How to add V3 telemetry is under discussion.

swatDong (approved):

HuihuiWu-Microsoft (approved):

PR #7287, opened 2022-12-20
chore: upgrate restify to support node 18

swatDong:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/16598471/
https://github.com/restify/node-restify/releases/tag/v10.0.0

Update all bot templates to use `restify:10` which supports node 18

### DIFF
--- a/templates/bot/js/default/package.json
+++ b/templates/bot/js/default/package.json
@@ -18,7 +18,7 @@
   "dependencies": {
     "@microsoft/adaptivecards-tools": "^1.0.0",
     "botbuilder": "^4.18.0",
-    "restify": "~8.5.1"
+    "restify": "~10.0.0"
   },
   "devDependencies": {
     "env-cmd": "^10.1.0",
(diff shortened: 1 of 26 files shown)


kuojianlu (dismissed):

templates/bot/js/default/package.json
```diff
@@ -18,7 +18,7 @@
   "dependencies": {
     "@microsoft/adaptivecards-tools": "^1.0.0",
     "botbuilder": "^4.18.0",
-    "restify": "~8.5.1"
+    "restify": "~10.0.0"
```
yukun-dong, line 21: Some are using "~" and some are using "^". How about making them consistent?

swatDong: Updated.
I'm not sure why some were using `~`, but it seems safe to use `^` for all templates.

yukun-dong (approved):

kuojianlu (approved):

PR #7293, opened 2022-12-20
chore: improve TDP & VSC

yuqizhou77:
- fxcore: add checking platform to avoid overwriting values of input from VS
- vscode: add learn more link and icon for "publish in developer portal", and update description
- m365login: tolowercase()

### DIFF
--- a/packages/vscode-extension/src/commonlib/m365Login.ts
+++ b/packages/vscode-extension/src/commonlib/m365Login.ts
@@ -309,6 +309,7 @@ export class M365Login extends BasicLogin implements M365TokenProvider {
       tokenRes = await this.signInWhenNoAccountInCache(tokenRequest, loginHint, true);
       VsCodeLogInstance.info(`Signed in with Microsoft 365 account: ${loginHint}`);
     } else if (
+      !!loginHint &&
       M365Login.codeFlowInstance.account.username.toLowerCase() === loginHint.toLowerCase()
     ) {
       // If the account in cache matched with the loginHint, we will try to get access token for the currently cached account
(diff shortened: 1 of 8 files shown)


packages/vscode-extension/src/commonlib/m365Login.ts
```diff
@@ -308,7 +308,9 @@ export class M365Login extends BasicLogin implements M365TokenProvider {
       // If no account in cache file, we will ask to sign in directly.
       tokenRes = await this.signInWhenNoAccountInCache(tokenRequest, loginHint, true);
       VsCodeLogInstance.info(`Signed in with Microsoft 365 account: ${loginHint}`);
-    } else if (M365Login.codeFlowInstance.account.username === loginHint) {
+    } else if (
+      M365Login.codeFlowInstance.account.username.toLowerCase() === loginHint.toLowerCase()
```
swatDong, line 312: would username or loginHint be undefined?

yuqizhou77: added check for loginHint.
We checked whether account is undefined or not in the if block above, so here the account is always AccountInfo type. And the type of AccountInfo in msal
export declare type AccountInfo = {
    homeAccountId: string;
    environment: string;
    tenantId: string;
    username: string;
    localAccountId: string;
    name?: string;
    idTokenClaims?: object;
};

jayzhang (approved):

swatDong (approved):

nliu-ms (approved):

HuihuiWu-Microsoft (approved):

PR #7301, opened 2022-12-20
chore(debug-migration): migrate fx-extension.pre-debug-check task

kuojianlu:


packages/fx-core/src/core/middleware/utils/debug/taskMigrator.ts
```diff
@@ -644,6 +644,96 @@ export async function migrateValidateLocalPrerequisites(
...
+        messagingEndpoint: `$\{{${context.placeholderMapping.botEndpoint}}}/api/messages`,
+      };
+    }
+
+    if (!context.appYmlConfig.configureApp) {
+      context.appYmlConfig.configureApp = {};
+    }
+    if (OldProjectSettingsHelper.includeTab(context.oldProjectSettings)) {
+      context.appYmlConfig.configureApp.tab = {
+        domain: "localhost:53000",
```
swatDong, line 685: When did we change tab port to 53000? Is it possible the old project uses 3000?

kuojianlu: From about v3.2.0. I think it is not a matter because after migration users can modify the port as  their needs.

swatDong (approved):

hund030 (approved):

PR #7307, opened 2022-12-21
chore: telemetry for package version

Siglud:


packages/fx-core/src/component/driver/script/npmBuildDriver.ts
```diff
@@ -33,4 +35,24 @@ export class NpmBuildDriver extends BaseBuildStepDriver {
...
+    try {
+      const packageJson = path.join(this.workingDirectory, "package.json");
+      if (fs.existsSync(packageJson)) {
+        const content = await fs.readFile(packageJson, "utf-8");
+        const json = JSON.parse(content);
+        const dependencies: { [key: string]: string } = {
+          ...json.dependencies,
+          ...json.devDependencies,
+        };
+        this.context.telemetryReporter?.sendTelemetryEvent("package-version", dependencies);
```
hund030, line 51: Do we really need to get all dependencies of user's project? I wonder how do you classify them?

Siglud: Ok, changed it to track our SDK version only

PR #7340, opened 2022-12-27
fix(migrationV3): add a missing name in convert list

frankqianms:
E2E TEST: N/A
Add fx-resource-bot.botWebAppResourceId in convert list


packages/fx-core/src/core/middleware/utils/MigrationUtils.ts
```diff
@@ -289,6 +290,10 @@ function provisionOutputNamingConverterV3(
             break;
           }
         } while ((outputNames = pluginRegex.exec(bicepContent)));
+      } else if ("fx-resource-bot" === pluginNames[0] && keyName.endsWith("AppResourceId")) {
+        // keep AppResourceId aligned in fx-resource-bot plugin
+        outputName = outputNames[1];
+        keyName = "functionAppResourceId";
```
blackchoey, line 296: can you add some comments to explain why we need to change the key name here?

frankqianms: First of all, I want to make sure that whether we need to change the key name to keep the placeholder of azure function resource id the same(PROVISIONOUTPUT__BOTOUTPUT__FUNCTIONAPPRESOURCEID).
What's your opinion?

hund030 (approved):

packages/fx-core/src/core/middleware/utils/MigrationUtils.ts
frankqianms: Now I changed the logic in app yml generating

PR #7346, opened 2022-12-28
chore(debug-migration): trigger migration for teamsfx tasks and commands

kuojianlu:


packages/vscode-extension/src/debug/teamsfxTaskProvider.ts
```diff
@@ -108,6 +108,16 @@ const customTasks = Object.freeze({
   },
 });
 
+async function triggerV3Migration(): Promise<void> {
```
swatDong, line 111: Why defining `triggerV3Migration()` again?

kuojianlu: They are a little different, i.e., different return values.

swatDong: Yep... that's why I asked... if the only difference is return value, still can reuse the above one - just ignore the return value.

kuojianlu: You're right, updated.

XiaofuHuang (approved):

HuihuiWu-Microsoft (approved):

PR #7350, opened 2022-12-29
fix: aadApp/update should keep manifestTemplatePath param nmae

nliu-ms:
teamsApp actions will use manifestPath as param name,
aadApp/update action will use manifestTemplatePath


nliu-ms:
E2E TEST: N/A

wenytang-ms (approved):

PR #7391, opened 2023-01-05
chore(core): correct localization usage part 1

jayzhang:


hund030 (approved):

packages/fx-core/src/common/deps-checker/constant/message.ts
```diff
@@ -4,51 +4,51 @@
 import { getDefaultString, getLocalizedString } from "../../localizeUtils";
 
 export const Messages = {
-  learnMoreButtonText: getLocalizedString("depChecker.learnMoreButtonText"),
-  defaultErrorMessage: [
+  // learnMoreButtonText: getLocalizedString("depChecker.learnMoreButtonText"),
```
kuojianlu, line 7: Why commentting this?

jayzhang: There is no reference on this variable

kuojianlu: I see this variable is referenced in https://github.com/OfficeDev/TeamsFx/blob/dev/packages/fx-core/src/component/utils/depsChecker/bicepChecker.ts#L446.

a1exwang (approved):

PR #7469, opened 2023-01-11
fix: update spfx template

nliu-ms:


nliu-ms:
E2E TEST: ./spfx/CreateSPFxProject.tests.ts

HuihuiWu-Microsoft (approved):

PR #7475, opened 2023-01-11
chore(migration): update backup

xzf0587:


jayzhang (approved):

packages/fx-core/src/core/middleware/utils/v3MigrationUtils.ts
```diff
@@ -176,12 +176,12 @@ export async function getProjectVersionFromPath(projectPath: string): Promise<Ve
...
   const abandonedPath = path.resolve(
     projectPath,
-    MetadataV3Abandoned.folder,
+    MetadataV3Abandoned.configFolder,
     MetadataV3Abandoned.configFile
   );
   if (await fs.pathExists(abandonedPath)) {
     return {
-      version: MetadataV3Abandoned.folder,
+      version: MetadataV3Abandoned.configFolder,
```
blackchoey, line 184: minor: it may be strange to use a configFolder name as version, we can see how to improve later.

PR #7476, opened 2023-01-11
chore: add prompt for aad

wenytang-ms:
issue:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/16740430

### DIFF
--- a/packages/fx-core/src/component/driver/aad/update.ts
+++ b/packages/fx-core/src/component/driver/aad/update.ts
@@ -27,7 +27,7 @@ import { MissingEnvUserError } from "./error/missingEnvError";
 
 const actionName = "aadApp/update"; // DO NOT MODIFY the name
 const helpLink = "https://aka.ms/teamsfx-actions/aadapp-update";
-const ViewAadAppHelpLink = "https://aka.ms/teamsfx-view-aad-app";
+const ViewAadAppHelpLink = "https://aka.ms/teamsfx-view-aad-app-v3";
 const driverConstants = {
   generateManifestFailedMessageKey: "driver.aadApp.error.generateManifestFailed",
 };
--- a/packages/fx-core/src/component/driver/aad/update.ts
+++ b/packages/fx-core/src/component/driver/aad/update.ts
@@ -11,7 +11,7 @@ import * as fs from "fs-extra";
 import * as path from "path";
 import { AadAppClient } from "./utility/aadAppClient";
 import axios from "axios";
-import { SystemError, UserError, ok, err, FxError, Result } from "@microsoft/teamsfx-api";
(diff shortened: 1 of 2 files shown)


packages/fx-core/src/component/driver/aad/update.ts
```diff
@@ -109,7 +110,15 @@ export class UpdateAadAppDriver implements StepDriver {
         getLocalizedString(logMessageKeys.successExecuteDriver, actionName)
       );
       await progressHandler?.end(true);
-
+      const msg = getLocalizedString("core.deploy.aadManifestSuccessNotice");
```
blackchoey, line 113: Since the bug is about "Teams: Update AAD manifest" command, can we add this logic to the command part (in `FxCoreImplementV3.ts`)? So when provision, users won't get this notification. I remember there's another bug that complaints provision pops up some notification. @jidddddd for your opinion.

packages/fx-core/src/component/driver/aad/update.ts
```diff
@@ -27,6 +27,7 @@ import { MissingEnvUserError } from "./error/missingEnvError";
 
 const actionName = "aadApp/update"; // DO NOT MODIFY the name
 const helpLink = "https://aka.ms/teamsfx-actions/aadapp-update";
+const ViewAadAppHelpLink = "https://aka.ms/teamsfx-view-aad-app";
```
blackchoey, line 30: This aka link is an old one, maybe we need a new aka link pointing to new position? Since V2&V3 will exists together for a while.

packages/fx-core/src/component/driver/aad/update.ts
wenytang-ms: I find that the teams app manifest logic, which does the same thing within component.

packages/fx-core/src/component/driver/aad/update.ts
wenytang-ms: thanks, create a new link https://aka.ms/teamsfx-view-aad-app-v3

packages/fx-core/src/component/driver/aad/update.ts
jidddddd: > Since the bug is about "Teams: Update AAD manifest" command, can we add this logic to the command part (in `FxCoreImplementV3.ts`)? So when provision, users won't get this notification. I remember there's another bug that complaints provision pops up some notification. @jidddddd for your opinion.

We could solve the pop up issue separately later, I'm still trying to figure out what to do with it because that other bug involves changing names for lifecycle stages.

wenytang-ms: sure, thanks. 
I will send the issue bug to your first. @jidddddd

PR #7480, opened 2023-01-11
chore: update cli m365 lock

xiaolang124:
When using command account logout m365, activate will try to get region information at the start time.
At the same time AccountLogout handler will try to logout account.
This will trigger getStatus->logout account->getStatus process switching and cause silent getting token in logout status issue.
This method provides a way to lock specific method and release to do other work.

Fix: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/16731952/


packages/cli/src/commonlib/codeFlowLogin.ts
```diff
@@ -253,16 +251,31 @@ export class CodeFlowLogin {
...
     }
+  }
 
-    await saveAccountId(this.accountName, undefined);
-    return true;
+  // when using command account logout m365, activate will try to get region information at the start time.
+  // At the same time AccountLogout handler will try to logout account.
+  // This will trigger getStatus->logout account->getStatus process switching and cause silent getting token in logout status issue.
+  // This method provides a way to lock specific method and release to do other work.
+  public async getLock(): Promise<MutexInterface.Releaser> {
```
swatDong, line 276: Is it possible to use lock in `codeFlowLogin` only?
E.g., add lock to `reloadCache`, `getTokenByScope`, `logout`...

One best practice to use lock is to ensure the lock scope is as small as possible.

packages/cli/src/activate.ts
```diff
@@ -33,7 +33,7 @@ export default async function activate(
     }
   }
 
-  M365TokenProvider.setStatusChangeMap(
+  await M365TokenProvider.setStatusChangeMap(
```
swatDong, line 36: Is it possible to just not call `setStatusChangeMap` in CLI? Just:
1. Try set region when activate
2. Try set region whenever login/logout

IMO, during one CLI call, login/logout at most happens only once.

packages/cli/src/commonlib/codeFlowLogin.ts
swatDong: Of course, another best practice is - do not use lock if possible :P

xiaolang124: I have tried to use lock in `codeFlowLogin` only, it does not address this problem. Because `getStatus` has used `reloadCache` and `getTokenByScope` methods. `logout` will insert the execution in the middle of two. I am trying to find a better way to fix this.

swatDong (dismissed):

swatDong (approved):

Alive-Fish (approved):

PR #7496, opened 2023-01-12
chore: revert previous method to check sdk versions and use filesystem watcher instead

yiqing-zhao:
E2E TEST: N/A


packages/vscode-extension/src/handlers.ts
```diff
@@ -369,6 +369,27 @@ export function addFileSystemWatcher(workspacePath: string) {
   backupConfigWatcher.onDidCreate(async (event) => {
     await openBackupConfigMd(workspacePath, event.fsPath);
   });
+
+  const packageLockFileWatcher = vscode.workspace.createFileSystemWatcher("**/package-lock.json");
```
tecton, line 373: I think we need to monitor `package-lock.json` only when it's a teamsfx project, does it make sense?

tecton (approved):

kuojianlu (approved):

PR #7499, opened 2023-01-12
chore(template): use explicit type for message extension ts templates

yukun-dong:


swatDong (dismissed):
Code change looks good to me.
BTW, what's this change for? Better code style or to fix anything?

yukun-dong:
> Code change looks good to me. BTW, what's this change for? Better code style or to fix anything?

Yeah, for better code style. Since it's a ts project, it's better to use explicit type rather than "any".
Also there existed a bug in message extension template for a long time that got fixed recently https://github.com/OfficeDev/TeamsFx/pull/7455. If it doesn't use "any" then this bug should have been noticed earlier.

Siglud (approved):

swatDong (approved):

PR #7559, opened 2023-01-18
chore: apply isLocalDebug for env

swatDong:
Align with #7558 to create `.env` file if not exist on F5.


swatDong:
Covered in #7558

PR #7560, opened 2023-01-18
chore: add env folder in migrated generated yml

frankqianms:


packages/fx-core/templates/core/v3Migration/csharp.app.yml
```diff
@@ -97,4 +99,4 @@ configureApp:
       appPackagePath: ./build/appPackage/appPackage.$\{{TEAMSFX_ENV}}.zip # Relative path to this file. This is the path for built zip file.
     # Output: following environment variable will be persisted in current environment's .env file.
     # TEAMS_APP_ID: the id of Teams app
-projectId: {{placeholderMappings.[projectId]}}
\ No newline at end of file
+projectId: {{projectId}}
```
blackchoey, line 102: can we put the projectId to the beginning of the yml file (together with environmentFolderPath)?

frankqianms: sure. good idea

blackchoey (dismissed):

packages/fx-core/src/core/middleware/utils/appYmlGenerator.ts
```diff
@@ -50,6 +52,8 @@ export class AppYmlGenerator extends BaseAppYmlGenerator {
       isWebAppBot: false,
       isTypescript: false,
       defaultFunctionName: undefined,
+      defaultEnvironmentFolder: undefined,
```
xzf0587, line 55: In generator, the key name environmentFolder is better than defaultEnvironmentFolder.
Actually, we will refine it to use another folder when there is existing user's env folder in project

frankqianms: changed

xzf0587 (approved):

PR #7573, opened 2023-01-19
chore: convert project before import

millerds:
Add the ability to convert a xml manifest based office add-in project when using it as a basis for a new project.


packages/fx-core/package.json
```diff
@@ -101,6 +101,7 @@
     "node-forge": "^1.0.0",
     "node-ts-uuid": "^1.0.8",
     "office-addin-manifest": "^1.10.4",
+    "office-addin-project": "0.5.0",
```
1yefuwang1, line 104: Please commit package-lock.json file as well, or CI/CD pipeline would fail. Becasue `npm ci` checks whether package.json and lock are in sync.

1yefuwang1:
Please add unit tests in https://github.com/OfficeDev/TeamsFx/blob/dev/packages/fx-core/tests/component/generator/officeAddinGenerator.test.ts.  
We have a strict policy blocking PRs that causes UT coverage to drop.

packages/fx-core/package.json
millerds: When I install the lock file gets udpate from version 1 format to version 2.  Is this OK?

FYI . . . This happen because I'm using an newer version of npm that goes with node 16 (which is a year old already).  The current version of npm actually has another update to the lock file version (v3).  I'm not sure what your policy is for supporting different versions of node and npm.

1yefuwang1: We are using node v14.17.0 (npm v6.14.17), which uses lockfersion 1. I think it better to keep it that way.
You can easily switch to node 14 using nvm or other nodejs version manager.

millerds: OK . . . I think I got the right setup and submitted the updated to the file.

1yefuwang1: Could you please to resolve the conflict?

packages/fx-core/tests/component/generator/officeAddinGenerator.test.ts
```diff
@@ -180,6 +181,58 @@ describe("OfficeAddinGenerator", function () {
...
+      .callsFake((from: string, to: string) => {
+        return;
+      });
+    const updateManifestStub = sinon
+      .stub(HelperMethods, "updateManifest")
+      .callsFake(async (destination: string, manifestPath: string) => {
+        return;
+      });
+    const convertProjectStub = sinon
+      .stub()
```
1yefuwang1, line 206: Is the stub() deliberately called with no input?

millerds: The example I followed for proxyquiry did that.  I didn't try doing something different.

PR #7587, opened 2023-01-29
chore: clear login cache

xiaolang124:
1.Clear account cache file in logout function.


packages/cli/src/commonlib/codeFlowLogin.ts
```diff
@@ -253,14 +253,8 @@ export class CodeFlowLogin {
...
   async logout(): Promise<boolean> {
-    const accountCache = await loadAccountId(this.accountName);
-    if (accountCache) {
-      const dataCache = await this.msalTokenCache.getAccountByHomeId(accountCache);
-      if (dataCache) {
-        this.msalTokenCache?.removeAccount(dataCache);
-      }
-    }
-
+    (this.msalTokenCache as any).storage.setCache({});
```
swatDong, line 256: Is it safe to directly access the private `storage` property? E.g., what if *msal-node* updates its implementation and uses another naming?

How about calling `msalTokenCache.getAllAccounts`, then removing all accounts? (I guess we do not support multi-accounts so it's safe to remove all accounts)

xiaolang124: This makes sense. If we remove any account in `msalTokenCache` MSAL will update cache file in the. So `clearCache` will not work in this scenario and I have deleted this code.

swatDong (approved):

PR #7594, opened 2023-01-30
chore(core): correct localization usage part 3

jayzhang:


hund030 (approved):

packages/fx-core/src/component/generator/spfxGenerator.ts
```diff
@@ -63,7 +64,7 @@ export class SPFxGenerator {
       const componentName = Utils.normalizeComponentName(webpartName);
       const componentNameCamelCase = camelCase(componentName);
 
-      await progressHandler?.next(ScaffoldProgressMessage.DependencyCheck);
```
HuihuiWu-Microsoft, line 66: If this doesn't need anymore, can we remove the constant?

HuihuiWu-Microsoft (approved):

PR #7630, opened 2023-02-01
chore: rename updateAppSettings to updateJson

swatDong:
Better name `file/updateAppSettings` to `file/updateJson` since it actually can output any settings to JSON format.
Then this action can be used by `init debug` for those Node.js projects which use JSON file as config (e.g., [app-hello-world](https://github.com/OfficeDev/Microsoft-Teams-Samples/tree/main/samples/app-hello-world/nodejs))

### DIFF
--- a/packages/fx-core/resource/package.nls.json
+++ b/packages/fx-core/resource/package.nls.json
@@ -1148,7 +1148,7 @@
   "driver.file.progressBar.title": "file/updateEnv",
   "driver.file.progressBar.generate": "Generating the environment variables to .env file.",
   "driver.file.updateJson.description": "Add or update appsettings to JSON file.",
-  "driver.file.updateJson.summary.withTarget": "The appsettings has been generated successfully to %s.",
+  "driver.file.updateJson.summary.withTarget": "The value for appsettings has been generated successfully to %s.",
   "driver.env.error.folderNotExist": "Folder %s does not exist.",
   "driver.deploy.error.deployToAzureRemoteFailed": "Remote service error, upload failed.",
   "driver.deploy.error.unknownError": "Unknown error happened: %s",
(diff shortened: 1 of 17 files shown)


kuojianlu (dismissed):

kuojianlu (approved):

packages/fx-core/resource/package.nls.json
```diff
@@ -1147,6 +1147,8 @@
   "driver.file.summary.withTarget": "The environment variables has been generated successfully to %s.",
   "driver.file.progressBar.title": "file/updateEnv",
   "driver.file.progressBar.generate": "Generating the environment variables to .env file.",
+  "driver.file.updateJson.description": "Add or update appsettings to JSON file.",
+  "driver.file.updateJson.summary.withTarget": "The appsettings has been generated successfully to %s.",
```
sffamily, line 1151: suggest to be more specific:  The appsettings -> The value for appsettings

and it is aligned with singular form as well.

packages/fx-core/tests/component/driver/file/generateAppsettings.test.ts
```diff
@@ -69,12 +68,13 @@ describe("AppsettingsGenerateDriver", () => {
       if (result.isErr()) {
         chai.assert(result.error instanceof InvalidParameterUserError);
         const message =
-          "Following parameter is missing or invalid for file/updateAppSettings action: appsettings.";
+          "Following parameter is missing or invalid for file/updateJson action: appsettings.";
```
sffamily, line 71: just to double confirm the last word is still appsettings?

packages/fx-core/resource/package.nls.json
swatDong: Updated.

packages/fx-core/tests/component/driver/file/generateAppsettings.test.ts
swatDong: Yes, since the parameter schema does not change. This action is to generate some appsettings, in JSON format.
``` yaml
  - uses: file/updateJson
    with:
      target: some.json
      appsettings: # still appsettings
        A: ${{SOME_ENV}}
        B: 
          Foo: Bar
        C:
          - true
          - false
```

kuojianlu (approved):

sffamily (approved):

jayzhang (approved):

PR #7638, opened 2023-02-01
chore(server+core): security vulnerability

Alive-Fish:


chagong (approved):

Alive-Fish:
Compared with [chore: update mocha and run lerna-audit for the fallback branch](https://github.com/OfficeDev/TeamsFx/pull/7603), this do not change the API and Manifest packages in other to not upgrade these package's versions.

PR #7659, opened 2023-02-02
chore(debug-migration): add comment in local yml

kuojianlu:


packages/fx-core/templates/core/v3Migration/js.ts.app.local.yml
```diff
@@ -106,7 +129,7 @@ deploy:
 
   {{/dotnetCommand}}
   {{#tab}}
-  - uses: file/updateEnv
+  - uses: file/updateEnv # Generate env to .env file
```
swatDong, line 132: For those with target `.env.teamsfx.local` or `.localSettings` or any target, prefer **Generate runtime environment variables** since the target is not TeamsFx context.

kuojianlu: Good suggestion, updated.

swatDong (approved):

PR #7660, opened 2023-02-02
chore(migration): update

xzf0587:
1. add telemetry when open v3 project in v2
2. replace migration with upgrade in ux


packages/fx-core/src/common/telemetry.ts
```diff
@@ -71,6 +71,7 @@ export enum TelemetryEvent {
   CheckSubscription = "check-subscription",
   CheckSideloading = "check-sideloading",
   EnvConfig = "env-config",
+  OpenAdvancedProjectV3InV2 = "open-advanced-project-v3-in-v2",
```
blackchoey, line 74: I'm not sure whether PM can easily understand what does this event mean. @KennethBWSong how's your feeling to this event name?

blackchoey: not sure whether `display-tooling-update-notification` or something else is better?

PR #7425, opened 2023-01-06
chore: fix some spelling mistakes and adjust phrasing.

therealjohn:
First pass at fixing some typos.

PR #7690, opened 2023-02-06
chore: shorter message in v3 upgrade dialog

frankqianms:


sffamily (approved):

therealjohn:
I have changes to this message and many others in #7425. However, this PR would make this message shorter than my suggestion - is there a technical reason it needs to be shorter?

PR #7710, opened 2023-02-08
chore: localize summary strings

1yefuwang1:
localize summary strings. 
Fixes https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_sprints/taskboard/Teams%20Extensibility%20E2E%20Team/Microsoft%20Teams%20Extensibility/CY23-2.1?workitem=17064428


packages/fx-core/resource/package.nls.json
```diff
@@ -781,6 +781,15 @@
   "core.common.OutlookWebClientName2": "Outlook web access client id 2",
   "core.NewOfficeAddinOptionVSC.label": "Start with an Outlook add-in",
   "core.NewOfficeAddinOptionVSC.detail": "Use the Teams Toolkit to create a new Office add-in application.",
+  "core.summary.actionDescription": "Action %s%s",
+  "core.summary.lifecycleDescription": "Running lifecycle stage: %s(%s step(s) in total) The following actions will be executed %s",
```
sffamily, line 785: Q: not sure about 'Running', do we really 'run' something for each lifecycle stage? or this is to say: "Running through lifecycle stages"?

packages/fx-core/resource/package.nls.json
```diff
@@ -781,6 +781,15 @@
...
+  "core.summary.lifecycleNotExecuted": "%s Lifecycle stage %s not executed.",
+  "core.summary.lifecycleSucceeded": "%s Lifecycle stage %s executed successfully.",
```
sffamily, line 787: "executed" -> "was executed"

similar as comment above, consider make 'L' lower case if the %s before it always has value.

packages/fx-core/resource/package.nls.json
```diff
@@ -781,6 +781,15 @@
...
+  "core.summary.lifecycleFailed": "%s Lifecycle stage %s failed.",
+  "core.summary.actionNotExecuted": "%s not executed.",
```
sffamily, line 789: not executed -> was not executed

packages/fx-core/resource/package.nls.json
```diff
@@ -781,6 +781,15 @@
...
+  "core.summary.actionFailed": "%s failed.",
+  "core.summary.actionSucceeded": "%s executed successfully.",
```
sffamily, line 791: executed -> was executed

packages/fx-core/resource/package.nls.json
```diff
@@ -781,6 +781,15 @@
...
+  "core.summary.createdEnvFile": "Created environment file at",
```
sffamily, line 792: no clear subject for the action so suggest to use passive structure:

"Environment file was created at"

sffamily (changes requested):

packages/fx-core/resource/package.nls.json
1yefuwang1: The %s before it will be either √, × or ! for success, failure and not executed.

packages/fx-core/resource/package.nls.json
1yefuwang1: It looks like this in ttk.
image
What should I change it to?

jayzhang (approved):

sffamily (approved):

jayzhang (approved):

xzf0587 (approved):

PR #7713, opened 2023-02-08
chore(vsc): rearrange welcome view of upgrade

tecton:
Related task: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/17024945

Effect:
image


packages/vscode-extension/package.json
```diff
@@ -164,6 +164,11 @@
           "name": "Deployment",
           "when": "fx-extension.isTeamsFx && fx-extension.customizedTreeview"
         },
+        {
+          "id": "teamsfx-project-and-check-upgradeV3",
+          "name": "Teams Toolkit",
```
HuihuiWu-Microsoft, line 169: Is it expected that both the view and the section is called 'Teams Toolkit'?

tecton: Thank you, it should be "UPGRADE". I think it should also be localized.

sffamily (approved):

HuihuiWu-Microsoft (approved):

PR #7721, opened 2023-02-09
chore(migration): add parameters

xzf0587:
The notifacation of opening project.
image
The popup confirm dialog when click upgrade project button in tree view
image


sffamily (dismissed):

wenytang-ms (approved):

sffamily (approved):

packages/fx-core/src/core/middleware/projectMigratorV3.ts
```diff
@@ -1,19 +1,19 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
+/**
+ * @author xzf0587 <zhaofengxu@microsoft.com>
```
blackchoey, line 5: I'm not sure what's the best practice, is it fine to add our emails to public repo? @LongOddCode

tecton (approved):

packages/fx-core/src/core/middleware/projectMigratorV3.ts
xzf0587: synced with @LongOddCode , it is ok to add the email as the emails can be found in git log as well

packages/vscode-extension/src/handlers.ts
```diff
@@ -1848,6 +1848,8 @@ export async function checkUpgrade(args?: any[]) {
     const input = getSystemInputs();
     if (triggerFrom?.[TelemetryProperty.TriggerFrom] === TelemetryTriggerFrom.Auto) {
       input["isNonmodalMessage"] = true;
+      core.phantomMigrationV3(input);
+      return;
     } else if (triggerFrom?.[TelemetryProperty.TriggerFrom] === TelemetryTriggerFrom.SideBar) {
       input["confirmOnly"] = true;
```
blackchoey, line 1854: Do we also need to apply this logic for "Teams: Upgrade project" command?

xzf0587: commands will show a modal dialog not nonmodal notification.
So it will await the result

packages/fx-core/src/core/middleware/projectMigratorV3.ts
```diff
@@ -156,13 +176,21 @@ export const ProjectMigratorMWV3: Middleware = async (ctx: CoreHookContext, next
...
     }
 
-    const skipUserConfirm = getParameterFromCxt(ctx, "skipUserConfirm");
-    if (!skipUserConfirm && !(await askUserConfirm(ctx, versionForMigration))) {
-      return;
+    const isRunMigration = await showNotification(ctx, versionForMigration);
+    if (isRunMigration) {
+      const isNonmodalMessage = getParameterFromCxt(ctx, Parameters.isNonmodalMessage);
+      if (isNonmodalMessage) {
+        const versionForMigration = await checkVersionForMigration(ctx);
```
blackchoey, line 183: why we need to check version again here?

packages/fx-core/src/core/middleware/projectMigratorV3.ts
```diff
@@ -176,12 +204,12 @@ export async function wrapRunMigration(
   exec: (context: MigrationContext) => void
 ): Promise<void> {
   try {
-    sendTelemetryEvent(Component.core, TelemetryEvent.ProjectMigratorMigrateStartV3);
+    sendTelemetryEvent(Component.core, TelemetryEvent.ProjectMigratorMigrateStart);
```
blackchoey, line 207: Are we reusing previous migration event? If yes, can we add a property to distinguish 3.0->4.0 upgrade, and 4.0->5.0 upgrade?

packages/fx-core/src/core/middleware/projectMigratorV3.ts
```diff
@@ -397,21 +425,49 @@ export async function azureParameterMigration(context: MigrationContext): Promis
...
+  if (isNonmodalMessage) {
+    return await showNonmodalNotification(ctx, versionForMigration);
+  }
+  const confirmOnly = getParameterFromCxt(ctx, Parameters.confirmOnly);
+  if (confirmOnly) {
+    return await showConfirmOnlyNotification(ctx);
+  }
+  const skipUserConfirm = getParameterFromCxt(ctx, Parameters.skipUserConfirm);
+  if (skipUserConfirm) {
+    sendTelemetryEvent(Component.core, TelemetryEvent.ProjectMigratorNotification, {
```
blackchoey, line 442: Is there any case that we need to skip confirmation?

packages/fx-core/src/core/middleware/projectMigratorV3.ts
```diff
@@ -422,18 +478,78 @@ export async function askUserConfirm(
...
+  sendTelemetryEvent(Component.core, TelemetryEvent.ProjectMigratorNotificationStart);
+  const res = await TOOLS?.ui.showMessage(
+    "info",
+    getLocalizedString("core.migrationV3.confirmOnly.Message"),
+    true,
+    "OK"
+  );
+  if (res?.isOk() && res.value === "OK") {
+    sendTelemetryEvent(Component.core, TelemetryEvent.ProjectMigratorNotification, {
+      [TelemetryPropertyKey.status]: TelemetryPropertyValue.ok,
```
blackchoey, line 520: the property name `status` may not be so accurate, can we use `clickedButton` or something else? Can talk with PM about whether there's a common terminology for this.

jayzhang (approved):

packages/vscode-extension/src/handlers.ts
blackchoey: I mean should "Teams: Upgrade project" command set "confirmOnly" to true?

packages/fx-core/src/core/middleware/projectMigratorV3.ts
xzf0587: nonmodal mode will show the upgrade notification. 
If user click upgrade button in notification after the project is v3, it will trigger migration error.
Double check to exclude this case.

packages/fx-core/src/core/middleware/projectMigratorV3.ts
xzf0587: vs platform has a case for it.

packages/vscode-extension/src/handlers.ts
xzf0587: No. It will trigger the modal dialog

packages/fx-core/src/core/middleware/projectMigratorV3.ts
xzf0587: OK. will use another pr for the telemetry property

packages/fx-core/src/core/middleware/projectMigratorV3.ts
xzf0587: OK. will use another pr for the telemetry property

PR #7733, opened 2023-02-09
chore(migration): update

xzf0587:


Alive-Fish (approved):

packages/fx-core/src/core/FxCoreImplementV3.ts
```diff
@@ -364,7 +366,17 @@ export class FxCoreV3Implement {
         return err(new InvalidProjectError());
       }
       const trackingId = await getTrackingIdFromPath(projectPath);
-      const isSupport = getVersionState(versionInfo);
+      let isSupport: VersionState;
+      // As projectVersionCheck is a v3 interface, v3 not enabled case is an exception and only called by vs platform
+      if (!isV3Enabled() && inputs.platform === Platform.VS) {
```
Alive-Fish, line 371: I think it is also correct when the platform is not VS.

xzf0587: Yes. Currently only VS platform will use the interface in v2.
If other platforms use the interface in v2, we should check the flow.

blackchoey (commented):
Overall looks good to me.

blackchoey:
@Alive-Fish @xzf0587 Have a small comment about the relationship between VS and fx-core: I see there's logic that checks the platform and do things differently, but is it better to let VS/VS Code/CLI to do the platform specific things? I personally feels fx-core should not aware which platform calls it. So maybe the logic can move to VS and becomes something like:
```
// the code in VS extension
var isV2 = fx.isValidV2Project(path);
var v3Enabled = readTeamsToolkitSettings("enableV3");
if (isV2 && v3Enabled) {
   VSUI.popupDialog("Please upgrade your project"); // any UI we want to show
}

```

Does it make sense to you? It's not urgent but maybe we can improve in the future.

jayzhang (approved):

xzf0587:
> @Alive-Fish @xzf0587 Have a small comment about the relationship between VS and fx-core: I see there's logic that checks the platform and do things differently in fx-core, but is it better to let VS/VS Code/CLI to do the platform specific things? I personally feels fx-core should not aware which platform calls it. So maybe the logic can move to VS and becomes something like:
> 
> ```
> // the code in VS extension
> var isV2 = fx.isValidV2Project(path);
> var v3Enabled = readTeamsToolkitSettings("enableV3");
> if (isV2 && v3Enabled) {
>    VSUI.popupDialog("Please upgrade your project"); // any UI we want to show
>    return;
> }
> if (isV2 && !v3Enabled) {
>    // nothing need to be done
>    return;
> }
> ```
> 
> Does it make sense to you? It's not urgent but maybe we can improve in the future.

It is a protective condition as this interface only called by VS in v2 scenario, which means other platform should not call it if v3 flag is not enabled. 
We will clean the code in v3 flag refactor

PR #7741, opened 2023-02-10
fix(vsc): show development commands after upgrade

tecton:
1. Fix commands list display
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/17115597
2. Add "learn more"
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/17027510
3. Fix feedback button.

E2E TEST: N/A


packages/vscode-extension/package.json
```diff
@@ -127,7 +127,7 @@
       {
         "view": "teamsfx-feedback",
         "contents": "%teamstoolkit.viewsWelcome.teamsfx-feedback.content%",
-        "enablement": "fx-extension.isTeamsFx && !fx-extension.canUpgradeV3"
+        "enablement": "fx-extension.initialized"
```
HuihuiWu-Microsoft, line 130: With this change, will 'feedback' panel be shown when current project can be upgraded to v3?

tecton: This is a misusage of "enablement", which means the button is clickable. This change fixes the bug that feedback button is shown but cannot be clicked. It should always be clickable. BTW, feedback is designed to show when project can upgrade to V3.

sffamily (approved):

tecton:
Replaced by #7747

PR #7756, opened 2023-02-13
fix: update tab localhost URL if app from TDP

yuqizhou77:
- the default localhost URL for a tab project in C# is different from the value put in JS/TS project, so need to remove hard-code part and get the value from downloaded template.

E2E TEST: N/A

### DIFF
--- a/packages/fx-core/src/component/developerPortalScaffoldUtils.ts
+++ b/packages/fx-core/src/component/developerPortalScaffoldUtils.ts
@@ -357,8 +357,8 @@ export function updateScope(scopes: string[]): string[] {
   return scopes.map((o) => o.toLowerCase());
 }
 
-export function isFromDevPortalInVSC(inputs: Inputs): boolean {
-  return !!inputs.teamsAppFromTdp && inputs.platform === Platform.VSCode;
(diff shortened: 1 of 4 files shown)


yuqizhou77:
work item: [Bug 17146135](https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/17146135): [VS] localhost URL is incorrect in project created from TDP

packages/fx-core/src/component/developerPortalScaffoldUtils.ts
```diff
@@ -122,6 +122,10 @@ async function updateManifest(
   }
   const existingManifestTemplate = manifestRes.value;
 
+  if (!existingManifestTemplate) {
+    return err(new ObjectIsUndefinedError("manifest.json downloaded from template"));
```
nliu-ms, line 126: should it be localized?

yuqizhou77: I don't think we do localization here usually. the whole message is something like this " error.ObjectIsUndefinedError": "Object %s is undefined, which is not expected" where we put object name.

nliu-ms (approved):

PR #7767, opened 2023-02-13
fix(debug): fix resolving msedge configuration

kuojianlu:
From practice, if debug configuration type is `msedge`, `resolveDebugConfiguration` will be called twice. The type of the configuration in first time is `msedge` but `pwa-msedge` in second time.

If `url` is not specified in the configuration, the `url` in first time is `undefined` but `null` in second time.


swatDong (approved):
The fix code looks good.
But for "*`msedge` resolves twice*", better to create issue on vscode repo to get more details... that's so weird.

kuojianlu:
> The fix code looks good. But for "_`msedge` resolves twice_", better to create issue on vscode repo to get more details... that's so weird.

Created a issue: https://github.com/microsoft/vscode/issues/174310.

PR #7777, opened 2023-02-14
chore(migration): update notification

xzf0587:
1. Set upgrade button to default
image
image

2. Remove migration v3 flag, using v3 flag instead.

### DIFF
--- a/packages/fx-core/src/core/middleware/projectMigratorV3.ts
+++ b/packages/fx-core/src/core/middleware/projectMigratorV3.ts
@@ -88,7 +88,7 @@ import { AppLocalYmlGenerator } from "./utils/debug/appLocalYmlGenerator";
 import { EOL } from "os";
 import { getTemplatesFolder } from "../../folder";
 import { MetadataV2, MetadataV3, VersionSource, VersionState } from "../../common/versionMetadata";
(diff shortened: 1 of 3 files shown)


jayzhang (approved):

blackchoey (approved):
minor: can add a test case or this if possible

tecton (approved):

packages/fx-core/src/core/middleware/projectMigratorV3.ts
```diff
@@ -564,11 +566,17 @@ export async function popupMessage(
   versionForMigration: VersionForMigration,
   isModal: boolean
 ): Promise<string | undefined> {
+  let buttons: string[];
+  if (!isModal) {
+    buttons = [upgradeButton, learnMoreText];
+  } else {
+    buttons = migrationMessageButtons;
```
tecton, line 573: Minor: seems like `migrationMessageButtons` are also reversed, it's for modal dialog and the default value is "learn more".

xzf0587: Will discuss with PM and use another PR  to update it if necessary

tecton (approved):

blackchoey (approved):

jayzhang (approved):

PR #7787, opened 2023-02-15
chore(m365): expose upgrade command for non-teamsfx project

kuojianlu:
Fix https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/16555275.

### DIFF
--- a/packages/vscode-extension/package.json
+++ b/packages/vscode-extension/package.json
@@ -604,11 +604,15 @@
       },
       {
         "command": "fx-extension.migrateTeamsManifest",
-        "title": "%teamstoolkit.commands.migrateManifest.title%"
+        "title": "%teamstoolkit.commands.migrateManifest.title%",
+        "enablement": "isWorkspaceTrusted",
+        "category": "Teams"
       },
       {
         "command": "fx-extension.migrateTeamsTabApp",
-        "title": "%teamstoolkit.commands.migrateApp.title%"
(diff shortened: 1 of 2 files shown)


swatDong (dismissed):

packages/vscode-extension/package.json
```diff
@@ -604,15 +604,11 @@
       },
       {
         "command": "fx-extension.migrateTeamsManifest",
-        "title": "%teamstoolkit.commands.migrateManifest.title%",
-        "enablement": "fx-extension.isTeamsFx && isWorkspaceTrusted",
```
tecton, line 608: Please keep the `isWorkspaceTrusted` condition and "category" property, otherwise the prefix Teams will be missing. isWorkspaceTrusted is suggested to preserve because extension will read content of workspace file, if user doesn't trust it, this may contain malicious codes.

tecton (changes requested):
I guess you'll need to move the registration of these 2 commands to non-teamsfx part in `extension.ts`. Currently it's only registered when a TeamsFx project is detected. Without registration the command will not work.

packages/vscode-extension/package.json
kuojianlu: Thanks for the explanation. Updated.

swatDong (dismissed):

tecton (dismissed):

tecton (approved):

PR #7802, opened 2023-02-16
chore(debug): add teams channel to create/update bot request body

kuojianlu:
@microsoft/teamsfx-cli	v1.2.2-rc.1
@microsoft/teamsfx-core	v1.23.0-rc.1
ms-teams-vscode-extension	v4.2.2-rc.1

CLI BREAKING CHANGE:
  

CLI feat commits:
  

### DIFF
--- a/packages/fx-core/src/component/resource/botService/appStudio/appStudioClient.ts
+++ b/packages/fx-core/src/component/resource/botService/appStudio/appStudioClient.ts
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
-import { IBotRegistration } from "./interfaces/IBotRegistration";
+import { BotChannelType, IBotRegistration } from "./interfaces/IBotRegistration";
 
 import { AxiosInstance, AxiosResponse, default as axios } from "axios";
 import {
@@ -131,6 +131,9 @@ export class AppStudioClient {
     }
 
     botReg.messagingEndpoint = endpoint;
+    if (botReg.configuredChannels === undefined || botReg.configuredChannels.length === 0) {
+      botReg.configuredChannels = [BotChannelType.MicrosoftTeams];
+    }
 
     await AppStudioClient.updateBotRegistration(token, botReg);
 
--- a/packages/fx-core/src/component/debugHandler/bot.ts
+++ b/packages/fx-core/src/component/debugHandler/bot.ts
@@ -30,7 +30,10 @@ import {
   getProjectSettingsPath,
   loadProjectSettingsByProjectPath,
 } from "../../core/middleware/projectSettingsLoader";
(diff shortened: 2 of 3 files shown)


kuojianlu:
This is to fix https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/16985827.

swatDong (dismissed):

packages/fx-core/src/component/resource/botService/appStudio/appStudioClient.ts
```diff
@@ -131,6 +131,9 @@ export class AppStudioClient {
     }
 
     botReg.messagingEndpoint = endpoint;
+    if (botReg.configuredChannels === undefined || botReg.configuredChannels.length === 0) {
+      botReg.configuredChannels = [BotChannelType.MicrosoftTeams];
```
IvanJobs, line 135: To make it robust, how about appending "Teams Channel" if "Teams Channel" doesn't exist in the channel list? Just considering len(configuredChannels) > 0, I'm not sure if there must have a Teams channel in it.

IvanJobs (approved):

packages/fx-core/src/component/resource/botService/appStudio/appStudioClient.ts
kuojianlu: This scenario will not appear. It means users manually deleted Teams channel and enabled M365 Extensions channel. Then let us just respect users' behavior.

swatDong (approved):

packages/fx-core/src/component/resource/botService/appStudio/appStudioClient.ts
eriolchan: The method is to update message endpoint, why do we add hidden logic to update configured channels here? I have concern to add this temporary workaround logic here, which is error-prone. Should this be in the create bot registration steps?

kuojianlu: Please see the following explantion of this issue. To fix it in TTK, we have to add Teams channel to the request body when updating messaging endpoint, i.e. updating bot.

During F5, TTK handles bot via 3 steps.
1. Creates a bot (POST /api/botframework). This API uses regional bot framework API to create a bot, and then enable Teams channel.
2. Gets the bot (GET /api/botframework/<botId>). This API uses non-regional bot framework API to get the bot.
3. Update the bot messaging endpoint (POST /api/botframework/<botId>). This API uses non-regional bot framework API to update the bot.

As the bot framework API used in step 2 is different from that in step 1, error may happen. Consider this case: in step 1, bot has been created and synchronized to non-regional API; Teams channel has been enabled but not yet synchronized to non-regional API. Then in step 2, the bot returned from non-regional API does not contain Teams channel. Then in step 3, the bot without Teams channel overrides the bot in step 1 which enables Teams channel. The final result is a bot without Teams channel.

PR #7821, opened 2023-02-17
chore(vsc): lazy load in non-teamsfx project

tecton:
1. Upgrade to VS Code 1.66 to leverage new feature [Output channel with custom language ID](https://code.visualstudio.com/updates/v1_66#_output-channel-with-custom-language-id)
2. Remove activationEvents "onStartupFinished" so that our extension will be activated automatically in non-teamsfx project.
3. Remove languages/mimetypes to prevent auto-onLanguage activation.

Related task: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/17137840

### DIFF
--- a/packages/vscode-extension/src/commonlib/log.ts
+++ b/packages/vscode-extension/src/commonlib/log.ts
@@ -14,7 +14,10 @@ export class VsCodeLogProvider implements LogProvider {
   private static instance: VsCodeLogProvider;
 
   private constructor() {
-    this.outputChannel = vscode.window.createOutputChannel(outputChannelDisplayName);
+    this.outputChannel = vscode.window.createOutputChannel(
+      outputChannelDisplayName,
+      "teamsfx-toolkit-output"
+    );
   }
 
   /**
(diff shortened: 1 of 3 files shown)


packages/vscode-extension/src/commonlib/log.ts
```diff
@@ -14,7 +14,10 @@ export class VsCodeLogProvider implements LogProvider {
   private static instance: VsCodeLogProvider;
 
   private constructor() {
-    this.outputChannel = vscode.window.createOutputChannel(outputChannelDisplayName);
+    this.outputChannel = vscode.window.createOutputChannel(
+      outputChannelDisplayName,
+      "teamsfx-toolkit-output"
```
swatDong, line 19: Is this a reusable constant or just to align with `package.json`?
If reusable, better to define in some `constants.ts` file.
If just to align with `package.json`, add comment such as *// Align with languages:id value in package.json*, for better maintenance.

tecton: Thank you! The comment is added.

swatDong (approved):

HuihuiWu-Microsoft (approved):

PR #7838, opened 2023-02-22
chore(core): delete unused strings

jayzhang:


sffamily (approved):
assume it is simplifying by consolidating repeating wording into a structured pattern. Since it is removal only, by default approve it.

PR #7859, opened 2023-02-23
chore: polish generator utils and tests

yukun-dong:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_sprints/taskboard/Compute%20Team/Microsoft%20Teams%20Extensibility/CY23-2.2?workitem=17264435

### DIFF
--- a/packages/fx-core/src/component/generator/utils.ts
+++ b/packages/fx-core/src/component/generator/utils.ts
@@ -16,26 +16,23 @@ import {
 import { SampleInfo, sampleProvider } from "../../common/samples";
 import AdmZip from "adm-zip";
 import axios, { AxiosResponse, CancelToken } from "axios";
-import { EOL } from "os";
 import templateConfig from "../../common/templates-config.json";
 import sampleConfig from "../../common/samples-config-v3.json";
 import semver from "semver";
 
-export const preRelease = (): string =>
-  process.env.TEAMSFX_TEMPLATE_PRERELEASE ? `0.0.0-${process.env.TEAMSFX_TEMPLATE_PRERELEASE}` : "";
-export const templateVersion = (): string => templateConfig.version;
-export const templateTagPrefix = templateConfig.tagPrefix;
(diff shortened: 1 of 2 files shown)


packages/fx-core/src/component/generator/utils.ts
```diff
@@ -153,8 +146,7 @@ export async function unzip(
   dstPath: string,
   nameReplaceFn?: (filePath: string, data: Buffer) => string,
   dataReplaceFn?: (filePath: string, data: Buffer) => Buffer | string,
-  relativePath?: string,
-  filesInAppendMode = [".gitignore"]
```
hund030, line 157: This is v2 legacy code. Thanks to clean up it.

packages/fx-core/src/component/generator/utils.ts
```diff
@@ -118,9 +111,9 @@ export async function fetchTemplateZipUrl(
   tryLimits = defaultTryLimits,
   timeoutInMs = defaultTimeoutInMs
 ): Promise<string> {
-  const selectedTag = await selectTemplateTag(async () =>
-    (await fetchTagList(templateTagListURL, tryLimits, timeoutInMs)).replace(/\r/g, "").split("\n")
-  );
+  const templateTagListURL = templateConfig.tagListURL;
+  const tags = await fetchTagList(templateTagListURL, tryLimits, timeoutInMs);
```
hund030, line 115: I suppose making `fetchTagList` as a callback function can improve performance.
When it is alpha version, generator does not need to fetch tag list and use fallback template immediately.
While it indeed hurts the code readability. You can make the decision.

hund030 (dismissed):

packages/fx-core/src/component/generator/utils.ts
yukun-dong: You are absolutely right. I will use a callback function here.

hund030 (approved):

PR #7873, opened 2023-02-27
chore: no need to include valid domain for non-sso app

hund030:


huimiu (approved):

hund030:
[Create a content page - Teams | Microsoft Learn](https://learn.microsoft.com/en-us/microsoftteams/platform/tabs/how-to/create-tab-pages/content-page?tabs=teamsjs-v2#valid-domains)
Teams recommends adding all URL domains used in the tabs to valid domain list.
Keep the valid domains list in non sso tab template even though the template works without valid domains.

PR #7874, opened 2023-02-27
chore: replace client id in codespaces

xiaolang124:
Use our client id in Code Spaces login. Add notify logic for m365 code spaces login.

https://www.eliostruyf.com/microsoft-authentication-provider-visual-studio-code/

### DIFF
--- a/packages/vscode-extension/src/commonlib/m365CodeSpaceLogin.ts
+++ b/packages/vscode-extension/src/commonlib/m365CodeSpaceLogin.ts
@@ -19,7 +19,11 @@ import {
 import { getDefaultString } from "../utils/localizeUtils";
 import { UserCancelError } from "./codeFlowLogin";
 
-const CodeSpacesScopes = [
+// By using `VSCODE_CLIENT_ID:<aad-app-id>` in the `vscode.authentication` API,
+// we can override the scope with our own first-party AAD app.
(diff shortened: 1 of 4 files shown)


packages/vscode-extension/src/commonlib/m365CodeSpaceLogin.ts
```diff
@@ -19,6 +19,11 @@ import {
 import { getDefaultString } from "../utils/localizeUtils";
 import { UserCancelError } from "./codeFlowLogin";
 
+const CodeSpacesScopes = [
```
dooriya, line 22: `TeamsFxAadScopes` would be more accurate here, and the M365 local login and codespaces login can be unified in our extension as suggested from VS Code team. 

Further more, better to add comment for this secret scope. for example,

```typescript
// By using `VSCODE_CLIENT_ID:<aad-app-id>` in the `vscode.authentication` API, 
// we can override the scope with our own first-party AAD app. 
// Check https://www.eliostruyf.com/microsoft-authentication-provider-visual-studio-code/ for more details.
```

xiaolang124: Thanks for the advice. Done.

dooriya (approved):

swatDong (dismissed):

swatDong (approved):

PR #7897, opened 2023-02-28
chore: add cache policy to remove the cache for link unfurling

JerryYangKai:


yuqizhou77:
guess similar change needed to be add here? https://github.com/OfficeDev/TeamsFx/blob/dev/templates/scenarios/js/non-sso-tab-default-bot/bot/teamsBot.js

JerryYangKai:
> guess similar change needed to be add here? https://github.com/OfficeDev/TeamsFx/blob/dev/templates/scenarios/js/non-sso-tab-default-bot/bot/teamsBot.js

This template's manifest only has tab and bot capability. I think the msg-ext part is redundant here.

JerryYangKai:
@MuyangAmigo Please help review the wording in the comment. I found the related link is already in the readme, so I haven't updated the readme.

yuqizhou77:
> > guess similar change needed to be add here? https://github.com/OfficeDev/TeamsFx/blob/dev/templates/scenarios/js/non-sso-tab-default-bot/bot/teamsBot.js
> 
> This template's manifest only has tab and bot capability. I think the msg-ext part is redundant here.

this is by design as this template won't be shown in UI but just for users from TDP.

templates/scenarios/csharp/message-extension/Bot/TeamsMessageExtension.cs.tpl
```diff
@@ -194,9 +194,17 @@ public class TeamsMessageExtension : TeamsActivityHandler
             Images = new List<CardImage> { new CardImage("https://raw.githubusercontent.com/microsoft/botframework-sdk/master/icon.png") },
         };
 
+        // By default the link unfurling result is cached in Teams for 30 minutes.
+        // The code has set a cache policy and removed the cache for the app.
+        var action = new CardAction
```
MuyangAmigo, line 199: Can we link the doc from here as well, e.g.,

```
// By default the link unfurling result is cached in Teams for 30 minutes.
 // The code has set a cache policy and removed the cache for the app. Learn more here: https://learn.microsoft.com/microsoftteams/platform/messaging-extensions/how-to/link-unfurling?tabs=dotnet%2Cadvantages#remove-link-unfurling-cache
```

templates/scenarios/ts/default-bot-message-extension/teamsBot.ts
```diff
@@ -165,11 +165,23 @@ export class TeamsBot extends TeamsActivityHandler {
...
+    // The code has set a cache policy and removed the cache for the app.
     return {
       composeExtension: {
         type: "result",
         attachmentLayout: "list",
         attachments: [attachment],
+        suggestedActions: {
+          actions: [
+            {
+              title: "default",
```
yukun-dong, line 179: Is this `title` needed?

JerryYangKai: It is a required, not an optional input, only in ts.

JerryYangKai:
@yukun-dong Since you are the owner, how's your suggestion about how to deal with the template non-sso-tab-default-bot, as @yuqizhou77 said. Do we need also remove cache in this template?

templates/scenarios/ts/default-bot-message-extension/teamsBot.ts
yukun-dong: Indeed `title` is required in `CardAction` type. If the doc doesn't give any suggestions for the `title` value, "default" is ok to me.

yukun-dong:
> @yukun-dong Since you are the owner, how's your suggestion about how to deal with the template non-sso-tab-default-bot, as @yuqizhou77 said. Do we need also remove cache in this template?



> @yukun-dong Since you are the owner, how's your suggestion about how to deal with the template non-sso-tab-default-bot, as @yuqizhou77 said. Do we need also remove cache in this template?

Yes, we should make them consistent.

SLdragon (approved):

templates/scenarios/csharp/message-extension/Bot/TeamsMessageExtension.cs.tpl
eriolchan: what's the recommendation of Production? Does ME suggest enable cache for Production code?
If that's only true for debugging, is it better to put this part in README instead of template code? @JerryYangKai  @MuyangAmigo

JerryYangKai: @eriolchan From official document there is no suggest environment for enable cache. It depends on the customers need.

SLdragon (approved):

PR #7929, opened 2023-03-02
chore: store artifact to pipeline instead of GitHub release page

wenytang-ms:


hund030 (dismissed):

qinezh (dismissed):

.github/workflows/cd.yml
```diff
@@ -412,3 +345,5 @@ jobs:
           path: |
             changed.txt
             versions.json
+            ./packages/vscode-extension/*.vsix
+            ./packages/server/lib/*.exe
```
Alive-Fish, line 349: The built server under "./packages/server/lib" has the version info as below: add server version as output to VS or change the "./packages/server/buildExe.js" to let the output named "server.exe".

image

Alive-Fish (approved):

LongOddCode (approved):

PR #7930, opened 2023-03-02
chore(m365): update m365 sideloading

kuojianlu:
- remove calling launchInfo api in sideloading
- get title id and app id from status api in sideloading
- persist M365_APP_ID in m365Title/acquire action
- use M365_APP_ID to launch office
- update templates comment
- update cli preview to use title

### DIFF
--- a/packages/fx-core/src/common/m365/packageService.ts
+++ b/packages/fx-core/src/common/m365/packageService.ts
@@ -40,7 +40,7 @@ export class PackageService {
     }
   }
 
-  public async sideLoading(token: string, manifestPath: string): Promise<string> {
+  public async sideLoading(token: string, manifestPath: string): Promise<[string, string]> {
     try {
       const data = await fs.readFile(manifestPath);
       const content = new FormData();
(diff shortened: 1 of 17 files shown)


packages/fx-core/src/common/m365/packageService.ts
```diff
@@ -90,32 +88,15 @@ export class PackageService {
         );
         const resCode = statusResponse.status;
         if (resCode === 200) {
-          complete = true;
+          const titleId: string = statusResponse.data.titleId;
+          const appId: string = statusResponse.data.appId;
+          this.logger?.info(`Acquire done. App TitleId: ${titleId}, AppId: ${appId}`);
```
swatDong, line 93: Office addin matches title id from the output, so need to keep the original titleid output format.
https://github.com/OfficeDev/Office-Addin-Scripts/blob/master/packages/office-addin-dev-settings/src/publish.ts#L19

Since there's no "Check acquired package" step, how about just output?
```
TitleId: ...
AppId: ...
Sideloading done.
```

kuojianlu: Updated.

swatDong (dismissed):

swatDong (approved):

PR #7938, opened 2023-03-02
feat: add debug task to launch Teams web client for codespaces

dooriya:
**Motivation and context:**
Currently Codespaces doesn't support launching browser by using `msedge` / `chrome` launch configuration. So we need a dedicate debug task to launch app in Teams web client for both local & remote environments.

**How to use in task.json**:
```
{
  "label": "Launch Teams for Codespaces",
  "type": "teamsfx",
  "command": "launch-web-client",
  "args": {
    "env": "local"
  }
}
```

**How it works**:
The task will launch the app in Teams web client with the following url: 
`https://teams.microsoft.com/l/app/${teamsAppId}?installAppPackage=true&webjoin=true&${accountHint}`
>Note: `teamsAppId` and `accountHint` will be resolved in this task according to the `env` specified in the task arguments.

**Work item**:
[User Story 17335275](https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/17335275): [Codespaces] Add built-in debug task for launching Teams web client

**To-do**:
- Add help document in [wiki page](https://github.com/OfficeDev/TeamsFx/wiki/%7BDebug%7D-Teams-Toolkit-v5-VS-Code-Tasks) for `launch-web-client` task.

E2E TEST: N/A


packages/fx-core/src/common/local/constants.ts
```diff
@@ -98,6 +98,7 @@ export const TaskCommand = Object.freeze({
   setUpBot: "debug-set-up-bot",
   setUpSSO: "debug-set-up-sso",
   prepareManifest: "debug-prepare-manifest",
+  launchTeamsWebClient: "debug-launch-teams-client",
```
swatDong, line 101: Will this be used to launch outlook and m365 app later?

swatDong: And, it seems this task is not for local only, and can also be used for remote preview. So how about removing the `debug-` prefix?

packages/vscode-extension/src/debug/taskTerminal/launchTeamsClientTerminal.ts
```diff
@@ -0,0 +1,156 @@
...
+
+  private openUrl(url: string): Promise<Result<Void, FxError>> {
+    return new Promise<Result<Void, FxError>>((resolve, reject) => {
+      const options: cp.SpawnOptions = {
+        cwd: globalVariables.workspaceUri?.fsPath ?? "",
+        shell: true,
+        detached: false,
+      };
+
+      const childProc = cp.spawn("npx", ["open-cli", url], options);
```
swatDong, line 91: What if `open-cli` is not installed?
Since the task is totally managed by ttk, how about directly using the [open](https://www.npmjs.com/package/open) package?

packages/fx-core/src/common/local/constants.ts
dooriya: - Removing `debug-` prefix make sense to me.
- We're still considering whether to support m365 app later, and we can first add support for Teams. If m365 support is needed we can add PRs later for launching other hubs (and the command name will change to something like `launch-m365-hub`)

swatDong: Rename a task command is a bit heavy... So I'd suggest the task command be something like `launch-web-client`, and make the task label in tasks.json more specific, e.g., `Launch Teams Web Client`.

packages/vscode-extension/src/debug/taskTerminal/launchTeamsClientTerminal.ts
dooriya: `npx` can execute package which was not previously installed. But I do agree with you that using [open](https://www.npmjs.com/package/open) package here will be more straightforward since it's in ttk not in the `task.json`. Will change to use `open` package. Thanks.

swatDong (approved):

swatDong (approved):

packages/vscode-extension/package.nls.json
```diff
@@ -366,6 +368,7 @@
     "teamstoolkit.taskDefinitions.command.startLocalTunnel.description": "Start the local tunneling service.\n See https://aka.ms/teamsfx-tasks/local-tunnel for details and how to customize the arguments.",
     "teamstoolkit.taskDefinitions.command.provision.description": "Execute provision lifecycle.\n See https://aka.ms/teamsfx-tasks/provision for details and how to customize the arguments.",
     "teamstoolkit.taskDefinitions.command.deploy.description": "Execute deploy lifecycle.\n See https://aka.ms/teamsfx-tasks/deploy for details and how to customize the arguments.",
+    "teamstoolkit.taskDefinitions.command.launchWebClient.description": "Launch Teams web client. \n See https://aka.ms/teamsfx-tasks/launch-web-client for details and how to customize the arguments.",
```
HuihuiWu-Microsoft, line 371: Better exclude link in strings since it's not needed for localization

packages/vscode-extension/package.nls.json
```diff
@@ -267,6 +267,8 @@
     "teamstoolkit.localDebug.tunnelServiceNotStartedError": "The tunnel service hasn't been started. Please wait for a moment or restart the local tunnel task.",
     "teamstoolkit.localDebug.tunnelEndpointNotFoundError": "Failed to find the tunnel endpoint. Teams toolkit try to get the first HTTPS URL from %s and failed.",
     "teamstoolkit.localDebug.tunnelEnvError": "Failed to save environment variables. %s",
+    "teamstoolkit.localDebug.launchTeamsWebClientError": "Failed to launch Teams web client. %s",
```
HuihuiWu-Microsoft, line 270: The %s placeholder will be quite confusing to localization team. If it's an independent string, separate it from this one.

packages/vscode-extension/package.nls.json
dooriya: Sure. will use placeholder for the link and resolve in the caller.

packages/vscode-extension/package.nls.json
dooriya: Remove %s for this case. Thanks for the suggestion.

dooriya: Just found that this string will be directly used in `package.json`, so just keep it.

packages/vscode-extension/src/debug/taskTerminal/launchTeamsClientTerminal.ts
```diff
@@ -0,0 +1,150 @@
...
+
+export class LaunchTeamsClientTerminal extends BaseTaskTerminal {
+  private readonly args: LaunchTeamsClientArgs;
+
+  constructor(taskDefinition: vscode.TaskDefinition) {
+    super(taskDefinition);
+    this.args = taskDefinition.args as LaunchTeamsClientArgs;
+  }
+
+  do(): Promise<Result<Void, FxError>> {
```
XiaofuHuang, line 39: Should the 'stop' method be implemented to terminate the child process? The 'stop' method will be invoked when the debug session ends or when the terminal is closed.

packages/vscode-extension/package.nls.json
```diff
@@ -267,6 +267,8 @@
...
+    "teamstoolkit.localDebug.launchTeamsWebClientStoppedError": "Launch Teams web client task stopped with exit code '%s'.",
```
sffamily, line 271: either "Task to launch Teams web client ..." 
or "Launch-Teams-web-client task ..."

packages/vscode-extension/src/debug/constants.ts
```diff
@@ -353,4 +353,12 @@ export const sideloadingDisplayMessages = Object.freeze({
     "The app supports hot reloading. If you have any code changes within the project, the app will be reloaded.",
 });
 
+export const launchingTeamsClientDisplayMessages = Object.freeze({
+  title: "Launching Teams web client.",
+  launchUrlMessage: (url: string) =>
+    `Teams web client is being launched for you to debug the Teams app: ${url}.`,
+  hotReloadingMessage:
+    "The app supports hot reloading. If you have any code changes within the project, the app will be reloaded.",
```
sffamily, line 361: within -> in

XiaofuHuang (approved):

packages/vscode-extension/src/debug/taskTerminal/launchTeamsClientTerminal.ts
dooriya: The child process here will exist immediately after execution, so here we don't need to kill the process in `stop()` method when debug session end or terminal is closed.

packages/vscode-extension/src/debug/constants.ts
dooriya: updated.

packages/vscode-extension/package.nls.json
dooriya: Updated.

HuihuiWu-Microsoft (approved):

XiaofuHuang (approved):

sffamily (approved):

PR #7945, opened 2023-03-03
chore(localdebug): delete the dev tunnel if it is not deleted in the last F5

XiaofuHuang:
detail: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/17358463

### DIFF
--- a/packages/vscode-extension/src/debug/taskTerminal/devTunnelTaskTerminal.ts
+++ b/packages/vscode-extension/src/debug/taskTerminal/devTunnelTaskTerminal.ts
@@ -170,7 +170,7 @@ export class DevTunnelTaskTerminal extends BaseTunnelTaskTerminal {
         clusterId: idArr[1],
       });
 
-      if (tunnelInstance) {
+      if (tunnelInstance?.tags?.includes(DevTunnelTag)) {
         await this.tunnelManagementClientImpl.deleteTunnel(tunnelInstance);
       }
     } catch {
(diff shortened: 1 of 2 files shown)


kuojianlu (approved):

packages/vscode-extension/src/debug/taskTerminal/devTunnelTaskTerminal.ts
```diff
@@ -132,7 +132,7 @@ export class DevTunnelTaskTerminal extends BaseTunnelTaskTerminal {
     await this.outputStartMessage(devTunnelDisplayMessages);
     await this.outputStartDevTunnelStepMessage(devTunnelDisplayMessages);
     await this.resolveArgs(this.args);
-    // TODO: delete the last debug tunnel if it is not deleted
+    await this.deleteExistingTunnel();
```
a1exwang, line 135: How long does delete operation usually take? If deletion is slow, should we do it asynchronously (without await)?

XiaofuHuang: > How long does delete operation usually take? If deletion is slow, should we do it asynchronously (without await)?
It is about 0.2s.

Once the maximum number of tunnels is reached, the existing tunnel should be deleted before they can create a new one.

XiaofuHuang: At present, Basis limits each user to a maximum of 10 tunnels, and in the future, the number of tunnels available for the free tier may be further reduced. To ensure that the previous tunnel is closed correctly and avoid reaching the maximum number of tunnels (may be similiar to ngrok, allow only 2 tunnel to be opened), I recommend deleting the tunnel before creating a new one, as long as a delay of 0.2 seconds is acceptable

a1exwang (approved):

PR #7950, opened 2023-03-03
chore: new spfx scaffolding question for package selection

yuqizhou77:
- add new questions for SPFx package selection (temporarily put it behind a feature flag)
image

- If user currently does not have any SPFx installed globally, we won't show version details
image

- Since getting versions of latest or installed SPFx will take some time, we show a progress indicator when loading (similar to the UI we have today when checking NPM and Node version)
image


packages/fx-core/resource/package.nls.json
```diff
@@ -187,6 +187,12 @@
   "plugins.spfx.questions.versionCheck.title": "Checking Node.js and NPM versions...",
   "plugins.spfx.questions.webpartName.error.duplicate": "The folder %s already exists. Choose a different name for your component.",
   "plugins.spfx.questions.webpartName.error.notMatch": "%s doesn't match pattern: %s",
+  "plugins.spfx.questions.packageSelect.title": "SharePoint Framework",
+  "plugins.spfx.questions.packageSelect.placeholder": "Select an option for scaffolding",
+  "plugins.spfx.questions.packageSelect.useGlobalPackage.withVersion.label": "Use globally installed SPFx (%s)",
```
HuihuiWu-Microsoft, line 192: Add 'v' prefix? Need PM confirm here.

packages/fx-core/src/component/resource/spfx/utils/constants.ts
```diff
@@ -26,6 +26,9 @@ export class Constants {
   public static readonly SUPPORTED_NPM_VERSION = ["7", "8"];
   public static readonly SPFX_HELP_LINK = "https://aka.ms/teamsfx-spfx-help";
   public static readonly TEMPLATE_NAME = "spfx-tab";
+  public static readonly LatestVersion = "latest";
+  public static readonly RecommenLowestSpfxVersion = "v1.14.0";
```
HuihuiWu-Microsoft, line 30: Typo: Recommend

HuihuiWu-Microsoft:
Since user is able to go back to last step in question flow, is everything OK in that case?

yuqizhou77:
> Since user is able to go back to last step in question flow, is everything OK in that case?

Yes, similar to checking Node and Npm Version, user won't go back to the question for us to load versions.

packages/fx-core/resource/package.nls.json
yuqizhou77: updated. thought it would be consistent with the description message we put.

yuqizhou77: updated the code again. and will add "v" in code when generating string. Please see the attached screenshot in PR description.

packages/fx-core/resource/package.nls.json
```diff
@@ -187,6 +187,13 @@
...
+  "plugins.spfx.questions.packageSelect.useGlobalPackage.noVersion.label": "Use globally installed SPFx",
```
sffamily, line 193: Please check whether this "SPFx" is publicly communicated to end users/developers. If not through their official documents or sources, we should always avoid using acronyms, but to use the full name instead whenever possible.

sffamily: The finding will apply to all occurences of SPFx.

packages/fx-core/resource/package.nls.json
```diff
@@ -187,6 +187,13 @@
...
+  "plugins.spfx.questions.packageSelect.useGlobalPackage.detail": "Recommend SPFx %s or later",
+  "plugins.spfx.questions.packageSelect.installLocally.withVersion.label": "Install latest SPFx (%s) locally in Teams Toolkit directory ",
```
sffamily, line 195: latest -> the latest

packages/fx-core/resource/package.nls.json
```diff
@@ -187,6 +187,13 @@
...
+  "plugins.spfx.questions.packageSelect.installLocally.noVersion.label": "Install latest SPFx locally in Teams Toolkit directory ",
```
sffamily, line 196: latest -> the latest

packages/fx-core/resource/package.nls.json
yuqizhou77: Updated to "SharePoint Framework" as what spfx doc uses.

sffamily (approved):

packages/fx-core/resource/package.nls.json
yuqizhou77: synced offline and we will confirm with SPFx team about whether "SPFx" is enough.

HuihuiWu-Microsoft (approved):

PR #7986, opened 2023-03-08
chore(core): script driver using new UX

jayzhang:


packages/fx-core/log
```diff
@@ -0,0 +1,3 @@
+123
```
Siglud, line 1: What's this?

jayzhang: generated by UT, deleted

Siglud (approved):

LongOddCode (approved):

PR #7991, opened 2023-03-08
chore(template): merge registerApp and configureApp into provision

jayzhang:


kuojianlu:
Why not update the templates for scaffold as well?

jayzhang:
> Why not update the templates for scaffold as well?

scaffold templates are owned by others

xzf0587 (approved):

PR #8011, opened 2023-03-10
chore: npm & dotnet & npx driver using new UX

Siglud:


packages/fx-core/src/component/code/utils.ts
```diff
@@ -27,6 +30,140 @@ function capitalizeFirstLetter(raw: string) {
...
+    if (redirectTo) {
+      appendFile = path.isAbsolute(redirectTo) ? redirectTo : path.join(projectPath, redirectTo);
+    }
+    const outputs = parseKeyValueInOutput(command);
+    if (outputs) {
+      resolve(ok(["", outputs]));
+      return;
+    }
+    if (shell === "cmd") {
+      run = `%ComSpec% /D /E:ON /V:OFF /S /C "CALL ${command}"`;
```
hund030, line 69: What's the flags for?

hund030 (approved):

packages/fx-core/src/component/code/utils.ts
Siglud: %ComSpec% means the location of cmd.
The other part of the arguments can see from [here](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/cmd)

yukun-dong (approved):

eriolchan (approved):

PR #8013, opened 2023-03-10
chore: no error message for cancel downloading

hund030:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/17420379/

### DIFF
--- a/packages/fx-core/src/component/generator/generatorAction.ts
+++ b/packages/fx-core/src/component/generator/generatorAction.ts
@@ -9,7 +9,6 @@ import { getTemplatesFolder } from "../../folder";
 import { MissKeyError } from "./error";
 import { FeatureFlagName } from "../../common/constants";
 import { LogProvider } from "@microsoft/teamsfx-api";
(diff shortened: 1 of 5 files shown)


packages/fx-core/src/component/generator/generatorAction.ts
```diff
@@ -21,6 +20,7 @@ export interface GeneratorContext {
   zipUrl?: string;
   zip?: AdmZip;
   fallbackZipPath?: string;
+  cancelDownloading?: boolean;
```
yukun-dong, line 23: Is there any problem if don't add this `cancelDownloading` property to context?

hund030: I think context is used for the communication between actions. We need to inform the action `fetchZipFromUrlAction` that the downloading process is canceled. Otherwise, it finds the zipUrl is not fulfilled and throw an unexpected error.

yukun-dong (approved):

PR #8018, opened 2023-03-13
chore: spfx install local pacakges and scaffold using local pacakges

yuqizhou77:
- install latest yeoman and sp generator package if user selects to let TTK install packages for them.
- TODO: error handling

### DIFF
--- a/packages/fx-core/src/component/resource/spfx/depsChecker/generatorChecker.ts
+++ b/packages/fx-core/src/component/resource/spfx/depsChecker/generatorChecker.ts
@@ -77,17 +77,13 @@ export class GeneratorChecker implements DependencyChecker {
     ctx: PluginContext | ContextV3
   ): Promise<Result<boolean, FxError>> {
     telemetryHelper.sendSuccessEvent(ctx, TelemetryEvents.EnsureLatestSharepointGeneratorStart);
-    let needInstall = false;
+
     try {
-      needInstall = !(await this.isLatestInstalled());
-      if (needInstall) {
-        this._logger.info(`${displayName} not found, installing...`);
-        await this.install();
-        this._logger.info(`Successfully installed ${displayName}`);
-      }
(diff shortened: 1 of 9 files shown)


packages/fx-core/src/component/resource/spfx/depsChecker/generatorChecker.ts
```diff
@@ -70,6 +73,40 @@ export class GeneratorChecker implements DependencyChecker {
     return ok(true);
   }
 
+  public async ensureLatestDependency(
+    ctx: PluginContext | ContextV3
+  ): Promise<Result<boolean, FxError>> {
+    telemetryHelper.sendSuccessEvent(ctx, TelemetryEvents.EnsureLatestSharepointGeneratorStart);
+    let needInstall = false;
+    try {
+      needInstall = !(await this.isLatestInstalled());
```
HuihuiWu-Microsoft, line 82: In our new design, why not just execute 'npm install <package>@latest' all the time since we don't care about what latest version is?

packages/fx-core/src/component/resource/spfx/depsChecker/yoChecker.ts
```diff
@@ -64,6 +66,49 @@ export class YoChecker implements DependencyChecker {
     return ok(true);
   }
 
+  public async ensureLatestDependency(
```
HuihuiWu-Microsoft, line 69: Same here.

packages/fx-core/src/component/resource/spfx/depsChecker/generatorChecker.ts
yuqizhou77: I think it will help save some time :)

packages/fx-core/src/component/resource/spfx/depsChecker/yoChecker.ts
yuqizhou77: same above.

HuihuiWu-Microsoft (approved):

PR #8031, opened 2023-03-13
chore: remove spfx configuration

yuqizhou77:


HuihuiWu-Microsoft:
You may also need to remove logics where the configuration is consumed.

PR #8032, opened 2023-03-14
test: update cli UT to suit for v5 feature flag turn on

wenytang-ms:


Alive-Fish (approved):
Need to clean up the test case later.

PR #8100, opened 2023-03-20
chore: align Teams manifest path question for add web part

HuihuiWu-Microsoft:
image


packages/fx-core/src/core/question.ts
```diff
@@ -87,9 +89,10 @@ export enum CoreQuestionNames {
   ReplaceWebsiteUrl = "replaceWebsiteUrl",
   AppPackagePath = "appPackagePath",
   ReplaceBotIds = "replaceBotIds",
-  TeamsAppManifestFilePath = "teamsAppManifestFilePath",
+  TeamsAppManifestFilePath = "manifest-path",
```
HuihuiWu-Microsoft, line 92: CLI parameter should follow the same pattern.

sffamily (dismissed):

packages/fx-core/src/core/question.ts
nliu-ms: The existed cli command update Teams/aad app uses "manifest-file-path" as param name
image

sffamily (approved):

jayzhang (approved):

Alive-Fish (approved):

PR #8103, opened 2023-03-20
chore: add writeToEnvironmentFile support to ConfigManager

1yefuwang1:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/17603245


packages/fx-core/src/component/configManager/lifecycle.ts
```diff
@@ -240,7 +244,13 @@ export class Lifecycle implements ILifecycle {
       let result: Result<Map<string, string>, FxError>;
       let summary: string[];
       if (driver.instance.execute) {
-        const r = await driver.instance.execute(driver.with, ctx);
+        const r = await driver.instance.execute(
+          driver.with,
+          ctx,
+          driver.writeToEnvironmentFile
+            ? new Map(Object.entries(driver.writeToEnvironmentFile))
```
blackchoey, line 251: I see driverInterface defines `writeToEnvironmentFile` as `Record<string,string>`. I don't have special preference on the type, the major reason to use `Map<string,string>` in action interface is `ExecutionOutput` uses Map. If you think `Record<string,string>` is better, maybe you can update the action interface so you don't need to do the type conversion here.

blackchoey (approved):

jayzhang (approved):

packages/fx-core/src/component/configManager/lifecycle.ts
1yefuwang1: Map is easier to work with. Let's use map for now.

PR #8110, opened 2023-03-21
chore(debug-upgrade): upgrade m365 projects

kuojianlu:
- remove install-app-in-teams task
- add m365Title/acquire action
- add m365-tab and m365-me test case


swatDong (approved):
The code logic looks good.
Any discussion history or work item? Curious on how necessary to migrate existing m365 projects to use new sideloading flow.

kuojianlu:
> The code logic looks good. Any discussion history or work item? Curious on how necessary to migrate existing m365 projects to use new sideloading flow.

The main consideration is that if we deprecate (I mean breaking change) `${teamsAppInternalId}` in `launch.json` then we have to migrate existing m365 projects. IMO, it is a good chance to simplify the debug provider, so I prefer a breaking change rather than compatibility. Wanna to know your thoughts.

swatDong:
> > The code logic looks good. Any discussion history or work item? Curious on how necessary to migrate existing m365 projects to use new sideloading flow.
> 
> The main consideration is that if we deprecate (I mean breaking change) `${teamsAppInternalId}` in `launch.json` then we have to migrate existing m365 projects. IMO, it is a good chance to simplify the debug provider, so I prefer a breaking change rather than compatibility. Wanna to know your thoughts.

Make sense to me. Better to confirm with @MuyangAmigo since it's breaking change.

kuojianlu:
> > > The code logic looks good. Any discussion history or work item? Curious on how necessary to migrate existing m365 projects to use new sideloading flow.
> > 
> > 
> > The main consideration is that if we deprecate (I mean breaking change) `${teamsAppInternalId}` in `launch.json` then we have to migrate existing m365 projects. IMO, it is a good chance to simplify the debug provider, so I prefer a breaking change rather than compatibility. Wanna to know your thoughts.
> 
> Make sense to me. Better to confirm with @MuyangAmigo since it's breaking change.

Sure.

PR #8114, opened 2023-03-21
chore: add retrieveAppId in package service

kuojianlu:


packages/fx-core/src/common/m365/packageService.ts
```diff
@@ -164,6 +164,58 @@ export class PackageService {
     }
   }
 
+  public async retrieveAppId(token: string, manifestId: string): Promise<string> {
```
swatDong, line 167: Any scenario to retrieve both title id and app id at the same time?
If so, how about merging `retrieveTitleId` and `retrieveAppId` into one method since they are calling the same API...

kuojianlu: Seems no scenario now. However, agree to merge them. How about `getLaunchInfoByManifestId`, corresponding with `getLaunchInfo(ByTitleId)`?

swatDong: Sounds good

swatDong (approved):

PR #8127, opened 2023-03-22
fix: fix userdata resolve logic in v5 migration

frankqianms:
E2E TEST: N/A
To avoid end of line charactor resolving error, we change the logic here. The bad resolving logic may lead to bot-password and bot-id generating error in some projects when provision: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/17496886


blackchoey:
please add some test cases for related bug to avoid regression

frankqianms:
> please add some test cases for related bug to avoid regression



> please add some test cases for related bug to avoid regression

done.

packages/fx-core/tests/core/middleware/projectMigrationV3.test.ts
```diff
@@ -1082,6 +1084,24 @@ describe("userdataMigration", () => {
...
+    sandbox
+      .stub(fs, "readFileSync")
+      .returns(
+        "fx-resource-aad-app-for-teams.clientSecret=abcd\nfx-resource-bot.botPassword=1234\n"
+      );
+
+    const migrationContext = await mockMigrationContext(projectPath);
+    await copyTestProject(Constants.happyPathTestProject, projectPath);
+    await userdataMigration(migrationContext);
+
```
blackchoey, line 1098: can we call sandbox.restore() here? The current code works fine, my only concern is the mock may impact `assertFileContent()` function if someone changed its implementation to use `readFileSync()`. Restore the mocked function in advance can avoid such kind of issues.

frankqianms: reasonable concern. modified in the latest commit

blackchoey (approved):

xzf0587 (approved):

blackchoey (approved):

PR #8129, opened 2023-03-22
chore: add metadata events at the start of debug

a1exwang:
Send manifest/yml metadata events at the first of local debug in case of preqrequisites check failure.

To determine project type/capabilities for V5 local debug:
  - For happy path: use metadata events in the actual action
  - For users blocked by prerequisites check: use this metadata events to determine project type 

tasks.json metadata event will be added in another PR.

https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/17518368/

### DIFF
--- a/packages/vscode-extension/src/debug/localTelemetryReporter.ts
+++ b/packages/vscode-extension/src/debug/localTelemetryReporter.ts
@@ -3,7 +3,7 @@
 
 import { performance } from "perf_hooks";
 
-import { FxError } from "@microsoft/teamsfx-api";
+import { FxError, TeamsAppManifest } from "@microsoft/teamsfx-api";
 import {
   LocalEnvManager,
   LocalTelemetryReporter,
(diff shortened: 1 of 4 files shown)


packages/vscode-extension/src/debug/localTelemetryReporter.ts
```diff
@@ -67,6 +91,145 @@ export async function sendDebugAllStartEvent(additionalProperties: {
...
+          manifest,
+        };
+      }
+    } catch (e) {
+      // fall through next case
+    }
+  }
+
+  // Case 3: Assume user zips & uploads app package manually
+  // Try default location of manifest: appPackage/manifest.json
```
swatDong, line 185: Do you mean the template path or the build output path? E.g., `build/appPackage/manifest.xxx.json`?

a1exwang: Template path. Because in this case, probably appPackage.zip isn't generated by the toolkit (user doesn't use "teamsApp/zipAppPackage"/"teamsApp/update").

Although template file may contain placeholder, it doesn't matter because we only use it to determine project type.

I will add a property later to indicate which is the method we get the manifest so they won't mix with normal cases and contaminate the data.
image

swatDong (approved):

PR #8143, opened 2023-03-23
chore: split env files

Siglud:


templates/scenarios/csharp/message-extension/env/.env.dev.tpl
```diff
@@ -9,7 +9,4 @@ RESOURCE_SUFFIX=
 # Generated during provision, you can also add your own variables.
```
blackchoey, line 9: Please reference https://github.com/OfficeDev/TeamsFx-Samples/pull/731/files to add some comments to the beginning of all the .env files (note content of every .env file is slightly different).
By the way, the `.gitignore` file also needs update to only ignore `.env.xxx.user` and `.env.local`. Above sample PR also involves the change.

templates/scenarios/csharp/non-sso-tab/env/.env.dev.user.tpl
```diff
@@ -0,0 +1 @@
+# Secrets. Keys prefixed with `SECRET_` will be masked in Teams Toolkit logs.
```
blackchoey, line 1: If the project does not save any content to .env.xxx.user, my personal opinion is remove the `.env.xxx.user` file from template.
@jidddddd what's your opinion?

blackchoey: But we still need to ignore the .user file incase user adds it manually

templates/scenarios/csharp/message-extension/env/.env.dev.tpl
Siglud: Fixed

blackchoey: Please also add comments to `.env.dev` and `.env.local` (note they're different)

templates/scenarios/csharp/non-sso-tab/env/.env.dev.user.tpl
jidddddd: @blackchoey yes we can remove the file from template

blackchoey (approved):

PR #8145, opened 2023-03-23
chore: change template placeholder delimiters since yaml format does not allow %

hund030:
1. Replace all {% and %} with {{ and }} in templates.
2. Escape undefined variable to avoid unexpected replacement when generating project from templates.
    For example:  `{{appName}}-${{TEAMSFX_ENV}}` is expected to become `realAppName-${{TEAMSFX_ENV}}`
3. Add enough unit tests to ensure the functionality correct.


swatDong (dismissed):

packages/fx-core/src/component/generator/constant.ts
```diff
@@ -15,4 +15,5 @@ export const templatePrereleasePrefix = "templates-";
 export const errorSource = "GEN";
 export const componentName = "generator";
 export const commonTemplateName = "common";
-export const placeholderDelimiters: [string, string] = ["{%", "%}"];
+export const placeholderDelimiters: [string, string] = ["{{", "}}"];
+export const oldPlaceholderDelimiters: [string, string] = ["{%", "%}"];
```
eriolchan, line 19: Is this for backward compatible? From which TTK version do we introduce %placeholder%?

Siglud (approved):

yukun-dong (approved):

packages/fx-core/src/component/generator/utils.ts
```diff
@@ -170,12 +172,39 @@ export function renderTemplateFileData(
...
+    const writer = new Writer();
+    const result = writer.renderTokens(token, new Context(variables));
+    // Be compatible with current stable templates, can be removed after new template released.
+    return Mustache.render(result, variables, {}, oldPlaceholderDelimiters);
   }
   // Return Buffer instead of string if the file is not a template. Because `toString()` may break binary resources, like png files.
   return fileData;
 }
 
+export function escapeEmptyVariable(
```
eriolchan, line 185: I want to know more background, will check with you offline tomorrow.

kuojianlu (approved):

PR #8148, opened 2023-03-23
chore: update aad app with manifest path and env selection

wenytang-ms:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_sprints/taskboard/AuthAndData/Microsoft%20Teams%20Extensibility/CY23-3.2?workitem=17630494


KennethBWSong (dismissed):

packages/fx-core/src/component/driver/aad/update.ts
```diff
@@ -161,3 +171,28 @@ export class UpdateAadAppDriver implements StepDriver {
     };
   }
 }
+
+export async function getQuestionForDeployAadManifest(
```
blackchoey, line 175: Personally, I would prefer to put these 2 functions to another place, because the driver itself does not care about the questions.

wenytang-ms: mvoe to questions.ts

blackchoey (dismissed):

xzf0587 (approved):

KennethBWSong (approved):

jayzhang (approved):

tecton (approved):

PR #8151, opened 2023-03-23
chore: update remote preview in treeview

kuojianlu:
- UX change: select a manifest file
- get teams app id from manifest
- get m365 app id by teams app id
- add M365TitleNotAcquiredError

remote-preview


packages/vscode-extension/src/handlers.ts
```diff
@@ -612,27 +627,90 @@ export async function treeViewLocalDebugHandler(args?: any[]): Promise<Result<nu
...
+      type: "singleSelect",
+      name: "hub",
+      title: localize("teamstoolkit.preview.platform.title"),
+      staticOptions: [constants.Hub.teams, constants.Hub.outlook, constants.Hub.office],
+      placeholder: localize("teamstoolkit.preview.platform.placeholder"),
+    };
+    const node = new QTreeNode(platformSingleSelect);
+    const inputs = getSystemInputs();
+    node.addChild(selectTeamsAppManifestQuestion(inputs));
+    const res = await traverse(node, inputs, VS_CODE_UI, ExtTelemetry.reporter);
```
swatDong, line 643: Does this have concurrent lock?
E.g., does remote preview work when there's a running provision or deploy progress?

packages/fx-core/resource/package.nls.json
```diff
@@ -467,6 +467,7 @@
   "error.debugHandler.InvalidExistingBotArgsError": "botId and botPassword need to be set together if you use existing bot.",
   "error.debugHandler.DebugArgumentEmptyError": "Debug argument '%s' is empty.",
   "error.debugHandler.InvalidTabBaseUrlError": "baseUrl must be a valid https url.",
+  "error.m365.M365TitleNotAcquiredError": "M365 title is not acquired yet, please use m365Title/acquire action to acquire M365 title.",
```
sffamily, line 470: M365 -> Microsoft 365

packages/vscode-extension/src/handlers.ts
kuojianlu: Good question. It does not have concurrent lock since all things are done in extension part (i.e., does not use `ConcurrentLockerMW`).
Not sure if there is a need to add concurrent lock because preview only reads env and does not write anything, like preview from F5.

swatDong: I would prefer not to add concurrent lock since it's a read-only operation.

swatDong (approved):

packages/fx-core/resource/package.nls.json
kuojianlu: Thanks, updated.

swatDong (dismissed):

packages/vscode-extension/src/utils/commonUtils.ts
```diff
@@ -373,6 +373,10 @@ export async function getResourceGroupNameFromEnv(env: string): Promise<string |
 }
 
 export async function getProvisionSucceedFromEnv(env: string): Promise<boolean | undefined> {
+  if (isV3Enabled()) {
```
tecton, line 376: I know it is hard to detect whether provisioned actually in V3. This will make all environment nodes except local always show `(Provisioned)` description like in GIF. Is this UI expected?

kuojianlu: Sorry, it is only my local test, committed this change by mistake, let me revert it.

swatDong (approved):

packages/vscode-extension/src/utils/commonUtils.ts
tecton: No problem, I'll change this part later.

sffamily (approved):

PR #8154, opened 2023-03-24
chore: support outputEnvVarNames in m365Title action

kuojianlu:


swatDong (approved):

blackchoey (approved):
LGTM. By the way, how will we adjust Teams Toolkit's behavior when user specifies another name for m365 app id?

kuojianlu:
> LGTM. By the way, how will we adjust Teams Toolkit's behavior when user specifies another name for m365 app id?

User needs to update `launch.json` as well. As there is no hard-code logic now, so there is no special need to handle M365 app id.

PR #8156, opened 2023-03-24
chore: support outputEnvVarNames in prerequisite action

kuojianlu:
sslCertFile, sslKeyFile, funcPath, dotnetPath are all optional

### DIFF
--- a/packages/fx-core/src/component/driver/prerequisite/installDriver.ts
+++ b/packages/fx-core/src/component/driver/prerequisite/installDriver.ts
@@ -30,15 +30,18 @@ import { DotnetInstallationUserError } from "./error/dotnetInstallationUserError
 import { FuncInstallationUserError } from "./error/funcInstallationUserError";
 import { InstallToolArgs } from "./interfaces/InstallToolArgs";
 import { InvalidActionInputError } from "../../../error/common";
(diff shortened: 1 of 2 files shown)


packages/fx-core/src/component/driver/prerequisite/installDriver.ts
```diff
@@ -30,16 +30,19 @@ import { DotnetInstallationUserError } from "./error/dotnetInstallationUserError
...
-const outputName = Object.freeze({
-  SSL_CRT_FILE: "SSL_CRT_FILE",
-  SSL_KEY_FILE: "SSL_KEY_FILE",
-  FUNC_PATH: "FUNC_PATH",
-  DOTNET_PATH: "DOTNET_PATH",
-});
 const helpLink = "https://aka.ms/teamsfx-actions/prerequisite-install";
 
+const outputKeys = {
+  sslCertFile: "sslCertFile",
```
swatDong, line 40: Although they are all optional, will we check mis-match? E.g., do not install cert but output has sslCertFile.

kuojianlu: I prefer not. We can let user know in yaml schema that `sslCertFile` only takes effect when `devCert` is specified.

swatDong (approved):

PR #8178, opened 2023-03-28
perf(aad): add input signInAudience in aadApp/create action

KennethBWSong:
can replace botAadApp/create with aadApp/create like following:
image


KennethBWSong:
E2E Test: None

packages/fx-core/src/component/driver/aad/create.ts
```diff
@@ -90,7 +90,8 @@ export class CreateAadAppDriver implements StepDriver {
           )
         );
         // Create new AAD app if no client id exists
-        const aadApp = await aadAppClient.createAadApp(args.name);
+        const signInAudience = args.signInAudience ?? "AzureADMyOrg";
```
adashen, line 93: how about have a variable for audience types to wrap the different types "AzureADMyOrg".....

adashen: And SignInAudience should be enum right?

KennethBWSong: Sure I will update here. One potential problem is that we need to update our code if AAD updates its schema.

KennethBWSong: Updated, thx!

packages/fx-core/src/component/driver/aad/create.ts
```diff
@@ -90,7 +92,17 @@ export class CreateAadAppDriver implements StepDriver {
           )
         );
         // Create new AAD app if no client id exists
-        const aadApp = await aadAppClient.createAadApp(args.name);
+        // Throw error if unexpected signInAudience
```
xzf0587, line 95: there is a validateArgs method in 79. It is better to move this check into it

packages/fx-core/src/component/driver/aad/create.ts
```diff
@@ -90,7 +92,17 @@ export class CreateAadAppDriver implements StepDriver {
...
+        if (
+          args.signInAudience &&
+          !(<any>Object).values(SignInAudience).includes(args.signInAudience)
+        ) {
+          throw new InvalidParameterUserError(actionName, "signInAudience", helpLink);
+        }
```
xzf0587, line 101: there is a validateArgs method in line 79. It is better to move this check into it

packages/fx-core/src/component/driver/aad/create.ts
```diff
@@ -90,7 +92,17 @@ export class CreateAadAppDriver implements StepDriver {
           )
         );
         // Create new AAD app if no client id exists
-        const aadApp = await aadAppClient.createAadApp(args.name);
+        // Throw error if unexpected signInAudience
+        if (
+          args.signInAudience &&
+          !(<any>Object).values(SignInAudience).includes(args.signInAudience)
```
xzf0587, line 98: can we use Object.values(...) here? It will be readable.

packages/fx-core/src/component/driver/aad/create.ts
KennethBWSong: Updated, thx!

packages/fx-core/src/component/driver/aad/create.ts
KennethBWSong: Resolved according to offline sync.

xzf0587 (approved):

packages/fx-core/src/component/driver/aad/interface/createAadAppArgs.ts
```diff
@@ -4,4 +4,5 @@
 export interface CreateAadAppArgs {
   name: string; // The name of AAD app
   generateClientSecret: boolean; // Whether generate client secret for the app
+  signInAudience?: string; // Specifies what Microsoft accounts are supported for the current application.
```
blackchoey, line 7: is it possible to mark signInAudience to be `SignInAudience` type?

KennethBWSong: Updated, thx!

blackchoey (approved):

xzf0587 (approved):

PR #8186, opened 2023-03-28
chore: log validation results to file

nliu-ms:
VSCode extension will write full validation results to log file.

The directory of extension logs can be acquired from extension context:
https://code.visualstudio.com/api/references/vscode-api#Uri

image

Sample logs:
image

CLI and VS doesn't support log file right now, CLI will write all log files to console.


jayzhang (dismissed):

jayzhang (approved):

packages/cli/src/commonlib/log.ts
```diff
@@ -36,6 +36,10 @@ export class CLILogProvider implements LogProvider {
     return CLILogProvider.instance;
   }
 
+  getLogFilePath(): string {
+    throw new NotImplementedError("CLILogProvider", "getLogFilePath");
```
swatDong, line 40: Does this mean CLI has different validation behavior from VSC?

Alive-Fish: Same worry about the server code (VS).

nliu-ms: CLI will output all the validation logs directly, won't output to log file. As the log folder path it's hard to generate, needs to consider different OSs, but it may be supported in future versions.
VS is the same, I will investigate if it's possible to get default extension log folder.

packages/vscode-extension/src/commonlib/log.ts
```diff
@@ -70,12 +86,17 @@ export class VsCodeLogProvider implements LogProvider {
...
    */
-  async log(logLevel: LogLevel, message: string): Promise<boolean> {
+  async log(logLevel: LogLevel, message: string, logToFile?: boolean): Promise<boolean> {
     try {
       if (logLevel < LogLevel.Info) return true;
       if (logLevel >= LogLevel.Warning) this.outputChannel.show();
       const dateString = new Date().toJSON();
-      this.outputChannel.appendLine(`[${dateString}] [${LogLevel[logLevel]}] - ${message}`);
+      const formattedMessage = `[${dateString}] [${LogLevel[logLevel]}] - ${message}`;
+      if (logToFile) {
```
HuihuiWu-Microsoft, line 95: Does this mean the validation result will be logged into file and doesn't display in output channel anymore?

HuihuiWu-Microsoft (dismissed):

packages/cli/src/commonlib/log.ts
swatDong: OK to me if that's a known point or as designed.

swatDong (approved):

packages/cli/src/commonlib/log.ts
nliu-ms: As discussed with @Alive-Fish , will provide the log folder under %TEMP% for CLI/server to avoid potential error thrown.

swatDong: Is that a hard-coded file name or with timestamp prefix/suffix?

nliu-ms: E.g. log file:
"C:\Users\xx\AppData\Local\Temp\.fx\server-log\20230330T030429.log"

Alive-Fish (approved):

swatDong (approved):

HuihuiWu-Microsoft (approved):

jayzhang (approved):

SLdragon (approved):

PR #8201, opened 2023-03-29
chore: add previewWithManifest in core to preview app

kuojianlu:
This method can then be used for treeview preivew, cli preview, codespaces preview, or possible mobile preview.

Why this change:
1. `ManifestUtils.getManifestV3` needs envs to be loaded to process.env. I would like to use `EnvLoaderMW` to handle envs injection and ejection.
2. Can be reused.
3. To be consistent with other functionalities like validating/building manifest.


swatDong (dismissed):
BTW, will there be scenario that skipping `selectM365HostQuestion` but only `selectTeamsAppManifestQuestion`?

swatDong (approved):

sffamily (approved):

jayzhang (approved):

PR #8232, opened 2023-03-31
feat: simplified basic tab template

hund030:
The basic tab template is a simple vanilla NodeJS web application that does not require any web frameworks or single sign-on. This template is designed to be compatible with Outlook and Office platform, but for the v5 TTK GA, we have made it exclusively for Teams to ensure user experience consistency.

I still need to update the yml to make it compatible with latest fx-core.
I will also add a TypeScript version after this done.

E2E TEST: https://github.com/OfficeDev/TeamsFx/actions/runs/4634850841

FEATURE ID: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/15664209/


huimiu (approved):

templates/scenarios/js/non-sso-tab/.webappignore
```diff
@@ -0,0 +1,26 @@
...
+.localSettings
+.vscode
+*.js.map
+*.ts.map
+*.ts
+.git*
+.tsbuildinfo
+CHANGELOG.md
+readme.md
+local.settings.json
```
eriolchan, line 13: I guess local.settings.json is for Az Func, can you help check if it is still valid for this tab scenario?

templates/scenarios/js/non-sso-tab/.webappignore
```diff
@@ -0,0 +1,26 @@
...
+test
+tsconfig.json
+.DS_Store
+teamsapp.yml
+teamsapp.*.yml
+/env/
+/script/
+/node_modules/.bin
+/node_modules/ts-node
+/node_modules/typescript
```
eriolchan, line 23: Since it is js template, do we need to include some ts file in this ignore file?

eriolchan (approved):

templates/scenarios/js/non-sso-tab/.webappignore
hund030: I copy this file from JavaScript default bot template. I think we don't need to maintain many versions of .webappignore file. So let's keep the content.

PR #8239, opened 2023-04-03
fix(vsc): regex to match link

tecton:
Related bug: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/17567115

E2E TEST: N/A

### DIFF
--- a/packages/vscode-extension/syntaxes/teamsfx-toolkit-output.tmLanguage
+++ b/packages/vscode-extension/syntaxes/teamsfx-toolkit-output.tmLanguage
@@ -106,7 +106,7 @@
                     </dict>
                 </dict>
                 <key>match</key>
-                <string>(https?:\/\/[a-zA-Z0-9_\-\/\.]*)</string>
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/syntaxes/teamsfx-toolkit-output.tmLanguage
```diff
@@ -106,7 +106,7 @@
                     </dict>
                 </dict>
                 <key>match</key>
-                <string>(https?:\/\/[a-zA-Z0-9_\-\/\.]*)</string>
+                <string>(https?:\/\/[^\s]*)</string>
```
kuojianlu, line 109: Consider using a stricter pattern to avoid possible corner cases.

tecton: Thanks, updated.

HuihuiWu-Microsoft (approved):

PR #8245, opened 2023-04-03
chore: sync up fallback template after new template released

hund030:
VSC extension copies local templates when webpack. So, we should make sure the local templates are up to date.
TEST: https://github.com/hund030/TeamsFx/actions/runs/4625258194/jobs/8180845521

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -127,6 +127,11 @@ jobs:
       - name: generate templates v3
         run: |
           .github/scripts/template-zip-autogen-v3.sh ${{ runner.temp }}/teamsfx_templates_v3
+          cd ./packages/fx-core
+          rm -rf ./templates/fallback
+          node ./scripts/download-templates-zip.js
+        env:
+          TEMPLATE_PATH: ${{ runner.temp }}/teamsfx_templates_v3
 
       - name: update templates prerelease tag
         uses: richardsimko/update-tag@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/cd.yml
```diff
@@ -191,6 +191,14 @@ jobs:
           tag: "template-tag-list"
           allowUpdates: true
 
+      - name: sync up fx-core fallback templates
+        run: |
+          rm -r ./templates/fallback
```
wenytang-ms, line 196: if no fallback exists before?

wenytang-ms: rm -rf ..

.github/workflows/cd.yml
```diff
@@ -191,6 +191,14 @@ jobs:
...
+          node ./scripts/download-templates-zip.js $(git tag -l templates*)
```
wenytang-ms, line 197: current vm branch may have such tag like templates@2.0.1-alpha.3498543894.0 
will this can work??

.github/workflows/cd.yml
hund030: Updated.

.github/workflows/cd.yml
hund030: Removed.

wenytang-ms (approved):

.github/workflows/cd.yml
```diff
@@ -127,6 +127,11 @@ jobs:
       - name: generate templates v3
         run: |
           .github/scripts/template-zip-autogen-v3.sh ${{ runner.temp }}/teamsfx_templates_v3
```
wenytang-ms, line 129: the input ${{ runner.temp }}/teamsfx_templates_v3 is duplicate with env.TEMPLATE_PATH, can merge together?

hund030: Good suggestion. Updated.

wenytang-ms (approved):

PR #8249, opened 2023-04-03
chore(debug-upgrade): improve error message

kuojianlu:
To fix https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/17603544.

### DIFF
--- a/packages/vscode-extension/src/handlers.ts
+++ b/packages/vscode-extension/src/handlers.ts
@@ -1462,14 +1462,12 @@ export async function validateSpfxDependenciesHandler(): Promise<string | undefi
  */
 export async function validateLocalPrerequisitesHandler(): Promise<string | undefined> {
   if (isV3Enabled()) {
-    if (isV3Enabled()) {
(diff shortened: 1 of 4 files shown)


packages/vscode-extension/src/handlers.ts
```diff
@@ -1456,7 +1462,15 @@ export async function validateSpfxDependenciesHandler(): Promise<string | undefi
  */
 export async function validateLocalPrerequisitesHandler(): Promise<string | undefined> {
   if (isV3Enabled()) {
-    return await commonUtils.triggerV3Migration();
+    if (isV3Enabled()) {
```
swatDong, line 1465: Duplicated `isV3Enabled`. Other changes as well.

kuojianlu: Nice catch. Updated.

swatDong (approved):

HuihuiWu-Microsoft (approved):

PR #8262, opened 2023-04-03
chore: update templates with yaml schema URI

1yefuwang1:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/17768149

### DIFF
--- a/templates/scenarios/csharp/command-and-response/teamsapp.yml.tpl
+++ b/templates/scenarios/csharp/command-and-response/teamsapp.yml.tpl
@@ -1,3 +1,4 @@
+# yaml-language-server: $schema=https://developer.microsoft.com/json-schemas/teams-toolkit/teamsapp-yaml/1.0.0/yaml.schema.json
 # Visit https://aka.ms/teamsfx-v5.0-guide for details on this file
 # Visit https://aka.ms/teamsfx-actions for details on actions
 version: 1.0.0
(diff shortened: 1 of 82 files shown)


templates/scenarios/csharp/command-and-response/teamsapp.yml.tpl
```diff
@@ -1,3 +1,4 @@
+# yaml-language-server: $schema=https://developer.microsoft.com/json-schemas/teams-toolkit/teamsapp-yaml/1.0.0/yaml.schema.json
```
swatDong, line 1: Does toolkit uses *resource/yaml.schema.json* to check format?
If so, what if *resource/yaml.schema.json* has difference from this remote one?
User may see error hint but actually it passes the TTK check, or user see no error but it fails the TTK check.

1yefuwang1: Currently, the bundled schema is used for schema validation.  This one is only for intellisense.
We have to carefully keep them in sync, or it's a logic error.
If there's a schema change, a new schema  and a new TTK will be published.

swatDong: So for each stable release we can keep them consistency. But will it be a problem if an older TTK opens a newer project, or a newer TTK opens an older project?

swatDong (approved):

PR #8280, opened 2023-04-06
chore: download directory for sample

yukun-dong:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/17713763/


packages/fx-core/src/component/generator/utils.ts
```diff
@@ -230,8 +234,69 @@ export function getSampleRelativePath(sampleName: string): string {
...
+  const filePrefixUrl = `https://raw.githubusercontent.com/${owner}/${repository}/${fileInfo.sha}/`;
+  let fileInfoTree = fileInfo.tree as any[];
+  fileInfoTree = fileInfoTree.filter(
+    (node) => node.path.startsWith(sampleName) && node.type != "tree"
+  );
+
+  //step 2: download files with limited concurrency
+  const downloadCallbacks = [];
+  for (const node of fileInfoTree) {
+    const downloadCallbak = async () => {
```
hund030: typo: downloadCallback

packages/fx-core/src/component/generator/utils.ts
```diff
@@ -230,8 +234,69 @@ export function getSampleRelativePath(sampleName: string): string {
...
+      const file = await sendRequestWithRetry(async () => {
+        return await axios.get(filePrefixUrl + node.path, { responseType: "arraybuffer" });
+      }, retryLimits);
+      const filePath = path.join(dstPath, node.path.replace(sampleName, ""));
+      await fs.ensureFile(filePath);
+      await fs.writeFile(filePath, Buffer.from(file.data as any));
```
hund030, line 282: writeFile create new file if it doesn't exist, I think you do not need to call ensureFile.

packages/fx-core/src/component/generator/utils.ts
```diff
@@ -230,8 +234,69 @@ export function getSampleRelativePath(sampleName: string): string {
...
+    };
+    downloadCallbacks.push(downloadCallback);
+  }
+  await limitConcurrency(downloadCallbacks, concurrencyLimits);
+}
+
+async function limitConcurrency(callbacks: any[], limit: number) {
```
hund030, line 289: The concurrent process seems excessive to me. I would suggest removing the `downloadCallbacks` array and simplifying the code as follows:
```
const callback = async (node) => { ... }
...
processing.push(callback(fileInfoTree[i]);
...
```

However, ultimately the decision is up to you.

packages/fx-core/tests/component/generator/generator.test.ts
```diff
@@ -247,6 +248,38 @@ describe("Generator utils", () => {
...
+  it("download directory get file info error", async () => {
+    const axiosStub = sandbox.stub(axios, "get");
+    axiosStub.onFirstCall().resolves({ status: 403 });
+    try {
+      await downloadDirectory(
+        "https://github.com/OfficeDev/TeamsFx-Samples/tree/dev/bot-sso",
+        tmpDir
+      );
+    } catch (e) {
+      assert.exists(e);
```
hund030, line 261: Consider asserting the error status and verifying that the error being stubbed is the one being tested for.

packages/fx-core/tests/component/generator/generator.test.ts
```diff
@@ -443,3 +476,30 @@ describe("Generator happy path", async () => {
...
+  afterEach(async () => {
+    sandbox.restore();
+    mockedEnvRestore();
+    if (await fs.pathExists(tmpDir)) {
+      await fs.rm(tmpDir, { recursive: true });
+    }
+  });
+
+  it("generate sample using download directory", async () => {
+    const result = await Generator.generateSample(ctx, path.join(tmpDir, "test1"), "bot-sso");
```
hund030, line 502: What if bot-sso sample is removed? Will this test case fails and break our CI on pull request?

packages/fx-core/src/common/tools.ts
```diff
@@ -416,6 +416,10 @@ export function isV3Enabled(): boolean {
   return process.env.TEAMSFX_V3 ? process.env.TEAMSFX_V3 === "true" : true;
 }
 
+export function isDownloadDirectoryEnabled(): boolean {
+  return process.env.DOWNLOAD_DIRECTORY === "true";
```
hund030, line 420: What's the purpose of the feature flag?

packages/fx-core/src/component/generator/utils.ts
```diff
@@ -230,8 +234,69 @@ export function getSampleRelativePath(sampleName: string): string {
...
+  const fileInfo = (
+    await sendRequestWithRetry(async () => {
+      return await axios.get(fileInfoUrl);
+    }, retryLimits)
+  ).data as any;
+
+  const filePrefixUrl = `https://raw.githubusercontent.com/${owner}/${repository}/${fileInfo.sha}/`;
+  let fileInfoTree = fileInfo.tree as any[];
+  fileInfoTree = fileInfoTree.filter(
+    (node) => node.path.startsWith(sampleName) && node.type != "tree"
```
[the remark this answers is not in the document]
yukun-dong: The folder's type will be "tree", so this step is to filter out folders.
I will update != to !==. Thanks for the info.

packages/fx-core/src/common/tools.ts
yukun-dong: Since the decision for improving downloading sample is not finalized. I'm considering using this feature-flag to turn it on or off.

packages/fx-core/src/component/generator/utils.ts
yukun-dong: Thanks! will update.

packages/fx-core/src/component/generator/utils.ts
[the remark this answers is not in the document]
hund030: If we only care `node.path`, what about:
```
samplePaths = fileInfoTree.filter((node) => ...).map((node) => node.path);
```

packages/fx-core/src/component/generator/utils.ts
yukun-dong: Double checked. The "writeFile" won't create new file, so I will keep "ensureFile" here.

packages/fx-core/tests/component/generator/generator.test.ts
yukun-dong: Thanks. Added.

packages/fx-core/tests/component/generator/generator.test.ts
yukun-dong: The axios calls are mocked now, so removing sample won't have influence.

packages/fx-core/src/component/generator/utils.ts
hund030: Really? I write below code and it works well even when `tmp.txt` does not exist.
```
```

packages/fx-core/src/component/generator/generator.ts
```diff
@@ -103,20 +108,28 @@ export class Generator {
     merge(actionContext?.telemetryProps, {
       [TelemetryProperty.SampleName]: sampleName,
     });
+    merge(actionContext?.telemetryProps, {
+      [TelemetryProperty.SampleDownloadDirectory]: isDownloadDirectoryEnabled() ? "true" : "false",
```
hund030, line 112:
```
merge(actionContext?.telemetryProps, {
    [TelemetryProperty.SampleName]: sampleName,
    [TelemetryProperty.SampleDownloadDirectory]: isDownloadDirectoryEnabled() ? "true" : "false",
});
```

packages/fx-core/src/component/generator/generatorAction.ts
```diff
@@ -77,6 +79,13 @@ export const fetchTemplateZipFromSourceCodeAction: GeneratorAction = {
   },
 };
 
+export const downloadDirectoryAction: GeneratorAction = {
+  name: GeneratorActionName.DownloadDirectory,
+  run: async (context: GeneratorContext) => {
+    await downloadDirectory(context.url!, context.destination);
```
hund030, line 85: Could you add null-check for context.url? We should avoid using non-null assertion as possible.

packages/fx-core/src/component/generator/utils.ts
```diff
@@ -230,8 +234,65 @@ export function getSampleRelativePath(sampleName: string): string {
...
+  const processing: any[] = [];
+  let i = 0;
+  while (i < limit && i < data.length) {
+    processing.push(callback(data[i]));
+    i++;
+  }
+  while (i < data.length) {
+    await Promise.race(processing);
+    processing.push(callback(data[i]));
+    i++;
```
hund030, line 295: Add unit test to cover this loop logic.

packages/fx-core/resource/package.nls.json
```diff
@@ -75,6 +75,7 @@
   "core.progress.generateTemplate.detail": "Downloading and unzipping Template",
   "core.progress.generateSample": "Generating Sample",
   "core.progress.generateSample.detail": "Downloading and unzipping Sample",
+  "core.progress.downloadDirectory.detail": "Downloading sample directory from GitHub",
```
eriolchan, line 78: Do we need to add new message for this since it is an internal engineering improvement?

yukun-dong: The original message is "Downloading and unzipping Sample". The new method doesn't involve unzip process so I think it's inappropriate to use original message for the new method.

packages/fx-core/src/component/generator/utils.ts
```diff
@@ -230,8 +234,65 @@ export function getSampleRelativePath(sampleName: string): string {
...
+    }, retryLimits)
+  ).data as any;
+
+  const filePrefixUrl = `https://raw.githubusercontent.com/${owner}/${repository}/${fileInfo.sha}/`;
+  const fileInfoTree = fileInfo.tree as any[];
+  const samplePaths = fileInfoTree
+    .filter((node) => node.path.startsWith(sampleName) && node.type !== "tree")
+    .map((node) => node.path);
+
+  //step 2: download files with limited concurrency
```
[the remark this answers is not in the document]
yukun-dong: IMO, it's okay to keep it in one function since these 2 steps are coupled. The step2 needs parsed results(L259) and samplePaths(269) from step1. If we want to split it into 2 functions then step2 functions will have more parameters.

packages/fx-core/src/component/generator/error.ts
```diff
@@ -25,6 +25,30 @@ export class UnzipError extends BaseComponentInnerError {
   }
 }
 
+export class DownloadSampleNetworkError extends BaseComponentInnerError {
+  constructor(url: string) {
+    super(
+      errorSource,
+      "SystemError",
```
eriolchan, line 32: I think the download error caused by network issue and api limit is "User error" because they can be recovered after user retry it.

packages/fx-core/src/component/generator/generator.ts
eriolchan: Since the value of isDownloadDirectoryEnabled() is boolean, can we use toString() to convert boolean to string?
```
isDownloadDirectoryEnabled() ? "true" : "false" => isDownloadDirectoryEnabled().toString()
```

packages/fx-core/resource/package.nls.json
yukun-dong: Updated. However, the final message shown will be "Generating sample: Downloading sample" which may be a little strange. I will discuss it with you offline.

packages/fx-core/src/component/generator/utils.ts
```diff
@@ -230,8 +234,69 @@ export function getSampleRelativePath(sampleName: string): string {
   return `${sampleConfig.baseFolderName}/${sampleName}/`;
 }
 
+export function getSampleUrl(sample: SampleInfo): string {
```
eriolchan, line 237: Can you help illustrate more about this method by example? I don't quite understand this logic, maybe offline check with you.

packages/fx-core/src/component/generator/utils.ts
```diff
@@ -230,8 +234,69 @@ export function getSampleRelativePath(sampleName: string): string {
...
+  const samplePaths = fileInfoTree
+    .filter((node) => node.path.startsWith(sampleName) && node.type !== "tree")
+    .map((node) => node.path);
+
+  //step 2: download files with limited concurrency
+  const downloadCallback = async (samplePath: string) => {
+    const file = await sendRequestWithRetry(async () => {
+      return await axios.get(filePrefixUrl + samplePath, { responseType: "arraybuffer" });
+    }, retryLimits);
+    const filePath = path.join(dstPath, samplePath.replace(sampleName, ""));
```
eriolchan, line 278: why do we replace sample name with ""? Is it safer to use some method in path library to get relative path.

packages/fx-core/src/component/generator/utils.ts
```diff
@@ -230,8 +234,69 @@ export function getSampleRelativePath(sampleName: string): string {
...
+    await fs.ensureFile(filePath);
+    await fs.writeFile(filePath, Buffer.from(file.data as any));
+  };
+  await limitConcurrency(samplePaths, downloadCallback, concurrencyLimits);
+}
+
+export async function limitConcurrency<T>(data: T[], callback: (arg0: T) => any, limit: number) {
```
eriolchan, line 285: Can we have a better name for this method, it is not easy to understand what this method do from name

packages/fx-core/src/component/generator/utils.ts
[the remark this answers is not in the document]
eriolchan: I have the some feeling that this method is too long. Let's check how to improve this code for better maintenance offline.

packages/fx-core/src/component/generator/utils.ts
eriolchan: It looks like this method is directly copied from Internet and we need to refactor it to make it easy to understand.

packages/fx-core/src/component/generator/utils.ts
```diff
@@ -230,8 +234,69 @@ export function getSampleRelativePath(sampleName: string): string {
...
+  const queue: any[] = [];
+  for (const ele of data) {
+    // fire the async function, add its promise to the queue, and remove
+    // it from queue when complete
+    const p = callback(ele).then((res: any) => {
```
eriolchan, line 290: what if some thread throws error due to network issue?

packages/fx-core/src/component/generator/utils.ts
```diff
@@ -230,8 +234,69 @@ export function getSampleRelativePath(sampleName: string): string {
...
 }
+
+export async function downloadDirectory(
+  sampleUrl: string,
+  dstPath: string,
+  concurrencyLimits = sampleConcurrencyLimits,
+  retryLimits = sampleDefaultRetryLimits
+): Promise<void> {
+  //step 1: get file info
+  const urlParserRegex = /https:\/\/github.com[/]([^/]+)[/]([^/]+)[/]tree[/]([^/]+)[/](.*)/;
```
Siglud, line 258: Can be simplify to `const urlParserRegex = /https:\/\/github.com\/([^/]+)\/([^/]+)\/tree\/([^/]+)\/(.*)/;`

packages/fx-core/src/component/generator/utils.ts
Siglud: Maybe you should try ``path.relative(samplePath, `/${sampleName}`)`` ?

packages/fx-core/src/component/generator/utils.ts
yukun-dong: The sample url is unified in PR https://github.com/OfficeDev/TeamsFx/pull/8376, so this function is no longer needed and I removed it.

packages/fx-core/src/component/generator/generatorAction.ts
```diff
@@ -19,6 +19,7 @@ export interface GeneratorContext {
   relativePath?: string;
   zipUrl?: string;
   zip?: AdmZip;
+  url?: string;
```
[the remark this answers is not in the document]
yukun-dong: I changed it to reuse "url".

packages/fx-core/src/component/generator/utils.ts
```diff
@@ -230,8 +234,69 @@ export function getSampleRelativePath(sampleName: string): string {
...
+  const [owner, repository, ref, sampleName] = urlParserRegex.exec(sampleUrl)!.slice(1);
```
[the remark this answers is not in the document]
yukun-dong: Add an error here.

packages/fx-core/src/component/generator/utils.ts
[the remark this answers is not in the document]
yukun-dong: This function has been splitted into 3 functions.

packages/fx-core/resource/package.nls.json
yukun-dong: After discussion, I changed this message to be "Generating sample: Downloading sample ${sampleId}".

packages/fx-core/src/component/generator/generatorAction.ts
```diff
@@ -112,13 +121,10 @@ export const fetchTemplateZipFromLocalAction: GeneratorAction = {
     if (context.zip) {
       return;
     }
-
-    if (!context.fallbackZipPath) {
-      context.fallbackZipPath = path.join(getTemplatesFolder(), "fallback");
-    }
-
+    context.fallback = true;
+    const fallbackPath = path.join(getTemplatesFolder(), "fallback");
```
hund030, line 125: What's the reason behind this modification? I am not convincd that it improves the code. By hard-coding the fallback path in this action, we are limiting our ability to customize it in the future.

packages/fx-core/src/component/generator/utils.ts
```diff
@@ -235,3 +239,97 @@ export function zipFolder(folderPath: string): AdmZip {
...
+    .map((node) => node.path);
+  const filePrefixUrl = `https://raw.githubusercontent.com/${urlInfo.owner}/${urlInfo.repository}/${fileInfo.sha}/`;
+  return { samplePaths, filePrefixUrl };
+}
+
+async function downloadSampleFiles(
+  filePrefixUrl: string,
+  samplePaths: string[],
+  dstPath: string,
+  dir: string,
```
hund030, line 295: The argument name is not clear enough. Please consider choosing a more descriptive name that clearly conveys the purpose of the argument.

hund030 (dismissed):
Overall, it looks good to me.

Siglud (dismissed):

packages/fx-core/src/component/generator/generatorAction.ts
yukun-dong: The reason is to try to make generatorContext concise. The original generatorContext.fallbackPath's value is never actually used after set and it's purpose is just to send telemetry, so we can change it to be a boolean.

hund030 (approved):

packages/fx-core/resource/package.nls.json
```diff
@@ -419,6 +419,9 @@
   "error.generator.FetchZipFromUrlError": "Unable to download zip package from %s.",
   "error.generator.UnzipError": "Unable to unzip templates and write to disk.",
   "error.generator.MissKeyError": "Unable to find key %s",
+  "error.generator.DownloadSampleApiLimitError": "Unable to download sample due to rate limit exceeded. Please wait for the rate limit reset and try again (This may take up to 1 hour). Alternatively, you can go to %s to git clone the repo manually",
```
sffamily, line 422: rate limit exceeded -> throttling

Please wait ... and try again (....) -> Retry later (....)

packages/fx-core/resource/package.nls.json
```diff
@@ -419,6 +419,9 @@
...
+  "error.generator.DownloadSampleNetworkError": "Unable to download sample due to network error. Please check your network connection and try again. Alternatively, you can go to %s to git clone the repo manually",
```
sffamily, line 423: Please check -> Check
try again -> retry

sffamily (approved):

hund030 (approved):

packages/fx-core/src/component/generator/generator.ts
```diff
@@ -171,8 +178,18 @@ export async function sampleDefaultOnActionError(
 ): Promise<void> {
   await context.logProvider.error(error.message);
   switch (action.name) {
+    case GeneratorActionName.DownloadDirectory:
+      if (await fs.pathExists(context.destination)) {
```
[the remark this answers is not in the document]
yukun-dong: Yes it's a new folder. The coordinator ensures that.

packages/fx-core/src/component/generator/utils.ts
```diff
@@ -55,9 +57,11 @@ export async function sendRequestWithRetry<T>(
       const res = await requestFn();
       if (res.status === 200 || res.status === 201) {
         return res;
+      } else if (res.status === 403) {
```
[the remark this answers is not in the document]
yukun-dong: Yes, you are right. I will change the string here.

packages/fx-core/src/component/generator/generatorAction.ts
```diff
@@ -17,9 +17,11 @@ export interface GeneratorContext {
   tryLimits?: number;
   timeoutInMs?: number;
   relativePath?: string;
-  zipUrl?: string;
+  // for downloading directory, the url will be the github directory url
```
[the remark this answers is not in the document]
yukun-dong: Ok, I will update this.

packages/fx-core/src/component/generator/generator.ts
```diff
@@ -171,8 +178,18 @@ export async function sampleDefaultOnActionError(
...
+        await fs.rm(context.destination, { recursive: true });
+      }
+      if (error instanceof BaseComponentInnerError) throw error.toFxError();
```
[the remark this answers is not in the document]
yukun-dong: This action could throw 3 types of error: parseGitHubUrlError, ApiLimitError, NetworkError. This line is tu throw the parseGitHubUrlError.
I will check if the error handling part could be improved.

packages/fx-core/src/component/generator/utils.ts
```diff
@@ -235,3 +239,97 @@ export function zipFolder(folderPath: string): AdmZip {
...
+  );
+}
+
+type SampleUrlInfo = {
+  owner: string;
+  repository: string;
+  ref: string;
+  dir: string;
+};
+function parseSampleUrl(url: string): SampleUrlInfo {
```
eriolchan, line 267: It is better to have a blank line between these two method. I wonder why this is checked by eslint/prettier?

PR #8323, opened 2023-04-10
chore: update cd name with branch name and preid

wenytang-ms:
image


wenytang-ms:
schedule skip the run-name display

IvanJobs (approved):

.github/workflows/cd.yml
```diff
@@ -1,5 +1,5 @@
 name: CD
-
+run-name: CD ${{ github.run_id }} on ${{ github.ref_name }} with preid ${{ inputs.preid }}
```
qinezh, line 2: no need to add `${{ github.run_id }}` as it will be included automatically?
image

wenytang-ms: this is not run_id

PR #8336, opened 2023-04-11
chore(e2e): upgrade

xzf0587:


Alive-Fish (approved):
Will solve the error when calling `teamsfx upgrade` on CI_ENABLED part.

PR #8345, opened 2023-04-11
fix: component property is redacted in action telemetry

blackchoey:
The string with `/` character will be treated as path and redacted by VSCode telemetry reporter. Remove the `/` character in component name to avoid the value being redacted.

E2E TEST: N/A. Covered by UT.


blackchoey:
The string with `/` character will be treated as path and redacted by VSCode telemetry reporter. Remove the `/` character in component name to avoid the value being redacted.

Related bug: https://dev.azure.com/msazure/Microsoft%20Teams%20Extensibility/_workitems/edit/17921633

E2E TEST: N/A. Covered by UT.

nliu-ms (dismissed):

swatDong (dismissed):

xzf0587 (approved):

chagong (approved):

PR #8369, opened 2023-04-12
chore(vsc): support custom Teams app id key

tecton:
Related bug: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/17902656


kuojianlu (approved):
LGTM. Curious if PM has any concerns.

HuihuiWu-Microsoft (approved):

PR #8376, opened 2023-04-13
chore: unify sample url for internal/external samples

HuihuiWu-Microsoft:


packages/fx-core/src/common/samples.ts
```diff
@@ -41,7 +41,7 @@ class SampleProvider {
             url:
               (sample as any).relativePath && (sample as any).url
                 ? (sample as any).url
-                : sampleConfigV3.baseUrl,
```
yukun-dong, line 44: Should L60 be changed as well?

HuihuiWu-Microsoft: Will we have other hotfix release for v4? We can update if yes.

yukun-dong: I see. As long as both v3 and v2 work, we can leave it alone.

packages/fx-core/src/common/samples.ts
```diff
@@ -41,7 +41,7 @@ class SampleProvider {
...
+                : `${sampleConfigV3.baseUrl}/${sample.id}`,
```
yukun-dong, line 44:
```suggestion
                : `${sampleConfigV3.baseUrl}${sample.id}`,
```
The baseUrl already has a '/' at the end.

packages/fx-core/src/common/samples.ts
HuihuiWu-Microsoft: Actually hotfix release for v2 will be shipped from last hotfix branch instead of dev branch. So this won't affect v2.

packages/fx-core/src/common/samples.ts
HuihuiWu-Microsoft: Nice catch! I wasn't aware of this since opening is successful with additional '/'.

yukun-dong:
There is 1 ut failed. Please help fix as well.

yukun-dong (approved):

PR #8404, opened 2023-04-14
ci: Update .eslintignore

Jvr2022:
In this example, we've added the following patterns:

tab/csharp/default/*: This pattern ignores all files in the tab/csharp/default/ directory. tab/csharp/non-sso/*: This pattern ignores all files in the tab/csharp/non-sso/ directory. **/*.css: This pattern ignores all CSS files.
Note that we've also added an exception for the package.json file in the root directory, since ESLint needs to be able to read that file to determine the configuration of the project.

### DIFF
--- a/templates/.eslintignore
+++ b/templates/.eslintignore
@@ -1,3 +1,12 @@
+# Ignore everything in the root directory except for package.json
+/*
+!package.json
+
+# Ignore all files in tab/csharp/default/
 tab/csharp/default/*
+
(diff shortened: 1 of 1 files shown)


templates/.eslintignore
```diff
@@ -1,3 +1,12 @@
+# Ignore everything in the root directory except for package.json
+/*
```
yiqing-zhao, line 2: This line may cause ESLint to ignore all the files, not only the files in the root directory.
image

PR #8408, opened 2023-04-17
fix(vs): not confirm manifest and fix the server api

Alive-Fish:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/17924560
E2E TEST: N/A
image


wenytang-ms:
use fix as PR title type instead of chore, thanks!

chagong (approved):

nliu-ms (approved):

PR #8424, opened 2023-04-18
fix: add necessary comment in generated userdata files in migration

frankqianms:
E2E TEST: N/A
fix bug: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/17956993

### DIFF
--- a/packages/fx-core/src/common/versionMetadata.ts
+++ b/packages/fx-core/src/common/versionMetadata.ts
@@ -1,6 +1,14 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
+/**
+ * @author xzf0587 <zhaofengxu@microsoft.com>
+ */
+import { EOL } from "os";
+
+/**
+ * @author xzf0587 <zhaofengxu@microsoft.com>
+ */
 export const MetadataV3 = {
   projectVersion: "1.0.0",
   platformVersion: {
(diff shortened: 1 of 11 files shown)


xzf0587 (dismissed):

packages/fx-core/src/common/versionMetadata.ts
```diff
@@ -18,6 +18,8 @@ export const MetadataV3 = {
   teamsManifestFileName: "manifest.json",
   aadManifestFileName: "aad.manifest.json",
   v3UpgradeWikiLink: "https://aka.ms/teams-toolkit-5.0-upgrade",
+  secretFileComment:
```
blackchoey, line 21: please also add comment for `.env.dev` and `.env.local`, which has different strings.

frankqianms: done

packages/fx-core/src/common/versionMetadata.ts
```diff
@@ -18,6 +26,13 @@ export const MetadataV3 = {
...
+    "# This file includes environment variables that will not be committed to git by default. You can set these environment variables in your CI/CD system for your project." +
+    EOL,
+  secretComment:
+    "# Secrets. Keys prefixed with `SECRET_` will be masked in Teams Toolkit logs." + EOL,
+  envFileComment:
```
blackchoey, line 34: .env.local has different comment than .env.dev, see https://github.com/OfficeDev/TeamsFx/blob/dev/templates/scenarios/js/sso-tab/env/.env.local.tpl#L1

frankqianms: got it

blackchoey (approved):

xzf0587 (approved):

packages/fx-core/src/common/versionMetadata.ts
```diff
@@ -1,6 +1,14 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
+/**
+ * @author xzf0587 <zhaofengxu@microsoft.com>
+ */
+import { EOL } from "os";
+
+/**
+ * @author xzf0587 <zhaofengxu@microsoft.com>
```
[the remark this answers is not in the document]
frankqianms: It is added automatically, I don't know if the format is correct

xzf0587 (approved):

PR #8433, opened 2023-04-18
fix(upgrade): add invalid project condition

xzf0587:
fix bug: https://dev.azure.com/msazure/Microsoft%20Teams%20Extensibility/_workitems/edit/17900027

### DIFF
--- a/packages/fx-core/src/core/middleware/projectMigratorV3.ts
+++ b/packages/fx-core/src/core/middleware/projectMigratorV3.ts
@@ -16,12 +16,11 @@ import {
   Platform,
   AzureSolutionSettings,
   ProjectSettingsV3,
-  Inputs,
 } from "@microsoft/teamsfx-api";
 import { Middleware, NextFunction } from "@feathersjs/hooks/lib";
 import { CoreHookContext } from "../types";
 import { backupFolder, MigrationContext } from "./utils/migrationContext";
-import { checkMethod, checkUserTasks, learnMoreText, upgradeButton } from "./projectMigrator";
+import { upgradeButton } from "./projectMigrator";
 import * as path from "path";
 import { loadProjectSettingsByProjectPathV2 } from "./projectSettingsLoader";
 import {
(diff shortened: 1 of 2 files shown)


packages/fx-core/src/core/middleware/projectMigratorV3.ts
```diff
@@ -179,11 +179,11 @@ export const ProjectMigratorMWV3: Middleware = async (ctx: CoreHookContext, next
...
-  } else if (versionForMigration.state === VersionState.upgradeable && checkMethod(ctx)) {
-    if (!checkUserTasks(ctx)) {
-      ctx.result = ok(undefined);
-      return;
-    }
+  }
+  const projectPath = getParameterFromCxt(ctx, "projectPath", "");
+  const isValid = await checkActiveResourcePlugins(projectPath);
+  const isUpgradeable = isValid && versionForMigration.state === VersionState.upgradeable;
+  if (isUpgradeable) {
```
blackchoey, line 186: misses `checkMethod(ctx)` in original if condition

packages/fx-core/src/core/middleware/projectMigratorV3.ts
```diff
@@ -923,6 +923,20 @@ export async function generateApimPluginEnvContent(context: MigrationContext): P
...
+  try {
+    const projectSettings = await loadProjectSettings(projectPath);
+    const solutionSettings = projectSettings.solutionSettings;
+    if (projectSettings && solutionSettings && solutionSettings.activeResourcePlugins) {
+      return true;
+    } else {
+      return false;
+    }
+  } catch (error) {
+    return false;
```
blackchoey, line 936: should we add some telemetry about why it fails?

blackchoey: Maybe also add some telemetry to tell how many projects have invalid projectSettings.json

packages/fx-core/src/core/middleware/projectMigratorV3.ts
xzf0587: CheckMethod condition is correct in v5. 
It excludes  getProjectConfig and checkPermission which are needed to add upgrade logic. So remove the condition.

blackchoey (approved):

blackchoey:
Please add some test cases for the change

jayzhang (approved):

PR #8450, opened 2023-04-19
feat: enable MetaOS for dashboard templates

huimiu:
FEATURE: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/18028376

Make the dashboard template compatible with MetaOS, so that the Teams Toolkit can have a unified option description when creating a new tab project.


hund030 (dismissed):
M365 platform does not support configurable tab. Should we remove configurable tab from dashboard tab's manifest?

eriolchan (approved):

PR #8451, opened 2023-04-19
test(generator): add ut for download sample

yukun-dong:


packages/fx-core/tests/component/generator/generator.test.ts
```diff
@@ -337,7 +331,20 @@ describe("Generator error", async () => {
...
+      title: "test-title",
+      shortDescription: "test-sd",
+      fullDescription: "test-fd",
+      tags: [],
+      time: "",
+      configuration: "test-configuration",
+      link: "test-link",
+      suggested: false,
+      url: "https://github.com/OfficeDev/TeamsFx-Samples/tree/dev/test",
+    };
```
hund030, line 345: It seems that you've repeated this object three times within different describe blocks. I think it could be beneficial to identify it as a common mocked object. 

Consider creating a Helper class that owns this type of mocked object and provides some common methods that are useful for your testing toolkit.

packages/fx-core/tests/component/generator/generator.test.ts
```diff
@@ -540,10 +573,33 @@ describe("Generate sample using download directory", () => {
...
+    const axiosStub = sandbox.stub(axios, "get");
+    const sampleName = "test";
+    const mockFileName = "test.txt";
+    const mockFileData = "test data";
+    const fileInfo = [
+      { type: "file", path: `${sampleName}/${mockFileName}` },
+      { type: "file", path: `${sampleName}/${mockFileName}_1` },
+      { type: "file", path: `${sampleName}/${mockFileName}_2` },
+    ];
+    axiosStub.onFirstCall().resolves({ status: 200, data: { tree: fileInfo } });
```
hund030, line 593: Is `onFirstCall` the same with `onCall(0)`? If so, I think using `onCall(0)` will reduce some confusion.

packages/fx-core/tests/component/generator/generator.test.ts
```diff
@@ -540,10 +573,33 @@ describe("Generate sample using download directory", () => {
...
+    axiosStub.onCall(1).resolves({ status: 200, data: mockFileData });
+    axiosStub.onCall(2).resolves({ status: 200, data: mockFileData });
+    axiosStub.onCall(3).resolves({ status: 502 });
+    axiosStub.onCall(4).resolves({ status: 502 });
```
hund030, line 597: There are only three files, why do we need to mock four axios calls?

packages/fx-core/tests/component/generator/generator.test.ts
yukun-dong: Yes, let me update.

packages/fx-core/tests/component/generator/generator.test.ts
yukun-dong: The failed call will be retried one more time

packages/fx-core/tests/component/generator/generator.test.ts
yukun-dong: I think changing this var to be a global var will suffice.

PR #8462, opened 2023-04-20
feat(vsc): show help and feedback section on Treeview UI

tecton:
Related Feature: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/17884980

### DIFF
--- a/packages/vscode-extension/package.json
+++ b/packages/vscode-extension/package.json
@@ -167,19 +167,19 @@
         {
           "id": "teamsfx-help-and-feedback",
           "name": "Help and feedback",
-          "when": "fx-extension.isTeamsFx && !fx-extension.canUpgradeV3 && !fx-extension.isV3Enabled",
+          "when": "fx-extension.isTeamsFx && !fx-extension.canUpgradeV3",
           "visibility": "visible"
         },
         {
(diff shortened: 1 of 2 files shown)


packages/vscode-extension/package.json
```diff
@@ -167,19 +167,19 @@
         {
           "id": "teamsfx-help-and-feedback",
           "name": "Help and feedback",
-          "when": "fx-extension.isTeamsFx && !fx-extension.canUpgradeV3 && !fx-extension.isV3Enabled",
+          "when": "fx-extension.isTeamsFx && !fx-extension.canUpgradeV3",
           "visibility": "visible"
         },
         {
           "id": "teamsfx-help",
```
HuihuiWu-Microsoft, line 174: Why not remove the section if not used anymore?

tecton: I think it's possible to be reused again when we need to publish a new prerelease version in future.

HuihuiWu-Microsoft: Do you mean that this feature is just a temporary change and will be reverted in later prerelease?

HuihuiWu-Microsoft (approved):

PR #8472, opened 2023-04-20
fix: upgrade valid domain in manifest

xzf0587:
https://dev.azure.com/msazure/Microsoft%20Teams%20Extensibility/_workitems/edit/17900020
https://dev.azure.com/msazure/Microsoft%20Teams%20Extensibility/_workitems/edit/17958872


packages/fx-core/src/core/middleware/projectMigratorV3.ts
```diff
@@ -406,6 +410,31 @@ export async function manifestsMigration(context: MigrationContext): Promise<voi
     );
   }
 
+  const oldManifestPath = path.join(oldAppPackageFolderPath, MANIFEST_TEMPLATE_CONSOLIDATE);
+  const oldManifestExists = await fs.pathExists(path.join(context.projectPath, oldManifestPath));
+  if (!oldManifestExists) {
+    // templates/appPackage/manifest.template.json does not exist
+    throw MigrationError(
+      new Error(getLocalizedString("core.migrationV3.manifestNotExist")),
```
blackchoey, line 418: is it possible to remove line 403-411 and throw this error istead? Seems we only care whether the manifest file exists, not whether the app package folder exists?

packages/fx-core/src/core/middleware/projectMigratorV3.ts
```diff
@@ -406,6 +410,31 @@ export async function manifestsMigration(context: MigrationContext): Promise<voi
...
+      errorNames.manifestTemplateNotExist,
+      learnMoreLink
+    );
+  } else {
+    const res = await manifestUtils._readAppManifest(
+      path.join(context.projectPath, oldManifestPath)
+    );
+    if (res.isErr()) {
+      throw MigrationError(
+        new Error("templates/appPackage/manifest.template.json is invalid"),
```
blackchoey, line 428: please add the string to package.nls.json

packages/fx-core/src/core/middleware/projectMigratorV3.ts
```diff
@@ -406,6 +410,31 @@ export async function manifestsMigration(context: MigrationContext): Promise<voi
...
+        errorNames.manifestTemplateInvalid
+      );
+    }
+    const teamsAppManifest = res.value;
+    if (teamsAppManifest.validDomains?.includes(validDomain.botWithValid)) {
```
blackchoey, line 433: seems `v3MigrationUtils` just adds `validDomain.bot`. When will `validDomain.botWithValid` be added?

packages/fx-core/src/core/middleware/utils/v3MigrationUtils.ts
```diff
@@ -340,6 +350,27 @@ export async function updateAndSaveManifestForSpfx(
   await context.fsWriteFile(localTemplatePath, localTemplate);
 }
 
+export async function handleValidDomainForManifest(
+  manifestPath: string,
+  tab: boolean,
+  bot: boolean
```
blackchoey, line 356: maybe give it a better name? like `addMissingValidDomainForManifest`.
By the way, since V4 has similar logic, is it possible to reuse that logic?

swatDong (dismissed):

packages/fx-core/tests/core/middleware/migration/migrationUtilsV3.test.ts
```diff
@@ -400,3 +405,32 @@ describe("Migration: upgrade cancel messages", () => {
...
+      validDomains: [],
+    } as unknown as TeamsAppManifest;
+    sandbox.stub(ManifestUtils.prototype, "_readAppManifest").resolves(ok(teamsAppManifest));
+    const stub = sandbox.stub(ManifestUtils.prototype, "_writeAppManifest");
+    await v3MigrationUtils.handleValidDomainForManifest("", true, true);
+    const res = {
+      validDomains: [validDomain.tab, validDomain.bot],
+    };
+    stub.calledOnceWith(res as TeamsAppManifest, "");
+    // assert.equal(messageArray.join(""), groundTruth.join(""));
```
blackchoey, line 434: Remove the comment

packages/fx-core/src/core/middleware/projectMigratorV3.ts
xzf0587: validDomain.botWithValid is used to control debug task.json output key name.
Currently, we found no project missed validDomain.botWithValid. So there's no need to add it.

swatDong (approved):

sffamily (dismissed):

packages/fx-core/src/core/middleware/utils/v3MigrationUtils.ts
xzf0587: The name has been updated.
The v4 has no logic for valid domain, so the logic can't be reused.

blackchoey (approved):

packages/fx-core/src/core/middleware/projectMigratorV3.ts
blackchoey: Can you add a P2 task to add `validDomain.botWithValid`? The priority is not high since no template missed this placeholder. But since users may modify the template and we have hidden logic to fix ths missing value in V4, it's better to fix the template during migration.

LongOddCode (approved):

swatDong (approved):

nliu-ms (approved):

packages/fx-core/src/core/middleware/projectMigratorV3.ts
xzf0587: sure. Add it in issue
https://dev.azure.com/msazure/Microsoft%20Teams%20Extensibility/_workitems/edit/19286648

sffamily (approved):

PR #8486, opened 2023-04-21
feat(sdk): notification sdk enhancement for .NET

qinezh:
Work item: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/17902085/

Updates in this PR:
1. Add new interface `IConversationReferenceStore`.
2. Add new option `store` in `NotificationBot`.
3. Deprecate the interface `INotificationTargetStorage` and the option `storage` in `NotificationBot`.
4. Add new pagaination API `GetPagedInstallationsAsync`, `GetPagedMembersAsync`.
5. Deprecated API `GetMembersAsync`.
6. Add API `BuildTeamsBotInstallation`.

Incoming PR:
1. Bump new version. (after the 2.0.0 is released)
2. Update README and Change log.
3. Update templates.

### DIFF
--- a/packages/dotnet-sdk/src/TeamsFx/Conversation/NotificationOptions.cs
+++ b/packages/dotnet-sdk/src/TeamsFx/Conversation/NotificationOptions.cs
@@ -14,6 +14,30 @@ namespace Microsoft.TeamsFx.Conversation
         public string BotAppId { get; set; } = string.Empty;
 
         /// <summary>
+        /// An optional store to persist bot notification connections.
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// If <c>Store</c> is not provided, a default local store will be used, which stores notification connections into:
+        /// </para>
+        /// <list type="bullet">
+        ///     <item>
+        ///         <description>"{$env:TEAMSFX_NOTIFICATION_LOCALSTORE_DIR}/.notification.localstore.json" if running locally.</description>
+        ///     </item>
+        ///     <item>
+        ///         <description>"{$env:TEMP}/.notification.localstore.json" if {$env:RUNNING_ON_AZURE} is set to "1".</description>
+        ///     </item>
+        ///     <item>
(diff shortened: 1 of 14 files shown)


packages/dotnet-sdk/src/TeamsFx/Conversation/PagedData.cs
```diff
@@ -0,0 +1,27 @@
...
+        /// <value>
+        /// The array of items.
+        /// </value>
+        public T[] Data { get; set; } = Array.Empty<T>();
+
+        /// <summary>
+        /// Gets or sets a token for retrieving the next page of results.
+        /// </summary>
+        /// <value>
+        /// The Continuation Token to pass to get the next page of results.
```
swatDong, line 23: Better to add something like "*Null or empty token means the page reaches the end*"

swatDong (approved):

packages/dotnet-sdk/src/TeamsFx/Conversation/NotificationOptions.cs
```diff
@@ -13,6 +13,30 @@ public class NotificationOptions
         /// </summary>
         public string BotAppId { get; set; } = string.Empty;
 
+        /// <summary>
```
dooriya, line 16: nit: Here we  use `notification connections`  to describe the `Store` member, while in the `IConversationReferenceStore` interface we use `notification target references`, it would be better to consistent the terminology in our SDK.

packages/dotnet-sdk/src/TeamsFx/Conversation/PagedData.cs
qinezh: Updated.

packages/dotnet-sdk/src/TeamsFx/Conversation/NotificationOptions.cs
qinezh: Updated.

swatDong (approved):

dooriya (approved):

PR #8492, opened 2023-04-22
fix(localdebug): specify func path in cli preview command

XiaofuHuang:
detail: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/17920588/


packages/cli/src/cmds/preview/previewEnv.ts
```diff
@@ -97,6 +97,12 @@ export default class PreviewEnv extends YargsCommand {
         description: 'Argument to pass to the browser (e.g. --browser-args="--guest")',
         string: true,
         array: true,
+      })
+      .options("exec-path", {
+        description:
+          'A path that will be added to the system environment variable PATH when the command is executed, defaults to "${folder}/devTools/func".',
```
swatDong, line 103: How about default to `devTools/` and all subfolders of `devTools/`?

XiaofuHuang: If `exec-path` means adding all subfolders of `exec-path` to the env.PATH, it can be difficult for users to specify a custom function path.
By default, the `exec-path` includes all default paths installed with Teams Toolkit. In the future, it may include additional paths, such as `${folder}/devTools/func;${folder}/devTools/videoFilter`.

swatDong: Also make sense, just make sure it includes multiple default paths.

swatDong (approved):

swatDong (approved):

PR #8504, opened 2023-04-23
docs: Yaml style change

Siglud:


templates/scenarios/csharp/command-and-response/teamsapp.yml.tpl
```diff
@@ -84,10 +84,13 @@ deploy:
     with:
       args: publish --configuration Release --runtime win-x86 --self-contained
   # Deploy your application to Azure App Service using the zip deploy feature.
-  # For additional details, please refer to https://aka.ms/zip-deploy-to-app-services.
+  # For additional details, refer to https://aka.ms/zip-deploy-to-app-services.
```
swatDong, line 87: It seems some are *refer to* while some are *see at*. Consolidate to one or ok to keep two?

Siglud: The AzureFunctions driver has a column limit of 80, but it is exceeding this limit at 82. We need to either wrap the text onto another line or change the wording.

swatDong: I see. Make sense.

swatDong (approved):

swatDong (approved):

hund030 (approved):

PR #8508, opened 2023-04-23
fix(localdebug): update the func checker telemetry and help link

XiaofuHuang:
1. Update the depschecker help links
2. Record the detail error message if func installation failed.
3. Add telemetries about func version and error messages.
4. Refactor the code after the return type of `checkFuncVersion` changed.


packages/fx-core/src/component/driver/devTool/installDriver.ts
```diff
@@ -249,7 +249,14 @@ export class ToolsInstallDriverImpl {
     this.context.addTelemetryProperties({
       [TelemetryProperties.driverArgs]: JSON.stringify({
         devCert: args.devCert,
-        func: args.func,
+        func: {
+          version: args.func?.version,
+          symlinkDir: args.func?.symlinkDir
+            ? args.func.symlinkDir === "./devTools/func"
```
swatDong, line 255: Do we treat exactly the same string as default, or the same folder as default? E.g., `devTools/func`, `devTools/func/`, `./devTools/func/` ...

XiaofuHuang: Good idea, updated.

swatDong (approved):

sffamily (approved):

PR #8518, opened 2023-04-24
test: add all-version-upgrade-test script ymls

frankqianms:


LongOddCode (approved):

.github/workflows/all-versions-upgrade-test.yml
```diff
@@ -0,0 +1,212 @@
...
+      if: ${{ always() }}
+
+  rerun:
+    needs:
+    - sampleTest
+    - templateTest
+    if: ${{ failure() && github.run_attempt < 3 }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: rerun
```
blackchoey, line 204: do we need rerun? I think some matrix will always fail (like some CLI version does not have certain capability), which will always result in rerun.

frankqianms: Yes we do, I think. Some cases will failed with time-out due to network delay. So we need to re-run them.

blackchoey (approved):

PR #8544, opened 2023-04-25
fix: cli upgrade

xzf0587:
https://dev.azure.com/msazure/Microsoft%20Teams%20Extensibility/_workitems/edit/17793404

### DIFF
--- a/packages/fx-core/src/core/FxCoreImplementV3.ts
+++ b/packages/fx-core/src/core/FxCoreImplementV3.ts
@@ -441,7 +441,7 @@ export class FxCoreV3Implement {
         return err(new InvalidProjectError());
       }
     }
-    return this.innerMigrationV3(inputs);
+    return await this.innerMigrationV3(inputs);
   }
 
   @hooks([ErrorHandlerMW, ProjectMigratorMWV3])
--- a/packages/fx-core/src/core/FxCoreImplementV3.ts
+++ b/packages/fx-core/src/core/FxCoreImplementV3.ts
@@ -97,6 +97,7 @@ import { pathToFileURL } from "url";
 import { VSCodeExtensionCommand } from "../common/constants";
 import { Hub } from "../common/m365/constants";
 import { LaunchHelper } from "../common/m365/launchHelper";
+import { NoNeedUpgradeError } from "../error/upgrade";
 
 export class FxCoreV3Implement {
   tools: Tools;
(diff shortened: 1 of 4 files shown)


sffamily (dismissed):

packages/fx-core/tests/core/FxCore.test.ts
```diff
@@ -447,6 +449,65 @@ describe("Core basic APIs", () => {
...
+    try {
+      const core = new FxCore(tools);
+      const appName = await mockV2Project();
+      const inputs: Inputs = {
+        platform: Platform.VSCode,
+        projectPath: path.join(os.tmpdir(), appName),
+        skipUserConfirm: true,
+      };
+      const res = await core.phantomMigrationV3(inputs);
+      assert.isTrue(res.isErr());
```
blackchoey, line 466: assert which error (NoNeedUpgrade or InvalidProject) it throws, so we can ensure users can get the correct message

blackchoey: please also assert error for other test cases

packages/fx-core/src/core/FxCoreImplementV3.ts
```diff
@@ -426,8 +427,25 @@ export class FxCoreV3Implement {
     return ok(ctx?.envVars);
   }
 
-  @hooks([ErrorHandlerMW, ProjectMigratorMWV3])
   async phantomMigrationV3(inputs: Inputs): Promise<Result<Void, FxError>> {
+    // If the project is invalid or upgraded, the ProjectMigratorMWV3 will not take action.
+    // Check invaliad/upgraded project here before call ProjectMigratorMWV3
+    const projectPath = (inputs.projectPath as string) || "";
```
blackchoey, line 433: I remember similar logic exists in other place. If it's true, can we find a way to reuse the common part?

xzf0587: Here are the error conditions for the specific command upgrade not for middleware. 
In the ProjectMigratorMWV3, we do this check as well and the handle result is not the same. It will not return error but skip upgrade flow for these cases.
The common logic is wrapped as functions already.

packages/fx-core/tests/core/FxCore.test.ts
xzf0587: Fix it, thanks.

blackchoey (approved):

jayzhang (approved):

sffamily (approved):

PR #8574, opened 2023-04-27
fix: using kudu REST endpoint with AAD Auth

Siglud:
fix: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/19691043

### DIFF
--- a/packages/fx-core/src/component/driver/deploy/azure/impl/azureDeployImpl.ts
+++ b/packages/fx-core/src/component/driver/deploy/azure/impl/azureDeployImpl.ts
@@ -108,7 +108,7 @@ export abstract class AzureDeployImpl extends BaseDeployImpl {
   public async checkDeployStatus(
     location: string,
     config: AzureUploadConfig,
-    logger?: LogProvider
+    logger: LogProvider
   ): Promise<DeployResult | undefined> {
     let res: AxiosDeployQueryResult;
     for (let i = 0; i < DeployConstant.DEPLOY_CHECK_RETRY_TIMES; ++i) {
@@ -116,7 +116,7 @@ export abstract class AzureDeployImpl extends BaseDeployImpl {
         res = await AzureDeployImpl.AXIOS_INSTANCE.get(location, config);
       } catch (e) {
         if (axios.isAxiosError(e)) {
-          await logger?.error(
+          await logger.error(
             `Check deploy status failed with response status code: ${
               e.response?.status ?? "NA"
             }, message: ${JSON.stringify(e.response?.data)}`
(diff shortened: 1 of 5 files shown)


packages/fx-core/src/component/driver/deploy/azure/impl/azureDeployImpl.ts
```diff
@@ -174,6 +174,34 @@ export abstract class AzureDeployImpl extends BaseDeployImpl {
...
+          headers: {
+            "Content-Type": "application/octet-stream",
+            "Cache-Control": "no-cache",
+            Authorization: `Bearer ${token.token}`,
+          },
+          maxContentLength: Infinity,
+          maxBodyLength: Infinity,
+          timeout: DeployConstant.DEPLOY_TIMEOUT_IN_MS,
+        };
+      } else {
```
yukun-dong, line 192: Since basic auth will be disabled, why don't we just delete the basic auth related code?
For example: just throw error if `token` is empty.

packages/fx-core/src/component/driver/deploy/azure/impl/azureDeployImpl.ts
```diff
@@ -174,6 +174,34 @@ export abstract class AzureDeployImpl extends BaseDeployImpl {
       azureCredential,
       azureResource.subscriptionId
     ));
+    try {
+      const defaultScope = "https://management.azure.com/.default";
+      const token = await azureCredential.getToken(defaultScope);
+      if (token) {
+        this.logger?.info("Get AAD token successful. Upload zip package through AAD Auth mode.");
```
yukun-dong, line 181: successfully.

eriolchan (approved):

packages/fx-core/src/component/driver/deploy/azure/impl/azureDeployImpl.ts
Siglud: Fixed

packages/fx-core/src/component/driver/deploy/azure/impl/azureDeployImpl.ts
Siglud: This method has not been thoroughly tested; it is unclear if any accounts are unable to access the information.

eriolchan (approved):

packages/fx-core/src/component/driver/deploy/azure/impl/azureDeployImpl.ts
```diff
@@ -170,6 +170,34 @@ export abstract class AzureDeployImpl extends BaseDeployImpl {
...
+        );
+      }
+    } catch (e) {
+      this.logger.info(
+        `Get AAD token failed with error: ${JSON.stringify(
+          e
+        )}. Upload zip package through basic auth mode.`
+      );
+    }
+
```
XiaofuHuang, line 200: Should we update the bicep file to disable basic authentication? If we make this change, the basic authentication logic that follows will no longer work.

PS: When we create an App Service with Basic Authentication enabled, it may trigger an alert.

Siglud: I will remove that code snippet after the ARM templates are changed.

PR #8605, opened 2023-05-05
fix: do not throw when remote deploy response failed, let user check it manually

Siglud:
fix: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/21015887

### DIFF
--- a/packages/fx-core/src/component/driver/deploy/azure/impl/azureDeployImpl.ts
+++ b/packages/fx-core/src/component/driver/deploy/azure/impl/azureDeployImpl.ts
@@ -145,7 +145,6 @@ export abstract class AzureDeployImpl extends BaseDeployImpl {
                 JSON.stringify(res.data)
               )
             );
-            // throw new DeployRemoteStartError(location, JSON.stringify(res.data), this.helpLink);
           }
           return res.data;
         } else {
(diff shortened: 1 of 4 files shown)


packages/fx-core/src/component/driver/deploy/azure/impl/azureDeployImpl.ts
```diff
@@ -139,10 +138,14 @@ export abstract class AzureDeployImpl extends BaseDeployImpl {
...
-              `Deployment is failed with error message: ${JSON.stringify(res.data)}`
+            await this.logger.warning(
+              getDefaultString(
+                "error.deploy.DeployRemoteStartError",
+                location,
+                JSON.stringify(res.data)
+              )
             );
-            throw new DeployRemoteStartError(location, JSON.stringify(res.data), this.helpLink);
+            // throw new DeployRemoteStartError(location, JSON.stringify(res.data), this.helpLink);
```
eriolchan, line 148: If we don't use this code, it is better to delete it instead of comment it out

eriolchan (approved):

packages/fx-core/src/component/driver/deploy/azure/impl/azureDeployImpl.ts
Siglud: Removed

eriolchan (approved):

PR #8637, opened 2023-05-08
fix: handle msedge/chrome resolved twice

kuojianlu:
- fix env not set on second-time resolving, causing wrong telemetry
- fix sideloading message print twice


kuojianlu:
- fix env not set on second-time resolving, causing wrong telemetry
- fix sideloading message print twice

Fix https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/21189964/.

packages/vscode-extension/src/debug/teamsfxDebugProvider.ts
```diff
@@ -166,17 +169,22 @@ export class TeamsfxDebugProvider implements vscode.DebugConfigurationProvider {
...
-      VsCodeLogInstance.outputChannel.appendLine("");
-      VsCodeLogInstance.outputChannel.appendLine(
-        sideloadingDisplayMessages.sideloadingUrlMessage(
-          debugConfiguration.teamsfxHub!,
-          debugConfiguration.url
-        )
-      );
-      if (isLocal) {
+      // NOTE: handle the case that msedge/chrome will be resolved twice
+      if (debugConfiguration.type.startsWith("pwa")) {
```
swatDong, line 173: Would there be risk if vscode updates its resolve logic...
How about checking *url is updated*? For those no-placeholder ones, just ignore since the url is already plain text.

a1exwang (dismissed):

packages/vscode-extension/src/debug/teamsfxDebugProvider.ts
kuojianlu: Url can contain no placeholders (like Outlook message extension url). Let me add an extra variable to indicate if already resolved.

packages/vscode-extension/src/debug/teamsfxDebugProvider.ts
```diff
@@ -166,18 +169,24 @@ export class TeamsfxDebugProvider implements vscode.DebugConfigurationProvider {
...
-      VsCodeLogInstance.outputChannel.appendLine("");
-      VsCodeLogInstance.outputChannel.appendLine(
-        sideloadingDisplayMessages.sideloadingUrlMessage(
-          debugConfiguration.teamsfxHub!,
-          debugConfiguration.url
-        )
-      );
-      if (isLocal) {
+      // NOTE: handle the case that msedge/chrome will be resolved twice
+      if (!debugConfiguration.resolved) {
```
swatDong, line 173: Better naming to `teamsfxResolved` to avoid potential conflict.

kuojianlu: Updated.

swatDong (approved):

XiaofuHuang (approved):

PR #8641, opened 2023-05-09
fix: missing privacy and termsofuse page

hund030:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/21321936/


templates/scenarios/js/non-sso-tab/appPackage/manifest.json.tpl
```diff
@@ -6,9 +6,9 @@
     "packageName": "com.microsoft.teams.extension",
     "developer": {
         "name": "Teams App, Inc.",
-        "websiteUrl": "${{TAB_ENDPOINT}}",
-        "privacyUrl": "${{TAB_ENDPOINT}}/privacy",
-        "termsOfUseUrl": "${{TAB_ENDPOINT}}/termsofuse"
+        "websiteUrl": "https://www.example.com",
+        "privacyUrl": "https://www.example.com/termofuse",
```
huimiu, line 10: These two values seem to be reversed.

hund030: Good Catch! These two values were copied from the default-bot template, and they were wrong from the beginning.

huimiu (approved):

PR #8721, opened 2023-05-16
fix(core): default env write behavior when env folder path is not available

jayzhang:
fix: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/22651915
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/22302675

### DIFF
--- a/packages/fx-core/resource/package.nls.json
+++ b/packages/fx-core/resource/package.nls.json
@@ -976,6 +976,7 @@
   "error.common.MissingRequiredInputError": "Missing required input: %s",
   "error.common.InputValidationError": "Input '%s' validation failed: %s",
   "error.common.NoEnvFilesError": "Unable to find .env files.",
+  "error.common.MissingRequiredFileError": "Missing %srequired file `%s`",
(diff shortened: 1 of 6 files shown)


LongOddCode (approved):

packages/fx-core/resource/package.nls.json
```diff
@@ -976,6 +976,7 @@
   "error.common.MissingRequiredInputError": "Missing required input: %s",
   "error.common.InputValidationError": "Input '%s' validation failed: %s",
   "error.common.NoEnvFilesError": "Unable to find .env files.",
+  "error.common.MissingRequiredFileError": "Missing %srequired file `%s`",
```
sffamily, line 979: missing a space?

otherwise it can be approved.

sffamily (approved):

packages/fx-core/resource/package.nls.json
jayzhang: For debug case, the %s="Debug " (with a space), for other case, we don't know the command executed, so we will pass empty string into the placeholder, which is by design

PR #8727, opened 2023-05-17
feat: add update launch url action for vs

xiaolang124:
1.Add update launch url action for vs

https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/23190906


packages/fx-core/src/component/driver/file/updateLaunchUrlInLaunchSettings.ts
```diff
@@ -0,0 +1,135 @@
...
+import { logMessageKeys } from "../aad/utility/constants";
+import { DriverContext } from "../interface/commonArgs";
+import { ExecutionResult, StepDriver } from "../interface/stepDriver";
+import { addStartAndEndTelemetry } from "../middleware/addStartAndEndTelemetry";
+import { updateProgress } from "../middleware/updateProgress";
+import { InvalidActionInputError, UnhandledError } from "../../../error/common";
+import { UpdateLaunchUrlInLaunchSettingsArgs } from "./interface/UpdateLaunchUrlInLaunchSettingsArgs";
+import { AppStudioScopes } from "../../../common/tools";
+
+const actionName = "file/updateLaunchUrlInLaunchSettings";
```
swatDong, line 21: To be more general, how about `file/createOrUpdateLaunchProfile`?
Not just text replacement, but to modify the whole launch profile.
E.g., input *name, hub, appid, ...*, and output the whole launch profile.

xiaolang124: In vs 'launchSettings.json', we have comment content. It is not easy to handle this part like json content. So, I just replace 'launchUrl' using regex. Or should I just use another style to generate and replace whole 'launchSettings.json' using an action?

swatDong: I remember we already have such json comment logic (v5 tasks.json migration?).

packages/fx-core/resource/yaml.schema.json
```diff
@@ -1133,6 +1134,54 @@
...
+          "description": "Update launch url in launchSettings.json. Refer to https://aka.ms/teamsfx-actions/file-updatelaunchurlinlaunchsettings for more details.",
+          "const": "file/updateLaunchUrlInLaunchSettings"
+        },
+        "with": {
+          "type": "object",
+          "additionalProperties": false,
+          "description": "parameters for this action",
+          "required": ["target", "profile", "launchUrl", "addLoginHint"],
+          "properties": {
+            "target": {
```
swatDong, line 1165: Is target changeable in vs? Or make this field optional?

packages/fx-core/resource/yaml.schema.json
```diff
@@ -1133,6 +1134,54 @@
...
+              "type": "string",
+              "description": "the target file"
+            },
+            "profile": {
+              "type": "string",
+              "description": "the profile to be updated"
+            },
+            "launchUrl": {
```
swatDong, line 1173: Instead of the whole url, how about fields like `appId`, `host`, ... to be able to extend for m365 or more features.

packages/fx-core/resource/yaml.schema.json
```diff
@@ -1133,6 +1134,54 @@
         }
       }
     },
+    "fileUpdateLaunchUrlInLaunchSettings": {
```
swatDong, line 1137: Make the action general to create a new profile or update an existing one, e.g.,
``` yaml
- uses: file/createOrUpdateDebugProfile
  with:
    name: Microsoft Teams (browser) # the name of launch profile
    appId: ${{TEAMS_APP_ID}}
    loginHint: true # can be optional
    host: teams # can be optional
    ...
```

If there's no such a profile, generate a new one.
If there's already such a profile, update its `launchUrl`.

packages/fx-core/src/component/driver/file/updateLaunchUrlInLaunchSettings.ts
swatDong: BTW, in the VS schema (https://json.schemastore.org/launchsettings.json) it seems to be called *Debug Profile*

xiaolang124: Got it, I will update this part using comment json.

xiaolang124: Done.

packages/fx-core/resource/yaml.schema.json
xiaolang124: Done.

packages/fx-core/resource/yaml.schema.json
xiaolang124: Done.

packages/fx-core/resource/yaml.schema.json
```diff
@@ -1133,6 +1134,54 @@
...
+            "appId": {
+              "type": "string",
+              "description": "the app id to be used in launch url"
+            },
+            "loginHint": {
+              "type": "boolean",
+              "description": "whether to add login hint"
+            },
+            "host": {
+              "type": "string",
```
swatDong, line 1178: enum? We can support only `teams` for now.

packages/fx-core/resource/yaml.schema.json
xiaolang124: Already update local yml file.

packages/fx-core/src/component/driver/file/createOrUpdateDebugProfile.ts
```diff
@@ -0,0 +1,146 @@
...
+  }> {
+    try {
+      let launchSettingsPath = "";
+      if (!args.target) {
+        launchSettingsPath = getAbsolutePath(launchSettingsFilePath, context.projectPath);
+      } else {
+        launchSettingsPath = getAbsolutePath(args.target, context.projectPath);
+      }
+      if (!(await fs.pathExists(launchSettingsPath))) {
+        throw new UserError(
```
swatDong, line 85: Instead of error, just create the launch settings file.

packages/fx-core/src/component/driver/file/createOrUpdateDebugProfile.ts
```diff
@@ -0,0 +1,146 @@
...
+          "LaunchSettingsFileNotExist",
+          getLocalizedString(
+            "driver.file.createOrUpdateDebugProfile.launchSettingsFileNotExist",
+            launchSettingsPath
+          ),
+          helpLink
+        );
+      }
+      const launchSettingsContent = fs.readFileSync(launchSettingsPath, "utf-8");
```
swatDong, line 94: Use async function

packages/fx-core/src/component/driver/file/createOrUpdateDebugProfile.ts
```diff
@@ -0,0 +1,146 @@
...
+        });
+        if (tokenObjectRes.isErr()) {
+          throw tokenObjectRes.error;
+        }
+        const tokenObject = tokenObjectRes.value;
+        if (tokenObject && "upn" in tokenObject) {
+          launchUrl += `&login_hint=${tokenObject.upn}`;
+        }
+      }
+      (data.profiles as any)[args.name].launchUrl = launchUrl;
```
swatDong, line 123: If *{name}* profile not exist, add new profile.

packages/fx-core/src/component/driver/file/interface/createOrUpdateDebugProfileArgs.ts
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+export interface CreateOrUpdateDebugProfileArgs {
+  target: string | undefined; // The relative path of the laucnSettings.json file
```
swatDong, line 5: When is `target` needed? Seems not in yaml schema.

packages/fx-core/resource/yaml.schema.json
xiaolang124: Done.

packages/fx-core/src/component/driver/file/createOrUpdateDebugProfile.ts
xiaolang124: Done.

packages/fx-core/src/component/driver/file/createOrUpdateDebugProfile.ts
xiaolang124: Done.

packages/fx-core/src/component/driver/file/createOrUpdateDebugProfile.ts
xiaolang124: Done.

packages/fx-core/src/component/driver/file/interface/createOrUpdateDebugProfileArgs.ts
xiaolang124: 'target' is removed from createOrUpdateDebugProfileArgs.

packages/fx-core/src/component/driver/file/createOrUpdateDebugProfile.ts
```diff
@@ -0,0 +1,139 @@
...
+    args: CreateOrUpdateDebugProfileArgs,
+    context: DriverContext
+  ): Promise<{
+    output: Map<string, string>;
+    summaries: string[];
+  }> {
+    try {
+      const launchSettingsPath = getAbsolutePath(launchSettingsFilePath, context.projectPath);
+      if (!(await fs.pathExists(launchSettingsPath))) {
+        await fs.writeFile(launchSettingsPath, "{}", "utf-8");
```
swatDong, line 80: Is it possible the parent folder not exist? Will that cause error?

packages/fx-core/src/component/driver/file/createOrUpdateDebugProfile.ts
```diff
@@ -0,0 +1,139 @@
...
+        );
+      }
+      let launchUrl = utils.format(
+        launchUrlTemplate,
+        args.appId,
+        process.env[internalOutputNames.teamsAppTenantId]
+      );
+      if (args.loginHint === undefined || args.loginHint === true) {
+        const tokenObjectRes = await context.m365TokenProvider.getJsonObject({
+          scopes: AppStudioScopes,
```
swatDong, line 101: What if user has not signed in? Prefer to return error instead of popup signin window.

packages/fx-core/src/component/driver/file/createOrUpdateDebugProfile.ts
swatDong: For the new profile, does it only contain `launchUrl`?
Prefer to generate a complete profile with command, browser, appurl, ...

packages/fx-core/src/component/driver/file/createOrUpdateDebugProfile.ts
xiaolang124: In vs, user need to choose m365 before running prepare Teams app. And there is no logic to pop up login page in running prepare Teams app. So I think we do not need to worry this.

packages/fx-core/src/component/driver/file/createOrUpdateDebugProfile.ts
xiaolang124: Done.

packages/fx-core/src/component/driver/file/createOrUpdateDebugProfile.ts
xiaolang124: Add logic to add full profile. Because it's hard to get app type, so I add all content that project may need and users can uncomment part they need.

swatDong: This is to cover more scenarios for existing app so a basic profile is enough. Looks good.

swatDong (approved):

xiaolang124:
This pr will be picked to main.

PR #8751, opened 2023-05-22
ci(api2teams): add ci for api2teams

KennethBWSong:
success run in my fork branch: https://github.com/KennethBWSong/TeamsFx/actions/runs/5042653111/jobs/9043507489?pr=6

### DIFF
--- a/.github/workflows/api2teams_ci.yml
+++ b/.github/workflows/api2teams_ci.yml
@@ -1,0 +1,37 @@
+name: Api2teams CI
+
+on:
+  push:
+    paths:
+      - "packages/api2teams/**"
+    branches: [main, dev, prerelease]
+  pull_request:
+    paths:
+      - "packages/api2teams/**"
+    branches: [main, dev, prerelease, api2teams]
+  workflow_dispatch:
+
+defaults:
+  run:
+    working-directory: packages/api2teams/
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Setup node
+        uses: actions/setup-node@v3
+        with:
+          node-version: 14
+
+      - name: Install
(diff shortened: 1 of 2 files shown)


.github/workflows/api2teams_ci.yml
```diff
@@ -0,0 +1,37 @@
...
+        uses: actions/checkout@v3
+
+      - name: Setup node
+        uses: actions/setup-node@v3
+        with:
+          node-version: 14
+
+      - name: Install
+        run: |
+          npm install
```
wenytang-ms, line 33: better to run "npm run setup"

wenytang-ms, line 33: it's better use "npm run setup"

.github/workflows/api2teams_ci.yml
```diff
@@ -0,0 +1,37 @@
...
+
+on:
+  push:
+    paths:
+      - "packages/api2teams/**"
+    branches: [main, dev, prerelease]
+  pull_request:
+    paths:
+      - "packages/api2teams/**"
+    branches: [main, dev, prerelease, api2teams]
```
wenytang-ms, line 11: is "api2teams" a feature branch?

.github/workflows/api2teams_ci.yml
KennethBWSong: This CI is only for api2teams now, and we don't know whether it will be a long-term support package, so we prefer to run npm install in this package instead of run npm run setup in root folder.

.github/workflows/api2teams_ci.yml
KennethBWSong: Yes, this package is available only in api2teams branch, and we still need to decide whether we will merge it into dev.

wenytang-ms (approved):

SLdragon (approved):

PR #8762, opened 2023-05-23
feat: turn on download directory feature flag

yukun-dong:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_sprints/taskboard/Compute%20Team/Microsoft%20Teams%20Extensibility/CY23-5.2?workitem=23704990

### DIFF
--- a/packages/fx-core/src/common/tools.ts
+++ b/packages/fx-core/src/common/tools.ts
@@ -401,7 +401,7 @@ export function isV3Enabled(): boolean {
 }
 
 export function isDownloadDirectoryEnabled(): boolean {
-  return process.env.DOWNLOAD_DIRECTORY === "true";
+  return true;
 }
 
 export function isVideoFilterEnabled(): boolean {
--- a/packages/fx-core/tests/core/FxCore.create.test.ts
+++ b/packages/fx-core/tests/core/FxCore.create.test.ts
@@ -21,6 +21,7 @@ import { SPFXQuestionNames } from "../../src/component/resource/spfx/utils/quest
 import { setTools } from "../../src/core/globalVars";
 import { environmentManager } from "../../src/core/environment";
 import * as templateActions from "../../src/common/template-utils/templatesActions";
+import { Generator } from "../../src/component/generator/generator";
 
 describe("Core basic APIs for v3", () => {
   const sandbox = sinon.createSandbox();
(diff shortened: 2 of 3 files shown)


packages/fx-core/src/common/tools.ts
```diff
@@ -417,7 +417,7 @@ export function isV3Enabled(): boolean {
 }
 
 export function isDownloadDirectoryEnabled(): boolean {
-  return process.env.DOWNLOAD_DIRECTORY === "true";
+  return process.env.DOWNLOAD_DIRECTORY ? process.env.DOWNLOAD_DIRECTORY === "true" : true;
```
eriolchan, line 420: Since we want to always turn on this feature flag to monitor and collect data in the next release. Why don''t just remove this feature flag?

In this code, it still can go to download from big zip file if the env var is not true.

Siglud: Can you remove this switch or add another one?  
If you want to keep this switch, consider adding a new controller that can be configured to revert back to the original download method. This will make end-users debug more easily.

yukun-dong: I think setting it default to be true is ok. Since user won't know about this feature flag, it's almost impossible for them to turn it off by coincidence. Also we have the flexbility to revert it back to old method.

packages/fx-core/tests/core/FxCore.create.test.ts
```diff
@@ -30,6 +30,16 @@ describe("Core basic APIs for v3", () => {
   beforeEach(() => {
     sandbox.restore();
     setTools(tools);
+    sandbox.stub(environmentManager, "listRemoteEnvConfigs").resolves(ok(["dev"]));
```
eriolchan, line 33: It looks there are a lot of mock http request/response in these test cases. I am think if below way is more simple and clear to test?
1. Generator has its own unit test case to cover.
2. FxCore test case focus more on create flow, and generator can be considered as a dependent library. Is it possible to mock generator API in this case, which keep the same test coverage but also simplify the test case?

yukun-dong: Thanks, updated.

hund030 (approved):

Siglud (approved):

eriolchan (approved):

PR #8784, opened 2023-05-24
refactor(core): cleanup code for unused components

jayzhang:


blackchoey:
This PR decreased the overall code coverage by **1.32%**, can you create a bug as follow up of code coverage improvement? I'm not sure whether the removed test cases also tested some other source code, and removing the test cases causes some code no longer being tested.

cc @LongOddCode , please share your comments if you have

HuihuiWu-Microsoft (approved):

jayzhang:
> 



> This PR decreased the overall code coverage by **1.32%**, can you create a bug as follow up of code coverage improvement? I'm not sure whether the removed test cases also tested some other source code, and removing the test cases causes some code no longer being tested.
> 
> cc @LongOddCode , please share your comments if you have

The removed test cases are for v2 code.

nliu-ms (approved):

Siglud (approved):

packages/fx-core/src/component/resource/aadApp/aadApp.ts
```diff
@@ -191,11 +183,6 @@ export class AadApp implements CloudResource {
     return res;
   }
 
-  @hooks([
```
KennethBWSong, line 194: We still will call collaboration/buildAadManifest for non-lifecycle commands. Will this change have any influence?
cc @wenytang-ms

jayzhang: No, because FxCore already has a middleware to replace this deleted one.

KennethBWSong (approved):

xzf0587 (approved):

packages/fx-core/src/component/resource/aadApp/aadApp.ts
jayzhang: I will make up for the telemetry by addStartAndEndTelemetry(actionName, actionName) middleware in another PR.

hund030 (approved):

SLdragon (approved):

sffamily (approved):

PR #8809, opened 2023-05-26
fix: convert token fail return empty object

xiaolang124:
Resolve the issue where login using dev tunnel scope or metaos scope returns a token that cannot be encoded by base64. In this scenario, we will return empty object and for UX we will try to get appStudio token.

[Bug 21597038: [Bug Bash] [v5.0.0] Dev Tunnel fails to start when running from "Tasks: Run Task" - Boards (visualstudio.com)](https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/21597038)

### DIFF
--- a/packages/vscode-extension/src/commonlib/m365Login.ts
+++ b/packages/vscode-extension/src/commonlib/m365Login.ts
@@ -269,7 +269,7 @@ export class M365Login extends BasicLogin implements M365TokenProvider {
         false
       );
       if (tokenRes.isOk()) {
-        let tokenJson = ConvertTokenToJson(tokenRes.value);
+        const tokenJson = ConvertTokenToJson(tokenRes.value);
         // if token is empty, try to get token by app studio scope, normally this should only affect UX
         if (Object.keys(tokenJson).length === 0) {
           const appStudioRes = await M365Login.codeFlowInstance.getTokenByScopes(
@@ -277,11 +277,11 @@ export class M365Login extends BasicLogin implements M365TokenProvider {
             false
           );
           if (appStudioRes.isOk()) {
-            tokenJson = ConvertTokenToJson(appStudioRes.value);
+            const appStudioJson = ConvertTokenToJson(appStudioRes.value);
             return ok({
               status: signedIn,
               token: appStudioRes.value,
-              accountInfo: tokenJson as any,
+              accountInfo: appStudioJson as any,
             });
           }
         }
--- a/packages/vscode-extension/src/commonlib/m365Login.ts
+++ b/packages/vscode-extension/src/commonlib/m365Login.ts
@@ -269,7 +269,22 @@ export class M365Login extends BasicLogin implements M365TokenProvider {
         false
       );
       if (tokenRes.isOk()) {
-        const tokenJson = ConvertTokenToJson(tokenRes.value);
+        let tokenJson = ConvertTokenToJson(tokenRes.value);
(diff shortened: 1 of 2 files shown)


XiaofuHuang (approved):

kuojianlu (approved):

packages/vscode-extension/src/commonlib/m365Login.ts
```diff
@@ -270,6 +270,21 @@ export class M365Login extends BasicLogin implements M365TokenProvider {
       );
       if (tokenRes.isOk()) {
         const tokenJson = ConvertTokenToJson(tokenRes.value);
+        // if token is empty, try to get token by app studio scope, normally this should only affect UX
```
swatDong, line 273: What if always get app studio token for `getStatus` (remove the `tokenRequest` input)?

It's a potential issue if input scope if A but the returned token is for scope B.

swatDong: Or, just return empty `accountInfo` but the token matches input scope.

xiaolang124: Like `getTeamsAppInternalId` will try to get graph scope token, I think always returning appStudio token is not a safe way.

swatDong: So returning appStudio token for empty json is also unsafe. I prefer just returning empty tokenjson, but the token itself is still the original scope.

xiaolang124: For now, it's not a good idea to return the original token. UX is using this to get sideloading permission and user information.
https://github.com/OfficeDev/TeamsFx/blob/bcfc8de2efa4685b79a59fb424acc1836d7c8aa7/packages/vscode-extension/src/treeview/account/accountTreeViewProvider.ts#L78
image

swatDong: Who calls the `m365AccountStatusChangeHandler`? Is it possible to always pass appstudio token to `m365AccountStatusChangeHandler`?

UX fix can be a wrapper with hard-coding logic. But for the general `getStatus` function, keep consistent behavior.

swatDong: [Here](https://github.com/OfficeDev/TeamsFx/blob/dev/packages/vscode-extension/src/treeview/account/accountTreeViewProvider.ts#L36) passes appstudio scope, then why `m365AccountStatusChangeHandler` receives token for other scope?

xiaolang124: I mean if I change this to return origin scope token and appStudio account info.

swatDong: I see. So for `getStatus`, keep the behavior meaningful - return token and json always matching the input scope. Since `getStatus` is a public api, we cannot assume all callers need the appstudio fallback.

For UX, better to fix in `setStatusChangeMap` - currently it seems ignore the input `tokenRequest` when notifying status.

xiaolang124: I will have two more prs. One will focus on saving the state for notification, while the other will involve using basic scope for login and updating getToken to utilize request scope.

swatDong: And may roll the `getStatus` workaround back in the notify fix~

PR #8830, opened 2023-05-29
build: add annotations to SDK protected methods

huimiu:
- Add annotations to the protected methods in the SDK so that they can be recognized when the SDK reference is automatically generated on learn.microsoft site.
- Change all `@beta` annotations to `@public`.


tecton (dismissed):
Could you please also help to change other annotations in sdk-react package to `@public`?

huimiu:
> Could you please also help to change other annotations in sdk-react package to `@public`?

@tecton Sure, I have changed all `@beta` annotations to `@public`, please help review again, thanks.

tecton (approved):

PR #8863, opened 2023-05-31
build: remove simpleauth download

wenytang-ms:


.github/workflows/cd.yml
```diff
@@ -197,15 +197,6 @@ jobs:
...
           allowUpdates: true
 
-      - name: download simpleauth to fx-core
-        uses: nick-invision/retry@v2
-        with:
-          timeout_minutes: 10
-          max_attempts: 10
-          retry_on: error
-          command: |
-            bash ./.github/scripts/download-simpleauth.sh
```
Siglud, line 207: I don't know about the detail of the github actions, but. Does this script attempt to run a file that was just deleted?

wenytang-ms: V5 will never use simpleauth, so no need to download simpleauth.

Siglud (approved):

PR #8865, opened 2023-05-31
ci: workflow to report code cleanup status

blackchoey:
Add workflow to detect code cleanup status, all checks should pass when code cleanup finished:
- There's no unused variables, functions, classes in source code
- There's no unused exports in source code
- All files should have a code owner

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -66,7 +66,7 @@
 
 /packages/fx-core/src/common/deps-checker @qinezh @a1exwang @kimizhu @swatDong
 /packages/fx-core/tests/common/deps-checker @qinezh @a1exwang @kimizhu @swatDong
- 
+
 /packages/fx-core/src/component/resource/aadApp @KennethBWSong @adashen @SLdragon
 /packages/fx-core/tests/component/resource/aadApp @KennethBWSong @adashen @SLdragon
 
--- a/.github/workflows/code-cleanup-report.yml
+++ b/.github/workflows/code-cleanup-report.yml
@@ -1,0 +1,95 @@
+name: Code Cleanup Report
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 0 * * *'
+
+jobs:
+  unused-vars:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        package:
+        - fx-core
+        - server
+        - vscode-extension
+        - manifest
+
+    steps:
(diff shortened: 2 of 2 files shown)


.github/CODEOWNERS
```diff
@@ -66,7 +66,7 @@
 
 /packages/fx-core/src/common/deps-checker @qinezh @a1exwang @kimizhu @swatDong
 /packages/fx-core/tests/common/deps-checker @qinezh @a1exwang @kimizhu @swatDong
- 
+
```
blackchoey, line 69: There's an extra space in this line. The tool used to detect files with no codeowner can't handle this line correctly, so remove this extra space.

.github/workflows/code-cleanup-report.yml
```diff
@@ -0,0 +1,95 @@
...
+    - name: Checkout
+      uses: actions/checkout@v2
+
+    - run: |
+        npm install -g @snyk/github-codeowners
+
+    - name: Find files with no codeowners
+      run: |
+        files=$(github-codeowners audit -u -r packages/${{matrix.package}})
+
```
wenytang-ms, line 88: Do other files besides src and test also need code-owner, such as README.md, package.json, webpack.config.js ..... ?

blackchoey: I think we need, but can have exceptions per further discussion. Let's find them first.

.github/workflows/code-cleanup-report.yml
```diff
@@ -0,0 +1,95 @@
...
+  unused-vars:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        package:
+        - fx-core
+        - server
+        - vscode-extension
+        - manifest
```
wenytang-ms, line 18: why select these 4 packages? why not contains cli, api, sdk or others?

blackchoey: I missed CLI and API, will add them.

The checks in this pipeline aim to detect missing part during code cleanup. For example, I deleted file A, but forgot to delete file B (or part of file B) which is only referenced by file A. For other packages, the case will not happen since we're not going to delete code in them. So excluded them to avoid distraction.
But I will check code owners for the whole repo (e.g. the templates). For same reason of avoiding distraction, the pipeline will only report code owner issues for selected packages.

wenytang-ms (approved):

adashen (approved):

PR #8907, opened 2023-06-02
perf(telemetry): load v4 project id if v5 project id is not available

KennethBWSong:


packages/fx-core/src/component/utils/settingsUtil.ts
```diff
@@ -12,7 +12,10 @@ import {
   TelemetryEvent,
   TelemetryProperty,
 } from "../../common/telemetry";
-import { getProjectSettingPathV3 } from "../../core/middleware/projectSettingsLoader";
+import {
+  getProjectSettingPathV2,
+  getProjectSettingPathV3,
+} from "../../core/middleware/projectSettingsLoader";
```
xzf0587, line 18: minor: this update of import is not be used.

xzf0587 (approved):

packages/fx-core/src/component/utils/settingsUtil.ts
KennethBWSong: updated, thx!

PR #8917, opened 2023-06-05
fix: remove unused ignore line

Siglud:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/22682022


templates/js/command-and-response/.appserviceIgnore
```diff
@@ -17,10 +17,8 @@ tsconfig.json
 teamsapp.yml
 teamsapp.*.yml
 /env/
-/script/
 /node_modules/.bin
 /node_modules/ts-node
```
yukun-dong, line 21: js templates don't have ts-node and typescript, should they be removed too?

swatDong (approved):

eriolchan (approved):

eriolchan (approved):

templates/js/command-and-response/.appserviceIgnore
Siglud: The main reason for removing these lines is to prevent conflicts and ambiguities with user-defined folders. For example, if a folder named "script" is inadvertently ignored when it needs to be uploaded, it will cause problems. On the other hand, folders like /node_modules/typescript are very targeted and rarely ambiguous in most cases.

PR #8919, opened 2023-06-05
fix(upgrade): sleep 2s before reloading after upgade to send telemetry

kuojianlu:
To fix https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/24133861.

### DIFF
--- a/packages/vscode-extension/src/debug/commonUtils.ts
+++ b/packages/vscode-extension/src/debug/commonUtils.ts
@@ -27,6 +27,7 @@ import { localize } from "../utils/localizeUtils";
 import * as util from "util";
 import { VS_CODE_UI } from "../extension";
 import * as vscode from "vscode";
+import { sleep } from "../utils/commonUtils";
 
 export async function getProjectRoot(
   folderPath: string,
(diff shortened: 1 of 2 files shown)


packages/vscode-extension/src/debug/commonUtils.ts
```diff
@@ -352,5 +353,7 @@ export async function triggerV3Migration(): Promise<void> {
     throw result.error;
   }
   // reload window to terminate debugging
+  // wait 2 seconds for the migration telemetry to be sent
+  await sleep(2000);
```
swatDong, line 357: Is 2s an estimated value or aligned with other upgrade process?
There's another 1s in `VS_CODE_UI.reload()`...

swatDong: BTW, is this a bug in `VS_CODE_UI.reload()` since the comment in `reload()` says it fixes the telemetry lost issue...

swatDong (approved):

tecton (approved):

xzf0587 (approved):

PR #8936, opened 2023-06-06
refactor: cleanup collaboration, add sso in vsc

KennethBWSong:


packages/vscode-extension/src/handlers.ts
```diff
@@ -913,73 +913,6 @@ function isLoginFailureError(error: FxError): boolean {
   return !!error.message && error.message.includes("Cannot get user login information");
 }
 
-function showWarningMessageWithProvisionButton(message: string): void {
-  window
-    .showWarningMessage(message, localize("teamstoolkit.handlers.provisionResourcesButton"))
```
tecton, line 918: If string is not used any more, could you please also remove it in `package.nls.json`? Same as other localized strings.

KennethBWSong: Updated, thx!

tecton (approved):

sffamily (approved):

PR #8969, opened 2023-06-08
ci: update template list that need to update deps

hund030:


templates/package.json
```diff
@@ -9,37 +9,34 @@
...
+        "ts/default-bot",
+        "ts/default-bot-message-extension",
         "ts/m365-tab",
-        "ts/sso-tab",
+        "ts/message-extension",
         "ts/non-sso-tab",
-        "ts/dashboard-tab",
         "ts/non-sso-tab-default-bot",
-        "ts/default-bot-message-extension",
-        "csharp/notification-http-timer-trigger",
```
eriolchan, line 35: what's the reason we remove csharp template list in this file?

PR #9089, opened 2023-06-21
perf(template): add js template for obo

KennethBWSong:


templates/js/sso-tab-with-obo-flow/env/.env.dev
```diff
@@ -0,0 +1,24 @@
+# This file includes environment variables that will be committed to git by default.
```
hund030, line 1: I think we should include .env.local and .env.*.user in the templates. Could you also update the ts template?

KennethBWSong: Updated, thx!

hund030 (approved):

SLdragon (approved):

PR #9125, opened 2023-06-26
refactor: add error stack to action telemetries

blackchoey:
Add error stack to help better troubleshoot system errors in our code
Also update vscode telemetry reporter to remove file path per suggestion.

### DIFF
--- a/packages/vscode-extension/src/commonlib/telemetry.ts
+++ b/packages/vscode-extension/src/commonlib/telemetry.ts
@@ -106,6 +106,12 @@ export class VSCodeTelemetryReporter extends vscode.Disposable implements Teleme
       );
     }
 
+    if (TelemetryProperty.ErrorStack in properties) {
+      properties[TelemetryProperty.ErrorStack] = anonymizeFilePaths(
+        properties[TelemetryProperty.ErrorStack]
+      );
+    }
+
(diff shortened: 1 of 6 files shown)


xzf0587 (approved):

chagong (approved):

KennethBWSong (approved):

swatDong (approved):

xzf0587 (approved):

packages/vscode-extension/src/commonlib/telemetry.ts
```diff
@@ -106,6 +106,12 @@ export class VSCodeTelemetryReporter extends vscode.Disposable implements Teleme
       );
     }
 
+    if (TelemetryProperty.ErrorStack in properties) {
+      properties[TelemetryProperty.ErrorStack] = anonymizeFilePaths(
```
tecton, line 110: One idea that can be improved later: Can we keep the filename and remove the file path? We have error stack in some error messages but all file paths are redacted. E.g. `some user stack trace at C:/fake_path/fake_file:1:1` can be anonymized to `some user stack trace at fake_file:1:1`. Since all these files are sourced from our extension, it's not related to user file and can help us better debug.

tecton (approved):

PR #9149, opened 2023-06-27
draft: add keytar to vscode extension

xiaolang124:
Due to the removal of keytar in VS Code in October, we need to add keytar for compatibility with Teamsfx/cli. If we change keytar to SecretStorage API, we cannot share m365 account in VS Code extension and CLI.

https://github.com/OfficeDev/TeamsFx/issues/9078


packages/vscode-extension/package.json
```diff
@@ -1299,6 +1299,9 @@
     "validator": "^13.7.0",
     "vscode-tas-client": "^0.1.47"
   },
+  "optionalDependencies": {
+    "keytar": "^7.7.0"
```
swatDong, line 1303: keytar is platform-dependent and our visx is built on linux. Does this work on all platforms?

The issue says *(not recommended) You can bundle the [keytar](https://github.com/atom/node-keytar) module with your extension. Keep in mind that keytar is a native node module which means that you will need to [publish a platform specific extension](https://code.visualstudio.com/api/working-with-extensions/publishing-extension#platformspecific-extensions) for each platform you want to support.*

xiaolang124: I'm going to test this branch on Windows, WSL, and macOS. Additionally, I will test it on Linux as well. The results will determine whether we can use this approach to handle keytar.

swatDong: Hope it works... but I remember the first time adding account encryption failed due to cross-platform issue (https://github.com/OfficeDev/TeamsFx/pull/1507). That's why we used vscode's built-in keytar...

For CLI, it works because users run `npm install` from their own machines, so keytar is installed at that time.

xiaolang124: Wenyu has helped build a VSIX in Linux, and I have verified it on Windows, WSL, and macOS. It works fine, so I think it may work in this way.

xiaolang124: We need to bundle keytar in every platform we need to support.

If a Linux user installs a VSIX package that is packaged in Windows, the Teams Toolkit extension will hang during the activation step. The same applies to a Windows user installing a VSIX package that is packaged in Linux.
image

xiaolang124: Change `webpack.exports.target = "electron-main"` may work. I will test it more.

PR #9162, opened 2023-06-28
feat: question to select a local file or enter value

yuqizhou77:
- add a question type allowing user to select a file or input a value for the file. 
image
image

- add VSC implementation. Confirmed with designer that it is recommended to have two steps rather than combine inputting a value and picking an option together.

[Task 24305677](https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/24305677): Question to select a file and enter Url

### DIFF
--- a/packages/api/src/qm/question.ts
+++ b/packages/api/src/qm/question.ts
@@ -314,7 +314,7 @@ export interface FuncQuestion extends BaseQuestion {
   func: LocalFunc<any>;
 }
 
-export interface SelectFileOrInputQuestion extends UserInputQuestion {
+export interface SingleFileOrInputQuestion extends UserInputQuestion {
   type: "singleFileOrText";
   inputOptionItem: OptionItem;
   inputBoxConfig: InputTextConfig;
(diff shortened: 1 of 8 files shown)


packages/api/src/qm/question.ts
```diff
@@ -97,7 +98,14 @@ export interface UserInputQuestion extends BaseQuestion {
...
    */
-  type: "singleSelect" | "multiSelect" | "singleFile" | "multiFile" | "folder" | "text";
+  type:
+    | "singleSelect"
+    | "multiSelect"
+    | "singleFile"
+    | "multiFile"
+    | "folder"
+    | "text"
+    | "selectLocalFileOrInput";
```
jayzhang, line 108: "singleSelectOrText" seems to be more consistant?

yuqizhou77: guess you mean "singleFileOrText"?

packages/vscode-extension/src/qm/vsc_ui.ts
```diff
@@ -725,6 +726,34 @@ export class VsCodeUI implements UserInteraction {
     });
   }
 
+  async singleFileOrInput(
+    config: singleFileOrInputConfig
+  ): Promise<Result<InputResult<string>, FxError>> {
+    const selectFileConfig: SelectFileConfig = {
+      possibleFiles: [config.inputOptionItem],
```
tecton, line 733: minor: this line can be switched with the one below to prevent overwrite.

tecton (approved):

jayzhang (approved):

tecton (approved):

PR #9172, opened 2023-06-29
build: remove useless dependencies

KennethBWSong:


templates/ts/sso-tab-with-obo-flow/package.json.tpl
```diff
@@ -9,8 +9,6 @@
     "@fluentui/react-components": "^9.18.0",
     "@microsoft/mgt-element": "^2.11.1",
     "@microsoft/mgt-react": "^2.11.1",
-    "@microsoft/mgt-teamsfx-provider": "^2.11.1",
```
SLdragon, line 12: Below is also not graph toolkit dependency, you should remove it if not used
```
"@microsoft/mgt-element": "^2.11.1",
"@microsoft/mgt-react": "^2.11.1",
```

KennethBWSong: Tested and they can be safely removed. Thanks!

SLdragon (approved):

PR #9184, opened 2023-06-30
build: merge main back to dev after publish

Siglud:


sffamily (approved):
I only checked the default English version of package.nls.json.

PR #9211, opened 2023-07-04
refactor: use secret storage to replace keytar in vscode extension

xiaolang124:
1.Use secret storage API in VS Code extension login.

### DIFF
--- a/packages/vscode-extension/src/commonlib/cacheAccess.ts
+++ b/packages/vscode-extension/src/commonlib/cacheAccess.ts
@@ -8,11 +8,11 @@ import { TokenCacheContext } from "@azure/msal-node";
 import { ConfigFolderName } from "@microsoft/teamsfx-api";
 import * as crypto from "crypto";
 import * as fs from "fs-extra";
-import * as keytarType from "keytar";
 import VsCodeLogInstance from "./log";
 import * as os from "os";
-import { env } from "vscode";
 import { localize } from "../utils/localizeUtils";
+import { context } from "../globalVariables";
+import { SecretStorage } from "vscode";
 
 const cacheDir = os.homedir + `/.${ConfigFolderName}/account`;
 const cachePath = os.homedir + `/.${ConfigFolderName}/account/token.cache.`;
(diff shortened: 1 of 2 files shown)


packages/vscode-extension/src/commonlib/cacheAccess.ts
```diff
@@ -8,11 +8,11 @@ import { TokenCacheContext } from "@azure/msal-node";
 import { ConfigFolderName } from "@microsoft/teamsfx-api";
 import * as crypto from "crypto";
 import * as fs from "fs-extra";
-import * as keytarType from "keytar";
 import VsCodeLogInstance from "./log";
 import * as os from "os";
-import { env } from "vscode";
 import { localize } from "../utils/localizeUtils";
+import { context } from "../globalVariables";
+import { SecretStorage } from "vscode";
```
swatDong, line 15: Any version requirement for vscode? E.g., the `engines:vscode` value (https://github.com/OfficeDev/TeamsFx/blob/dev/packages/vscode-extension/package.json#L24) in *package.json*?

packages/vscode-extension/src/commonlib/common/constant.ts
```diff
@@ -12,7 +12,7 @@ export const loggingIn = "LoggingIn";
 export const switching = "Switching";
 export const initializing = "Initializing";
 
-export const m365CacheName = "appStudio";
```
swatDong, line 15: Besides the vsc change, if user installs old ttk in new vsc, will that break CLI login?

packages/vscode-extension/src/commonlib/cacheAccess.ts
xiaolang124: Thanks for the reminder. I noticed that the 'secretStorage' API was added around 2020, and since VS Code 1.66.0 was released in 2022, we may not need to add an additional version check.

swatDong:
To `dev` or `main` branch?

packages/vscode-extension/src/commonlib/common/constant.ts
xiaolang124: After deleting Keytar in VS Code, users who are using the old VS Code extension and CLI may encounter these scenarios:
1. Login in the CLI -> CLI commands work fine -> User need to login in the VS Code extension -> if account is the same -> all work fine in VS Code extension and CLI

2. Login in the VS Code extension -> VS Code extension works fine -> User needs to login in the CLI -> if account is the same -> all work fine in the VS Code extension and the CLI

3. Login in the CLI -> CLI commands work fine -> User need to login in the VS Code extension -> if account is different -> CLI login info will be clear out, the user needs to login in the CLI again

4. Login in the VS Code extension -> VS Code extension works fine -> User needs to login in the CLI -> if account is different -> VS Code extension login info will be clear out after reloading, the user needs to login in CLI again (If the user does not reload VS Code, all teamsfx commands in VS Code extension work fine)

For users who are using the new VS Code extension and CLI.
Everything works fine because the account and token cache are the same in different files.

xiaolang124:
> To `dev` or `main` branch?

May cherry pick to main later or directly merge to main.

swatDong (approved):

xiaolang124:
I will merge this into the 'dev' branch as VS Code has pushed back the 'remove keytar' plan.

PR #9232, opened 2023-07-06
fix: adaptive cards code lens fit all folder

Siglud:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_sprints/taskboard/Compute%20Team/Microsoft%20Teams%20Extensibility/CY23-6.2?workitem=21741056

### DIFF
--- a/packages/vscode-extension/src/extension.ts
+++ b/packages/vscode-extension/src/extension.ts
@@ -6,7 +6,6 @@
 import * as vscode from "vscode";
 
 import {
-  AdaptiveCardsFolderName,
   AppPackageFolderName,
   BuildFolderName,
   ConfigFolderName,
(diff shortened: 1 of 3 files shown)


eriolchan (approved):

packages/vscode-extension/src/extension.ts
```diff
@@ -678,7 +677,7 @@ function registerCodelensAndHoverProviders(context: vscode.ExtensionContext) {
   };
 
   const adaptiveCardCodeLensProvider = new AdaptiveCardCodeLensProvider();
-  const adaptiveCardFilePattern = `**/${AdaptiveCardsFolderName}/*.json`;
+  const adaptiveCardFilePattern = `**/*.json`;
```
tecton, line 680: minor: it might search `node_modules` folder.

tecton (approved):

packages/vscode-extension/src/extension.ts
Siglud: This code lens will only be active when the user opens the file. So we do not need more restrictions on the folder's name.

PR #9285, opened 2023-07-13
perf: add validate function in spec parser library

SLdragon:


nliu-ms (dismissed):

packages/fx-core/src/common/spec-parser/specParser.ts
```diff
@@ -72,7 +138,8 @@ export class SpecParser {
    */
   async list(): Promise<string[]> {
     try {
-      const apiMap = await this.getAllSupportedApi(this.specPath);
+      const spec = (await SwaggerParser.validate(this.specPath)) as OpenAPIV3.Document;
```
yuqizhou77, line 141: If the path provided by the user is a remote URL, then we will do multiple GET request in scaffolding when loading operations for the user. During loading, we will call validate() first and then list(), which in total will lead to two calls. If the path parsed from OpenAI plugin manifest, then we will do three calls in total.

SLdragon: If we do not want to multiple requests, we can cache the result inside the class, I can update the code to support the cache

packages/fx-core/src/common/spec-parser/specParser.ts
```diff
@@ -100,22 +167,37 @@ export class SpecParser {
     // TODO: implementation
   }
 
+  private getAllRemoteRefs(spec: any, root = ""): Map<string, string> {
```
yuqizhou77, line 170: Looks like there is a paths() method (https://apitools.dev/swagger-parser/docs/swagger-parser.html#refs) provided by SwaggerParser. Just wondering if that would help so we don't have to handle this by ourselves.

SLdragon: Thanks, I will take a look

packages/fx-core/src/common/spec-parser/constants.ts
```diff
@@ -4,6 +4,14 @@
...
 export class ConstantString {
   static readonly CancelledMessage = "Operation cancelled";
+  static readonly SpecVersionNotSupported = "OpenAPI version lower than 3.0.0 is not supported";
+  static readonly NoServerInformation =
+    "No server information found in the OpenAPI specification file";
+  static readonly MultipleServerInformation =
+    "Multiple server information found in the OpenAPI specification file";
+  static readonly RemoteRefNotSupported = "Remote reference is not supported: %s";
+  static readonly MissingOperationId = "Missing operationId: %s";
+  static readonly NoSupportedApi = "No supported API found in the OpenAPI specification file";
```
Alive-Fish, line 14: Should we localize these strings?

SLdragon: Yeah, this is a good question and I want to discuss it in the next meeting, and if it needs localization, I can change the code in the future PR

yuqizhou77 (approved):

PR #9288, opened 2023-07-13
ci: update cd for api2teams

KennethBWSong:


packages/api2teams/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@microsoft/api2teams",
-  "version": "0.0.1",
+  "version": "0.1.0",
```
SLdragon, line 3: do we need to use version 0.1.0-rc

KennethBWSong: rc should be added in npm version step in cd @wenytang-ms

SLdragon (approved):

wenytang-ms (approved):

packages/api2teams/package.json
KennethBWSong: @SLdragon As sync with @wenytang-ms , we need to add -rc to publish 0.1.0-rc.0

wenytang-ms (approved):

PR #9314, opened 2023-07-17
ci: onboard template maintenance utils

hund030:


templates/scripts/yamlSolver.js
```diff
@@ -0,0 +1,131 @@
...
+  VERIFY: "verify",
+};
+
+// The constraints are defined in mustache files.
+// default constraints folders
+const mustacheFolder = path.resolve(__dirname, "..", "constraints", "yml", "templates");
+const solutionFolder = path.resolve(__dirname, "..");
+
+function strToObj(str) {
+  var properties = str.split(",");
```
Siglud, line 22: Maybe use let or const will be better?

templates/scripts/yamlSolver.js
```diff
@@ -0,0 +1,131 @@
...
+      path.dirname(path.relative(mustacheFolder, mustachePath)),
+      path.basename(mustachePath, ".mustache") + ".yml.tpl"
+    );
+    switch (action) {
+      case Action.APPLY:
+        fs.writeFileSync(solutionPath, solution);
+        break;
+      case Action.VERIFY:
+        const expected = fs.readFileSync(solutionPath, "utf8");
+        const assertion = solution.replaceAll("\r\n", "\n") === expected.replaceAll("\r\n", "\n");
```
Siglud, line 116: Run `npm run verify` will throw an error here.  The `expected` is a Buffer not string.

hund030: I think by specifying the encoding `utf8`, the return value of `fs.readFileSync` should be string. Do you mean you get an error here?

templates/scripts/yamlSolver.js
hund030: Updated. Thanks~

templates/scripts/yamlSolver.js
hund030: The `replaceAll` method is only available in node@16+, I use regex instead to support node 14.

templates/constraints/yml/snippets/azureAppServiceZipDeploy.mustache
```diff
@@ -0,0 +1,16 @@
...
+  # For additional details, refer to https://aka.ms/zip-deploy-to-app-services.
+  - uses: azureAppService/zipDeploy
+    with:
+    {{#workingDirectory}}
+      workingDirectory: {{workingDirectory}}
+    {{/workingDirectory}}
+      # Deploy base folder
+      artifactFolder: .
+      # Ignore file location, leave blank will ignore nothing
+      ignoreFile: .webappignore
```
eriolchan, line 11: Do we use webappignore or appserviceignore?

templates/constraints/yml/snippets/devToolInstall.mustache
```diff
@@ -0,0 +1,28 @@
+  # Install development tool(s)
+  - uses: devTool/install
+    with:
+{{#devCert}}
```
eriolchan, line 4: Why not combine the content in one place for {{#xxx}?

templates/scripts/utils.js
```diff
@@ -0,0 +1,51 @@
+const path = require("path");
+const { readdirSync, lstatSync, existsSync } = require("node:fs");
+const Mustache = require("mustache");
+
+function filterFiles(dir, fileList = [], filter = () => true) {
```
eriolchan, line 5: Do we need to add unit tests to make sure core functionality/utility works fine?

templates/scripts/yamlSolver.js
```diff
@@ -0,0 +1,132 @@
...
+            return utils.renderMustache(mustache, strToObj(text)).trimEnd();
+          };
+        },
+      },
+    };
+  });
+  return result;
+}
+
+function* solveMustache(mustachePaths) {
```
eriolchan, line 59: What's the special purpose of adding * after function key word?

templates/scripts/yamlSolver.js
```diff
@@ -0,0 +1,132 @@
...
+  };
+}
+
+function main({ action, mustachePaths }) {
+  let SAT = Action.VERIFY === action;
+  for (const { mustachePath, solution } of solveMustache(mustachePaths)) {
+    const solutionPath = path.resolve(
+      solutionFolder,
+      path.dirname(path.relative(mustacheFolder, mustachePath)),
+      path.basename(mustachePath, ".mustache") + ".yml.tpl"
```
eriolchan, line 109: Maybe we can put some string to constant in case there is error prone in string value across files.

templates/scripts/yamlSolver.js
```diff
@@ -0,0 +1,132 @@
...
+    );
+    switch (action) {
+      case Action.APPLY:
+        writeFileSync(solutionPath, solution);
+        break;
+      case Action.VERIFY:
```
eriolchan, line 115: For the verification, it ends when the first violation happens, or it will list all violations in the file?

hund030: It will list all violations.

templates/constraints/yml/snippets/azureAppServiceZipDeploy.mustache
hund030: I think the correct one is appserviceignore. I will confirm it and update it in another PR.

templates/constraints/yml/snippets/devToolInstall.mustache
hund030: Do you mean something like the following?
```
{{#devCert}}
  # Install development tool(s)
  - uses: devTool/install
    with:
      devCert:
        trust: true
{{/devCert}}
{{#funcTool}}
  # Install development tool(s)
  - uses: devTool/install
    with:
      func:
        version: ~4.0.4670
        symlinkDir: ./devTools/func
{{/funcTool}}
```
There are apps that need both devCert and funcTool, We want they can be installed in one action.

templates/scripts/utils.js
hund030: We did not add unit test for any script that work for CICD. Since We plan to run verify in CI, I think CI ensures the process works fine overall.

templates/scripts/yamlSolver.js
hund030: function* declares it returns an iterator. I may refactor this function to make the code style more consistent later.

templates/scripts/yamlSolver.js
hund030: Sure, let me make this refactor in next PR.

eriolchan (approved):

eriolchan (approved):

PR #9333, opened 2023-07-19
perf: generate new spec for copilot plugin

SLdragon:


nliu-ms (approved):

packages/fx-core/src/common/spec-parser/specFilter.ts
```diff
@@ -0,0 +1,59 @@
...
+        delete (newPaths[path] as any)[m];
+      }
+    }
+
+    if (supportedMethodNames.includes(methodName)) {
+      (newPaths[path] as any)[methodName] = (unResolveSpec.paths[path] as any)[methodName];
+
+      // Add the operationId if missing
+      if (!(newPaths[path] as any)[methodName].operationId) {
+        (newPaths[path] as any)[methodName].operationId = `${methodName}${convertPathToCamelCase(
```
nliu-ms, line 36: should here be a whitespace between method name and path?

yuqizhou77 (approved):

packages/fx-core/src/common/spec-parser/specFilter.ts
SLdragon: No, it's used for operationId, should not contain whitespace

Alive-Fish (approved):

wenytang-ms (approved):

PR #9338, opened 2023-07-19
feat: support sample tag in yaml schema and send sample tag telemetry

chagong:
feat: support sample tag in yaml schema and send sample tag telemetry

image

https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/24585467
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/24585468

### DIFF
--- a/packages/fx-core/src/component/configManager/parser.ts
+++ b/packages/fx-core/src/component/configManager/parser.ts
@@ -39,6 +39,13 @@ function parseRawProjectModel(obj: Record<string, unknown>): Result<RawProjectMo
     return err(new YamlFieldMissingError("version"));
   }
 
+  if ("sampleTag" in obj) {
+    if (typeof obj["sampleTag"] !== "string") {
+      return err(new YamlFieldTypeError("sampleTag", "string"));
+    }
+    result.sampleTag = obj["sampleTag"] as string;
+  }
+
   for (const name of LifecycleNames) {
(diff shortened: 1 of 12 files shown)


1yefuwang1 (approved):

xzf0587 (approved):

hund030 (approved):

jayzhang (approved):

packages/fx-core/src/component/configManager/parser.ts
```diff
@@ -39,6 +39,13 @@ function parseRawProjectModel(obj: Record<string, unknown>): Result<RawProjectMo
     return err(new YamlFieldMissingError("version"));
   }
 
+  if ("sampleTag" in obj) {
+    if (typeof obj["sampleTag"] !== "string") {
+      return err(new YamlFieldTypeError("sampleTag", "string"));
```
blackchoey, line 44: @chagong Since sampleTag is optional and it doesn't block anything when the value is invalid, I personally recommend ignore this schema error instead of throw error.
cc @1yefuwang1

1yefuwang1: This is good observation which I missed. 
The validation error is dismissed in https://github.com/OfficeDev/TeamsFx/pull/9533

PR #9376, opened 2023-07-24
refactor: add `listening` to default running pattern

hund030:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/24571805/
image

### DIFF
--- a/packages/cli/src/cmds/preview/constants.ts
+++ b/packages/cli/src/cmds/preview/constants.ts
@@ -64,7 +64,7 @@ export const funcHostedBotStartPattern =
   /Worker process started and initialized|Host lock lease acquired by instance ID/g;
 export const funcHostedBotAzuritePattern = /successfully listening/g;
 export const gulpServePattern = /^.*Finished subtask 'reload'.*/g;
-export const defaultRunningPattern = /started|successfully|finished|crashed|failed/i;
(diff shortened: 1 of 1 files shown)


packages/cli/src/cmds/preview/constants.ts
```diff
@@ -64,7 +64,7 @@ export const funcHostedBotStartPattern =
   /Worker process started and initialized|Host lock lease acquired by instance ID/g;
 export const funcHostedBotAzuritePattern = /successfully listening/g;
 export const gulpServePattern = /^.*Finished subtask 'reload'.*/g;
-export const defaultRunningPattern = /started|successfully|finished|crashed|failed/i;
+export const defaultRunningPattern = /started|successfully|finished|crashed|failed|listening/i;
```
kuojianlu, line 67: LGTM. @swatDong Any concerns?

swatDong: The CLI source code looks good to me.
BTW, @hund030 - any template or sample's `tasks.json` needs update?

swatDong (approved):

packages/cli/src/cmds/preview/constants.ts
hund030: We added `listening` in tasks.json on the initial day. But the CLI preview was unintentionally omitted.

PR #9379, opened 2023-07-25
perf(specParser): save file with original format, only support get with 1 param without auth

SLdragon:
- Save new OpenAPI spec file based on its original format
- Update filter logic:
   - Only GET operation with 1 param without auth
   - Doesn't support array/object type
   - Only support parameter in query or path, and in header or cookie are not supported

### DIFF
--- a/packages/fx-core/src/common/spec-parser/specParser.ts
+++ b/packages/fx-core/src/common/spec-parser/specParser.ts
@@ -18,7 +18,7 @@ import { ConstantString } from "./constants";
 import jsyaml from "js-yaml";
 import fs from "fs-extra";
 import { specFilter } from "./specFilter";
-import { isSupportedApi, isYamlSpecFile } from "./utils";
+import { isSupportedApi } from "./utils";
 
 /**
  * A class that parses an OpenAPI specification file and provides methods to validate, list, and generate artifacts.
@@ -173,7 +173,7 @@ export class SpecParser {
     const newUnResolvedSpec = await specFilter(filter, this.unResolveSpec!);
     if (outputSpecPath) {
       let resultStr;
-      if (await isYamlSpecFile(this.specPath)) {
+      if (this.specPath.endsWith(".yaml")) {
         resultStr = jsyaml.dump(newUnResolvedSpec);
       } else {
         resultStr = JSON.stringify(newUnResolvedSpec, null, 2);
--- a/packages/fx-core/src/common/spec-parser/specParser.ts
+++ b/packages/fx-core/src/common/spec-parser/specParser.ts
@@ -18,6 +18,7 @@ import { ConstantString } from "./constants";
 import jsyaml from "js-yaml";
 import fs from "fs-extra";
 import { specFilter } from "./specFilter";
+import { isSupportedApi, isYamlSpecFile } from "./utils";
 
 /**
  * A class that parses an OpenAPI specification file and provides methods to validate, list, and generate artifacts.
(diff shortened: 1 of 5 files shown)


packages/fx-core/src/common/spec-parser/utils.ts
```diff
@@ -0,0 +1,52 @@
...
+// Licensed under the MIT license.
+"use strict";
+
+import axios from "axios";
+import fs from "fs-extra";
+import { ConstantString } from "./constants";
+import { OpenAPIV3 } from "openapi-types";
+
+export async function isYamlSpecFile(specPath: string): Promise<boolean> {
+  const isRemoteFile = specPath.startsWith("http:") || specPath.startsWith("https:");
```
adashen, line 11: Can we just depend on the extension name to identify yml or json?

packages/fx-core/src/common/spec-parser/utils.ts
```diff
@@ -0,0 +1,52 @@
...
+}
+
+export function isSupportedApi(method: string, path: string, spec: OpenAPIV3.Document): boolean {
+  debugger;
+  const pathObj = spec.paths[path];
+  method = method.toLocaleLowerCase();
+  if (pathObj) {
+    if (method === ConstantString.GetMethod && !pathObj[method]?.security) {
+      const operationObject = pathObj[method] as OpenAPIV3.OperationObject;
+      if (operationObject.parameters?.length === 1) {
```
nliu-ms, line 31: Is there the case for zero parameter? E.g. list all may not require param.

nliu-ms (dismissed):

packages/fx-core/src/common/spec-parser/utils.ts
SLdragon: For remote url, maybe there is no extension name. So just use extension name is not enough, it is better to parse it for accurate result.

packages/fx-core/src/common/spec-parser/specParser.ts
```diff
@@ -171,8 +172,13 @@ export class SpecParser {
     await this.loadSpec();
     const newUnResolvedSpec = await specFilter(filter, this.unResolveSpec!);
     if (outputSpecPath) {
-      const ymlString = jsyaml.dump(newUnResolvedSpec);
-      await fs.writeFile(outputSpecPath, ymlString);
+      let resultStr;
+      if (await isYamlSpecFile(this.specPath)) {
```
yuqizhou77, line 176: I was expecting this method was exposed so that we could determine the outputSpecPath (whether end with .json or .yml)

SLdragon: Sure.
As discussed, yuqi will use this function to determine the outputSpecPath's extension name, and pass it into generate function, so inside the generate function, check the extension name is enough.

packages/fx-core/src/common/spec-parser/utils.ts
SLdragon: image

Thanks, from the design doc, it said that only 1 paramter, however, I think your suggestions is correct, we should also support 0 parameter

Alive-Fish (dismissed):

packages/fx-core/src/common/spec-parser/utils.ts
SLdragon: I can optimize this function, only parse the file when the extension name cannot determine the format.

packages/fx-core/src/common/spec-parser/utils.ts
SLdragon: Updated

nliu-ms (approved):

yuqizhou77 (approved):

Alive-Fish (approved):

wenytang-ms (approved):

LongOddCode (approved):

adashen (approved):

PR #9424, opened 2023-07-28
refactor: update promise handle

xzf0587:


packages/api/src/utils/login.ts
```diff
@@ -231,9 +231,9 @@ export abstract class BasicLogin {
     return ok(true);
   }
 
-  async removeStatusChangeMap(name: string): Promise<Result<boolean, FxError>> {
+  removeStatusChangeMap(name: string): Promise<Result<boolean, FxError>> {
     this.statusChangeMap.delete(name);
-    return ok(true);
+    return Promise.resolve(ok(true));
```
blackchoey, line 236: What's the effort if we also update the return type from `Promise<Result<boolean, FxError>>` to `Result<boolean, FxError>`? Seems there's no need to return a promise here.

xzf0587: This method is in the M365TokenProvider interface.
There are concrete implementations in vscode/cli.
We need to change them to keep the method interface consistant.

SLdragon (approved):

yukun-dong (approved):

blackchoey (approved):

jayzhang (approved):

packages/fx-core/src/core/middleware/utils/v3MigrationUtils.ts
```diff
@@ -278,7 +278,7 @@ export async function readAndConvertUserdata(
 ): Promise<string> {
   let returnAnswer = "";
 
-  const userdataContent = fs.readFileSync(path.join(context.projectPath, filePath), "utf8");
+  const userdataContent = await fs.readFile(path.join(context.projectPath, filePath), "utf8");
```
frankqianms, line 281: The same question as Chaoyi's: why we need to make this change, they look the same

xzf0587: It is ok to update this function to sync function. 
If so, we need to handle the thrown out error of readFile which the caller of the function should update as well.
If we use await inside the function, the caller of the function has no need to change.

PR #9427, opened 2023-07-28
refactor(vsc): eslint checks

tecton:


packages/vscode-extension/src/handlers.ts
```diff
@@ -1389,12 +1387,11 @@ async function showLocalPreviewMessage() {
       openFolderCommand
     );
   }
-  vscode.window.showInformationMessage(message, localPreview).then((selection) => {
-    if (selection?.title === localize("teamstoolkit.handlers.localPreviewTitle")) {
-      ExtTelemetry.sendTelemetryEvent(TelemetryEvent.ClickLocalPreview);
-      selection.run();
-    }
-  });
+  const selection = await vscode.window.showInformationMessage(message, localPreview);
```
HuihuiWu-Microsoft, line 1390: Is it needed to await here? User may not click any of the buttons on notification window?

tecton: I think it has no problem here because there's no other logic after it and `showLocalPreviewMessage()` is not awaited. If we don't write `await` here, we still need to use `then` and `catch`.

tecton: Thanks, fixed related invocations.

HuihuiWu-Microsoft (dismissed):

PR #9430, opened 2023-07-28
feat(cli): cli new help UX （new, new template）

jayzhang:
feature link: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/24539544

New CLI -help UI (for "new", "new template" command):

image

**Note: this PR will not affect current CLI behavior, it will be enabled under feature flag in the future.**


packages/cli/src/commonlib/logger.ts
```diff
@@ -0,0 +1,96 @@
+// Copyright (c) Microsoft Corporation.
```
swatDong, line 1: Is this a renaming of *cli/src/commonlib/**log.ts***?

jayzhang: logger.ts is a refactored log provider which will replace log.ts in the future

swatDong: So are you going to do the refactor and own the new logger? If so, feel free to change the CODEOWNERS then no need to wait unnecessary reviewers each time~

Alive-Fish (approved):

Alive-Fish (approved):
Check if "npx teamsfx xxx" works.

swatDong (approved):

LongOddCode (approved):

sffamily (approved):

PR #9443, opened 2023-07-31
fix(core): remove dependency of base url in samples config

tecton:
Related bug: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/24610139


HuihuiWu-Microsoft:
Please also help remove the `baseUrl` in sample config of Teamsfx-samples repo.

HuihuiWu-Microsoft (dismissed):

jayzhang (approved):

PR #9455, opened 2023-08-01
refactor: log summary of warnings after scaffolding

yuqizhou77:
E2E: https://github.com/OfficeDev/TeamsFx/actions/runs/5722268165

image

PR #9475, opened 2023-08-02
perf(specParser): generate adaptive card support

SLdragon:
- Generate adaptive card support
- Add period at the end of the message: https://github.com/OfficeDev/TeamsFx/pull/9455#discussion_r1280316026
- Fix validate remote reference does not work

### DIFF
--- a/packages/fx-core/src/common/spec-parser/adaptiveCardGenerator.ts
+++ b/packages/fx-core/src/common/spec-parser/adaptiveCardGenerator.ts
@@ -20,7 +20,7 @@ export function generateAdaptiveCard(operationItem: OpenAPIV3.OperationObject): 
   if (cardBody.length === 0 && (json.examples || json.example)) {
     cardBody = [
       {
-        type: "TextBlock",
+        type: ConstantString.TextBlockType,
         text: "${jsonStringify($root)}",
         wrap: true,
       },
@@ -31,7 +31,7 @@ export function generateAdaptiveCard(operationItem: OpenAPIV3.OperationObject): 
   if (cardBody.length === 0) {
     cardBody = [
       {
-        type: "TextBlock",
+        type: ConstantString.TextBlockType,
         text: "success",
         wrap: true,
       },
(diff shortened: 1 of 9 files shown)


packages/fx-core/src/common/spec-parser/adaptiveCardGenerator.ts
```diff
@@ -0,0 +1,133 @@
...
+        text: "success",
+        wrap: true,
+      },
+    ];
+  }
+
+  const fullCard: AdaptiveCard = {
+    type: "AdaptiveCard",
+    $schema: "http://adaptivecards.io/schemas/adaptive-card.json",
+    version: "1.5",
```
adashen, line 44: better not hard code the version?

packages/fx-core/src/common/spec-parser/specParser.ts
```diff
@@ -192,13 +195,19 @@ export class SpecParser {
 
     await fs.writeJSON(manifestPath, updatedManifest, { spaces: 2 });
 
-    // TODO: other implementations
+    for (const url in newSpec.paths) {
+      const getOperation = newSpec.paths[url]?.get;
+      const card: AdaptiveCard = generateAdaptiveCard(getOperation!);
+      const fileName = path.join(adaptiveCardFolder, `${getOperation!.operationId!}.json`);
```
yuqizhou77, line 201: just want to confirm about this. Looks like you will assume this folder always exists before this method was called?

SLdragon: I am using fs.outputJSON, this will generate folder if it does not exist: https://github.com/jprichardson/node-fs-extra/blob/master/docs/outputJson.md#outputjsonfile-object-options-callback

packages/fx-core/src/common/spec-parser/adaptiveCardGenerator.ts
SLdragon: Sure, will move it to constants

yuqizhou77 (approved):

nliu-ms (approved):

Alive-Fish (approved):

adashen (approved):

LongOddCode (approved):

PR #9482, opened 2023-08-02
feat: replace message extension with message extension action template

tecton:
Related task: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/24719427


packages/fx-core/src/component/coordinator/index.ts
```diff
@@ -81,7 +81,7 @@ export enum TemplateNames {
   CommandAndResponse = "command-and-response",
   Workflow = "workflow",
   DefaultBot = "default-bot",
-  MessageExtension = "message-extension",
+  MessageExtension = "message-extension-action",
```
yukun-dong, line 84: How about renaming this variable to "MessageExtensionAction" to avoid confusion.

tecton: Sure, will fix.

packages/fx-core/src/component/coordinator/index.ts
```diff
@@ -107,7 +107,7 @@ const Feature2TemplateName: any = {
   [`${CapabilityOptions.commandBot().id}:undefined`]: TemplateNames.CommandAndResponse,
   [`${CapabilityOptions.workflowBot().id}:undefined`]: TemplateNames.Workflow,
   [`${CapabilityOptions.basicBot().id}:undefined`]: TemplateNames.DefaultBot,
-  [`${CapabilityOptions.me().id}:undefined`]: TemplateNames.MessageExtension,
+  [`${CapabilityOptions.me().id}:undefined`]: TemplateNames.MessageExtensionAction,
```
yukun-dong, line 110: Based on below, should here be `CapabilityOptions.collectFormMe().id` ?https://github.com/OfficeDev/TeamsFx/blob/dea6b9e8aae05e82dca0dedecfadd0fd22ea591f/packages/fx-core/src/question/create.ts#L391

I think the reason it still works is that `CapabilityOptions.collectFormMe()` and `CapabilityOptions.me()` share the same id.

tecton: Yes, I agree. Thanks.

yukun-dong (approved):

jayzhang (approved):

PR #9488, opened 2023-08-02
refactor: fix eslint error for bot sso

SLdragon:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/24669461


packages/sdk/src/conversation/sso/defaultBotSsoExecutionActivityHandler.browser.ts
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
+/* eslint-disable @typescript-eslint/require-await */
```
swatDong, line 4: Would directly removing all `async` be better? Since all methods are just throwing exception, no side-effect if removing `async`?

swatDong: Anyway, also no side-effect to disable the rule since bot is server-side sdk.

swatDong (approved):

KennethBWSong (dismissed):

packages/sdk/src/conversation/sso/defaultBotSsoExecutionActivityHandler.browser.ts
SLdragon: Yes, this file is running inside a browser environment (not supported) and is used to align the function signature.

KennethBWSong (approved):

swatDong (approved):

dooriya (approved):

XiaofuHuang (approved):

PR #9494, opened 2023-08-02
fix(core): fix question model traverse algorithm

jayzhang:
1.The order of question tree node condition visit should be in DFS order.
2.Fix validation logic when the value is undefined.
fix: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/24723979


nliu-ms (approved):

xzf0587 (approved):

packages/fx-core/src/ui/visitor.ts
```diff
@@ -285,6 +289,132 @@ const questionVisitor: QuestionTreeVisitor = async function (
     )
   );
 };
+// export async function traverse(
```
wenytang-ms, line 292: why not remove this?

jayzhang: keep the history

wenytang-ms (approved):

PR #9500, opened 2023-08-03
feat(api): improve api to support cli new feature

jayzhang:
feature link: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/24539544

sub PR of this: #9499

### DIFF
--- a/packages/api/src/cli.ts
+++ b/packages/api/src/cli.ts
@@ -1,0 +1,79 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { Result } from "neverthrow";
+import { FxError } from "./error";
+
+export type OptionValue = string | boolean | string[] | undefined;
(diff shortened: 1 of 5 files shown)


LongOddCode (approved):

packages/api/src/cli.ts
```diff
@@ -0,0 +1,79 @@
...
+import { Result } from "neverthrow";
+import { FxError } from "./error";
+
+export type OptionValue = string | boolean | string[] | undefined;
+export type CLIOptionType = "boolean" | "string" | "array";
+
+export interface CLICommand {
+  name: string;
+  fullName?: string;
+  version?: string;
```
Alive-Fish, line 13: What does `version` and `fullName` point to?

jayzhang: Version is the CLI command version, only root node need this
fullName is like: "teamsfx new template"

PR #9526, opened 2023-08-07
fix(vsc): activation sync tasks and background jobs

tecton:
Related bug: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/24568693


packages/vscode-extension/src/handlers.ts
```diff
@@ -306,7 +306,7 @@ export function addFileSystemWatcher(workspacePath: string) {
 }
 
 export async function refreshSPFxTreeOnFileChanged() {
```
HuihuiWu-Microsoft, line 308: Remove 'async' here?

tecton: Thanks, fixed

HuihuiWu-Microsoft (approved):

PR #9527, opened 2023-08-07
ci: add init functionality for template maintenance tool

hund030:


templates/package.json
```diff
@@ -9,6 +9,7 @@
         "verify": "node ./scripts/yamlSolver.js verify",
         "apply": "node ./scripts/yamlSolver.js apply",
         "watch": "node ./scripts/watch.js",
+        "init": "node ./scripts/yamlSolver init",
```
huimiu, line 12: Only one argument is passed here, but the YamlSolver class initialization should accept three arguments.

huimiu (approved):

templates/package.json
hund030: Yes. The rest arguments are expected to be specify by users.

PR #9528, opened 2023-08-07
refactor: update tests promise error

xzf0587:


packages/tests/src/commonlib/frontendValidator.ts
```diff
@@ -127,6 +127,7 @@ export class FrontendValidator {
     fs.access(indexPath, fs.constants.F_OK, (err) => {
       chai.assert.isNull(err);
     });
+    return Promise.resolve();
```
hund030, line 130: It appears that some functions are returning void while others are returning Promise.resolve(). What's the distinction between the two? Could we make them consistent?

hund030 (approved):

packages/tests/src/commonlib/frontendValidator.ts
xzf0587: It depends on the function declaration.
If the function declaration has the async keyword, we should return promise resolve, else return void.
There will be an eslint promise rule to check it.

wenytang-ms:
please paste the runner result of relative E2E and ui test cases

PR #9560, opened 2023-08-08
refactor: use app name entered during scaffolding for name field in manifest

yuqizhou77:
- no need to update name in manifest.json. We will use the appName entered by users during scaffolding.

### DIFF
--- a/packages/fx-core/src/common/spec-parser/interfaces.ts
+++ b/packages/fx-core/src/common/spec-parser/interfaces.ts
@@ -121,11 +121,6 @@ export interface PartialManifest {
   composeExtensions: ComposeExtension[];
 }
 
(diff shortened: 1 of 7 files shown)


Alive-Fish (dismissed):

packages/fx-core/src/common/spec-parser/interfaces.ts
```diff
@@ -117,7 +117,6 @@ export interface AdaptiveCard {
 }
 
 export interface PartialManifest {
-  name: Name;
```
SLdragon, line 120: if Name interface is not used, you can remove it.

SLdragon (approved):

Alive-Fish (approved):

PR #9596, opened 2023-08-11
feat: new log api to support verbose

jayzhang:
feat link: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/24539544

1. add new `verbose` API in LogProvider
2. remove unused `fatal` and `trace` API
3. change return type from `Promise<>` to `void`


packages/cli/src/commonlib/log.ts
```diff
@@ -46,50 +46,46 @@ export class CLILogProvider implements LogProvider {
     return path.join(os.tmpdir(), `.${ConfigFolderName}`, "cli-log", this.logFileName);
   }
 
-  trace(message: string): Promise<boolean> {
-    return this.log(LogLevel.Trace, message);
+  verbose(message: string): void {
```
swatDong, line 49: Is there more detailed reason on changing `Promise` to `void`?
IMO, `Promise` is more extensible than `void`, and changing `Promise` to `void` may have potential performance downgrade (e.g., async file operation to sync one)

packages/fx-core/tests/component/driver/arm/utils.test.ts
```diff
@@ -379,7 +379,7 @@ describe("getDeploymentError", () => {
     deploymentName: "mockDeployName",
     resourceGroupName: "mockRG",
     deploymentStartTime: Date.now(),
-  } as DeployContext;
+  } as any;
```
blackchoey, line 382: Can you share the reason behind the change?

jayzhang: I don't knwo the exact reason why the error emerges when I change LogProvider API. I found the original case is type wrong, I don't know why they can pass the type check.

swatDong (dismissed):

swatDong (approved):

Alive-Fish (approved):

HuihuiWu-Microsoft (approved):

hund030 (approved):

yuqizhou77 (approved):

nliu-ms (approved):

xzf0587 (approved):

LongOddCode (approved):

JerryYangKai (approved):

Siglud (approved):

PR #9672, opened 2023-08-21
perf(template): update sso template to obo

KennethBWSong:


templates/csharp/sso-tab/Components/Graph.razor
```diff
@@ -49,7 +52,18 @@
     private async Task ShowProfile()
     {
         IsLoading = true;
-        var graph = GetGraphServiceClient();
+
+        var config = Configuration.Get<ConfigOptions>();
+        var tenantId = config.TeamsFx.Authentication.OAuthAuthority.TrimStart("https://login.microsoftonline.com/".ToCharArray());
+        AccessToken ssoToken = await teamsUserCredential.GetTokenAsync(new TokenRequestContext(null), new CancellationToken());
+        var tokenCredential = new OnBehalfOfCredential(
```
blackchoey, line 59: Overall looks good to me. One only question is: if user does not consent MS Graph permission, and OBO failed, will the app correctly handle the exception and ask user login? Not sure what exception will OBOCredential throw, and whether we need to catch the new exception in the login logic.

templates/csharp/sso-tab/Components/Graph.razor
```diff
@@ -77,12 +80,13 @@
         IsLoading = true;
         try
         {
-            await teamsUserCredential.GetTokenAsync(new TokenRequestContext(new string[] { _scope }), new System.Threading.CancellationToken());
+            var graph = await GetGraphServiceClient();
+            Profile = await graph.Me.GetAsync();
```
blackchoey, line 84: can we just use OBOCredential to acquire a graph token instead of calling graph API here? This code works but may not easy to understand if user does not have context about what we're tring to do (get a graph token) here.

KennethBWSong: Updated. Now we are using masl.net to get graph token with specified scope.

blackchoey (approved):

templates/csharp/sso-tab/Components/Graph.razor
```diff
@@ -95,10 +101,24 @@
         IsLoading = false;
         return false;
     }
+
+    private async Task<OnBehalfOfCredential> GetTokenCredential()
+    {
+        var config = Configuration.Get<ConfigOptions>();
+        var tenantId = config.TeamsFx.Authentication.OAuthAuthority.TrimStart("https://login.microsoftonline.com/".ToCharArray());
+        AccessToken ssoToken = await teamsUserCredential.GetTokenAsync(new TokenRequestContext(null), new CancellationToken());
+        var tokenCredential = new OnBehalfOfCredential(
```
hund030, line 110: I prefer:
```
return new OnBehalfOfCredential(...);
```

templates/csharp/sso-tab/Components/Graph.razor
```diff
@@ -95,10 +101,24 @@
         IsLoading = false;
         return false;
     }
+
+    private async Task<OnBehalfOfCredential> GetTokenCredential()
```
hund030, line 105: `GetOnBehalfOfCredential` may be a better naming.

hund030 (approved):

PR #9695, opened 2023-08-23
perf(specParser): support partial generate ac, support empty schema

SLdragon:
1. support partial generate adaptive card
2. support empty schema
3. support response code 20X
4. filter api without json response
5. support required parameter <=1

### DIFF
--- a/packages/fx-core/src/common/spec-parser/manifestUpdater.ts
+++ b/packages/fx-core/src/common/spec-parser/manifestUpdater.ts
@@ -79,17 +79,17 @@ export async function generateCommands(
             operationItem.operationId! + ".json"
           );
 
-          if (await fs.pathExists(adaptiveCardPath)) {
-            const command: Command = {
-              context: ["compose"],
-              type: "query",
-              title: operationItem.summary ?? "",
-              id: operationItem.operationId!,
-              parameters: parameters,
-              apiResponseRenderingTemplate: getRelativePath(manifestPath, adaptiveCardPath),
-            };
-            commands.push(command);
-          }
+          const command: Command = {
+            context: ["compose"],
+            type: "query",
+            title: operationItem.summary ?? "",
+            id: operationItem.operationId!,
+            parameters: parameters,
+            apiResponseRenderingTemplate: (await fs.pathExists(adaptiveCardPath))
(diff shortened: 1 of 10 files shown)


adashen (approved):

packages/fx-core/src/common/spec-parser/manifestUpdater.ts
```diff
@@ -79,15 +79,17 @@ export function generateCommands(
...
-          const command: Command = {
-            context: ["compose"],
-            type: "query",
-            title: operationItem.summary ?? "",
-            id: operationItem.operationId!,
-            parameters: parameters,
-            apiResponseRenderingTemplate: getRelativePath(manifestPath, adaptiveCardPath),
-          };
-          commands.push(command);
+          if (await fs.pathExists(adaptiveCardPath)) {
```
yuqizhou77, line 82: Does this mean if we cannot generate card, then we will not include the command in the manifest?

SLdragon: Yes, when generate card failed, it would not generate adaptive card, then will not include the command in the manifest

yuqizhou77: I am thinking about which way would be better:
1. Generate command, but leave the field for card template path empty, so that user won't need to fill in other fields?
2. Keep the command and card path in manifest, and user have to generate a card template with the exact name
3. Skip generating command. User have to add the command manually and fill in each field?

SLdragon: I think skip generate command would be better, this will make the manifest workable without needing to update something. We can show warnings to user that some cards are failed to generate, and user can optionally fix it by himself, even he doesn't know how to fix it, the manifest still works for other api.

yuqizhou77 (approved):

packages/fx-core/src/common/spec-parser/manifestUpdater.ts
SLdragon: 1. As discussed, we plan to generate command with empty path if generate adaptive card failed.
2. Use operationId for the warning data

adashen (approved):

PR #9743, opened 2023-08-28
feat(core): enable feature flag when downloading samples

tecton:
Related task: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/24610127

Download design:
1. stable version uses bundled tag.
2. prerelease version uses bundled branch head for prerelease.
3. rc version uses bunded tag.
4. daily release uses 'dev' branch.
5. stable version uses environment variable _TEAMSFX_SAMPLE_CONFIG_BRANCH_ branch if available.

### DIFF
--- a/packages/fx-core/src/common/samples.ts
+++ b/packages/fx-core/src/common/samples.ts
@@ -2,15 +2,20 @@
 // Licensed under the MIT license.
 
 import axios from "axios";
+
 import { parseSampleUrl, sendRequestWithTimeout } from "../component/generator/utils";
-import sampleConfigV3 from "./samples-config-v3.json";
+import { FeatureFlagName } from "./constants";
 import { isVideoFilterEnabled } from "./featureFlags";
+import sampleConfigV3 from "./samples-config-v3.json";
+
 const packageJson = require("../../package.json");
 
 const SampleConfigOwner = "OfficeDev";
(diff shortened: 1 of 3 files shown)


packages/fx-core/src/common/samples.ts
```diff
@@ -37,34 +42,42 @@ interface SampleCollection {
...
-
-      if (fileResponse && fileResponse.data) {
-        this.samplesConfig = fileResponse.data as { samples: Array<Record<string, unknown>> };
+    const version: string = packageJson.version;
+    if (version.includes("alpha")) {
+      // daily build version always use 'dev' branch
+      this.branchOrTag = "dev";
+    } else if (version.includes("rc")) {
+      // rc version always use prerelease tag
+      this.branchOrTag = SampleConfigTagForPrerelease;
```
HuihuiWu-Microsoft, line 54: I suppose rc version is prerelease version before stable version, we should use v3 for testing?

tecton: This is designed for breaking changes in future. If we need to release sample changes now, I think we should test the samples using the stable version TTK and set feature flag to 'v3' instead of using prerelease TTK.

HuihuiWu-Microsoft (approved):

LongOddCode (approved):

PR #9761, opened 2023-08-29
feat: support preview copilot plugin from treeview

kuojianlu:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/23094134

### DIFF
--- a/packages/fx-core/src/common/m365/launchHelper.ts
+++ b/packages/fx-core/src/common/m365/launchHelper.ts
@@ -26,7 +26,8 @@ export class LaunchHelper {
     hub: HubTypes,
     teamsAppId: string,
     capabilities: string[],
-    withLoginHint = true
+    withLoginHint = true,
+    isCopilotPlugin = false
   ): Promise<Result<string, FxError>> {
     const loginHint = withLoginHint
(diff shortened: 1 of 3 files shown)


packages/fx-core/src/common/m365/launchHelper.ts
```diff
@@ -26,15 +26,27 @@ export class LaunchHelper {
...
+          isCopilotPlugin &&
+          !capabilities.includes("staticTab") &&
+          !capabilities.includes("configurableTab") &&
+          !capabilities.includes("Bot")
+        ) {
+          installAppPackage = false;
+        }
+        const baseUrl = installAppPackage
+          ? `https://teams.microsoft.com/l/app/${teamsAppId}?installAppPackage=true&webjoin=true`
+          : "https://teams.microsoft.com";
```
swatDong, line 49: webjoin?

kuojianlu: I tried `webjoin` but found that it worked for `l/app/${teamsAppId}?installAppPackage=true` but took no effect for `https://teams.microsoft.com`, i.e. the below window will not be poped out without `webjoin` or even with `webjoin=false`.
image

swatDong (approved):

jayzhang (approved):

PR #9763, opened 2023-08-29
perf(spec-parser): support post and update server validation

SLdragon:
1. Support post operation
2. Update server validation

### DIFF
--- a/packages/fx-core/src/common/spec-parser/manifestUpdater.ts
+++ b/packages/fx-core/src/common/spec-parser/manifestUpdater.ts
@@ -8,6 +8,7 @@ import fs from "fs-extra";
 import path from "path";
 import { getRelativePath, updateFirstLetter } from "./utils";
 import { SpecParserError } from "./specParserError";
+import { ConstantString } from "./constants";
 
 export async function updateManifest(
   manifestPath: string,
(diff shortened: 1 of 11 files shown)


packages/fx-core/src/common/spec-parser/manifestUpdater.ts
```diff
@@ -55,40 +88,51 @@ export async function generateCommands(
...
+                  parameters.push({
+                    name: param.name,
+                    title: updateFirstLetter(param.name),
+                    description: param.description ?? "",
+                  });
+                });
+              }
+
+              if (operationItem.requestBody) {
+                const requestJson = (operationItem.requestBody as OpenAPIV3.RequestBodyObject)
```
yuqizhou77, line 110: Do we check content to be "application/json" when validating? or any requirements about content type? Just trying to think about possible values when we run generate()

SLdragon: From Josh's validation rule, the request body should be "application/json", other type is not supported

yuqizhou77: so we will have plan to update our validate() then?

yuqizhou77 (dismissed):

packages/fx-core/src/common/spec-parser/manifestUpdater.ts
yuqizhou77: noticed that we've already have this in this PR. Resolved.

yuqizhou77 (approved):

adashen (approved):

PR #9793, opened 2023-08-31
perf(spec-parser): only allow single parameter to manifest for post

SLdragon:
1. Each post body property should be a parameter
2. Optional parameter should not map to manifest
3. Resolve server url environment variables

### DIFF
--- a/packages/fx-core/src/common/spec-parser/utils.ts
+++ b/packages/fx-core/src/common/spec-parser/utils.ts
@@ -66,10 +66,9 @@ export function checkRequiredParameters(
   return false;
 }
 
-export function isSupportedSchema(schema: OpenAPIV3.SchemaObject): boolean {
-  // we support schema: {}
+export function checkPostBodyRequiredParameters(schema: OpenAPIV3.SchemaObject): number {
   if (Object.keys(schema).length === 0) {
-    return true;
+    return 0;
   }
 
   if (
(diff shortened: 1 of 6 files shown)


packages/fx-core/src/common/spec-parser/utils.ts
```diff
@@ -78,20 +77,19 @@ export function isSupportedSchema(schema: OpenAPIV3.SchemaObject): boolean {
...
+      const result = checkPostBodyRequiredParameters(
+        properties[property] as OpenAPIV3.SchemaObject
+      );
+      count += result;
     }
-    return true;
+    return count;
   } else {
-    return false;
+    return NaN;
```
yuqizhou77, line 92: Is this for the case type:array?

SLdragon: yes, schema.array, schema.allOf, schema.oneOf, schema.not, and schema.anyOf

packages/fx-core/src/common/spec-parser/utils.ts
```diff
@@ -78,20 +77,19 @@ export function isSupportedSchema(schema: OpenAPIV3.SchemaObject): boolean {
...
+    let count = 0;
     for (const property in properties) {
-      const result = isSupportedSchema(properties[property] as OpenAPIV3.SchemaObject);
-      if (!result) {
-        return false;
-      }
+      const result = checkPostBodyRequiredParameters(
+        properties[property] as OpenAPIV3.SchemaObject
+      );
+      count += result;
```
yuqizhou77, line 88: thought there might cause exception? the value can be NaN returned?

SLdragon: NaN + any number would still NaN, so it would work

yuqizhou77: do we validate such cases in validate() today?

yuqizhou77: oh, realized we will filter out these APIs.

yuqizhou77 (dismissed):

adashen (approved):

PR #9834, opened 2023-09-04
perf(spec-parser): add warning and ignore api only contains optional param

SLdragon:


yuqizhou77 (dismissed):

packages/fx-core/src/common/spec-parser/constants.ts
```diff
@@ -30,6 +30,8 @@ export class ConstantString {
     "Server url is not correct: relative server url is not supported.";
   static readonly ResolveServerUrlFailed =
     "Resolve server Url failed: environment variable %s is not defined";
+  static readonly OperationOnlyContainsOptionalParam =
+    "Operation %s %s only contains optional parameters which is not supported";
```
yuqizhou77, line 34: better to add " or ' for %s

packages/fx-core/src/common/spec-parser/manifestUpdater.ts
```diff
@@ -124,6 +134,19 @@ export async function generateCommands(
...
 
+              if (parameters.length === 0) {
+                warnings.push({
+                  type: WarningType.OperationOnlyContainsOptionalParam,
+                  content: format(
+                    ConstantString.OperationOnlyContainsOptionalParam,
+                    method.toUpperCase(),
+                    pathUrl
+                  ),
+                  data: method.toUpperCase() + " " + pathUrl,
```
yuqizhou77, line 145: Can we have operationId here instead?

packages/fx-core/src/common/spec-parser/constants.ts
SLdragon: OK

packages/fx-core/src/common/spec-parser/manifestUpdater.ts
SLdragon: sure

yuqizhou77: or even both operationId and "Method + Path"

SLdragon: As discussed, will use operationId

adashen (dismissed):

yuqizhou77 (approved):

adashen (approved):

PR #9841, opened 2023-09-05
feat(cli): filter working samples from list

tecton:
Add property `maximumCliVersion` and `minimumCliVersion` for CLI.

Related task: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/25061407


packages/cli/src/utils.ts
```diff
@@ -200,7 +203,19 @@ export interface Sample {
 
 export async function getTemplates(): Promise<Sample[]> {
```
wenytang-ms, line 204: maybe the name should be like getSamples?

Alive-Fish (approved):

packages/cli/src/utils.ts
```diff
@@ -200,7 +203,19 @@ export interface Sample {
...
   await sampleProvider.fetchSampleConfig();
-  const samples = sampleProvider.SampleCollection.samples.map((sample) => {
+  const version = getVersion();
+  const availableSamples = sampleProvider.SampleCollection.samples.filter(
```
wenytang-ms, line 207: do we need consider alpha or beta, not sure gte or lte works fine with version with prerelease postfix.

jayzhang (approved):

HuihuiWu-Microsoft (approved):

packages/cli/src/utils.ts
tecton: Good point, I tested the `gte()` function with following versions, so I think it should work as expected.
* gte("2.1.0-rc.0", "2.1.0") is false
* gte("2.1.0-rc.0", "2.0.5") is true
* gte("2.1.0-alpha.0", "2.1.0") is false
* gte("2.1.0-alpha.0", "2.0.5") is true

packages/cli/src/utils.ts
tecton: It looks like the API name is related to CLI argument name "template-name", I think maybe @jayzhang could help to take a look and decide whether to rename it.

wenytang-ms (approved):

PR #9845, opened 2023-09-05
fix: show error when no extra API can be added

nliu-ms:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/25052215
image

### DIFF
--- a/packages/fx-core/src/component/generator/copilotPlugin/helper.ts
+++ b/packages/fx-core/src/component/generator/copilotPlugin/helper.ts
@@ -163,7 +163,7 @@ export async function listOperations(
         if (operations.length == 0) {
           const errors = [
             {
-              type: ApiSpecErrorType.NoSupportedApi,
+              type: ApiSpecErrorType.NoExtraAPICanBeAdded,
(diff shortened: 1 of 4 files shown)


packages/fx-core/src/component/generator/copilotPlugin/helper.ts
```diff
@@ -159,6 +160,12 @@ export async function listOperations(
         operations = operations.filter(
           (operation: string) => !existingOperations.includes(operation)
         );
+        // No extra API can be added
+        if (operations.length == 0) {
+          return err([
+            { type: ApiSpecErrorType.NoSupportedApi, content: ConstantString.NoSupportedApi },
```
yuqizhou77, line 166: The content of this stinrg here will likely be changed to later to something like  "No supported API found. Only GET and POST methods with single parameters and no auth are supported." Not sure whether it will impact you or not.

yuqizhou77 (dismissed):

sffamily (approved):

SLdragon (approved):

PR #9847, opened 2023-09-05
perf(spec-parser): add support for swagger 2.0

SLdragon:


SLdragon:
Some check failed due to Github service issue

packages/fx-core/src/common/spec-parser/specParser.ts
```diff
@@ -273,6 +268,13 @@ export class SpecParser {
   private async loadSpec(): Promise<void> {
     if (!this.spec) {
       this.unResolveSpec = (await this.parser.parse(this.specPath)) as OpenAPIV3.Document;
+      // Convert swagger 2.0 to openapi 3.0
+      if (!this.unResolveSpec.openapi && (this.unResolveSpec as any).swagger === "2.0") {
```
adashen, line 272: if lower then 2.0 do we have error message?

SLdragon: Yes, it will show invalid spec error, because current `@apidevtools/swagger-parser` library only support 2.0 and 3.x, when parse 1.0 spec, the library will throw error and the validate function will catch the error and show invalid spec error

packages/fx-core/src/common/spec-parser/specParser.ts
```diff
@@ -273,6 +268,13 @@ export class SpecParser {
...
+        const specObj = await converter.convert(this.unResolveSpec as any, {});
```
yuqizhou77, line 273: wondering which version will the spec file be converted to?

yuqizhou77 (approved):

adashen (approved):

LongOddCode (approved):

PR #9859, opened 2023-09-06
feat: add "testtool" env in core

a1exwang:
- separate old `environmentManager` into `environmentNameManager` (only contains constants) & `environmentManager` (do actual file ops) to prevent import loop
- unify all "local" hardcoding to use `environmentNameManager`
- accepts "testtool" as a non-remote environment which is similar to current "local" env

TODO: core telemetry / VSC / CLI


packages/cli/src/cmds/preview/previewEnv.ts
```diff
@@ -131,7 +132,7 @@ export default class PreviewEnv extends YargsCommand {
 
     cliTelemetry.withRootFolder(workspaceFolder);
     this.telemetryProperties[TelemetryProperty.PreviewType] =
-      env.toLowerCase() === environmentManager.getLocalEnvName() ? "local" : `remote-${env}`;
+      environmentNameManager.isRemoteEnvironment(env.toLowerCase()) ? `remote-${env}` : env;
```
swatDong, line 135: `env.toLowerCase()`, or will `env` contains upper case?

swatDong: Same for all telemetry changes

packages/fx-core/src/core/environmentName.ts
```diff
@@ -0,0 +1,39 @@
...
+  public getTestToolEnvName() {
+    return this.testToolEnvName;
+  }
+
+  public getNonRemoteEnvNames(): string[] {
+    return [this.localEnvName, this.testToolEnvName];
+  }
+
+  public isRemoteEnvironment(env: string) {
+    return this.getNonRemoteEnvNames().indexOf(env) === -1;
```
swatDong, line 35: To lower case? BTW, is env name case-sensitive?

packages/cli/src/cmds/preview/previewEnv.ts
a1exwang: Core doesn't treat env specially, it just rely on OS file system behavior.
https://github.com/OfficeDev/TeamsFx/blob/fef761eba616243b7cff601080407af6878702fe/packages/fx-core/src/component/utils/pathUtils.ts#L53
So on Windows it's case-insensitive. Probably case sentive on macOS if using case-sensitive file system.

But CLI preview treats env name as case-insensitive despite of OS.

Core and CLI are already inconsistent. 

The goal of this PR is not to solve this existing problem. Just be consistent with current impl (in core don't do conversion, in cli `toLowerCase()`)

swatDong: Agreed to be consistent with current impl.
But this line's change uses `env` instead of `"local"`. Is it possible to send "**LoCal**" (any case) as telemetry? Will that impact telemetry query?

swatDong (approved):

1yefuwang1 (approved):

Alive-Fish (approved):

HuihuiWu-Microsoft (approved):

tecton (approved):

xzf0587 (approved):

jayzhang (approved):

PR #9882, opened 2023-09-08
feat(vsc): add version compare result to each sample for different UI

tecton:
Task: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/25093360

UI of sample that doesn't work in current TTK will be added later.

### DIFF
--- a/packages/vscode-extension/src/controls/webviewPanel.ts
+++ b/packages/vscode-extension/src/controls/webviewPanel.ts
@@ -8,6 +8,7 @@ import * as vscode from "vscode";
 import { Inputs } from "@microsoft/teamsfx-api";
 import { Correlator, sampleProvider } from "@microsoft/teamsfx-core";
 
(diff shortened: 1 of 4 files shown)


packages/vscode-extension/src/controls/webviewPanel.ts
```diff
@@ -174,10 +177,31 @@ export class WebviewPanel {
 
   private async LoadSampleCollection() {
     await sampleProvider.fetchSampleConfig();
+    const sampleCollection = sampleProvider.SampleCollection;
```
HuihuiWu-Microsoft, line 180: What about the experience for CLI?

tecton: It's covered in this PR: #9841 
Samples not working in current CLI version are excluded.

HuihuiWu-Microsoft (approved):

PR #9885, opened 2023-09-08
feat: add v1.3 schema for test tool

a1exwang:
Add devTool/install argument schema for test tool 

image

Diff with version 1.2 for easier viewing:

image

https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/16357625

### DIFF
--- a/packages/fx-core/resource/yaml-schema/v1.3/yaml.schema.json
+++ b/packages/fx-core/resource/yaml-schema/v1.3/yaml.schema.json
@@ -1,0 +1,1500 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+    "projectId": {
+      "type": "string",
+      "description": "The projectId used for telemetry."
+    },
+    "environmentFolderPath": {
+      "type": "string",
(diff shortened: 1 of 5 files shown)


packages/fx-core/resource/yaml-schema/v1.3/yaml.schema.json
```diff
@@ -0,0 +1,1500 @@
...
+                "symlinkDir"
+              ],
+              "properties": {
+                "version": {
+                  "oneOf": [
+                    {
+                      "type": "string"
+                    },
+                    {
+                      "type": "integer"
```
1yefuwang1, line 1235: Do we need to support integer versions if it's semver?

a1exwang: by default, yaml will treat

```
version: 3
```

as an integer. This is to prevent user from making such mistakes.

a1exwang: Without type integer, user has to write
```
version: "3"
```

1yefuwang1 (approved):

xzf0587 (approved):

PR #9904, opened 2023-09-11
fix: add a param for the repair api

huimiu:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/25124615#


templates/ts/copilot-plugin-from-scratch/repair/index.ts
```diff
@@ -25,18 +25,33 @@ export default async function run(context: Context, req: HttpRequest): Promise<R
...
-    id: 1,
-    title: "Oil change",
-    description:
-      "Need to drain the old engine oil and replace it with fresh oil to keep the engine lubricated and running smoothly.",
-    assignedTo: "Karin Blair",
-    date: "2023-05-23",
-    image: "https://www.howmuchisit.org/wp-content/uploads/2011/01/oil-change.jpg",
-  };
+  // Get the assignedTo query parameter.
+  const assignedTo = req.query.assignedTo;
```
hund030, line 29: Could we use Fail-fast approach here? I think it makes the code simpler.
```
if (!assignedTo) {
  return res;
}
```

templates/csharp/copilot-plugin-from-scratch/Repair.cs.tpl
```diff
@@ -8,28 +8,36 @@ using Microsoft.Extensions.Logging;
...
+                    AssignedTo = "Karin Blair",
+                    Date = "2023-05-23",
+                    Image = "https://www.howmuchisit.org/wp-content/uploads/2011/01/oil-change.jpg"
+                }
             };
 
-            // Return a response with the RepairModel object
-            return new OkObjectResult(repair);
+            // Filter the repair records by the assignedTo query parameter
+            var repair = repairRecords.FirstOrDefault(r => r.AssignedTo.Equals(assignedTo, StringComparison.InvariantCultureIgnoreCase));
```
hund030, line 37: No default value is provided here? I suppose `repair` would be null if no record match `assignedTo`. Is this behavior expected?

huimiu: Yeah, the adaptive card renders the response returned by the API have verified the data, so it should have no impact on the display. In order to unify it with the ts/js version, I have also added a default value. Thanks for pointing out this.

templates/ts/copilot-plugin-from-scratch/repair/index.ts
huimiu: Thanks for your suggestion, I've updated.

templates/js/copilot-plugin-from-scratch/repair/index.js
```diff
@@ -31,15 +31,15 @@ module.exports = async function (context, req) {
...
+    return res;
+  }
 
   // Filter the repair information by the assignedTo query parameter.
-  if (assignedTo) {
-    repair = repairRecords.find(
-      (item) => item.assignedTo.toLocaleLowerCase() === assignedTo.toLocaleLowerCase()
-    );
-  }
+  let repair = repairRecords.find(
```
hund030, line 40: You may want to use `const`

hund030 (dismissed):

hund030 (approved):

PR #9925, opened 2023-09-13
refactor: add debug log

KennethBWSong:


packages/fx-core/src/component/driver/aad/utility/aadAppClient.ts
```diff
@@ -53,25 +55,38 @@ export class AadAppClient {
   @hooks([ErrorContextMW({ source: "Graph", component: "AadAppClient" })])
   public async createAadApp(
     displayName: string,
-    signInAudience = SignInAudience.AzureADMyOrg
+    signInAudience = SignInAudience.AzureADMyOrg,
+    logProvider?: LogProvider
```
blackchoey, line 59: can we pass the log provider in constructor? So we don't need to update every function in this class.

packages/fx-core/src/component/driver/aad/utility/aadAppClient.ts
```diff
@@ -53,25 +55,38 @@ export class AadAppClient {
...
     const requestBody: IAADDefinition = {
       displayName: displayName,
       signInAudience: signInAudience,
     }; // Create an AAD app without setting anything
 
+    logProvider?.debug(
+      getLocalizedString("core.common.SentApiRequest", `${this.baseUrl}/applications`, JSON.stringify(requestBody))
+    );
     const response = await this.axios.post("applications", requestBody);
+    logProvider?.debug(
```
blackchoey, line 70: axios will throw exception when received 4xx/5xx response. May need to wrap the log in a try catch block

blackchoey: By the way, a minor improvement: axios has interceptors (middlewares), so maybe we can add request/response interceptor to  log the requests and reduce some duplicate code. The try-catch block should no longer required in this case.

No need to improve in this PR. You can see which effort is smaller.

KennethBWSong: Updated. Now using interceptors in aad clients and keep directly logging in app studio client to avoid big changes. One problem is that axios@0.21.2 has a bug that will skip response interceptors when request interceptors exists, so I updated to 0.21.4, while updating the version will cause another bug in per request try: "Object literal may only specify known properties, and '"axios-retry"' does not exist in type 'AxiosRequestConfig'.". I added following code to fix build error:
···
declare module 'axios' {
  export interface AxiosRequestConfig {
    'axios-retry'?: IAxiosRetryConfig;
  }
}
···
@blackchoey  Do you have any idea on the reason?

blackchoey (dismissed):

packages/fx-core/src/component/driver/aad/utility/aadAppClient.ts
KennethBWSong: @blackchoey Update version of retry axios can not fix this error, so I will keep the above code.

blackchoey (approved):

jayzhang (dismissed):

packages/fx-core/src/component/driver/teamsApp/clients/appStudioClient.ts
```diff
@@ -541,8 +541,18 @@ export namespace AppStudioClient {
       let response;
       if (region) {
         try {
+          logProvider.debug(
+            getLocalizedString(
+              "core.common.SentApiRequest",
+              `${region ?? baseUrl}/api/appdefinitions/{teamsAppId}/owner`,
```
SLdragon, line 547: remove region?

SLdragon (dismissed):

packages/fx-core/resource/package.nls.json
```diff
@@ -401,6 +401,8 @@
   "core.common.ResolveServerUrlFailed": "Unable to resolve the server URL: please make sure that the environment variable %s is defined.",
   "core.common.OperationOnlyContainsOptionalParam": "Operation %s only contains optional parameters which are not supported",
   "core.common.ConvertSwaggerToOpenAPI": "The Swagger 2.0 file has been converted to OpenAPI 3.0.",
+  "core.common.SentApiRequest": "Sending API request: %s. Request body: %s",
```
sffamily, line 404: please check whether it is "Sending" or supposed to be "Sent"?

KennethBWSong: Should be sending, and I have rename the key name. Thanks!

blackchoey (approved):

sffamily (dismissed):

jayzhang (dismissed):

wenytang-ms (approved):

sffamily (approved):

SLdragon (approved):

wenytang-ms (approved):

xzf0587 (approved):

nliu-ms (approved):

PR #9941, opened 2023-09-14
perf(spec-parser): wrap adaptive card

SLdragon:
Wrap adaptive card as below for SME:
```
 {
    version: "devPreview",
    $schema: "<URL_REFERENCE_TO_SCHEMA>",
    responseLayout: "list",
    responseCardTemplate: card,
    previewCardTemplate: {
      title: title,
      subtitle: subtitle,
    },
  };
```

### DIFF
--- a/packages/fx-core/src/common/spec-parser/adaptiveCardWrapper.ts
+++ b/packages/fx-core/src/common/spec-parser/adaptiveCardWrapper.ts
@@ -11,8 +11,8 @@ export function wrapAdaptiveCard(
   subtitle: string
 ): WrappedAdaptiveCard {
   const result: WrappedAdaptiveCard = {
-    version: ConstantString.AdaptiveCardVersion,
-    $schema: ConstantString.AdaptiveCardSchema,
+    version: ConstantString.WrappedCardVersion,
+    $schema: ConstantString.WrappedCardSchema,
     responseLayout: ConstantString.WrappedCardResponseLayout,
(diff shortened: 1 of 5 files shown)


huimiu (approved):

packages/fx-core/src/common/spec-parser/adaptiveCardWrapper.ts
```diff
@@ -0,0 +1,24 @@
...
+import { AdaptiveCard, WrappedAdaptiveCard } from "./interfaces";
+
+export function wrapAdaptiveCard(
+  card: AdaptiveCard,
+  title: string,
+  subtitle: string
+): WrappedAdaptiveCard {
+  const result: WrappedAdaptiveCard = {
+    version: "devPreview",
+    $schema: "<URL_REFERENCE_TO_SCHEMA>",
```
yuqizhou77, line 14: would be better if we could save the constants here in the "constants.ts"? thought the value will be changed later.

yuqizhou77 (dismissed):
Let's do an internal bug bash together after this PR is merged!

packages/fx-core/src/common/spec-parser/adaptiveCardWrapper.ts
SLdragon: Sure

adashen (approved):

yuqizhou77 (approved):

PR #9957, opened 2023-09-15
feat: auto update for test tool

a1exwang:
auto check and update when f5 at given interval

### DIFF
--- a/packages/fx-core/src/common/deps-checker/depsChecker.ts
+++ b/packages/fx-core/src/common/deps-checker/depsChecker.ts
@@ -56,6 +56,8 @@ export interface TestToolInstallOptions {
   symlinkDir: string;
   projectPath: string;
   versionRange: string;
+  // in milliseconds
+  updateInterval?: number;
 }
 
 export type InstallOptions = BaseInstallOptions | FuncInstallOptions | TestToolInstallOptions;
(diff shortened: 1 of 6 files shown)


packages/fx-core/src/common/deps-checker/depsChecker.ts
```diff
@@ -56,6 +56,8 @@ export interface TestToolInstallOptions {
   symlinkDir: string;
   projectPath: string;
   versionRange: string;
+  // in milliseconds
```
swatDong, line 59: Could this be customized by user? If so, millisecond is a bit ... weird - I need to calculate how many milliseconds a week has, or 1000.000 how many zeros I need to add.
Could it be days and integer only?

a1exwang: Added mainly for disable auto update. I will remove this property and let user use a fix version number `1.2.3` to skip auto update. This will align with func install

swatDong: sounds good.

swatDong (approved):

swatDong (approved):

hund030 (approved):

swatDong (approved):

PR #9935, opened 2023-09-13
fix: throw error when github is not available

jayzhang:
fix: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/25152566

PR #9966, opened 2023-09-18
fix(vsc): sample gallery catches exception

tecton:
Related PR: https://github.com/OfficeDev/TeamsFx/pull/9935

### DIFF
--- a/packages/fx-core/src/common/samples.ts
+++ b/packages/fx-core/src/common/samples.ts
@@ -68,10 +68,14 @@ class SampleProvider {
       this.branchOrTag = SampleConfigTag;
       const branch = process.env[FeatureFlagName.SampleConfigBranch];
       if (branch) {
-        const data = await this.fetchRawFileContent(branch);
-        if (data !== undefined) {
+        try {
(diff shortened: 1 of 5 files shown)


HuihuiWu-Microsoft (approved):

packages/fx-core/src/common/samples.ts
```diff
@@ -68,11 +68,11 @@ class SampleProvider {
...
       const branch = process.env[FeatureFlagName.SampleConfigBranch];
       if (branch) {
-        const data = await this.fetchRawFileContent(branch);
-        if (data !== undefined) {
+        try {
+          const data = await this.fetchRawFileContent(branch);
           this.branchOrTag = branch;
           this.samplesConfig = data as { samples: Array<Record<string, unknown>> };
-        }
+        } catch (e: unknown) {}
```
wenytang-ms, line 75: maybe record the telemetry data?

tecton: This usually happens when user sets wrong env variable. So I think telemetry event is redundant.

PR #9972, opened 2023-09-18
refactor: modify the logic to switch to local zip when remote zip processing fails

Yimin-Jin:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/25200387/


huimiu:
Kindly reminder: if your PR title starts with `feat:`, you need include a AZDO feature link in the description.

Yimin-Jin:
> Share a PR rule: if your PR title starts with `feat:`, you may need include a AZDO feature link in the description.

hund030:
1. I don't think this is a feature. You may want to use `refactor:` or `fix:` instead.
2. The unit test coverage does not meet the required threshold. You can check the details of the CI action and add a UT to increase the coverage. https://app.codecov.io/gh/OfficeDev/TeamsFx/pull/9972

eriolchan:
> 1. I don't think this is a feature. You may want to use `refactor:` or `fix:` instead.
> 2. The unit test coverage does not meet the required threshold. You can check the details of the CI action and add a UT to increase the coverage. https://app.codecov.io/gh/OfficeDev/TeamsFx/pull/9972
 
Good catch, can you share Yimin with the guidance of commit convention in our team?

packages/fx-core/src/component/generator/generatorAction.ts
```diff
@@ -143,16 +136,35 @@ export const unzipAction: GeneratorAction = {
...
     }
-    context.outputs = await unzip(
-      context.zip,
-      context.destination,
-      context.fileNameReplaceFn,
-      context.fileDataReplaceFn,
-      context.filterFn
-    );
+    const tryLimits = 2;
+    for (let i = 0; i < tryLimits; i++) {
```
hund030, line 140: I would prefer to add a new action that checks the `context.outputs` and fetches the fallback if the output is empty, instead of modifying the `unzipAction`. We can discuss this further offline.

packages/fx-core/src/component/generator/generatorAction.ts
```diff
@@ -143,16 +136,35 @@ export const unzipAction: GeneratorAction = {
...
       throw new MissKeyError("zip");
     }
-    context.outputs = await unzip(
-      context.zip,
-      context.destination,
-      context.fileNameReplaceFn,
-      context.fileDataReplaceFn,
-      context.filterFn
-    );
+    const tryLimits = 2;
```
eriolchan, line 139: Let's talk about the design offline since this change break the overall design of downloading actions. It violate the rule of single responsibility for this action and make it not be reused anymore.

packages/fx-core/src/component/generator/generatorAction.ts
```diff
@@ -133,13 +133,6 @@ export const fetchTemplateZipFromLocalAction: GeneratorAction = {
...
     const data: Buffer = await fs.readFile(zipPath);
     context.zip = new AdmZip(data);
-  },
-};
-
-export const unzipAction: GeneratorAction = {
-  name: GeneratorActionName.Unzip,
-  run: async (context: GeneratorContext) => {
-    context.logProvider.debug(`Unzipping: ${JSON.stringify(context)}`);
     if (!context.zip) {
```
hund030, line 136: This condition is meaningless. Let's remove it.

packages/fx-core/src/component/generator/generatorAction.ts
```diff
@@ -153,12 +146,28 @@ export const unzipAction: GeneratorAction = {
...
+  },
+};
+
 export const TemplateActionSeq: GeneratorAction[] = [
   fetchTemplateZipFromSourceCodeAction,
   fetchTemplateUrlWithTagAction,
   fetchZipFromUrlAction,
-  fetchTemplateZipFromLocalAction,
   unzipAction,
+  fetchTemplateZipFromLocalAction,
```
hund030, line 170: Let's update the action name. Maybe `fetchTemplateFromLocalAction`? You can think of a better name.

packages/fx-core/tests/component/generator/generator.test.ts
```diff
@@ -539,6 +539,25 @@ describe("Generator happy path", async () => {
...
+    await fs.writeFile(path.join(inputDir, foobarTemplateName, "test.txt.tpl"), fileData);
+    const zip = new AdmZip();
+    zip.addLocalFolder(inputDir);
+    zip.writeZip(path.join(tmpDir, "foobar.zip"));
+    sandbox.stub(generatorUtils, "fetchTemplateZipUrl").resolves("foobar.zip");
+    sandbox
+      .stub(generatorUtils, "fetchZipFromUrl")
+      .resolves(new AdmZip(path.join(tmpDir, "foobar.zip")));
+    const result = await Generator.generateTemplate(context, tmpDir, realTemplateName, language);
+    assert.isTrue(result.isOk());
```
hund030, line 559: Please also assert the results are expected.

packages/fx-core/tests/component/generator/generator.test.ts
```diff
@@ -539,6 +539,25 @@ describe("Generator happy path", async () => {
...
   });
+
+  it("template from fallback", async () => {
+    const foobarTemplateName = "foobarTemplate";
+    const realTemplateName = "non-sso-tab";
+    const language = "ts";
+    const inputDir = path.join(tmpDir, "input");
+    await fs.ensureDir(path.join(inputDir, foobarTemplateName));
+    const fileData = "{{appName}}";
+    await fs.writeFile(path.join(inputDir, foobarTemplateName, "test.txt.tpl"), fileData);
```
hund030, line 550: Seems you just want to create an AdmZip object. Is it necessary to write the file to the file system?

Yimin-Jin: Thanks for the feedback, I'll add a success check for generating real template and a check to make sure no foobar files are generated.

packages/fx-core/tests/component/generator/generator.test.ts
Yimin-Jin: Thanks for the feedback, I'll assert more results.

packages/fx-core/tests/component/generator/generator.test.ts
Yimin-Jin: Yes, you're right, because foobarTemplateName is only needed to make FilterFn unable to pass the prefix checksum, no need to add the contents of the file.

Yimin-Jin: I'll remove the code for adding file content.

packages/fx-core/src/component/generator/generatorAction.ts
Yimin-Jin: Yeah, I think that name would be better.

packages/fx-core/src/component/generator/generatorAction.ts
```diff
@@ -153,12 +146,28 @@ export const unzipAction: GeneratorAction = {
   },
 };
 
+export const unzipAction: GeneratorAction = {
+  name: GeneratorActionName.Unzip,
+  run: async (context: GeneratorContext) => {
+    context.logProvider.debug(`Unzipping: ${JSON.stringify(context)}`);
+    if (context.zip) {
```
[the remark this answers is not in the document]
Yimin-Jin: Yes, I think it would be better this way.

packages/fx-core/src/component/generator/generatorAction.ts
Yimin-Jin: All right.

packages/fx-core/tests/component/generator/generator.test.ts
```diff
@@ -13,7 +13,7 @@ import {
   renderTemplateFileData,
   renderTemplateFileName,
 } from "../../../src/component/generator/utils";
-import { assert } from "chai";
+import { assert, expect } from "chai";
```
hund030, line 16: Remove unused import `expect`.

packages/fx-core/tests/component/generator/generator.test.ts
```diff
@@ -541,21 +541,17 @@ describe("Generator happy path", async () => {
   });
 
   it("template from fallback", async () => {
-    const foobarTemplateName = "foobarTemplate";
     const realTemplateName = "non-sso-tab";
```
hund030, line 544: The best practice of unit test is to mock all external dependencies. You may consider when we remove `non-sso-tab` template, this test case will fail unexpectedly.

hund030: Anyway, this test case looks much "cleaner" after the refactor. It will be better if you get rid of the dependency of a real template.

packages/fx-core/tests/component/generator/generator.test.ts
Yimin-Jin: All right.

packages/fx-core/tests/component/generator/generator.test.ts
```diff
@@ -541,15 +542,25 @@ describe("Generator happy path", async () => {
...
-    const realTemplateName = "non-sso-tab";
+    const templateName = "test";
+    const mockFileName = "test.txt";
+    const mockFileData = "test data";
     const language = "ts";
+    const inputDir = path.join(tmpDir, "input");
+    const fallbackDir = path.join(tmpDir, "fallback");
+    await fs.ensureDir(path.join(inputDir, templateName));
+    await fs.ensureDir(fallbackDir);
+    await fs.writeFile(path.join(inputDir, templateName, mockFileName), mockFileData);
```
hund030, line 553: Writing temp files to fs in UT is not a good practice. Be aware that you can always mock the fs operation to get rid of the dependency of disk. 
While I understand mocking fs operation takes much more effort, it's also acceptable to do so when you feel it is too tedious to do it perfectly.
In this case, I do not think it is necessary to write this file. Try using AdmZip.addFile.

packages/fx-core/tests/component/generator/generator.test.ts
```diff
@@ -38,6 +38,7 @@ import templateConfig from "../../../src/common/templates-config.json";
 import { placeholderDelimiters } from "../../../src/component/generator/constant";
 import sampleConfigV3 from "../../common/samples-config-v3.json";
 import Mustache from "mustache";
+import * as folder from "../../../../fx-core/src/folder";
```
hund030, line 41: This import is a little confusing. I prefer using `import * as folderUtils from ...`. You can make the decision by yourself.

Yimin-Jin: Thanks for the feedback. It's true that it's less confusing to write that way.

packages/fx-core/tests/component/generator/generator.test.ts
Yimin-Jin: Thanks for the feedback. I'll try to minimize the fs code.

packages/fx-core/tests/component/generator/generator.test.ts
Yimin-Jin: Thanks for the feedback. I‘d replace real template with mock template.

hund030 (approved):

PR #10009, opened 2023-09-21
perf(spec-parser): treat required + default parameter as optional parameter

SLdragon:
[Bug 25203352](https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/25203352): [API ME] Treat required + default parameter as optional when list supported apis

### DIFF
--- a/packages/fx-core/src/common/spec-parser/utils.ts
+++ b/packages/fx-core/src/common/spec-parser/utils.ts
@@ -46,7 +46,7 @@ export function checkParameters(paramObject: OpenAPIV3.ParameterObject[]): Check
     const schema = param.schema as OpenAPIV3.SchemaObject;
 
     if (param.in === "header" || param.in === "cookie") {
-      if (param.required && schema.default === undefined) {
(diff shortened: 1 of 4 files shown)


yuqizhou77 (dismissed):

packages/fx-core/src/common/spec-parser/utils.ts
```diff
@@ -43,28 +43,29 @@ export function checkParameters(paramObject: OpenAPIV3.ParameterObject[]): Check
...
 
-    const schema = param.schema as OpenAPIV3.SchemaObject;
     if (
       schema.type !== "boolean" &&
       schema.type !== "string" &&
       schema.type !== "number" &&
       schema.type !== "integer"
     ) {
-      if (param.required) {
+      if (param.required && schema.default === undefined) {
```
adashen, line 61: can we have a utility method to extract this logic?

SLdragon: sure

adashen (approved):

PR #10015, opened 2023-09-21
refactor: update vs copilot template to support tunnel

xiaolang124:


huimiu:
You may also need to modify the [template constrain](https://github.com/OfficeDev/TeamsFx/blob/dev/templates/constraints/yml/templates/csharp/copilot-plugin-from-scratch/teamsapp.local.yml.tpl.mustache)

hund030 (approved):

PR #10020, opened 2023-09-21
fix: rename m365 command

jayzhang:
fix: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/25202429

### DIFF
--- a/packages/cli/src/commands/models/m365Sideloading.ts
+++ b/packages/cli/src/commands/models/m365Sideloading.ts
@@ -10,7 +10,7 @@ export const sideloadingServiceEndpoint =
   process.env.SIDELOADING_SERVICE_ENDPOINT ?? serviceEndpoint;
 
 export const m365SideloadingCommand: CLICommand = {
-  name: "sideloading",
+  name: "install",
   description:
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/models/m365Sideloading.ts
```diff
@@ -10,7 +10,7 @@ export const sideloadingServiceEndpoint =
   process.env.SIDELOADING_SERVICE_ENDPOINT ?? serviceEndpoint;
 
 export const m365SideloadingCommand: CLICommand = {
-  name: "sideloading",
```
swatDong, line 13: What's the requirement to rename this? Or, is it possible to mark `sideloading` as deprecated but can still be executed, to keep backward compatibility?

There's already existing projects using `teamsfx m365 sideloading ...`, e.g., office-addin, MOS test, documentation... This rename is a breaking change and may break those projects.

swatDong: Offline discussed. This is for CLI v3 only. The whole CLI v3 is a breaking change so just ensure dependents and docs are upgraded as well.

swatDong (approved):

PR #10030, opened 2023-09-22
perf(spec-parser): add jsonPath support

SLdragon:
[Bug 25216653](https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/25216653): jsonPath support

The code will attempt to identify if the array property in the root contains any of the well-known substrings below. If found, it will use the property as a JSON path:

```js
  static readonly WellknownResultNames = [
    "result",
    "data",
    "items",
    "root",
    "matches",
    "queries",
    "list",
    "output",
  ];
```

Yuqi and I have[ an investigation](https://microsoft-my.sharepoint.com/personal/yuqzho_microsoft_com/_layouts/15/Doc.aspx?sourcedoc={41143a23-0931-4bf9-90be-f2e47ee91491}&action=edit&wd=target%28New%20Section%201.one%7C1ccf704e-b58b-4db3-aaa6-c39ee69bbe0d%2FAPI%20Spec%20Response%7Ce3279249-e16c-4ba0-ba6f-de1af41a5af4%2F%29&wdorigin=703) about the names in spec file.

TODO: we may add some additional WellKnownResultNames in future.

### DIFF
--- a/packages/fx-core/src/common/spec-parser/utils.ts
+++ b/packages/fx-core/src/common/spec-parser/utils.ts
@@ -340,3 +340,12 @@ export function validateServer(spec: OpenAPIV3.Document): ErrorResult[] {
   }
   return errors;
 }
+
+export function isWellknownResultPropertyName(name: string): boolean {
+  for (let i = 0; i < ConstantString.WellknownResultNames.length; i++) {
+    if (name.toLowerCase().includes(ConstantString.WellknownResultNames[i])) {
+      return true;
+    }
+  }
+  return false;
+}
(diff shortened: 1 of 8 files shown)


packages/fx-core/src/common/spec-parser/utils.ts
```diff
@@ -340,3 +340,12 @@ export function validateServer(spec: OpenAPIV3.Document): ErrorResult[] {
   }
   return errors;
 }
+
+export function isWellknownResultPropertyName(name: string): boolean {
+  for (let i = 0; i < ConstantString.WellknownResultNames.length; i++) {
```
yuqizhou77, line 345: we might can do ConstantString.WellknownResultNames.includes(name.tolowercase()) directly?

SLdragon: If name's substring contains wellknown name, I also want to match it. For example, if a name is "xxxItems", I also want to match it.

yuqizhou77: Make sense.

yuqizhou77 (approved):

adashen (approved):

PR #10084, opened 2023-10-07
refactor: vs bot local template

xiaolang124:
Update env driver to update process after updating env file.


packages/fx-core/src/component/driver/file/createOrUpdateEnvironmentFile.ts
```diff
@@ -70,8 +70,12 @@ export class CreateOrUpdateEnvironmentFileDriver implements StepDriver {
       );
       context.logProvider?.debug(`Updated envs: ${JSON.stringify(updatedEnvs)}`);
       await fs.writeFile(target, updatedEnvs.join(os.EOL));
+      const output = new Map<string, string>();
+      for (const [key, value] of Object.entries(args.envs)) {
+        output.set(key, value);
+      }
       return {
-        output: new Map<string, string>(),
```
swatDong, line 74: This change has side-effect - if I'm updating my own `.env` file, e.g., `some-runtime.env`, those variables will be injected into ttk process as well.

To directly modify ttk's env, how about using `::set-output` script action?

PR #10097, opened 2023-10-09
refactor: modify the logic to switch to local zip when remote zip processing fails

Yimin-Jin:


packages/fx-core/tests/component/generator/generator.test.ts
```diff
@@ -539,6 +540,47 @@ describe("Generator happy path", async () => {
...
+    const templateName = "test";
+    const mockFileName = "test.txt";
+    const mockFileData = "test data";
+    const language = "ts";
+    const fallbackDir = path.join(tmpDir, "fallback");
+    await fs.ensureDir(fallbackDir);
+    const foobarTemplateZip = new AdmZip();
+    const templateZip = new AdmZip();
+    templateZip.addFile(path.join(templateName, mockFileName), Buffer.from(mockFileData));
+    templateZip.writeZip(path.join(fallbackDir, "ts.zip"));
```
hund030, line 554: I found there are same code building a zip file between Unit tests. Could you help refactor it by adding a common function? For example, `buildFakeTemplateZip()`.

hund030 (approved):

PR #10123, opened 2023-10-12
test: clean up m365 apps

hellyzh:


packages/tests/src/scripts/clean.ts
```diff
@@ -140,6 +143,29 @@ async function main() {
     Env.password
   );
   await devTunnelCleanHelper.deleteAll();
+
+  const m365TitleCleanService = await M365TitleCleanHelper.create(
+    Env.cleanTenantId,
+    "7ea7c24c-b1f6-4a20-9d11-9ae12e9e7ac0",
```
wenytang-ms, line 149: will this value change?

hellyzh: No, it will not change.

wenytang-ms (approved):

PR #10125, opened 2023-10-12
fix: enable empty string env value

jayzhang:
fix: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/25425932


1yefuwang1 (approved):
Changes to config manager LGTM

LongOddCode (approved):

PR #10141, opened 2023-10-16
fix: only APP_NAME_SUFFIX accepts empty string

jayzhang:
fix: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/25496058

### DIFF
--- a/packages/fx-core/src/component/configManager/lifecycle.ts
+++ b/packages/fx-core/src/component/configManager/lifecycle.ts
@@ -82,7 +82,7 @@ function resolve(
   }
 }
 
-function resolveString(
+export function resolveString(
   val: string,
   resolved: ResolvedPlaceholders,
   unresolved: UnresolvedPlaceholders
--- a/packages/fx-core/src/component/configManager/lifecycle.ts
+++ b/packages/fx-core/src/component/configManager/lifecycle.ts
@@ -93,7 +93,7 @@ function resolveString(
   while (matches != null) {
     const envVar = matches[1];
     const envVal = process.env[envVar];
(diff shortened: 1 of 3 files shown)


Siglud (approved):

packages/fx-core/src/component/configManager/lifecycle.ts
```diff
@@ -93,11 +93,20 @@ function resolveString(
   while (matches != null) {
     const envVar = matches[1];
     const envVal = process.env[envVar];
-    if (envVal === undefined || envVal === null) {
-      unresolved.push(envVar);
+    if (envVar === "APP_NAME_SUFFIX") {
+      if (envVal === undefined || envVal === null) {
```
1yefuwang1, line 97: It seems like this line will evaluate to true.

LongOddCode (approved):

packages/fx-core/src/component/utils/common.ts
```diff
@@ -123,8 +123,14 @@ export function expandEnvironmentVariable(content: string): string {
     for (const placeholder of placeholders) {
       const envName = placeholder.slice(3, -2).trim(); // removes `${{` and `}}`
       const envValue = process.env[envName];
-      if (envValue !== undefined && envValue !== null) {
-        content = content.replace(placeholder, envValue);
+      if (envName === "APP_NAME_SUFFIX") {
+        if (envValue !== undefined && envValue !== null) {
```
1yefuwang1, line 127: It seems like this line will always evaluate to true.

1yefuwang1 (approved):

PR #10133, opened 2023-10-13
feat: auto recommend user to install ACP extension

a1exwang:
- when there are adaptive card files in workspace, pop up to install ACP
- when user clicks button in treeview, pop up to install ACP

image

PR #10142, opened 2023-10-16
refactor: remove ac card codelens

a1exwang:
Remove codelens for adaptive cards since codelens is in new ACP extension


packages/vscode-extension/src/codeLensProvider.ts
```diff
@@ -147,6 +147,7 @@ export class CryptoCodeLensProvider implements vscode.CodeLensProvider {
   }
 }
 
+// Only used to detect adaptive cards in project. Codelens for ac is not used.
 export class AdaptiveCardCodeLensProvider implements vscode.CodeLensProvider {
```
dooriya, line 151: Since we already have a function to detect Adaptive Card in workspace (pr #10133), do we still need to keep this class?

tecton (approved):

PR #10156, opened 2023-10-18
feat(vsc): add multi-select tag filters to sample gallery

tecton:
Related task: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/25297769

- Use multi-select dropdown in fluent-ui.
- Use filter options in sample config json file. Related PR: https://github.com/OfficeDev/TeamsFx-Samples/pull/1066
- Filter by tag first using `or` logic and search by text then.
- light theme support.

Effect GIF:
tag-filters


packages/vscode-extension/src/controls/sampleGallery/sampleFilter.tsx
```diff
@@ -16,49 +16,101 @@ import {
 } from "../../telemetry/extTelemetryEvents";
 import { Commands } from "../Commands";
 import { Grid } from "../resources";
-import { SampleFilterProps, SampleFilterState, SampleInfo } from "./ISamples";
+import { SampleFilterProps } from "./ISamples";
+
+export default class SampleFilter extends React.Component<SampleFilterProps, unknown> {
+  private sampleTypes = ["Tab", "Bot", "Message extension"];
+  private sampleLanguages = ["TS", "JS"];
+  private sampleTechniques = ["Azure", "Adaptive Cards", "SSO", "SPFx", "Outlook", "Graph"];
```
HuihuiWu-Microsoft, line 24: Will we need to extend 'sampleTechniques' field when more samples are onboarded?

tecton: Yes, these are the options that I chose firstly, I think this could be updated later according to user behavior. We can keep the most representative ones in the list. Also @summzhan could help to review this list.

HuihuiWu-Microsoft: Yes, I think we'd better make it general and complete, or there'll be dependency between samples and TTK. cc @summzhan

summzhan: Do the values of all these tags, not only techniques, but also type and languages, could be stored in some config file but not hard code? We should as flexible as possible to add new values. It's quite hard to predict new techniques/types appear in future...., at least the type tag is about to add a new value which is copilot plugin in near future right?

tecton: @summzhan Do you mean to store the option values in an online config file like `sample-config.json`?

tecton: Updated code to add filter options in sample config file after offline discussion. cc @HuihuiWu-Microsoft @summzhan

HuihuiWu-Microsoft (approved):

jayzhang (approved):

PR #10162, opened 2023-10-18
feat: support sideload xml manifest for outlook addin

1yefuwang1:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/25520005


packages/cli/src/commands/models/m365Sideloading.ts
```diff
@@ -19,14 +20,22 @@ export const m365SideloadingCommand: CLICommand = {
       name: "file-path",
       description: "Path to the App manifest zip package.",
       type: "string",
-      required: true,
+    },
+    {
+      name: "xml-path",
```
swatDong, line 25: Does this only apply to cli v3?
For v2 there's other code at *cmds/m365/m365.ts*.

packages/fx-core/src/common/m365/packageService.ts
```diff
@@ -74,6 +74,71 @@ export class PackageService {
     }
   }
 
+  @hooks([ErrorContextMW({ source: M365ErrorSource, component: M365ErrorComponent })])
+  public async sideLoadXmlManifest(token: string, manifestPath: string): Promise<[string, string]> {
+    try {
+      const data = await fs.readFile(manifestPath);
+      const content = new FormData();
+      content.append("package", data);
+      content.append("culture", "en-us");
```
swatDong, line 83: Does culture depend on the xml content, or this hard-coded en-us already enough?

swatDong (approved):
Code looks good to me. Leave some UX questions to avoid further changes.

packages/fx-core/src/common/m365/packageService.ts
1yefuwang1: @millerds told me that culture "en-us" needs to be included in the form data. I'll double check with him.

packages/cli/src/commands/models/m365Sideloading.ts
1yefuwang1: That's a good point. I'll talk with huajie about it.

swatDong (approved):

packages/cli/src/commands/models/m365Sideloading.ts
1yefuwang1: According to @jayzhang , CLI v3 has not been enabled by default.
However, since the command is only used by office-addin tooling, I think it's OK to ask them to enable it when using xml sideload.

jayzhang (approved):

swatDong (approved):

PR #10165, opened 2023-10-18
refactor: API ME message fix and refactor for spec parser lib

yuqizhou77:
[Bug 25506195](https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/25506195): [API-ME] Provide a solution when failed to download OpenAPI Spec
[Bug 25506207](https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/25506207): [API-ME] Better error guidance when the format of OpenAPI spec is incorrect

Also format validation error string based on error type and data, which is for spec parser lib refactor.


nliu-ms (dismissed):

jayzhang (approved):

SLdragon (dismissed):

packages/fx-core/resource/package.nls.json
```diff
@@ -406,6 +406,8 @@
   "core.common.ResolveServerUrlFailed": "Unable to resolve the server URL: please make sure that the environment variable %s is defined.",
   "core.common.OperationOnlyContainsOptionalParam": "Operation %s contains multiple optional parameters. The first optional parameter is used for this command.",
   "core.common.ConvertSwaggerToOpenAPI": "The Swagger 2.0 file has been converted to OpenAPI 3.0.",
+  "core.common.ErrorFetchApiSpec": "Please make sure your OpenAPI description document can be accessed without authentication or you can download a copy of the document to your local disk and upload the local file.",
```
sffamily, line 409: Your OpenAPI specification should be accessible without authentication, otherwise download and start from a local copy.

packages/fx-core/resource/package.nls.json
```diff
@@ -295,7 +295,7 @@
   "core.createProjectQuestion.apiSpecInputUrl.label": "Enter OpenAPI Description Document Location",
   "core.createProjectQuestion.OpenAIPluginDomain": "OpenAI Plugin Manifest",
   "core.createProjectQuestion.OpenAIPluginDomain.placeholder": "Enter your website domain or manifest URL",
-  "core.createProjectQuestion.invalidUrl.message": "Please enter a valid URL",
+  "core.createProjectQuestion.invalidUrl.message": "Please enter a valid HTTP URL without authentication to access your OpenAPI description document.",
```
sffamily, line 298: typically in message we don't add "Please", so can remove it.

description document -> specification?

packages/fx-core/resource/package.nls.json
yuqizhou77: Updated.

packages/fx-core/resource/package.nls.json
yuqizhou77: According to [Bug 25257315](https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/25257315): [API ME] Update "OpenAPI Specfication" to "OpenAPI Description Document", OpenAPI description document is what we should use.  “OpenAPI Specification is a standard, OpenAPI Description document is to describe an API that conforms to that standard“

SLdragon (dismissed):

nliu-ms (approved):

jayzhang (approved):

sffamily (approved):

SLdragon (approved):

PR #10171, opened 2023-10-19
feat: m365 install/uninstall example

jayzhang:
feat link: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/25124858

### DIFF
--- a/packages/cli/src/commands/models/m365Sideloading.ts
+++ b/packages/cli/src/commands/models/m365Sideloading.ts
@@ -24,7 +24,7 @@ export const m365SideloadingCommand: CLICommand = {
   ],
   examples: [
     {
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/models/m365Sideloading.ts
```diff
@@ -24,7 +24,7 @@ export const m365SideloadingCommand: CLICommand = {
   ],
   examples: [
     {
-      command: `${process.env.TEAMSFX_CLI_BIN_NAME} m365 sideloading --file-path appPackage.zip`,
```
swatDong, line 27: Any command description and document change?

swatDong (approved):

PR #10183, opened 2023-10-20
refactor(sdk): export `handleMessageExtensionLinkQueryWithSSO` for link unfurling

wenytang-ms:


packages/sdk/src/index.ts
```diff
@@ -89,7 +89,7 @@ export { CardActionBot } from "./conversation/cardAction";
 export { MessageBuilder } from "./conversation/messageBuilder";
 export { InvokeResponseFactory } from "./conversation/invokeResponseFactory";
 export {
-  handleMessageExtensionQueryWithToken,
+  executionWithTokenAndConfig,
```
blackchoey, line 92: maybe not remove exported function? this should be a breaking change

blackchoey: some additional thoughts about the method name:
1. Can we use `executionWithToken` as the name? I see there's already an `executionWithToken` function in the source code and being deprecated. But since it's not exported to users, maybe we can use that name?
2. Another option maybe using `handleMessageExtensionQueryLinkWithSSO` (we can refer BotFramework to determine the name), so developers can easily understand this function. The name `executionWithToken` is somehow too general and might cause confusions.

wenytang-ms: 1. restore the `handleMessageExtensionQueryWithToken`
2. export `handleMessageExtensionLinkQueryWithSSO` instead of `executionWithToken`, which in botframework named as `handleTeamsAppBasedLinkQuery` for such type.

blackchoey (approved):

SLdragon (approved):

yiqing-zhao (approved):

PR #10201, opened 2023-10-24
feat: auto install dependencies after scaffolding/importing SPFx projects

HuihuiWu-Microsoft:
image

### DIFF
--- a/packages/vscode-extension/src/constants.ts
+++ b/packages/vscode-extension/src/constants.ts
@@ -27,6 +27,7 @@ export enum GlobalKey {
   ShowLocalDebugMessage = "ShowLocalDebugMessage",
   CreateWarnings = "CreateWarnings",
   SampleGalleryLayout = "teamsToolkit:sampleGallery:layout",
+  AutoInstallDependency = "fx-extension.autoInstallDependency",
 }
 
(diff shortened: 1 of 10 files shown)


jayzhang (approved):

packages/vscode-extension/src/constants.ts
```diff
@@ -27,6 +27,7 @@ export enum GlobalKey {
   ShowLocalDebugMessage = "ShowLocalDebugMessage",
   CreateWarnings = "CreateWarnings",
   SampleGalleryLayout = "teamsToolkit:sampleGallery:layout",
+  AutoInstallDependency = "fx-extension.autoInstallDependency",
```
tecton, line 30: It's better to use `teamsToolkit` as prefix instead of `fx-extension` which is confusing.

tecton (dismissed):

packages/vscode-extension/src/constants.ts
HuihuiWu-Microsoft: Updated

tecton (approved):

jayzhang (approved):

tecton (approved):

sffamily (approved):

PR #10208, opened 2023-10-25
refactor: integrate with API Key registration APIs

nliu-ms:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/25577741

### DIFF
--- a/packages/fx-core/src/component/driver/teamsApp/interfaces/ApiSecretRegistration.ts
+++ b/packages/fx-core/src/component/driver/teamsApp/interfaces/ApiSecretRegistration.ts
@@ -32,7 +32,7 @@ export interface ApiSecretRegistration {
 
 export enum ApiSecretRegistrationAppType {
   SpecificApp = "SpecificApp",
-  AnyApps = "AnyApp",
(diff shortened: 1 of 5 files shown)


packages/fx-core/src/component/driver/teamsApp/interfaces/ApiSecretRegistration.ts
```diff
@@ -0,0 +1,31 @@
...
+  tenantId?: string;
+  /**
+   * Currently max length 1
+   */
+  targetUrlsShouldStartWith?: string[];
+  /**
+   * Teams app Id associated with the ApiSecretRegistration
+   */
+  specificAppId?: string;
+  applicableToApps?: ApiSecretRegistrationAppType[];
```
KennethBWSong, line 25: According to [here](https://domoreexp.visualstudio.com/Teamspace/_git/teams-devapp?path=/Portal/src/interfaces/IApiKeyRegistration.ts), applicableToApps shouldn't be an array?

jayzhang (approved):

PR #10214, opened 2023-10-25
perf: add draft for apiSecret/create action

KennethBWSong:
still in draft

### DIFF
--- a/packages/fx-core/src/component/driver/apiSecret/create.ts
+++ b/packages/fx-core/src/component/driver/apiSecret/create.ts
@@ -1,0 +1,160 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { Service } from "typedi";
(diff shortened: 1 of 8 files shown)


packages/fx-core/src/component/driver/apiSecret/create.ts
```diff
@@ -0,0 +1,160 @@
...
+    args: CreateApiSecretArgs
+  ): Promise<ApiSecret> {
+    const currentUserRes = await tokenProvider.getJsonObject({ scopes: GraphScopes });
+    if (currentUserRes.isErr()) {
+      throw currentUserRes.error;
+    }
+    const currentUser = currentUserRes.value;
+    const userId = currentUser["oid"] as string;
+
+    const secrets = this.parseSecret(args.apiSecret!);
```
SLdragon, line 125: should check the secret 1<= lens <=2

PR #10227, opened 2023-10-26
perf(spec-parser): add support for api key in validate/generate function

SLdragon:
- Generate function will add authorization property in manifest when api contains api key auth
- Generate function will throw error if there are multiple api keys for all APIs
- List function will support APIs with only one api key auth
- List function will filter one API needs multiple api key auth


packages/fx-core/src/common/spec-parser/utils.ts
```diff
@@ -129,16 +129,27 @@ export function isSupportedApi(
...
     if (
       (method === ConstantString.PostMethod || method === ConstantString.GetMethod) &&
-      pathObj[method] &&
-      !pathObj[method]?.security
+      pathObj[method]
     ) {
+      const securities = pathObj[method]!.security;
+      const apiKeyAuth = getAPIKeyAuth(securities, spec);
+
+      if (!allowAPIKeyAuth && securities) {
```
adashen, line 145: The logic may have problem later when more securities type supported. What we want to do is only security type equal to apiKey right?

packages/fx-core/src/common/spec-parser/utils.ts
```diff
@@ -129,16 +129,27 @@ export function isSupportedApi(
...
+        return false;
+      }
+
+      if (allowAPIKeyAuth && securities && !apiKeyAuth) {
+        return false;
```
adashen, line 150: If the api key and logic will still be shown?

packages/fx-core/src/common/spec-parser/utils.ts
SLdragon: Yes, in the future if we support aad, the logic would be update to:
if (!allowAPIKeyAuth && !allowAadAuth && securities) {
   return false;
}

packages/fx-core/src/common/spec-parser/utils.ts
```diff
@@ -193,6 +204,27 @@ export function isSupportedApi(
...
 
+export function getAPIKeyAuth(
+  securities: OpenAPIV3.SecurityRequirementObject[] | undefined,
+  spec: OpenAPIV3.Document
+): OpenAPIV3.ApiKeySecurityScheme | undefined {
+  const securitySchemas = spec.components?.securitySchemes;
+  if (securities && securitySchemas) {
+    for (let i = 0; i < securities.length; i++) {
+      const security = securities[i];
+      if (Object.keys(security).length == 1) {
```
adashen, line 215: Based on the method name it is just get the APIKey Auth configuration. So we may not filter here. And how about later we support multiple and logic？ will we change the interface?

yuqizhou77 (dismissed):

sffamily (dismissed):

packages/fx-core/src/common/spec-parser/utils.ts
SLdragon: This condition will filter other auth such as aad for current stage

packages/fx-core/src/common/spec-parser/utils.ts
SLdragon: Thanks, will update the logic to make it more general

nliu-ms (dismissed):

yuqizhou77 (approved):

yuqizhou77 (approved):

adashen (approved):

PR #10259, opened 2023-11-01
fix: relative path check

nliu-ms:
Bug: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/24879976

1) Files outside directory is not allowed:
image

2) Deal with "../", this cannot be the filename in zip directly
image

### DIFF
--- a/packages/fx-core/resource/package.nls.json
+++ b/packages/fx-core/resource/package.nls.json
@@ -600,6 +600,7 @@
   "error.teamsApp.validate.details": "File path: %s, title: %s",
   "error.teamsApp.AppIdNotExistError": "Teams app with ID %s does not exist in Teams Developer Portal.",
   "error.teamsApp.InvalidAppIdError": "Teams app ID %s is invalid, must be a GUID.",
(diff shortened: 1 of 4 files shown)


packages/fx-core/resource/package.nls.json
```diff
@@ -600,6 +600,7 @@
   "error.teamsApp.validate.details": "File path: %s, title: %s",
   "error.teamsApp.AppIdNotExistError": "Teams app with ID %s does not exist in Teams Developer Portal.",
   "error.teamsApp.InvalidAppIdError": "Teams app ID %s is invalid, must be a GUID.",
+  "error.teamsApp.createAppPackage.invalidFile": "%s is invalid, it should be in the same or subdirectory of manifest.json",
```
sffamily, line 603: the same or subdirectory of manifest.json. -> the same directory as manifest.json or a subdirectory of it.

jayzhang (approved):

sffamily (approved):

jayzhang (approved):

PR #10264, opened 2023-11-02
fix: upgrade dependencies

huimiu:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/25386097
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/25593617

### DIFF
--- a/templates/ts/message-extension-copilot/package.json.tpl
+++ b/templates/ts/message-extension-copilot/package.json.tpl
@@ -3,7 +3,7 @@
     "version": "1.0.0",
     "description": "Microsoft Teams Toolkit message extension search sample",
     "engines": {
-        "node": "16 || 18"
+        "node": "16 || 18 || 20"
     },
     "author": "Microsoft",
     "license": "MIT",
--- a/templates/ts/message-extension-copilot/package.json.tpl
+++ b/templates/ts/message-extension-copilot/package.json.tpl
@@ -25,15 +25,15 @@
         "adaptivecards-templating": "^2.3.1",
         "adaptivecards": "^3.0.1",
         "botbuilder": "^4.20.0",
-        "restify": "^10.0.0"
+        "restify": "^11.1.0"
     },
     "devDependencies": {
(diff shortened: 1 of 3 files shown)


templates/ts/message-extension-copilot/package.json.tpl
```diff
@@ -25,15 +25,15 @@
         "adaptivecards-templating": "^2.3.1",
         "adaptivecards": "^3.0.1",
         "botbuilder": "^4.20.0",
-        "restify": "^10.0.0"
+        "restify": "^11.1.0"
     },
     "devDependencies": {
         "@types/restify": "^8.5.5",
-        "@types/node": "^16.0.0",
+        "@types/node": "^20.8.9",
```
hund030, line 32: Does this require node@20?

huimiu: While there isn't a strict requirement, considering this template doesn't have any other specific dependencies and @types/node is a dev dependency, I've tested it without issues after upgrading to node@20. To comprehensively address the issue mentioned in the bug, I'm planning to upgrade to node@20.

hund030: Should we also update the engine in package.json?
https://github.com/OfficeDev/TeamsFx/blob/b1cd95889caaf7c41e86376a0d16ab209537ac21/templates/ts/message-extension-copilot/package.json.tpl#L6

huimiu: Updated, thx! I'll double check with Xiaofu, as she's working on the node@20 upgrade for TTK.

hund030 (approved):

templates/ts/message-extension-copilot/package.json.tpl
huimiu: Confirmed with Xiaofu, the template needs to ensure that the engines has a minimum version of 16.

PR #10267, opened 2023-11-02
perf(apikey): add question model for api key

KennethBWSong:


packages/fx-core/src/component/middleware/questionMW.ts
```diff
@@ -7,9 +7,12 @@ import { TOOLS } from "../../core/globalVars";
 import { QuestionNodes, questionNodes } from "../../question";
 import { traverse } from "../../ui/visitor";
 
-export function QuestionMW(key: keyof QuestionNodes): Middleware {
+export function QuestionMW(key: keyof QuestionNodes, fromAction = false): Middleware {
   return async (ctx: HookContext, next: NextFunction) => {
     const inputs = ctx.arguments[0] as Inputs;
```
jayzhang, line 12: if fromAction is true, the inputs here is DriverContext, is there any issue if we use it as Inputs type?

KennethBWSong: Seems no error when building and running. Do you think we need to mock an input and set platform, xxx into it?

SLdragon (approved):

jayzhang (dismissed):

packages/fx-core/resource/package.nls.json
```diff
@@ -293,6 +293,8 @@
   "core.createProjectQuestion.apiSpecInputUrl.label": "Enter OpenAPI Description Document Location",
   "core.createProjectQuestion.OpenAIPluginDomain": "OpenAI Plugin Manifest",
   "core.createProjectQuestion.OpenAIPluginDomain.placeholder": "Enter your website domain or manifest URL",
+  "core.createaProjectQuestion.ApiKey": "Client secret for Api key in OpenAPI spec",
```
sffamily, line 296: Api -> API
spec -> specification

packages/fx-core/resource/package.nls.json
```diff
@@ -293,6 +293,8 @@
...
+  "core.createProjectQuestion.invalidApiKey.message": "Client secret is invalid. Length of secret should >= 10 and <= 128",
```
sffamily, line 297: Length -> The length
should -> should be

sffamily (approved):

sffamily (commented):
revoke approval, ask for changes

packages/fx-core/resource/package.nls.json
KennethBWSong: Updated, thanks!

packages/fx-core/resource/package.nls.json
KennethBWSong: Updated, thanks!

sffamily (approved):

SLdragon (approved):

jayzhang (approved):

PR #10287, opened 2023-11-07
feat(core): bump sample tag version

tecton:
Related task: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/25518911

Sample config has breaking change.
'v2.4.0' tag will be published before next RC & stable TTK release.

### DIFF
--- a/packages/fx-core/src/common/samples.ts
+++ b/packages/fx-core/src/common/samples.ts
@@ -15,9 +15,8 @@ const packageJson = require("../../package.json");
 const SampleConfigOwner = "OfficeDev";
 const SampleConfigRepo = "TeamsFx-Samples";
 const SampleConfigFile = ".config/samples-config-v3.json";
-export const SampleConfigTag = "v2.3.0";
-// rc and prerelease tag is only different with stable tag when there will a breaking change.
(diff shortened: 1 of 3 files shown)


packages/fx-core/src/common/samples.ts
```diff
@@ -71,8 +70,8 @@ class SampleProvider {
       // prerelease build version always use branch head for prerelease.
       this.branchOrTag = SampleConfigBranchForPrerelease;
     } else if (version.includes("rc")) {
-      // rc version(before next stable TTK) always use prerelease tag
-      this.branchOrTag = SampleConfigTagForRc;
+      // if there is a breaking change, the tag is not used by any stable version.
```
HuihuiWu-Microsoft, line 73: Do we ship v2.4.0 in rc build?

tecton: Yes, if I understand correctly, rc is the version that we will test before stable release. It's not published to marketplace.

HuihuiWu-Microsoft (approved):

yukun-dong (approved):

PR #10289, opened 2023-11-07
refactor: add graph permission in metadata telemetry

xzf0587:


packages/fx-core/src/component/utils/metadataGraphPermssion.ts
```diff
@@ -0,0 +1,89 @@
...
+// Licensed under the MIT license.
+
+import path from "path";
+import fs from "fs-extra";
+import isUUID from "validator/lib/isUUID";
+import { MetadataV3 } from "../../common/versionMetadata";
+import { ProjectModel } from "../configManager/interface";
+import { AADManifest } from "../driver/aad/interface/AADManifest";
+import { getDetailedGraphPermissionMap, graphAppId, graphAppName } from "../driver/aad/permissions";
+import { actionName } from "../driver/teamsApp/configure";
```
blackchoey, line 11: should it be aadApp/update's action name?

xzf0587: Fixed it. Thanks.

packages/fx-core/src/component/utils/metadataGraphPermssion.ts
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    let hasRole = false;
+    let hasAdminScope = false;
+    const scopes: string[] = [];
+    graphPermission.resourceAccess?.forEach((access) => {
+      if (access.type === "Role") {
+        hasRole = true;
+      } else {
+        const id = isUUID(access.id) ? access.id : graphPermissionMap.scopes[access.id];
```
blackchoey, line 72: What will happen if user writes a wrong scope name here?

xzf0587: We will use the id/name to find the entry in graph permission map.
If the scope name is wrong, the id in line72 will be undefine and it will be skipped by line 73

blackchoey (approved):

jayzhang (approved):

PR #10294, opened 2023-11-07
fix: error component name for teamsapp management

jayzhang:
fix: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/25743883


xzf0587 (approved):

xzf0587:
minor: use a constant for the duplicated `TeamsAppMgr` string.

PR #10295, opened 2023-11-07
fix: not to install certification for doctor command

jayzhang:
fix: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/25743898

### DIFF
--- a/packages/fx-core/src/common/local/localCertificateManager.ts
+++ b/packages/fx-core/src/common/local/localCertificateManager.ts
@@ -66,10 +66,7 @@ export class LocalCertificateManager {
    * - Check cert store if trusted (thumbprint, expiration)
    * - Add to cert store if not trusted (friendly name as well)
    */
-  public async setupCertificate(
-    needTrust: boolean,
-    failIfNotFound = false
-  ): Promise<LocalCertificate> {
+  public async setupCertificate(needTrust: boolean, checkOnly = false): Promise<LocalCertificate> {
     const certFilePath = `${this.certFolder}/${LocalDebugCertificate.CertFileName}`;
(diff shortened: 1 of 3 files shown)


packages/fx-core/src/common/local/localCertificateManager.ts
```diff
@@ -90,22 +90,25 @@ export class LocalCertificateManager {
...
         if (failIfNotFound) {
-          localCert.found = false;
           return localCert;
         }
         // generate cert and key
         certThumbprint = await this.generateCertificate(certFilePath, keyFilePath);
       }
 
       if (needTrust) {
-        if (certThumbprint && (await this.verifyCertificateInStore(certThumbprint))) {
```
swatDong, line 103: Add this back? There's corner case `certThumbprint` undefined here due to generate error.

packages/fx-core/src/common/local/localCertificateManager.ts
```diff
@@ -90,22 +90,25 @@ export class LocalCertificateManager {
...
+        if (await this.verifyCertificateInStore(certThumbprint)) {
           // already trusted
           localCert.isTrusted = true;
           localCert.alreadyTrusted = true;
         } else {
           localCert.alreadyTrusted = false;
+          if (failIfNotFound) {
```
swatDong, line 109: It seems the param name does not match the case, how about:
- renaming to `readonly` (indicate the setup is a read-only operation, no generate and no trust)?
- or, adding another optional param `failIfNotVerified`?

jayzhang: renamed by: checkOnly

swatDong (approved):

Alive-Fish (approved):

PR #10301, opened 2023-11-08
refactor: add inner error into download sample error

yukun-dong:


packages/fx-core/src/component/generator/error.ts
```diff
@@ -50,29 +51,48 @@ export class UnzipError extends BaseComponentInnerError {
...
 }
 
+export function isApiLimitError(error: Error): boolean {
+  //https://docs.github.com/en/rest/overview/rate-limits-for-the-rest-api?apiVersion=2022-11-28#exceeding-the-rate-limit
+  return (
+    axios.isAxiosError(error) &&
+    error.response !== undefined &&
+    (error.response.status === 403 || error.response.status === 429) &&
+    error.response.headers !== undefined &&
+    error.response.headers["x-ratelimit-remaining"] === "0"
```
hund030, line 92: I think this condition can be simplified as
```
    axios.isAxiosError(error) &&
    [403, 429].includes(error.response?.status) &&
    error.response?.headers?.["x-ratelimit-remaining"] === "0"
```

packages/fx-core/src/component/generator/error.ts
```diff
@@ -50,29 +51,48 @@ export class UnzipError extends BaseComponentInnerError {
...
       errorSource,
       "UserError",
       "DownloadSampleApiLimitError",
       "error.generator.DownloadSampleApiLimitError",
-      [url]
+      [url],
+      undefined,
+      undefined,
+      undefined,
+      error
```
hund030, line 80: Have you verified that the inner error is included in telemetry?

packages/fx-core/src/component/generator/generator.ts
```diff
@@ -204,10 +211,10 @@ export async function sampleDefaultOnActionError(
         await fs.rm(context.destination, { recursive: true });
       }
       if (error instanceof BaseComponentInnerError) throw error.toFxError();
-      else if (error.message.includes("403")) {
-        throw new DownloadSampleApiLimitError(context.url!).toFxError();
+      else if (isApiLimitError(error)) {
+        throw new DownloadSampleApiLimitError(convertToUrl(context.sampleInfo!), error).toFxError();
```
hund030, line 215: What's different between context.url and convertToUrl(context.sampleInfo) ?

packages/fx-core/src/component/generator/error.ts
yukun-dong: The inner error is included but only it's message, stack info are included, I will see how to include other properties.

packages/fx-core/src/component/generator/generator.ts
yukun-dong: There are some changes for downloading sample https://github.com/OfficeDev/TeamsFx/pull/10238 to support '/' in branch, so now the context.url will be empty for downloading sample and we need to concat sample info to get the url.

hund030 (approved):

Siglud (approved):

sffamily (approved):

PR #10384, opened 2023-11-16
fix: api me from scratch cannot debug after upgrading to .net 8

huimiu:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/25770678

- Migrating .NET apps from the in-process model to the [isolated worker model](https://learn.microsoft.com/en-us/azure/azure-functions/migrate-dotnet-to-isolated-model?tabs=net8). Currently, our Azure Function runtime uses the in-process model, but .NET 8 is not yet supported on the in-process model.
- Change the data type of the "Id" field in the repair data model to `string` for consistency with VSC.

### DIFF
--- a/templates/csharp/copilot-plugin-from-scratch/Program.cs
+++ b/templates/csharp/copilot-plugin-from-scratch/Program.cs
@@ -1,0 +1,14 @@
+using Microsoft.Azure.Functions.Worker;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+
+var host = new HostBuilder()
+    .ConfigureFunctionsWorkerDefaults()
+    .ConfigureServices(services =>
+    {
+        services.AddApplicationInsightsTelemetryWorkerService();
+        services.ConfigureFunctionsApplicationInsights();
+    })
+    .Build();
+
+host.Run();
--- a/templates/csharp/copilot-plugin-from-scratch/Program.cs
+++ b/templates/csharp/copilot-plugin-from-scratch/Program.cs
@@ -1,14 +1,7 @@
-using Microsoft.Azure.Functions.Worker;
-using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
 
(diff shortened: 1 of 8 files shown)


templates/csharp/copilot-plugin-from-scratch/Program.cs
```diff
@@ -0,0 +1,14 @@
+using Microsoft.Azure.Functions.Worker;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+
+var host = new HostBuilder()
+    .ConfigureFunctionsWorkerDefaults()
+    .ConfigureServices(services =>
+    {
+        services.AddApplicationInsightsTelemetryWorkerService();
```
hund030, line 9: Do we create application insights for user?

templates/csharp/copilot-plugin-from-scratch/{{ProjectName}}.csproj.tpl
```diff
@@ -4,6 +4,7 @@
     <TargetFramework>{{TargetFramework}}</TargetFramework>
     <ImplicitUsings>enable</ImplicitUsings>
     <AzureFunctionsVersion>v4</AzureFunctionsVersion>
+    <OutputType>Exe</OutputType>
```
hund030, line 7: Does this strict that this project can only be run in Windows OS. What if we don't add this configuration?

huimiu: No, this does not mean that the function can only run on Windows OS, this indicates that we want to generate a .NET console app executable running in a separate process. If we remove this configuration, it will be a `.NET library` that can be used as a reference in other projects.

templates/csharp/copilot-plugin-from-scratch/Program.cs
huimiu: Removed all configurations for application insights, thanks!

hund030 (approved):

PR #10421, opened 2023-11-22
feat: detect project type in extension

jayzhang:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/25861452


packages/vscode-extension/src/extension.ts
```diff
@@ -73,7 +74,9 @@ export async function activate(context: vscode.ExtensionContext) {
   context.subscriptions.push(new ExtTelemetry.Reporter(context));
 
   VS_CODE_UI = new VsCodeUI(context);
-  initializeGlobalVariables(context);
+
+  await initializeGlobalVariables(context);
```
tecton, line 78: It's better to keep it as a synchronized function.

PR #10488, opened 2023-12-05
fix: upgrade button localization

Alive-Fish:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/24931765

image

### DIFF
--- a/packages/fx-core/src/core/middleware/projectMigratorV3.ts
+++ b/packages/fx-core/src/core/middleware/projectMigratorV3.ts
@@ -139,9 +139,9 @@ export const errorNames = {
   manifestTemplateInvalid: "ManifestTemplateInvalid",
   aadManifestTemplateNotExist: "AadManifestTemplateNotExist",
 };
-const upgradeButton = getLocalizedString("core.option.upgrade");
-export const moreInfoButton = getLocalizedString("core.option.moreInfo");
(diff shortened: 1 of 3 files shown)


packages/fx-core/src/core/middleware/projectMigratorV3.ts
```diff
@@ -139,9 +139,9 @@ export const errorNames = {
   manifestTemplateInvalid: "ManifestTemplateInvalid",
   aadManifestTemplateNotExist: "AadManifestTemplateNotExist",
 };
-const upgradeButton = getLocalizedString("core.option.upgrade");
-export const moreInfoButton = getLocalizedString("core.option.moreInfo");
-const migrationMessageButtons = [upgradeButton, moreInfoButton];
+const upgradeButton = () => getLocalizedString("core.option.upgrade");
```
blackchoey, line 142: Overall looks good to me, can you share some background about why we need to make it a function?

jayzhang (approved):

blackchoey (approved):

packages/fx-core/src/core/middleware/projectMigratorV3.ts
Alive-Fish: The `locale` is passed when running, so use `const string` will always use `locale=en` and cannot modify when `locale` has been changed.

PR #10513, opened 2023-12-12
feat: onboard API ME with key auth template

huimiu:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/25643763

When user scaffold an API-ME project from a new API, there will be an additional question to choose the authentication type.
image


packages/fx-core/src/question/create.ts
```diff
@@ -1586,6 +1605,20 @@ export function openAIPluginManifestLocationQuestion(): TextInputQuestion {
...
+  return {
+    type: "singleSelect",
+    name: QuestionNames.ApiMEAuth,
+    title: getLocalizedString("core.createProjectQuestion.apiMessageExtensionAuth.title"),
+    placeholder: getLocalizedString(
+      "core.createProjectQuestion.apiMessageExtensionAuth.placeholder"
+    ),
+    cliDescription: "The authentication type for the API.",
+    staticOptions: ApiMessageExtensionAuthOptions.all(),
+    default: "none",
```
yuqizhou77, line 1618: might be better to get value from option id directly rather than hard code here.

yuqizhou77 (approved):

packages/fx-core/src/question/create.ts
huimiu: Updated, thanks

yuqizhou77 (approved):

LongOddCode (approved):

sffamily (approved):

hund030 (approved):

PR #10637, opened 2024-01-05
perf(spec-parser): add adaptive card support

SLdragon:


packages/fx-core/src/common/spec-parser/generateTeamsAiMaterial.ts
```diff
@@ -45,13 +46,25 @@ export interface Augmentation {
...
 
+export interface AdaptiveCardResult {
+  name: string;
+  pathUrl: string;
+  data: AdaptiveCard;
+}
+
 const codeTemplate = `
 app.ai.action("{{operationId}}", async (context: TurnContext, state: ApplicationTurnState, parameter: any) => {
+  const petIdCard = require("./adaptiveCards/{{operationId}}.json");
```
KennethBWSong, line 57: rename petIdCard to something like cardTemplate?

KennethBWSong (approved):

PR #10638, opened 2024-01-05
refactor: update loading component for VS template

Yimin-Jin:


templates/csharp/sso-tab-ssr/Components/Pages/Welcome.razor
```diff
@@ -54,11 +54,12 @@
             var user = await teamsUserCredential.GetUserInfoAsync();
             userName = user.DisplayName;
         }
+        loading = false;
```
hund030, line 57: Will `loading` be true when failed to get username?

Yimin-Jin: No, if we want to report an error and end the loading, we need to add a try catch to solve the problem.

hund030 (approved):

PR #10661, opened 2024-01-09
fix: fix e2e dotnet SDK error

Siglud:
fix: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/26369819

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -171,9 +171,11 @@ jobs:
           func --version
 
       - name: Setup .net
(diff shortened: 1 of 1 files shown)


wenytang-ms (approved):

BasetsanaNjody:
@wenytang-ms @yinaa 
 The label's attribute doesn't match any element id. This might prevent the  browser from correctly auto-filing the form and accessibility tools from working correctly.
To fix this issue make sure the label's for attribute references the correct id of a form field.

PR #10702, opened 2024-01-17
refactor: update generator and ut based on new template versioning policy

Yimin-Jin:


packages/fx-core/src/component/generator/generatorAction.ts
```diff
@@ -116,19 +116,29 @@ export const downloadDirectoryAction: GeneratorAction = {
...
+    const url = await fetchTemplateZipUrl(
       context.language!,
       context.tryLimits,
       context.timeoutInMs
     );
+    let parts = url.split("/");
+    const selectedTag = parts[parts.length - 2];
+    parts = selectedTag.split(".");
+    // hotfix generates template based on url
+    if (parts[parts.length - 1] !== "0") {
```
yukun-dong, line 136: How about  `if(semver.patch(selectedTag)!==0)`

packages/fx-core/src/component/generator/generatorAction.ts
```diff
@@ -116,19 +116,29 @@ export const downloadDirectoryAction: GeneratorAction = {
...
+      context.url = url;
+    } else {
+      // stable generates templates based on the fallback
```
yukun-dong, line 139: Does this mean we will stop releasing stable version templates?

packages/fx-core/src/component/generator/generatorAction.ts
Yimin-Jin: That's a good idea.

packages/fx-core/src/component/generator/generatorAction.ts
Yimin-Jin: This means that if it's a stable release, templates will not be downloaded from github, but instead will use the local template(fallback).

hund030: @yukun-dong Yes. we only release templates when there is hotfix.

packages/fx-core/src/component/generator/generatorAction.ts
```diff
@@ -174,33 +202,12 @@ export const fetchTemplateFromLocalAction: GeneratorAction = {
...
-  },
-};
-
 export const TemplateActionSeq: GeneratorAction[] = [
   fetchTemplateZipFromSourceCodeAction,
-  fetchTemplateUrlWithTagAction,
+  fetchUrlForHotfixOnlyAction,
   fetchZipFromUrlAction,
   unzipAction,
   fetchTemplateFromLocalAction,
```
hund030, line 211: I think the new flow could be:
1. Set templates zip in context if debug flag enabled (We will deprecate this step later).
2. (Skip if zip set or useLocal flag is true) Get latest templates version from GitHub tag list and set tag in context if there is a hotfix.
3. (Skip if zip set or tag not set or useLocal flag is true) Download templates zip from GitHub tag and set zip.
4. (Skip if zip not set) unzip, set outputs.
5. (Skip if outputs set) unzip local templates zip.

You can make more changes in the source code if necessary.

packages/fx-core/src/component/generator/generatorAction.ts
```diff
@@ -116,19 +116,29 @@ export const downloadDirectoryAction: GeneratorAction = {
...
+export const fetchUrlForHotfixOnlyAction: GeneratorAction = {
   name: GeneratorActionName.FetchTemplateUrlWithTag,
   run: async (context: GeneratorContext) => {
     if (context.zip || context.url || context.cancelDownloading) {
       return;
     }
 
     context.logProvider.debug(`Fetching template url with tag: ${JSON.stringify(context)}`);
-    context.url = await fetchTemplateZipUrl(
+    const url = await fetchTemplateZipUrl(
```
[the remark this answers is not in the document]
Yimin-Jin: The reason for getting the url is that if it's the hotfix version, then we still need to call this function when getting the url.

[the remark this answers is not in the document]
Yimin-Jin: > You don't need to get the zip url in order to check if there is hotfix. Consider about
> 
> ```
> const latestVer = await selectTemplateTag(async () =>
>     (await fetchTagList(templateTagListURL, tryLimits, timeoutInMs)).replace(/\r/g, "").split("\n")
>   );
> if(semver.patch(latestVer)!=0) {...}
> ```
I think I can modify the fetchTemplateZipUrl function and subdivide it into two functions.

packages/fx-core/src/component/generator/generatorAction.ts
```diff
@@ -118,24 +119,26 @@ export const downloadDirectoryAction: GeneratorAction = {
...
+  name: GeneratorActionName.FetchUrlForHotfixOnly,
   run: async (context: GeneratorContext) => {
-    if (context.zip || context.url || context.cancelDownloading) {
+    if (context.zip || context.cancelDownloading) {
       return;
     }
 
     context.logProvider.debug(`Fetching template url with tag: ${JSON.stringify(context)}`);
-    const url = await fetchTemplateZipUrl(
+    const latestVer = await getTemplateLatestVersion(
```
hund030, line 129: This function returns a tag, not a function. So you either change the function name or update it to return a version.

packages/fx-core/src/component/generator/utils.ts
```diff
@@ -117,6 +117,15 @@ export async function fetchTemplateZipUrl(
...
   }
+  return selectedTag;
+}
+
+export async function fetchTemplateZipUrl(
+  name: string,
+  tryLimits = defaultTryLimits,
+  timeoutInMs = defaultTimeoutInMs
+): Promise<string> {
+  const selectedTag = await getTemplateLatestVersion(name, tryLimits, timeoutInMs);
```
hund030, line 128: This function sends two HTTP requests in one call. It's better to avoid calling this function twice.

packages/fx-core/src/component/generator/generatorAction.ts
hund030: To simplify the flow, maybe you can consider merging `fetchUrlForHotfixOnlyAction`, `fetchZipFromUrlAction`, `unzipAction` into one action.

packages/fx-core/src/component/generator/utils.ts
Yimin-Jin: I will optimize this function.

packages/fx-core/src/component/generator/generatorAction.ts
Yimin-Jin: I'll change the function name.

packages/fx-core/src/component/generator/generatorAction.ts
eriolchan: Agree to do some refactor to simplify Generator as well.

packages/fx-core/src/component/generator/generator.ts
```diff
@@ -177,7 +177,7 @@ export function templateDefaultOnActionError(
   error: Error
 ): Promise<void> {
   switch (action.name) {
-    case GeneratorActionName.FetchTemplateUrlWithTag:
```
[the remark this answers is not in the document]
Yimin-Jin: In this pr, there is no reduction in the number of actions, so there are no unused actions.

packages/fx-core/src/component/generator/generatorAction.ts
```diff
@@ -117,26 +120,25 @@ export const downloadDirectoryAction: GeneratorAction = {
...
-    if (parts[parts.length - 1] !== "0") {
-      context.url = url;
+    const localVer = getTemplateLocalVersion();
+    const latestVer = latestTag.split("@")[1];
+    // git tag version is higher than the local version, download template from github
+    if (semver.gt(latestVer, localVer)) {
+      context.url = getTemplateZipUrlByTag(context.language!, latestTag);
     } else {
-      // stable generates templates based on the fallback
+      // download template from fallback
```
hund030, line 141: The comment is inaccurate. We don't `download` template when using fallback.

Yimin-Jin: I'll use "generate" instead of "download".

packages/fx-core/src/component/generator/utils.ts
```diff
@@ -117,6 +117,23 @@ export async function fetchTemplateZipUrl(
...
     throw new Error(`Failed to find valid template for ${name}`);
   }
+  return selectedTag;
+}
+
+export function getTemplateLocalVersion(): string {
+  return templateConfig.localVersion;
+}
+
+export async function fetchTemplateZipUrl(
```
hund030, line 127: Is this function unused? Let's remove it if so.

Yimin-Jin: Yes, this function is only used in the ut. I will remove this function and modify the ut.

hund030 (approved):

packages/fx-core/src/component/generator/utils.ts
```diff
@@ -117,6 +117,14 @@ export async function fetchTemplateZipUrl(
   if (!selectedTag) {
     throw new Error(`Failed to find valid template for ${name}`);
   }
+  return selectedTag;
+}
+
+export function getTemplateLocalVersion(): string {
+  return templateConfig.localVersion;
```
eriolchan, line 124: Is there any backward compatibility need to consider?

PR #10743, opened 2024-01-23
refactor: update github workflow permission

xzf0587:


wenytang-ms (approved):

.github/workflows/check-assignee.yml
```diff
@@ -12,6 +12,8 @@ on:
 jobs:
   main:
     runs-on: ubuntu-latest
+    permissions:
+      issues: read
```
wenytang-ms, line 16: contents: read is also required
https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-azure

xzf0587: After testing in [zhaofeng/github/permission-field](https://github.com/OfficeDev/TeamsFx/tree/zhaofeng/github/permission-field) branch, I found it can success without the content permission. If this is not working after Feb 1, I will fix it then.

PR #10751, opened 2024-01-24
test: adjust cli preview in command-bot workflow-bot samples

ayachensiyuan:


packages/tests/src/ui-test/localdebug/localdebug-command-and-response.test.ts
```diff
@@ -50,80 +88,126 @@ describe("Command And Response Bot Local Debug Tests", function () {
...
+          Env.username,
+          Env.password
+        );
+        await validateBot(page, {
+          botCommand: "helloWorld",
+          expected: "Your Hello World App is Running",
+        });
+
+        // cli preview
+        console.log("======= debug with cli ========");
```
wenytang-ms, line 170: can we extract these codes into a common method? just pass the env, folderpath.. ?

wenytang-ms: i'm not sure how many these codes will be reused?

ayachensiyuan: cool... I took some of the common code and wrapped it into a reusable function. basically happy path case will add this step.

wenytang-ms (approved):

PR #10775, opened 2024-01-29
fix: regular expression performance in stack mask

jayzhang:
fix: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/26592411

### DIFF
--- a/packages/vscode-extension/test/extension/commonUtils.test.ts
+++ b/packages/vscode-extension/test/extension/commonUtils.test.ts
@@ -439,7 +439,7 @@ describe("CommonUtils", () => {
       chai.assert.equal(result, "");
     });
 
-    it("happy path", async () => {
+    it("happy path 1", async () => {
       const result = await commonUtils.anonymizeFilePaths(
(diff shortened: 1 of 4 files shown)


packages/vscode-extension/test/extension/commonUtils.test.ts
```diff
@@ -425,4 +425,37 @@ describe("CommonUtils", () => {
...
+      chai.assert.equal(result, "");
+    });
+
+    it("happy path 1", async () => {
+      const result = await commonUtils.anonymizeFilePaths(
+        "at Object.require.extensions.<computed> [as .ts] (C:\\Users\\AppData\\Roaming\\npm\\node_modules\\ts-node\\src\\index.ts:1621:12)"
+      );
+      chai.assert.equal(
+        result,
+        "at Object.require.extensions.<computed> [as .ts] (<REDACTED: user-file-path>:1621:12)"
```
tecton, line 448: Can we preserve the filename like `<REDACTED: user-file-path>index.ts:1621:12`? It should be helpful to our debugging from telemetry events.

tecton (approved):

PR #10778, opened 2024-01-29
ci: add permission to all-version-upgrade-test

wenytang-ms:


frankqianms (approved):
add github permission action read

PR #10779, opened 2024-01-29
ci: do not commit useLocal in template-config

hund030:
1. separate template version update and useLocal flag update to two steps and do not commit useLocal flag.
2. auto update localVersion.
3. remove fs-extra deps

### DIFF
--- a/.github/scripts/fxcore-sync-up-version.js
+++ b/.github/scripts/fxcore-sync-up-version.js
@@ -1,6 +1,6 @@
 const path = require("path");
 const semver = require("semver");
-const fse = require("fs-extra");
+const fs = require("fs");
 
 const templatePath = path.join(__dirname, "../../templates");
 const templateVersion = require(path.join(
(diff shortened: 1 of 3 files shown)


.github/scripts/fxcore-sync-up-version.js
```diff
@@ -1,44 +1,63 @@
...
+
+function syncTemplateVersion(templateVersion, templateConfigs) {
+  console.log(
+    `================== template version: ${templateVersion} ==================`
+  );
+
+  console.log(
+    `================== template version in fx-core configurate as ${templateConfigs.version} ==================`
+  );
+  templateConfigs.localVersion = templateVersion;
```
Yimin-Jin, line 39: Does it mean that we can use the templateVersion in package.json instead of the localVersion in template-config?

hund030: No. We cannot get package.json of templates from fx-core since they are different packages.

Yimin-Jin (approved):

.github/scripts/fxcore-sync-up-version.js
```diff
@@ -1,44 +1,63 @@
...
+      );
+      templateConfigs.version = templateVersionRange;
+    }
+  } else if (templateVersion.includes("rc")) {
+    console.log("sync up template in fx-core as 0.0.0-rc");
+    templateConfigs.version = "0.0.0-rc";
+  }
+}
+
+function updateUseLocalFlag(templateVersion, templateConfigs) {
```
eriolchan, line 55: According to new release principle, do we still need useLocal flag? Let me offline get more details from you.

PR #10786, opened 2024-01-30
refactor: tools may not be initialized for some CLI commands

nliu-ms:
For some CLI commands, TOOLS may be undefined without initialization, cause following error:

```
(×) Error: m365.UnhandledError: An unexpected error has occurred while performing the m365 task. {"stack":"TypeError: Cannot read properties of undefined (reading 'method')\n    at
onRejected (C:\\DevDiv\\Microsoft.Teams.Extensibility\\TeamsFx\\packages\\fx-core\\build\\common\\wrappedAxiosClient.js:73:38)\
n   
```

### DIFF
--- a/packages/fx-core/src/common/wrappedAxiosClient.ts
+++ b/packages/fx-core/src/common/wrappedAxiosClient.ts
@@ -55,7 +55,7 @@ export class WrappedAxiosClient {
       eventName = TelemetryEvent.DependencyApi;
     }
 
-    TOOLS.telemetryReporter?.sendTelemetryEvent(`${eventName}-start`, properties);
+    TOOLS?.telemetryReporter?.sendTelemetryEvent(`${eventName}-start`, properties);
     return request;
   }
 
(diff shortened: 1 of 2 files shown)


jayzhang (approved):

swatDong:
BTW, I saw in https://github.com/OfficeDev/TeamsFx/pull/10766 there are fixes for `this` -> `WrappedAxiosClient`. But in current code (https://github.com/OfficeDev/TeamsFx/blob/dev/packages/fx-core/src/common/wrappedAxiosClient.ts#L152) there are still some `this` calls. Are those expected?

nliu-ms:
> BTW, I saw in #10766 there are fixes for `this` -> `WrappedAxiosClient`. But in current code (https://github.com/OfficeDev/TeamsFx/blob/dev/packages/fx-core/src/common/wrappedAxiosClient.ts#L28, https://github.com/OfficeDev/TeamsFx/blob/dev/packages/fx-core/src/common/wrappedAxiosClient.ts#L152) there are still some `this` calls. Are those expected?

Those are expected. `this context` will be different inside a callback, so I updated reference in these 3 callback functions, other functions don't need it.
image

Another way is to use arrow functions.

PR #10789, opened 2024-01-30
perf: add test

KennethBWSong:


SLdragon (approved):

packages/spec-parser/tsconfig.json
```diff
@@ -1,10 +1,10 @@
 {
   "compileOnSave": false,
   "compilerOptions": {
-    "module": "ES2015",
+    "module": "ES6",
     "outDir": "dist",
     "declarationDir": "dist",
-    "target": "es2016",
+    "target": "ES6",
```
SLdragon, line 7: Do we need to change it back?

KennethBWSong: I will check and change back if necessary in next pr.

PR #10791, opened 2024-01-30
feat: open sample from url support

hermanwenhe:
Support using Uri to launch TTK and open designated sample detail page in sample gallery.

feature link: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/26879784

example link: [vscode://TeamsDevApp.ms-teams-vscode-extension?referrer=officedoc&sampleId=hello-world-teams-tab-and-outlook-add-in](url)

### DIFF
--- a/packages/vscode-extension/src/controls/sampleGallery/SampleGallery.tsx
+++ b/packages/vscode-extension/src/controls/sampleGallery/SampleGallery.tsx
@@ -70,13 +70,10 @@ export default class SampleGallery extends React.Component<unknown, SampleGaller
     );
     if (this.state.loading) {
       return <div className="sample-gallery">{titleSection}</div>;
-    } else if (this.state.selectedSampleId) {
-      const selectedSample = this.samples.filter(
-        (sample: SampleInfo) => sample.id == this.state.selectedSampleId
-      )[0];
+    } else if (this.selectedSample) {
       return (
         <SampleDetailPage
-          sample={selectedSample}
+          sample={this.selectedSample}
           selectSample={this.onSampleSelected}
           createSample={this.onCreateSample}
           viewGitHub={this.onViewGithub}
@@ -173,6 +170,16 @@ export default class SampleGallery extends React.Component<unknown, SampleGaller
     }
   }
 
+  private get selectedSample(): SampleInfo | null {
+    if (!this.state.selectedSampleId) {
+      return null;
+    }
+    const selectedSamples = this.samples.filter(
+      (sample: SampleInfo) => sample.id == this.state.selectedSampleId
+    );
+    return selectedSamples.length > 0 ? selectedSamples[0] : null;
+  }
+
   private messageHandler = (event: any) => {
     const message = event.data.message;
     switch (message) {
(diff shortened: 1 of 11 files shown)


packages/vscode-extension/src/uriHandler.ts
```diff
@@ -63,6 +65,14 @@ export class UriHandler extends vscode.EventEmitter<vscode.Uri> implements vscod
...
           }
         );
+    } else if (queryParamas.referrer === Referrer.OfficeDoc) {
+      if (!queryParamas.sampleId) {
+        void vscode.window.showErrorMessage(
+          localize("teamstoolkit.devPortalIntegration.invalidLink")
+        );
+        return;
+      }
+      void vscode.commands.executeCommand("fx-extension.openSamples", false, queryParamas.sampleId);
```
tecton, line 75: You could use `ExternalUrl` instead of `false` which is the `trigger-from` property in telemetry event. Otherwise, it will be treated as command palette.

packages/vscode-extension/src/controls/webviewPanel.ts
```diff
@@ -33,7 +33,7 @@ export class WebviewPanel {
   private panelType: PanelType = PanelType.SampleGallery;
   private disposables: vscode.Disposable[] = [];
 
-  public static createOrShow(panelType: PanelType, isToSide?: boolean) {
+  public static createOrShow(panelType: PanelType, isToSide?: boolean, args?: any[]) {
```
tecton, line 36: minor: merging `isToSide` and `args` might be better because `isToSide` can be derived from `args[0]`.

packages/vscode-extension/src/controls/sampleGallery/SampleGallery.tsx
```diff
@@ -194,6 +201,9 @@ export default class SampleGallery extends React.Component<unknown, SampleGaller
           });
         }
         break;
+      case Commands.OpenDesignatedSample:
+        this.onSampleSelected(event.data.sampleId, TelemetryTriggerFrom.ExternalUrl);
```
tecton, line 205: If the `sampleId` in external link is incorrect, there won't be any warnings. Users will see the sample gallery page instead of detail page.

tecton (approved):

packages/vscode-extension/src/controls/sampleGallery/SampleGallery.tsx
hermanwenhe: Yeah, I would see this as a fallback behavior. We will validate our links before putting them in docs. If anything goes wrong, directing to gallery page seems like the right behavior.

packages/vscode-extension/src/controls/webviewPanel.ts
hermanwenhe: Done

packages/vscode-extension/src/uriHandler.ts
hermanwenhe: Done

tecton (approved):

PR #10806, opened 2024-02-01
feat: delete aad app

jayzhang:
feat: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/25990580


packages/fx-core/src/component/driver/aad/utility/aadAppClient.ts
```diff
@@ -89,6 +89,33 @@ export class AadAppClient {
 
     return <AADApplication>response.data;
   }
+
```
KennethBWSong, line 92: Should we add a TODO to make sure will not forget to remove these parts when we make sure we will not use list api?

KennethBWSong (approved):

PR #10810, opened 2024-02-02
feat: add teams agent

1openwindow:
add teams agent extension


packages/teams-agent/src/codeToCloud.ts
```diff
@@ -0,0 +1,433 @@
+import axios from "axios";
```
wenytang-ms, line 1: please add copyright here

packages/teams-agent/src/extension.ts
```diff
@@ -0,0 +1,10 @@
+import * as vscode from "vscode";
```
wenytang-ms, line 1: copyright

PR #10828, opened 2024-02-05
Feature/customized tree view

xuruiyao-msft:


packages/fx-core/src/core/environment.ts
```diff
@@ -59,6 +60,17 @@ class EnvironmentManager {
       environmentNameManager.getLocalEnvName(),
     ];
   }
+
+  public async getOfficeLocalEnv(projectPath: string): Promise<Result<string[], FxError>> {
+    if (!(await fs.pathExists(projectPath))) {
+      return err(new FileNotFoundError("EnvironmentManager", projectPath));
+    }
+    const officeLocalManifest = path.join(projectPath, "manifest.xml");
```
hermanwenhe, line 68: Better use regex to match "manifest*.xml"

PR #10915, opened 2024-02-26
feat: add llm related placehodlers in generator

kuojianlu:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/26356837

### DIFF
--- a/packages/fx-core/src/component/generator/utils.ts
+++ b/packages/fx-core/src/component/generator/utils.ts
@@ -199,7 +199,16 @@ function escapeEmptyVariable(
 ): string[][] {
   const parsed = Mustache.parse(template, tags) as string[][];
   const tokens = deepCopy(parsed); // Mustache cache the parsed result. Modify the result in place may cause unexpected issue.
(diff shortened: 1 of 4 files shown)


packages/fx-core/src/component/generator/utils.ts
```diff
@@ -209,10 +218,14 @@ function escapeEmptyVariable(
       token[0] = "text";
       token[1] = tags[0] + value + tags[1];
       accShift += shift;
+    } else if (token[0] === "#") {
```
hund030, line 221: Could you add a unit test for this logic? 
An example: https://github.com/OfficeDev/TeamsFx/blob/2dd9386184e52c0d8531dbd4a20ea6c8087481ae/packages/fx-core/tests/component/generator/generator.test.ts#L658-L670

kuojianlu: Good point, added.

hund030 (approved):

PR #10974, opened 2024-03-04
refactor: remove unused error message

xiaolang124:
Remove 'silentAcquireToken' error, as the Teams toolkit automatically clears the cache when unable to obtain the token.

### DIFF
--- a/packages/vscode-extension/src/commonlib/codeFlowLogin.ts
+++ b/packages/vscode-extension/src/commonlib/codeFlowLogin.ts
@@ -332,14 +332,6 @@ export class CodeFlowLogin {
             }
           })
           .catch(async (error) => {
-            VsCodeLogInstance.error(
-              "[Login] " +
-                stringUtil.format(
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/commonlib/codeFlowLogin.ts
```diff
@@ -332,14 +332,6 @@ export class CodeFlowLogin {
             }
           })
           .catch(async (error) => {
-            VsCodeLogInstance.error(
```
swatDong, line 335: Instead of totally removing, is it possible to log as debug or verbose? At least we can debug ourselves...

swatDong (approved):

packages/vscode-extension/src/commonlib/codeFlowLogin.ts
swatDong: Anyway, ok to remove as well.

PR #11021, opened 2024-03-08
perf(spec parser): only scaffold adaptice cards for me projects

KennethBWSong:


packages/spec-parser/src/specParser.ts
```diff
@@ -328,28 +328,31 @@ export class SpecParser {
...
-              );
-              await fs.outputJSON(dataFileName, {}, { spaces: 2 });
-            } catch (err) {
-              result.allSuccess = false;
-              result.warnings.push({
-                type: WarningType.GenerateCardFailed,
-                content: (err as Error).toString(),
-                data: operation.operationId!,
-              });
+      if (isMe === undefined || isMe === true) {
```
SLdragon, line 331: How about to make adaptiveCardFolder optional?  If adaptiveCardFolder is undefined, then it will not generate adaptive cards.

Another suggestion is that currently we plan to support 3 different project types: copilot, teams ai, and SME, so it is better to use an Enum for project types?

SLdragon: I think you can just merge it, and I will optimize the whole spec parser lib

SLdragon (approved):

PR #11023, opened 2024-03-08
perf(sme): update to support local debug for api key

SLdragon:


packages/fx-core/src/component/driver/apiKey/create.ts
```diff
@@ -268,7 +268,7 @@ export class CreateApiKeyDriver implements StepDriver {
     const apiKey: ApiSecretRegistration = {
       description: args.name,
       targetUrlsShouldStartWith: domain,
-      applicableToApps: ApiSecretRegistrationAppType.SpecificApp,
+      applicableToApps: ApiSecretRegistrationAppType.AnyApp,
       specificAppId: args.appId,
```
KennethBWSong, line 272: Need to remove this line

KennethBWSong (approved):

PR #11087, opened 2024-03-14
fix: add treeview telemetry rebase to main branch

richard6094:
fix: add treeview telemetry rebase to main branch

### DIFF
--- a/packages/fx-core/src/common/projectTypeChecker.ts
+++ b/packages/fx-core/src/common/projectTypeChecker.ts
@@ -35,7 +35,7 @@ export interface ProjectTypeResult {
   manifestVersion?: string;
   dependsOnTeamsJs?: boolean;
   isSPFx?: boolean;
-  isOfficeAddin?: boolean;
+  officeProjectType?: string;
   lauguages: ("ts" | "js" | "csharp" | "java" | "python" | "c")[];
 }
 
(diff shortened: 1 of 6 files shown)


packages/fx-core/src/common/projectTypeChecker.ts
```diff
@@ -34,6 +35,7 @@ export interface ProjectTypeResult {
   manifestVersion?: string;
   dependsOnTeamsJs?: boolean;
   isSPFx?: boolean;
+  officeProjectType?: string;
```
jayzhang, line 38: Better to use `officeAddinProjectType` is more precise, because there are another concept that M365 app can run in office platform, which will cause confusion with the concept of Office Add-in.

packages/fx-core/src/common/telemetry.ts
```diff
@@ -140,6 +141,7 @@ export enum ProjectTypeProps {
   TeamsManifestCapabilities = "manifest-capabilities",
   TeamsJs = "teams-js",
   Lauguages = "languages",
+  OfficeProjectType = "office-project-type",
```
jayzhang, line 144: office-add-in-project-type

packages/fx-core/src/common/projectTypeChecker.ts
richard6094: Modified.

packages/fx-core/src/common/telemetry.ts
richard6094: Modified.

PR #11129, opened 2024-03-19
fix: fix ts templates source code location and stable the NODE version

Siglud:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/27110117/?view=edit

### DIFF
--- a/templates/ts/sso-tab-with-obo-flow/package.json.tpl
+++ b/templates/ts/sso-tab-with-obo-flow/package.json.tpl
@@ -2,7 +2,7 @@
   "name": "{{SafeProjectNameLowerCase}}",
   "version": "0.1.0",
   "engines": {
-    "node": "16 || 18"
+    "node": "18 || 20"
   },
   "private": true,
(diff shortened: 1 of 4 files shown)


templates/ts/sso-tab-with-obo-flow/package.json.tpl
```diff
@@ -2,7 +2,7 @@
   "name": "{{SafeProjectNameLowerCase}}",
   "version": "0.1.0",
   "engines": {
-    "node": "16 || 18"
+    "node": "18 || 20"
```
huimiu, line 5: Do we need to update the @types/node dependency version?

Siglud: I tested it on my local machine and it worked without any errors or warnings in the editor. If we had used any features from Node 20, we should have upgraded this dependency, but we didn't.

hund030 (approved):

PR #11130, opened 2024-03-19
fix: cli flickering

anchenyi:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/27092892


anchenyi:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/27092892

packages/cli/package.json
```diff
@@ -113,6 +114,7 @@
     "express": "^4.18.2",
     "figures": "^3.2.0",
     "fs-extra": "^9.1.0",
+    "inquirer": "^7.3.3",
```
jayzhang, line 117: new version of inquirer is 9.2.x, can we use new version?

packages/cli/src/userInteraction.ts
```diff
@@ -431,14 +435,16 @@ class CLIUserInteraction implements UserInteraction {
     if (config.validation || config.additionalValidationOnAccept) {
       validationFunc = async (input: string) => {
         let res: string | undefined = undefined;
+        const spinner = new CustomizedSpinner();
+        spinner.start();
```
jayzhang, line 439: Make sure the UX is OK in terminals with different backgroud color: especially for white and black.

jayzhang (dismissed):

packages/cli/package.json
anchenyi: New version causes the flicks, so we need to downgrade the version to 7.x to fix this issue.

packages/cli/src/userInteraction.ts
anchenyi: The UX works for white, black, and blue background color

jayzhang (approved):

PR #11152, opened 2024-03-21
refactor: code owner

jayzhang:
1. clean up code owners and sort by file path
2. add office add-in code owners (delegate by huajie and ning tang)


MSFT-yiz (commented):
should have at least two IC owners as backup for each other for code files. IMO, no need to put manager as file owner, as managers have admin privilege that can overwirte the policy.

jayzhang:
> should have at least two IC owners as backup for each other for code files. IMO, no need to put manager as file owner, as managers have admin privilege that can overwirte the policy.

Resolved

MSFT-yiz (approved):

Issue #11142, opened 2024-03-20
.NET 8 Teams Tab Config Save button never registered


kylejuliandev:
**Describe the bug**
.NET 8 Teams Tab Config Save button never registered. I believe this is to do with the new Blazor Rendering mode(s) that have come in .NET 8. The template(s) do not specify a rendering mode on TabConfig.cs (see [Template code](https://github.com/OfficeDev/TeamsFx/blob/dev/templates/csharp/sso-tab/Pages/TabConfig.razor)), but when I specify a rendermode of `InteractiveServer` the Save button is registered. I'm not sure I fully understand it myself, still getting my head round the rendermodes.

**To Reproduce**
1. Follow documentation on creating [Personal Tab project](https://learn.microsoft.com/en-us/microsoftteams/platform/tabs/how-to/create-personal-tab?tabs=vs&pivots=blazor-app#create-your-tab-project)
2. Debug the Tab app in Visual Studio
3. Click the Add to a chat button
4. Enter the name of the Person or Chat you want to add the Personal tab too
5. Click the Set up a tab button
6. The Save button is disabled

**Expected behavior**
Save button to be registered and Tab set up as expected.

**Screenshots**

Add Tab Application to Chat

![Add Custom Tab Application to Teams Chat](https://github.com/OfficeDev/TeamsFx/assets/38759683/89009a7e-0a07-4860-8290-edc0f30bb7ed)

Set up a tab

![Set up a tab](https://github.com/OfficeDev/TeamsFx/assets/38759683/cc1e69c1-3e87-497d-a5b0-bf76acf60e50)

Save button remains unregistered and disabled

![Save button remains unregistered and disabled](https://github.com/OfficeDev/TeamsFx/assets/38759683/c84c73cc-60e0-4dbd-ad82-c37880550b15)

kylejuliandev:
Additionally, I had to adjust the ContentUrl and WebsiteUrl to avoid duplicate slashes.

```c#
protected override async Task OnAfterRenderAsync(bool firstRender)
{
    if (firstRender)
    {
        var settings = new TeamsInstanceSettings
        {
            SuggestedDisplayName = "Tab",
            EntityId = _entityId.ToString(),
            ContentUrl = $"{NavigationManager.BaseUri}tab",
            WebsiteUrl = $"{NavigationManager.BaseUri}tab"
        };

        await MicrosoftTeams.InitializeAsync();
[...]

PR #11155, opened 2024-03-21
fix: fix .net8 Teams TabConfig save button never registered

Yimin-Jin:
git issue: https://github.com/OfficeDev/TeamsFx/issues/11142
Making the `Interactive Server` as the default mode globally.

### DIFF
--- a/templates/csharp/non-sso-tab-ssr/Components/Pages/TabConfig.razor
+++ b/templates/csharp/non-sso-tab-ssr/Components/Pages/TabConfig.razor
@@ -18,12 +18,13 @@
     {
         if(firstRender)
         {
+            var baseUri = new Uri(NavigationManager.BaseUri);
             var settings = new TeamsInstanceSettings
             {
                 SuggestedDisplayName = "My Tab",
                 EntityId = _entityId.ToString(),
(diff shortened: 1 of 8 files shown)


templates/csharp/non-sso-tab-ssr/Components/Pages/TabConfig.razor
```diff
@@ -22,8 +22,8 @@
             {
                 SuggestedDisplayName = "My Tab",
                 EntityId = _entityId.ToString(),
-                ContentUrl = $"{NavigationManager.BaseUri}/tab",
-                WebsiteUrl = $"{NavigationManager.BaseUri}/tab"
+                ContentUrl = $"{NavigationManager.BaseUri}tab",
```
hund030, line 25: Could you try the Uri class to combine the basic uri and path?
https://learn.microsoft.com/en-us/dotnet/api/system.uri.-ctor?view=net-8.0#system-uri-ctor(system-uri-system-string)

Yimin-Jin: Sure.

hund030 (approved):

PR #11189, opened 2024-03-26
fix: rollback to Azure storage to avoid deploy error

Siglud:
fix: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/27230918


templates/ts/office-addin/infra/azure.bicep
```diff
@@ -1,25 +1,27 @@
 @maxLength(20)
 @minLength(4)
 param resourceBaseName string
-param staticWebAppSku string
+param storageSku string
```
jayzhang, line 4: `templates/ts/office-json-addin` is copied from `office-addin`, should update too

jayzhang (approved):

PR #11190, opened 2024-03-26
fix: retire error-message

jayzhang:
fix: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/27316534

retire `error-message`, `inner-error`, `error-stack` property in telemetry for now and we will use another property name for `error-message` after we can make sure the content will not contains any 3rd party message.


packages/fx-core/tests/component/driver/aad/create.test.ts
```diff
@@ -482,9 +482,9 @@ describe("aadAppCreate", async () => {
     expect(endTelemetry.properties.success).to.equal("no");
     expect(endTelemetry.properties["error-code"]).to.equal("aadAppCreate.HttpClientError");
     expect(endTelemetry.properties["error-type"]).to.equal("user");
-    expect(endTelemetry.properties["error-message"]).to.equal(
-      'A http client error happened while performing the aadApp/create task. The error response is: {"error":{"code":"Request_BadRequest","message":"Invalid value specified for property \'displayName\' of resource \'Application\'."}}'
-    );
+    // expect(endTelemetry.properties["error-message"]).to.equal(
```
blackchoey, line 485: will there be another property to store error message?

jayzhang: We will use another property name after we can make sure that the error message is in our total control. Retire this for now.

blackchoey (approved):

tecton (approved):

xzf0587 (approved):

PR #11195, opened 2024-03-26
perf: update schema for apiKey

KennethBWSong:


SLdragon (dismissed):

blackchoey:
@jayzhang  do we also need to update `yaml-schema/yaml.schema.json`?

packages/fx-core/resource/yaml-schema/v1.5/yaml.schema.json
```diff
@@ -72,7 +72,8 @@
           { "$ref": "#/definitions/teamsAppCopyAppPackageToSPFx" },
```
jayzhang, line 72: make sure `packages/fx-core/resource/yaml-schema/yaml.schema.json` is also synced.

jayzhang (approved):

PR #11198, opened 2024-03-27
feat: enhance async validation logs

anchenyi:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/27231503


supkasar (approved):

packages/fx-core/src/component/configManager/constant.ts
```diff
@@ -9,6 +9,7 @@ export enum SummaryConstant {
   Succeeded = "(√) Done:",
   Failed = "(×) Error:",
   NotExecuted = "(!) Warning:",
+  Warning = "(!) Warning:",
```
nliu-ms, line 12: It looks like the same as NotExecuted?

anchenyi: "NotExecuted" is used elsewhere, and the term "NotExecuted" does not precisely match our validation warning scenario. Therefore, I have added a new enum called "Warning" with the same string value as "NotExecuted".

anchenyi:
New Output UI:
async_validation1

anchenyi:
Mocked some validation results with all types.
async_validation2

nliu-ms (approved):

jayzhang (approved):

supkasar (approved):

PR #11209, opened 2024-03-28
feat: add flag for controlling chat participant and disable it when release rc/beta

Alive-Fish:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/26624499

Add `TEAMSFX_CHAT_PARTICIPANT` flag and set it to `true` in the code. When releasing `rc/stable`, change it to false for hiding other entries of chat participant.

Set `contributes.chatParticipants` to `[]` to hide chat participant when releasing `rc/stable`.

Use `fx-extension.isChatParticipantEnabled` in `package.json` to control the welcome page content and webview content.

Test in https://github.com/OfficeDev/TeamsFx/actions/runs/8462480742.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -188,6 +188,23 @@ jobs:
           git add packages/fx-core/src/common/m365/serviceConstant.ts
           git commit -m "build: replace sideloading placeholders"
 
+      - name: disable chat participant in package.json
+        if: ${{ github.event_name == 'workflow_dispatch' && contains(fromJson('[\"rc\", \"stable\"]'), github.event.inputs.preid) }}
+        uses: jossef/action-set-json-field@v1
+        with:
+          file: ./packages/vscode-extension/package.json
+          field: contributes.chatParticipants
+          value: '[]'
+
+      - name: disable chat participant environment variable
+        if: ${{ github.event_name == 'workflow_dispatch' && contains(fromJson('[\"rc\", \"stable\"]'), github.event.inputs.preid) }}
+        run: bash .github/scripts/chat-participant-disabled.sh
+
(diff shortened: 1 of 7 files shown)


packages/vscode-extension/src/extension.ts
```diff
@@ -126,6 +132,15 @@ export async function activate(context: vscode.ExtensionContext) {
   // UI is ready to show & interact
   await vscode.commands.executeCommand("setContext", "fx-extension.isTeamsFx", isTeamsFxProject);
 
+  // control whether to show chat participant entries
+  await vscode.commands.executeCommand(
+    "setContext",
+    "fx-extension.isChatParticipantEnabled",
```
tecton, line 138: Will we use this variable in `package.json`?

.github/workflows/cd.yml
```diff
@@ -188,6 +188,23 @@ jobs:
           git add packages/fx-core/src/common/m365/serviceConstant.ts
           git commit -m "build: replace sideloading placeholders"
 
+      - name: disable chat participant in package.json
```
tecton, line 191: We could also determine whether the extension is prerelease based on fx-core version and vscode-extension version. Is that one more convenient?

packages/vscode-extension/src/extension.ts
Alive-Fish: Yes

.github/workflows/cd.yml
tecton: Confirmed that we also need to remove contributes.chatparticipants in package.json from script.

tecton (approved):

wenytang-ms (approved):

PR #11210, opened 2024-03-28
refactor: disable create office xml addin

ZYUN-MSFT:
refactor: disable create office xml addin

### DIFF
--- a/packages/fx-core/src/common/featureFlags.ts
+++ b/packages/fx-core/src/common/featureFlags.ts
@@ -59,7 +59,7 @@ export function isMultipleParametersEnabled(): boolean {
 }
 
 export function isOfficeXMLAddinEnabled(): boolean {
-  return isFeatureFlagEnabled(FeatureFlagName.OfficeXMLAddin, true);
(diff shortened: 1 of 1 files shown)


LongOddCode (dismissed):

packages/fx-core/src/question/inputs/CreateProjectInputs.ts
```diff
@@ -14,12 +14,7 @@ export interface CreateProjectInputs extends Inputs {
   /** @description Teams Toolkit: select runtime for your app */
   runtime?: "node" | "dotnet";
   /** @description New Project */
-  "project-type"?:
```
jayzhang, line 17: Suggest not change this file manually, which is generated automatically.

jayzhang: Can apply Yue's change in dev and run `npm run build` in `fx-core`, make sure static options contains all values. Only dynamic options depends on feature flag:
image

ZYUN-MSFT: DONE

jayzhang (approved):

PR #11217, opened 2024-03-28
feat: office content add-in[MetaOS]

YueLi-MSFT:
### Background:

- [PM Spec](https://microsoft.sharepoint.com/:w:/t/OfficePlatform/EQTzlom9CSFGgAOIro3-sDsBTsmVkXxz6IWCUoJVLp6L9w?e=vW0Dla)

- [ADO Link](https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/27130363)

When isOfficeJSONAddinEnabled(default false) is true:
image

### DIFF
--- a/packages/fx-core/src/component/generator/officeAddin/generator.ts
+++ b/packages/fx-core/src/component/generator/officeAddin/generator.ts
@@ -25,6 +25,7 @@ import { getLocalizedString } from "../../../common/localizeUtils";
 import { assembleError } from "../../../error";
 import {
   CapabilityOptions,
+  OfficeAddinHostOptions,
   ProjectTypeOptions,
   getOfficeAddinFramework,
 } from "../../../question/create";
@@ -102,12 +103,21 @@ export class OfficeAddinGenerator {
       | "typescript";
     const projectType = inputs[QuestionNames.ProjectType];
     const capability = inputs[QuestionNames.Capabilities];
-    const host: string =
-      projectType === ProjectTypeOptions.outlookAddin().id
-        ? "outlook"
-        : projectType === ProjectTypeOptions.officeAddin().id
-        ? "wxpo" // wxpo - support word, excel, powerpoint, outlook
-        : inputs[QuestionNames.OfficeAddinHost];
+    const inputHost = inputs[QuestionNames.OfficeAddinHost];
+    let host: string = inputHost;
+    if (
+      projectType === ProjectTypeOptions.outlookAddin().id ||
+      (projectType === ProjectTypeOptions.officeXMLAddin().id &&
+        inputHost === OfficeAddinHostOptions.outlook().id)
+    ) {
+      host = "outlook";
+    } else if (projectType === ProjectTypeOptions.officeAddin().id) {
+      if (capability === "json-taskpane") {
+        host = "wxpo"; // wxpo - support word, excel, powerpoint, outlook
+      } else if (capability === CapabilityOptions.officeContentAddin().id) {
+        host = "xp"; // content add-in support excel, powerpoint
+      }
+    }
     const workingDir = process.cwd();
     const importProgressStr =
       projectType === ProjectTypeOptions.officeAddin().id
--- a/packages/fx-core/src/question/create.ts
+++ b/packages/fx-core/src/question/create.ts
@@ -1360,13 +1360,14 @@ export function officeAddinHostingQuestion(): SingleSelectQuestion {
   };
 }
 
-export function OfficeAddinFrameworkQuestion(): SingleSelectQuestion {
+export function officeAddinFrameworkQuestion(): SingleSelectQuestion {
   return {
     type: "singleSelect",
     name: QuestionNames.OfficeAddinFramework,
     cliShortName: "f",
     cliDescription: "Framework for WXP extension.",
     title: getLocalizedString("core.createProjectQuestion.projectType.officeAddin.framework.title"),
+    dynamicOptions: getAddinFrameworkOptions,
     staticOptions: [
       { id: "default", label: "Default" },
       { id: "react", label: "React" },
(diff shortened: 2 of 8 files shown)


jayzhang (approved):
LGTM.

hermanwenhe (approved):
LGTM

packages/fx-core/src/question/create.ts
```diff
@@ -2495,14 +2524,16 @@ export function capabilitySubTree(): IQTreeNode {
...
+        // Office addin framework for json manifest
+        data: officeAddinFrameworkQuestion(),
         condition: (inputs: Inputs) => {
           return (
-            inputs[QuestionNames.ProjectType] === ProjectTypeOptions.officeAddin().id &&
-            inputs[QuestionNames.Capabilities] !== CapabilityOptions.officeAddinImport().id
+            (inputs[QuestionNames.ProjectType] === ProjectTypeOptions.officeAddin().id &&
+              inputs[QuestionNames.Capabilities] !== CapabilityOptions.officeAddinImport().id) ||
+            (inputs[QuestionNames.ProjectType] === ProjectTypeOptions.officeAddin().id &&
+              inputs[QuestionNames.Capabilities] !== CapabilityOptions.officeContentAddin().id)
```
sigmarising, line 2534: These conditions can be simplified.

ZYUN-MSFT, line 2534: These conditions can be simplify.

packages/fx-core/src/question/create.ts
```diff
@@ -1357,22 +1357,51 @@ export function officeAddinHostingQuestion(): SingleSelectQuestion {
   };
 }
 
-export function OfficeAddinFrameworkQuestion(): SingleSelectQuestion {
-  return {
+export function officeAddinFrameworkQuestion(): SingleSelectQuestion {
+  const officeAddinFrameworkQuestion: SingleSelectQuestion = {
```
ZYUN-MSFT, line 1361: Just wonder that why just follow the original style to return it directly to keep the change as samll as possible?

packages/fx-core/src/question/create.ts
```diff
@@ -1357,22 +1357,51 @@ export function officeAddinHostingQuestion(): SingleSelectQuestion {
...
       { id: "default", label: "Default" },
       { id: "react", label: "React" },
     ],
     placeholder: getLocalizedString(
       "core.createProjectQuestion.projectType.officeAddin.framework.placeholder"
     ),
-    default: "default",
+    default: (inputs: Inputs) => {
+      return getAddinFrameworkOptions(inputs)[0].id;
+    },
```
ZYUN-MSFT, line 1377: Is this really necessary? I see this code will always return `"default"`.

packages/fx-core/src/component/generator/officeAddin/generator.ts
```diff
@@ -102,12 +103,19 @@ export class OfficeAddinGenerator {
...
+    let host: string = inputHost;
+    if (
+      projectType === ProjectTypeOptions.outlookAddin().id ||
+      (projectType === ProjectTypeOptions.officeXMLAddin().id &&
+        inputHost === OfficeAddinHostOptions.outlook().id)
+    ) {
+      host = "outlook";
+    } else if (capability === CapabilityOptions.officeContentAddin().id) {
+      host = "xp"; // content add-in support excel, powerpoint
+    } else if (capability === "json-taskpane") {
```
ZYUN-MSFT, line 116: Just curious that, why just make `content add-in` (but not include `json-taskpane`) avaliable in `CapabilityOptions`?

ZYUN-MSFT (changes requested):
If the create/import outlook add-in works normal when OfficeXMLAddin is enabled. I think no more change is needed. 

But I suggest to improve the comment area to improve code readability.

packages/fx-core/src/component/generator/officeAddin/generator.ts
YueLi-MSFT: `json-taskpane` exists by default. It is defined in `OfficeAddinProjectConfig` after Huajie's refactor. See it on Create.ts 560 lines.

ZYUN-MSFT: What I means is that why not add `json-taskpane` as a static function in `CapabilityOptions` (file create.ts).

Keep the condition to both use static class or string will look more readbility.

packages/fx-core/src/question/create.ts
YueLi-MSFT: Yeah, it always return "default", just only not to write it by hard code.

packages/fx-core/src/component/generator/officeAddin/generator.ts
ZYUN-MSFT: I see it. But in my opinion, these part look really confuse since the capability come from static class and dynamic push item.

Consider refactor it later on your demand.

These two condtions use the same right judgement format from static class or string will look better.

packages/fx-core/src/question/create.ts
YueLi-MSFT: Will do.

packages/fx-core/src/question/create.ts
ZYUN-MSFT: Sure, If this logic is keep for further use, please just resolve this comment.

packages/fx-core/src/question/create.ts
YueLi-MSFT: Just only as consistent as possible with other functions.

ZYUN-MSFT (approved):

jayzhang (approved):

jayzhang (approved):

PR #11229, opened 2024-03-29
feat: avoid aync validation conflicts

anchenyi:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/27132725


anchenyi:
If a user submits another async validation request while the previous one is still in progress, a warning will be shown in the output, and the second submission will be cancelled.


image

packages/fx-core/src/component/driver/teamsApp/validateTestCases.ts
```diff
@@ -99,7 +99,39 @@ export class ValidateWithTestCasesDriver implements StepDriver {
...
       const appStudioToken = appStudioTokenRes.value;
-
+      // Check if the app has ongoing validation
+      const existingValidationResponse = await AppStudioClient.getAppValidationRequestList(
+        manifest.id,
+        appStudioToken
+      );
+      if (existingValidationResponse.appValidations) {
+        for (const validation of existingValidationResponse.appValidations) {
+          if (validation.status === AsyncAppValidationStatus.InProgress) {
```
nliu-ms, line 109: Other status may be considered, e.g. Created

anchenyi: Yes, I will add it. Thanks

nliu-ms (approved):

supkasar (approved):

PR #11250, opened 2024-04-02
perf(spec-parser): separate api validation logic

SLdragon:


KennethBWSong (approved):

yuqizhou77 (approved):
not sure if list() needed to be updated to use the validateServer but it also has some validation against server url.

SLdragon:
> not sure if list() needed to be updated to use the validateServer but it also has some validation against server url.

Thanks, seems validateServer can be removed since it is already contains in the validator, will update it in next pr for updating spec validation

adashen (approved):

PR #11254, opened 2024-04-02
test: use cli to provision

hellyzh:


packages/tests/src/utils/cleanHelper.ts
```diff
@@ -573,7 +573,7 @@ export async function createResourceGroup(
 
 export async function createResourceGroupByName(
   name: string,
-  location = "eastus"
```
wenytang-ms, line 576: why change here?

hellyzh: reverted the change for default region and set the region in provision function. TTK provision also use westus.
image

wenytang-ms (approved):

PR #11296, opened 2024-04-07
fix: hide test tool from local debug message if tab app is scaffolded

qinezh:
work item: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/27230806/

When a Teams app is scaffolded, TTK will display a message in the right bottom corner to prompt user to local debug. 

In this PR, `debug in Test Tool` will be hide from the message if Test Tool is not enabled in the scaffolded Teams app.


packages/vscode-extension/src/handlers.ts
```diff
@@ -1410,6 +1404,33 @@ export async function showLocalDebugMessage() {
   });
 }
 
+export async function getLocalDebugMessageTemplate(isWindows: boolean): Promise<string> {
```
tecton, line 1407: Could these 2 functions be moved to another file like `utils/*.ts`? I'm trying to keep the `handlers.ts` small.

tecton (approved):

supkasar (approved):

PR #11298, opened 2024-04-08
fix: confirm before adding API

yuqizhou77:
- confirm with user before adding apis
image


packages/fx-core/resource/package.nls.json
```diff
@@ -1,4 +1,6 @@
 {
+  "core.addApi.confirm":"Teams Toolkit will modify files in your \"%s\" folder based on the new OpenAPI description document you provided. To prevent unexpected change loss, we recommend you back up your files or use git for change tracking before proceeding",
```
supkasar, line 2: Proposed: Teams Toolkit will modify files in your \"%s\" folder based on the new OpenAPI document you provided. To avoid losing unexpected changes, back up your files or use git for change tracking before proceeding.

yuqizhou77: Thanks, updated

supkasar (approved):

jayzhang (approved):

PR #11317, opened 2024-04-09
test: fix ut and add cases

yiqing-zhao:


packages/vscode-extension/test/extension/handlers.test.ts
```diff
@@ -592,6 +592,7 @@ describe("handlers", () => {
       });
 
       const res = await handlers.openConfigStateFile([]);
+      await fs.remove(tmpDir);
```
tecton, line 595: minor: It's better to put this in afterEach() or use a library in case the test case fails.

tecton (approved):

PR #11353, opened 2024-04-11
test: Add e2e test support for contianer app

Yimin-Jin:
Successful workflow:
https://github.com/OfficeDev/TeamsFx/actions/runs/8642767495

### DIFF
--- a/packages/tests/src/commonlib/containeraAppValidator.ts
+++ b/packages/tests/src/commonlib/containeraAppValidator.ts
@@ -1,0 +1,72 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { AzureScopes } from "@microsoft/teamsfx-core/build/common/tools";
+import * as chai from "chai";
+
+import MockAzureAccountProvider from "@microsoft/teamsapp-cli/src/commonlib/azureLoginUserPassword";
+import { EnvConstants } from "./constants";
+
+import {
+  getContainerAppProperties,
+  getSubscriptionIdFromResourceId,
+  getResourceGroupNameFromResourceId,
+} from "./utilities";
+import { Executor } from "../utils/executor";
+import { Env } from "../utils/env";
+
+export class ContainerAppValidator {
+  private ctx: any;
+  private subscriptionId: string;
+  private rg: string;
+  private containerAppName: string;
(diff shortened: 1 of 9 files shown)


packages/tests/src/commonlib/containeraAppValidator.ts
```diff
@@ -0,0 +1,72 @@
...
+  private containerAppName: string;
+
+  constructor(ctx: any) {
+    console.log("Start to init validator for Azure Container App.");
+
+    this.ctx = ctx;
+
+    const resourceId = ctx[EnvConstants.AZURE_CONTAINER_APP_RESOURCE_ID];
+    chai.assert.exists(resourceId);
+    this.subscriptionId = getSubscriptionIdFromResourceId(resourceId);
```
hund030, line 31: I think we can get subscription id from ctx directly. Is it necessary to parse the resourceId?

hund030 (dismissed):

packages/tests/src/commonlib/containeraAppValidator.ts
Yimin-Jin: I've tried to retrieve the resourceId directly from ctx, but I wasn't able to find it. The content included in ctx is as shown in the screenshot.
image

LongOddCode (dismissed):

hund030 (approved):

packages/tests/src/utils/executor.ts
```diff
@@ -49,9 +51,14 @@ export class Executor {
     }
   }
 
-  static login() {
-    const command = `az login --service-principal -u ${process.env.AZURE_CLIENT_ID} -p ${process.env.AZURE_CLIENT_SECRET} -t ${process.env.AZURE_TENANT_ID}`;
-    return this.execute(command, process.cwd());
+  static async login() {
+    const command = `az login -u ${Env["azureAccountName"]} -p '${Env["azureAccountPassword"]}'`;
+    await Executor.execute(command, process.cwd());
```
Yimin-Jin, line 56: The reason the return value here is false is that no subscription was specified, and I'm not sure if this will have any effect.
image

hund030 (approved):

LongOddCode (approved):

PR #11358, opened 2024-04-12
refactor: add default template generator

hund030:


packages/fx-core/src/component/generator/templates/templateGenerator.ts
```diff
@@ -0,0 +1,223 @@
...
+      openAIKey: openAIKey ?? "",
+      azureOpenAIKey: azureOpenAIKey ?? "",
+      azureOpenAIEndpoint: azureOpenAIEndpoint ?? "",
+      isNewProjectTypeEnabled: isNewProjectTypeEnabled() ? "true" : "",
+      NewProjectTypeName: process.env.TEAMSFX_NEW_PROJECT_TYPE_NAME ?? "TeamsApp",
+      NewProjectTypeExt: process.env.TEAMSFX_NEW_PROJECT_TYPE_EXTENSION ?? "ttkproj",
+    };
+  }
+
+  private getTemplateName(inputs: Inputs) {
```
yukun-dong, line 134: How about moving this function into `templateNames.ts`, so that when onboarding new templates this `templateGenerator.ts` can remain unchanged.

yukun-dong (dismissed):

packages/fx-core/src/component/generator/templates/templateGenerator.ts
hund030: Good idea. I agree that we should decouple the `getTemplateName` logic from generator, and I plan to refactor this logic in next PR.

yukun-dong (approved):

packages/fx-core/src/component/coordinator/index.ts
```diff
@@ -325,6 +209,25 @@ class Coordinator {
         } else {
           warnings = res.value.warnings;
         }
+      } else if (isNewGeneratorEnabled()) {
+        const generators = Generators.filter((g) => g.activate(context, inputs));
```
Siglud, line 213: Use `find` instead of `filter` will be more efficient and easier to understand.

packages/fx-core/src/component/generator/templates/templateGenerator.ts
Yimin-Jin, line 1: I'm curious as to why the TemplateGenerator remains as an interface. 
For instance, suppose I have a new Generator and I just want to carry out some actions post-generation. In the case of inheritance via 'extend', I might only need to override the 'activate' and 'post' methods. Could you explain the reasoning behind this design choice?

hund030: Good question, I have the same feeling that the interface is unnecessary. So, I removed it.

packages/fx-core/src/component/generator/templates/templateGenerator.ts
```diff
@@ -0,0 +1,223 @@
...
+  @hooks([
+    ActionExecutionMW({
+      enableProgressBar: true,
+      progressTitle: ProgressTitles.create,
+      progressSteps: 1,
+      enableTelemetry: true,
+      telemetryEventName: TelemetryEvent.GenerateTemplate,
+    }),
+  ])
+  public async run(
```
eriolchan, line 48: It seems that run() is the coordinator of all functions of generator provider and there are only 3 extension points: activate(), mapTemplateInfos() and post(). Is there a better way to manage this so that developers will not override run() flow by accident?

packages/fx-core/src/component/generator/templates/templateGenerator.ts
```diff
@@ -0,0 +1,223 @@
...
+export class DefaultTemplateGenerator implements TemplateGenerator {
+  componentName = componentName;
+
+  public activate(ctx: Context, inputs: Inputs): boolean {
+    return Object.keys(Feature2TemplateName).some((feature) =>
+      feature.startsWith(inputs.capabilities)
+    );
+  }
+
+  public getTemplateInfos(
```
eriolchan: what about mapTemplateInfos()?

packages/fx-core/src/component/generator/templates/templateGenerator.ts
```diff
@@ -0,0 +1,223 @@
...
+  public post(
+    ctx: Context,
+    inputs: Inputs,
+    destinationPath: string,
+    actionContext?: ActionContext
+  ): Promise<Result<undefined, FxError>> {
+    return Promise.resolve(ok(undefined));
+  }
+
+  public getDefaultReplaceMap(inputs: Inputs): { [key: string]: string } {
```
eriolchan: how will developers use/extend this function? Does this need to be public?

packages/fx-core/src/component/generator/templates/templateGenerator.ts
eriolchan: +1

packages/fx-core/src/component/generator/templates/templateGenerator.ts
```diff
@@ -0,0 +1,223 @@
...
+
+  private getTemplateName(inputs: Inputs) {
+    const language = inputs[QuestionNames.ProgrammingLanguage];
+    const capability = inputs.capabilities as string;
+    const meArchitecture = inputs[QuestionNames.MeArchitectureType] as string;
+    const apiMEAuthType = inputs[QuestionNames.ApiMEAuth] as string;
+    const trigger = inputs[QuestionNames.BotTrigger] as string;
+    let feature = `${capability}:${trigger}`;
+    if (
+      capability === CapabilityOptions.m365SsoLaunchPage().id ||
```
eriolchan, line 132: Since all templates has support both Teams and M365, we can take a look at if this flag is still necessary for app scenarios, and we can refactor it as well.

packages/fx-core/src/component/generator/templates/templateGenerator.ts
```diff
@@ -0,0 +1,223 @@
...
+    if (meArchitecture) {
+      feature = `${feature}:${meArchitecture}`;
+    }
+    if (
+      inputs.targetFramework &&
+      inputs.targetFramework !== "net6.0" &&
+      inputs.targetFramework !== "net7.0" &&
+      (capability === CapabilityOptions.nonSsoTab().id || capability === CapabilityOptions.tab().id)
+    ) {
+      feature = `${capability}:ssr`;
```
eriolchan, line 155: Looks most feature map use capability and custom options, is it easy to be refactor as real map table?

packages/fx-core/src/component/generator/templates/templateGenerator.ts
```diff
@@ -73,35 +46,48 @@ export class DefaultTemplateGenerator implements TemplateGenerator {
     }),
   ])
   public async run(
```
Yimin-Jin, line 48: The entry point for overwrite methods is the `run()` method. Therefore, would it be feasible to consolidate parameters such as context, inputs, and actionContext during the execution of the `run(`) method? This could ensure a uniform context.

packages/fx-core/src/component/generator/templates/templateNames.ts
```diff
@@ -0,0 +1,126 @@
...
+  CustomCopilotRagCustomize = "custom-copilot-rag-customize",
+  CustomCopilotRagAzureAISearch = "custom-copilot-rag-azure-ai-search",
+  CustomCopilotRagCustomApi = "custom-copilot-rag-custom-api",
+  CustomCopilotRagMicrosoft365 = "custom-copilot-rag-microsoft365",
+  CustomCopilotAssistantNew = "custom-copilot-assistant-new",
+  CustomCopilotAssistantAssistantsApi = "custom-copilot-assistant-assistants-api",
+}
+
+export const Feature2TemplateName = {
+  [`${CapabilityOptions.nonSsoTab().id}:undefined`]: TemplateNames.Tab,
```
eriolchan, line 54: Look the format of key is like "${capability}:${option}", can we make it as a real map table to avoid string concatenation issue

packages/fx-core/src/common/constants.ts
```diff
@@ -68,4 +68,5 @@ export class FeatureFlagName {
   static readonly AsyncAppValidation = "TEAMSFX_ASYNC_APP_VALIDATION";
   static readonly NewProjectType = "TEAMSFX_NEW_PROJECT_TYPE";
   static readonly ChatParticipant = "TEAMSFX_CHAT_PARTICIPANT";
+  static readonly NewGenerator = "TEAMSFX_NEW_GENERATOR";
```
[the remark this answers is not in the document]
hund030: After all generator refactoring done, we can enable the feature flag by default in next prerelease and remove the legacy code in the prerelease after the next prerelease.

packages/fx-core/src/component/coordinator/index.ts
```diff
@@ -74,122 +71,9 @@ import { metadataUtil } from "../utils/metadataUtil";
...
-  }`]: TemplateNames.CustomCopilotRagMicrosoft365,
-  [`${CapabilityOptions.customCopilotAssistant().id}:undefined:${
-    CustomCopilotAssistantOptions.new().id
-  }`]: TemplateNames.CustomCopilotAssistantNew,
-  [`${CapabilityOptions.customCopilotAssistant().id}:undefined:${
-    CustomCopilotAssistantOptions.assistantsApi().id
-  }`]: TemplateNames.CustomCopilotAssistantAssistantsApi,
-};
+import { Generators } from "../generator/generatorProvider";
+import { Feature2TemplateName } from "../generator/templates/templateNames";
```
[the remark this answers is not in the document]
hund030: Yes. I will refine the map logic in next PR. This PR focus on the implementation of new template generator class.

packages/fx-core/src/component/coordinator/index.ts
```diff
@@ -325,6 +209,25 @@ class Coordinator {
...
           warnings = res.value.warnings;
         }
+      } else if (isNewGeneratorEnabled()) {
+        const generators = Generators.filter((g) => g.activate(context, inputs));
+        if (generators.length === 0) {
+          return err(new MissingRequiredInputError(QuestionNames.Capabilities, "coordinator"));
+        }
+        const res = await generators[0].run(context, inputs, projectPath);
+        if (res.isErr()) return err(res.error);
+        if (inputs[QuestionNames.CustomCopilotRag] === CustomCopilotRagOptions.customApi().id) {
```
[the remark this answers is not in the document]
hund030: I think it will not be. I will refine the logic in next PR.

packages/fx-core/src/component/generator/templates/templateGenerator.ts
hund030: In TypeScript, there is no way to prevent a child class from overriding a public method of its parent class. I think this can only be done by documentation.

packages/fx-core/src/component/generator/templates/templateGenerator.ts
hund030: Yes. Let me remove it in next PR.

packages/fx-core/src/component/generator/templates/templateNames.ts
hund030: Yes. I will refine the map logic in next PR. This PR focus on the implementation of new template generator class.

packages/fx-core/src/component/generator/templates/templateGenerator.ts
hund030: Yes. I will refine the map logic in next PR. This PR focus on the implementation of new template generator class.

packages/fx-core/src/component/coordinator/index.ts
hund030: Updated, thanks~

packages/fx-core/src/component/generator/templates/templateGenerator.ts
hund030: I make this method a common function, so generators can call it to get a default replace map.

packages/fx-core/src/component/generator/templates/templateGenerator.ts
hund030: `mapTemplateInfos` does not seem to be better for me.

packages/fx-core/src/component/generator/templates/templateGenerator.ts
hund030: I didn't get what you mean. Could you provide an example?

Yimin-Jin: Could we possibly consider writing it as: 
`run(
    context: Context,
    inputs: Inputs,
    destinationPath: string,
    actionContext?: ActionContext
){
    this.ctx = context;
    this.inputs = inputs;
}`

hund030: TTK does not block users from creating two projects at the same time. Making the generator stateful may cause unexpected issues.

yukun-dong (approved):

PR #11377, opened 2024-04-16
perf(spec-parser): add namespace for plugin.json

SLdragon:


packages/spec-parser/src/manifestUpdater.ts
```diff
@@ -212,6 +212,7 @@ export class ManifestUpdater {
         schema_version: "v2",
```
yuqizhou77, line 212: should we change it to 2.1?

yuqizhou77 (dismissed):

packages/spec-parser/src/manifestUpdater.ts
SLdragon: OK,sure~

yuqizhou77 (approved):

KennethBWSong (approved):

PR #11413, opened 2024-04-19
perf(oauth): add question model for oauth

KennethBWSong:


packages/fx-core/src/question/other.ts
```diff
@@ -994,3 +994,58 @@ export function apiSpecApiKeyQuestion(): IQTreeNode {
...
+export function OauthClientSecretConfirmQestion(): ConfirmQuestion {
+  return {
+    name: QuestionNames.OauthClientSecretConfirm,
+    title: getLocalizedString("core.createProjectQuestion.OauthClientSecretConfirm"),
+    type: "confirm",
+    default: true,
+  };
+}
+
+export function oauthClientSecretQuestion(): IQTreeNode {
```
SLdragon, line 1007: Maybe we should also ask user for clientId?

packages/fx-core/src/question/other.ts
```diff
@@ -994,3 +994,58 @@ export function apiSpecApiKeyQuestion(): IQTreeNode {
...
+      cliDescription: "Oauth client secret for OpenAPI spec.",
+      forgetLastValue: true,
+      validation: {
+        validFunc: (input: string): string | undefined => {
+          const pattern = /^(\w){10,128}/g;
+          const match = pattern.test(input);
+
+          const result = match
+            ? undefined
+            : getLocalizedString("core.createProjectQuestion.invalidApiKey.message");
```
SLdragon, line 1023: rename invalidApiKey.message to invalidSecret?

KennethBWSong: Kept here since the message is the same. Will update if we want to have difference message.

SLdragon (approved):

jayzhang (dismissed):

packages/fx-core/resource/package.nls.json
```diff
@@ -363,6 +363,9 @@
   "core.createProjectQuestion.OpenAIPluginDomain.placeholder": "Enter your website domain or manifest URL",
   "core.createProjectQuestion.ApiKey": "Enter API Key in OpenAPI Description Document",
   "core.createProjectQuestion.ApiKeyConfirm": "Teams Toolkit will upload the API key to Teams Developer Portal. The API key will be used by Teams client to securely access your API in runtime when API-Based Message Extension is triggered. Teams Toolkit will not store your API key.",
+  "core.createProjectQuestion.OauthClientId": "Enter client id for Oauth registration in OpenAPI Description Document",
+  "core.createProjectQuestion.OauthClientSecret": "Enter client secret for Oauth registration in OpenAPI Description Document",
+  "core.createProjectQuestion.OauthClientSecretConfirm": "Teams Toolkit will upload the client id/secret for Oauth Registration to Teams Developer Portal. The client id/secret will be used by Teams client to securely access your API in runtime when API-Based Message Extension is triggered. Teams Toolkit will not store your client id/secret.",
```
supkasar, line 368: Revised:
Teams Toolkit uploads the client id/secret for Oauth Registration to Teams Developer Portal. It is used by Teams client to securely access your API at runtime when API-Based Message Extension is triggered. Teams Toolkit doesn't store your client id/secret.

KennethBWSong: Updated, thanks!

supkasar (approved):

SLdragon (approved):

jayzhang (approved):

PR #11443, opened 2024-04-23
fix: record localdebug status

xuruiyao-msft:
Record the local debug success/failure in .fx/projectStatus.json. It would be used in the nextStep for Office GitHub Agent project.

### DIFF
--- a/packages/vscode-extension/src/debug/officeTaskHandler.ts
+++ b/packages/vscode-extension/src/debug/officeTaskHandler.ts
@@ -1,0 +1,275 @@
+import * as vscode from "vscode";
+import * as globalVariables from "../globalVariables";
+import { Correlator, isValidOfficeAddInProject } from "@microsoft/teamsfx-core";
+import { DebugNoSessionId, endLocalDebugSession, getLocalDebugSession, getLocalDebugSessionId } from "./commonUtils";
+import { updateProjectStatus } from "../utils/projectStatusUtils";
+import { CommandKey } from "../constants";
+import { UserError, err, ok } from "@microsoft/teamsfx-api";
+import { ExtensionErrors, ExtensionSource } from "../error";
+import * as commonUtils from "./commonUtils";
+import { DebugSessionExists } from "./constants";
+
+export const allRunningOfficeTasks: Map<string, number> = new Map<string, number>();
+export const allRunningDebugSessions: Set<string> = new Set<string>();
+
+export const OfficeTaskName = Object.freeze({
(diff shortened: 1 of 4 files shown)


packages/vscode-extension/src/debug/officeTaskHandler.ts
```diff
@@ -0,0 +1,191 @@
...
+        const debugConfig = event.configuration;
+        if (
+            debugConfig &&
+            debugConfig.name &&
+            !debugConfig.postDebugTask
+        ) {
+            if (await commonUtils.checkAndSkipDebugging()) {
+                throw new Error(DebugSessionExists);
+            } else {
+                commonUtils.startLocalDebugSession();
```
swatDong, line 113: Will `startLocalDebugSession` be called twice? One from `teamsfxTaskHandler` and another from here.
I would suggest to totally separate the debug session handler from the original teamsfx one, e.g., record office addin's own sessions here.

packages/vscode-extension/src/debug/officeTaskHandler.ts
```diff
@@ -0,0 +1,191 @@
+import { ok } from "@microsoft/teamsfx-api";
```
swatDong, line 1: Copyright header.
And also adjust the code owner in `.github/CODEOWNERS`

packages/vscode-extension/src/debug/officeTaskHandler.ts
xuruiyao-msft: @swatDong  Thanks for your comments. I went through the extension.ts activate function :

export async function activate(context: vscode.ExtensionContext) {
  ........
  if (isTeamsFxProject) {
    activateTeamsFxRegistration(context);
  }

  if (isOfficeAddInProject) {
    activateOfficeDevRegistration(context);
  }
  ......
  }
The teamsfxTaskHandler is registered in activateTeamsFxRegistration, but the officeTaskHandler is registered in activateOfficeDevRegistration. They are mutually exclusive for each other. startLocalDebugSession will not be called twice here.

swatDong: Thanks for the explanation, make sense to me.

packages/vscode-extension/src/debug/officeTaskHandler.ts
xuruiyao-msft: Thanks for your comments. Updated in the latest iteration. Would you please help review it again? Thanks for your effort.

swatDong (approved):

tecton (approved):

MSFT-yiz (approved):

PR #11450, opened 2024-04-24
refactor(generator): mapping user inputs to template names

hund030:
1. Mapping inputs to template names. Templates onboarding process is simplified to registering new item in the map.
2. Customizing generator for ASP.NET SSR tab templates to keep default template generator simple.

### DIFF
--- a/packages/fx-core/src/component/generator/templates/templateNames.ts
+++ b/packages/fx-core/src/component/generator/templates/templateNames.ts
@@ -1,5 +1,6 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
+import { Inputs } from "@microsoft/teamsfx-api";
 import {
   ApiMessageExtensionAuthOptions,
   CapabilityOptions,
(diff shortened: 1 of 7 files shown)


packages/fx-core/src/component/generator/templates/templateNames.ts
```diff
@@ -124,3 +128,254 @@ export const Feature2TemplateName = {
...
+export function getTemplateName(inputs: Inputs): TemplateNames {
+  const templateName = tryGetTemplateName(inputs);
+  if (!templateName) {
+    throw new Error("Template name not found");
+  }
+  return templateName;
+}
+
+// When multiple template name matches, only the top one will be picked.
+export const inputsToTemplateName: { inputs: { [key: string]: any }; name: TemplateNames }[] = [
```
Siglud, line 147: Can we just use a Map instead of find something in a list?

yukun-dong (approved):

packages/fx-core/src/component/generator/templates/templateNames.ts
hund030: Actually, we can't benefit from using a Map because we can't get a simple key from user inputs. We have to iterate the Map and it does not improve the performance.
But I still update to use a Map because it looks more compact than a list.

yukun-dong (approved):

jayzhang (approved):

PR #11462, opened 2024-04-24
perf(oauth): add scaffold for copilot plugin

KennethBWSong:
image
image
image


packages/fx-core/src/component/generator/generator.ts
```diff
@@ -48,7 +48,12 @@ export class Generator {
     safeProjectNameFromVS?: string,
     targetFramework?: string,
     placeProjectFileInSolutionDir?: boolean,
-    apiKeyAuthData?: { authName: string; openapiSpecPath: string; registrationIdEnvName: string },
+    apiKeyAuthData?: {
```
yuqizhou77, line 51: might be better to rename this also as it contains more info not limited to API Key only.

yuqizhou77 (approved):

packages/fx-core/src/component/generator/copilotPlugin/helper.ts
```diff
@@ -1027,3 +1028,12 @@ export async function updateForCustomApi(
   // 4. update code
   await updateCodeForCustomApi(specItems, language, destinationPath, openapiSpecFileName, needAuth);
 }
+
+const EnvNameMapping: { [authType: string]: string } = {
+  apiKey: "REGISTRATION_ID",
+  oauth2: "CONFIGURATION_ID",
```
SLdragon, line 1034: if you want to use different postfix for the registration id, you should also update the logic in spec-parser/manifestUpdater.ts

packages/fx-core/src/component/generator/generator.ts
KennethBWSong: Updated, thanks!

packages/fx-core/src/component/generator/copilotPlugin/helper.ts
KennethBWSong: Using configuration id since we need to align with TDP. Will update spec parser in another pr.

templates/common/api-plugin-existing-api/teamsapp.yml.tpl
```diff
@@ -17,6 +17,19 @@ provision:
     writeToEnvironmentFile:
       teamsAppId: TEAMS_APP_ID
 
+{{#AuthType}}
```
SLdragon, line 20: Seems it only check the authType whether exist, and how distinguish api key auth and oauth?

KennethBWSong: Updated. Thanks!

yukun-dong (approved):

templates/common/api-plugin-existing-api/teamsapp.yml.tpl
```diff
@@ -17,6 +17,35 @@ provision:
...
+    # Write the registration information of API Key into environment file for
+    # the specified environment variable(s).
+    writeToEnvironmentFile:
+      registrationId: {{ApiSpecAuthRegistrationIdEnvName}}
+{{/ApiKey}}
+
+{{#OAuth}}
+  - uses: oauth/register
+    with:
+      name: {{ApiSpecAuthName}}
```
SLdragon, line 39: Do we need clientId

SLdragon (approved):

yuqizhou77 (approved):

PR #11468, opened 2024-04-25
perf(oauth): update suffix for oauth in spec parser

KennethBWSong:


SLdragon (approved):

packages/spec-parser/src/constants.ts
```diff
@@ -52,7 +52,12 @@ export class ConstantString {
   static readonly TextBlockType = "TextBlock";
   static readonly ImageType = "Image";
   static readonly ContainerType = "Container";
-  static readonly RegistrationIdPostfix = "REGISTRATION_ID";
+  static readonly RegistrationIdPostfix: { [key: string]: string } = {
+    apiKey: "REGISTRATION_ID",
+    oauth2: "CONFIGURATION_ID",
+    http: "REGISTRATION_ID",
+    openIdConnect: "REGISTRATION_ID",
```
SLdragon, line 59: Question: why add openIdConnect? Seems no plan to support it in the future

KennethBWSong: I add all supported schema in OpenAPIv3 interface to avoid return undefined/throw error if we support this and forget to update here.

PR #11486, opened 2024-04-26
refactor: add declarativeCopilot.json template for copilot gpt

huimiu:


templates/common/copilot-gpt-basic/appPackage/declarativeCopilot.json.tpl
```diff
@@ -0,0 +1,4 @@
+{
+    "$schema": "https://graphdevxdata.blob.core.windows.net/data/schemas/CopilotGPTManifestSchema-1.0.json",
+    "actions": []
```
yuqizhou77, line 3: I don't think we need actions now. This template is only for GPT basic now.

templates/common/copilot-gpt-basic/appPackage/declarativeCopilot.json.tpl
```diff
@@ -0,0 +1,4 @@
+{
+    "$schema": "https://graphdevxdata.blob.core.windows.net/data/schemas/CopilotGPTManifestSchema-1.0.json",
```
yuqizhou77, line 2: Do we also need to include other fields like name, description?

templates/common/copilot-gpt-basic/appPackage/manifest.json.tpl
```diff
@@ -28,6 +28,9 @@
         "identity",
         "messageTeamMembers"
     ],
-    "copilotGpts" [],
+    "copilotGpts" [
+        "id": "gpt_1",
```
yuqizhou77, line 32: maybe not gpt_1 as we are not going to call it GPT anymore?

templates/common/copilot-gpt-basic/appPackage/declarativeCopilot.json.tpl
huimiu: I'm considering whether the `actions` should be left empty like this by default, meaning we won't include any action. This could let users know that they can add some actions using this field.

huimiu: Also, user can use a command to add an action based on our this template.

yuqizhou77: Yeah, user can add action which we will generate for them. "actions should be left empty like this by default, meaning we won't include any action" - I think we should ask feedback from PM about this.

templates/common/copilot-gpt-basic/appPackage/declarativeCopilot.json.tpl
huimiu: Yes, we need. But I'm unsure about what should be written here. I'll update the content based on what PM shares next.

templates/common/copilot-gpt-basic/appPackage/manifest.json.tpl
huimiu: Yes, not just gpt_1, but also `copilotGps`. Once the naming is finalized, I will update everything together.

templates/common/copilot-gpt-basic/appPackage/declarativeCopilot.json.tpl
huimiu: Sure, let me check with PM

templates/common/copilot-gpt-basic/appPackage/declarativeCopilot.json.tpl
huimiu: Based on the document Sébastien just shared, I think I know what to write now. Let me update this content.

templates/js/copilot-gpt-from-scratch-plugin/appPackage/manifest.json.tpl
```diff
@@ -31,8 +31,8 @@
   ],
   "copilotGpts": [
     {
-        "id": "repairGPT",
```
yuqizhou77, line 34: maybe we also want to check if we want to add the plugin as Copilot plugin also in this template. 
I don't want to block this PR, so maybe in another PR later :)

yuqizhou77 (approved):

templates/js/copilot-gpt-from-scratch-plugin/appPackage/manifest.json.tpl
huimiu: Good catch, let me update in next PR!

hund030 (approved):

PR #11491, opened 2024-04-26
refactor: update scaffolding declarative Copilot

yuqizhou77:
- update scaffolding based on latest decision.
  - rename "GPT" to "declarative copilot".
  -  We will provide two options for user to start with, a basic declarative copilot or a declarative copilot with a plugin from on Azure function

New scaffold flow:
- basic Declarative Copilot
image
- Declarative Copilot with Plugin from scratch
image

TODO: will add a CLI command ("teamsapp add plugin" for example) so that user can add plugin/action from an existing API spec to a Declarative Copilot project.


packages/fx-core/src/question/create.ts
```diff
@@ -921,7 +922,7 @@ export function capabilityQuestion(): SingleSelectQuestion {
         case ProjectTypeOptions.customCopilot().id:
           return getLocalizedString("core.createProjectQuestion.projectType.customCopilot.title");
         case ProjectTypeOptions.customizeGpt().id:
-          return "Choose the GPT type";
+          return "Choose GPT type";
```
huimiu, line 925: Do we also need to replace this GPT with declarative copilot?

packages/fx-core/src/question/create.ts
```diff
@@ -864,17 +864,18 @@ export class CapabilityOptions {
   // customize GPT
   static customizeGptBasic(): OptionItem {
     return {
-      id: "customize-gpt-basic",
-      label: "Basic GPT",
-      detail: "A simple GPT skeleton you can author without any plugin",
+      id: "basic-declarative-copilot",
+      label: "Basic Declarative Copilot",
```
tecton, line 868: Please remember to do localization.

jayzhang (approved):

tecton (approved):

packages/fx-core/src/question/create.ts
yuqizhou77: sure! going to do that later. Currently this option is for CLI only.

huimiu (dismissed):

packages/fx-core/src/question/create.ts
yuqizhou77: Thanks, updated.

huimiu (approved):

tecton (approved):

Alive-Fish (approved):

LongOddCode (approved):

PR #11543, opened 2024-05-07
fix: bump api version

jayzhang:
fix: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/27982329


wenytang-ms (dismissed):
please set the version as 0.23.0-rc, then will ship 0.23.0 stable version
if set 0.23.0, will ship 0.23.1 stable version

LongOddCode (approved):

PR #11554, opened 2024-05-08
fix: hot fix for Python templates to remove Test Tool from the default debug choice

frankqianms:
workitem: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/27894013

E2E TEST: N/A

### DIFF
--- a/templates/python/custom-copilot-assistant-new/teamsapp.local.yml.tpl
+++ b/templates/python/custom-copilot-assistant-new/teamsapp.local.yml.tpl
@@ -8,7 +8,7 @@ provision:
   - uses: teamsApp/create
     with:
       # Teams app name
-      name: {{appName}}-${{TEAMSFX_ENV}}
+      name: {{appName}}${{APP_NAME_SUFFIX}}
     # Write the information of created resources into environment file for
     # the specified environment variable(s).
     writeToEnvironmentFile: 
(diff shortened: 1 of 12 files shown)


templates/python/custom-copilot-assistant-new/teamsapp.local.yml.tpl
```diff
@@ -8,7 +8,7 @@ provision:
   - uses: teamsApp/create
     with:
       # Teams app name
-      name: {{appName}}-${{TEAMSFX_ENV}}
+      name: {{appName}}${{APP_NAME_SUFFIX}}
```
blackchoey, line 11: What will happen if `.env.xxx` files do not contain this environment variable? Do we need to update the env files to include this new environment variable?

frankqianms: all .env.XXX files include APP_NAME_SUFFIX if the related .yml files use this variable

blackchoey: Per frank, all .env files contain the variable. Resolving.

blackchoey (approved):

adashen (approved):

PR #11559, opened 2024-05-08
fix: update proposed chat api

GavinGu07:
Fix the participant breaks triggered by the latest chat participant API update: https://github.com/microsoft/vscode/pull/212148
image


tecton (approved):
approve for create command part.

packages/vscode-extension/src/chat/utils.ts
```diff
@@ -50,7 +50,11 @@ export function countMessageTokens(message: LanguageModelChatMessage): number {
     if (!value) {
       continue;
     }
-    numTokens += tokenizer.tokenLength(value);
+    if (key === "role") {
```
lijie-lee, line 53: There is  no need to count token size of `role`.  The `BaseTokensPerMessage` just stands for that. 
> By the way: if wanna count the size of `role`, you should use the real role value like "system", "user" and "assistant"

lijie-lee: Just notice the ChatParticipant SDK has exposed the [interface](https://github.com/microsoft/vscode/blob/5f78b58b57b7cf84d28d801fed6bb4a48f908601/src/vscode-dts/vscode.proposed.languageModels.d.ts#L166) for counting token size. 

Maybe we could adopt to that later.

GavinGu07: Thanks. I've removed the token counting for the `role`. 

`computeTokenLength` will be supported later in another PR.

lijie-lee (dismissed):

lijie-lee (dismissed):
approve the tokenize part

supkasar (dismissed):

lijie-lee (dismissed):

supkasar (approved):

1openwindow (approved):

PR #11601, opened 2024-05-14
fix: proposed api breaking changes

Alive-Fish:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/28055324

### DIFF
--- a/packages/vscode-extension/src/chat/utils.ts
+++ b/packages/vscode-extension/src/chat/utils.ts
@@ -13,7 +13,13 @@ export async function verbatimCopilotInteraction(
   response: ChatResponseStream,
   token: CancellationToken
 ) {
-  const chatRequest = await lm.sendChatRequest(model, messages, {}, token);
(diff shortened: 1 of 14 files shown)


packages/vscode-extension/src/chat/utils.ts
```diff
@@ -13,7 +13,13 @@ export async function verbatimCopilotInteraction(
   response: ChatResponseStream,
   token: CancellationToken
 ) {
-  const chatRequest = await lm.sendChatRequest(model, messages, {}, token);
+  const [vendor, family] = model.split(/-(.*)/s);
```
tecton, line 16: Is it better to change the `model` parameter in signature?

Alive-Fish: Yeah, I think the better way is to manage the models together. This PR is to do the minimal change for unblocking our chat.

tecton (approved):

PR #11621, opened 2024-05-15
fix: check vscode.version to enable teams participant

Alive-Fish:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/28055653


.github/workflows/cd.yml
```diff
@@ -189,11 +189,11 @@ jobs:
           git commit -m "build: replace sideloading placeholders"
 
       - name: disable chat participant environment variable
-        if: ${{ github.event_name == 'workflow_dispatch' && (github.event.inputs.preid != 'alpha') }}
+        if: ${{ github.event_name == 'workflow_dispatch' && (github.event.inputs.preid != 'alpha' && github.event.inputs.preid != 'beta') }}
```
tecton, line 192: It seems that the script could be removed since we use function instead of const variable?

Alive-Fish: If we need to hide them in the stable release TTK, we also need to check the preid.

tecton (approved):

LongOddCode (approved):

wenytang-ms (approved):

PR #11721, opened 2024-05-28
fix: launch the api ME with AAD auth template using manifest

huimiu:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/28133154

### DIFF
--- a/packages/fx-core/src/common/m365/launchHelper.ts
+++ b/packages/fx-core/src/common/m365/launchHelper.ts
@@ -32,7 +32,7 @@ export class LaunchHelper {
       ? (await this.getUpnFromToken()) ?? "login_your_m365_account" // a workaround that user has the chance to login
       : undefined;
     let url: URL;
-    const copilotCapabilities = ["plugin", "copilotGpt"];
+    const copilotCapabilities = ["plugin", "copilotGpt", "apiMeAAD"];
     switch (hub) {
       case HubTypes.teams: {
         let installAppPackage = true;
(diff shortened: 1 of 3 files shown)


packages/fx-core/src/common/m365/launchHelper.ts
```diff
@@ -38,7 +38,9 @@ export class LaunchHelper {
         let installAppPackage = true;
         if (
           capabilities.length > 0 &&
-          capabilities.filter((capability) => !copilotCapabilities.includes(capability)).length == 0
+          (capabilities.filter((capability) => !copilotCapabilities.includes(capability)).length ==
+            0 ||
+            capabilities.includes("apiMeAAD"))
```
yuqizhou77, line 43: just want to double confirm：
- if the app has both tab and apiMeAAD capability, should we skip installAppPackage or direct user to installAppPackage URL?

yuqizhou77 (approved):

packages/fx-core/src/common/m365/launchHelper.ts
huimiu: Thanks, I updated the condition. It will only skip installAppPackage if it doesn't include a tab and it's an API ME project with AAD auth.

jayzhang (dismissed):

yuqizhou77 (approved):

kuojianlu (approved):

jayzhang (approved):

PR #11741, opened 2024-05-30
fix: set perScriptSourcemaps to stabilize the debugger

kptdobe:
Fixes https://github.com/OfficeDev/teams-toolkit/issues/11732

Notes:
- I do not use the other debugging mode (like the `Launch Remote (Chrome)`) - might be needed there too
- I did not use the Typescript version neither but I assume the same problem might occur.


kptdobe:
@microsoft-github-policy-service agree

PR #11771, opened 2024-06-04
refactor: add devproxy in local debug telemetry

xzf0587:


packages/vscode-extension/src/debug/teamsfxTaskHandler.ts
```diff
@@ -93,6 +93,15 @@ function isNpmInstallTask(task: vscode.Task): boolean {
   return false;
 }
 
+function isDevProxyTask(task: vscode.Task): boolean {
+  if (task) {
+    const name = task.name.trim();
```
swatDong, line 98: Does this new task have any specific identifier like `type: "teamsfx"` or `teamsfx: ` prefix?

xzf0587: The new tasks are shell commands which are used to check and start dev proxy process in some samples.
image

packages/vscode-extension/src/debug/teamsfxTaskHandler.ts
```diff
@@ -93,6 +93,15 @@ function isNpmInstallTask(task: vscode.Task): boolean {
...
+    return name === "Check dev proxy" || name === "Start dev proxy";
```
kuojianlu, line 99: Can this task be checked with type/command rather than name? since name may be changed by user.

packages/vscode-extension/src/debug/teamsfxTaskHandler.ts
swatDong: The "Start dev proxy" seems already be covered since it's command is "npm run *:teamsfx" (see https://github.com/OfficeDev/teams-toolkit/blob/dev/packages/vscode-extension/src/debug/teamsfxTaskHandler.ts#L139)

For "Check dev proxy", how about adding more checks like:
- Task type is shell
- cwd is */proxy

swatDong: Checked that "Start dev proxy" does exist in current telemetry. So you may only need to add minimal scope for "Check dev proxy".

swatDong (approved):

packages/vscode-extension/src/debug/teamsfxTaskHandler.ts
xzf0587: Thanks. It is updated.

packages/vscode-extension/src/debug/teamsfxTaskHandler.ts
xzf0587: Thanks. It is updated.

PR #11774, opened 2024-06-04
feat: add auth options for the api plugin capability

huimiu:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/27822940
This PR is for UI part, template code PR is [here](https://github.com/OfficeDev/teams-toolkit/pull/11747)


LongOddCode (approved):

hund030 (approved):

packages/fx-core/src/question/create.ts
```diff
@@ -1091,18 +1093,26 @@ export function apiSpecLocationQuestion(includeExistingAPIs = true): SingleFileO
...
-    staticOptions: ApiMessageExtensionAuthOptions.all(),
-    dynamicOptions: () => ApiMessageExtensionAuthOptions.all(),
-    default: ApiMessageExtensionAuthOptions.none().id,
+    staticOptions: ApiAuthOptions.all(),
+    dynamicOptions: (inputs: Inputs) => {
+      const options: OptionItem[] = [ApiAuthOptions.none()];
+      if (inputs[QuestionNames.MeArchitectureType] == MeArchitectureOptions.newApi().id) {
+        options.push(ApiAuthOptions.apiKey(), ApiAuthOptions.microsoftEntra());
+      } else if (inputs[QuestionNames.Capabilities] == CapabilityOptions.copilotPluginNewApi().id) {
+        options.push(ApiAuthOptions.apiKey(), ApiAuthOptions.oauth());
```
yuqizhou77, line 1111: Do we need feature for these two options? We have "isCopilotAuthEnabled" for Copilot plugin with auth from existing API. @SLdragon please correct me if I am wrong.

packages/fx-core/src/question/create.ts
```diff
@@ -1091,18 +1093,26 @@ export function apiSpecLocationQuestion(includeExistingAPIs = true): SingleFileO
...
       "core.createProjectQuestion.apiMessageExtensionAuth.placeholder"
     ),
     cliDescription: "The authentication type for the API.",
-    staticOptions: ApiMessageExtensionAuthOptions.all(),
-    dynamicOptions: () => ApiMessageExtensionAuthOptions.all(),
-    default: ApiMessageExtensionAuthOptions.none().id,
+    staticOptions: ApiAuthOptions.all(),
+    dynamicOptions: (inputs: Inputs) => {
+      const options: OptionItem[] = [ApiAuthOptions.none()];
+      if (inputs[QuestionNames.MeArchitectureType] == MeArchitectureOptions.newApi().id) {
```
yuqizhou77, line 1108: ===. Same for below.

huimiu: Since the API ME authentication is GA, I think we don't need to add a feature flag for it

yuqizhou77: Oh I mean if we should use "===" rather than "=="

huimiu: I see, good catch, let me update in next PR

packages/fx-core/src/question/create.ts
huimiu: Let me add the `isCopilotAuthEnabled` feature flag in next PR, thanks!

PR #11776, opened 2024-06-05
perf(sme): add separate env for SME OAuth

SLdragon:


KennethBWSong (approved):

huimiu (approved):

yuqizhou77 (approved):

LongOddCode (approved):

packages/fx-core/src/common/featureFlags.ts
```diff
@@ -62,6 +62,10 @@ export function isCopilotAuthEnabled(): boolean {
   return featureFlagManager.getBooleanValue(FeatureFlags.CopilotAuth);
 }
 
+export function isSMEOAuthEnabled(): boolean {
```
jayzhang, line 65: Suggest to use unified global feature flag manager directly: featureFlagManager.getBooleanValue(FeatureFlags.SMEOAuth)

jayzhang (approved):

packages/fx-core/src/common/featureFlags.ts
SLdragon: As discussed, we plan to update it in another PR

PR #11791, opened 2024-06-07
refactor: update global variables and resolve dependency cycle issues

yiqing-zhao:


tecton (approved):
Could also update codeowner file for new files later.

swatDong (approved):

PR #11805, opened 2024-06-12
feat(vs): add empty project template

Alive-Fish:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/28358211


templates/csharp/empty/teamsapp.yml.tpl
```diff
@@ -0,0 +1,45 @@
+# yaml-language-server: $schema=https://aka.ms/teams-toolkit/1.1.0/yaml.schema.json
```
hund030, line 1: Please consider updating yaml schema version to v1.5

hund030 (approved):

templates/csharp/empty/teamsapp.yml.tpl
Alive-Fish: All right. Will update it next PR.

PR #11870, opened 2024-06-20
refactor: move account related functions to separate files

yiqing-zhao:


tecton (approved):
Please remember to reuse the test codes in `handlers.test.ts` in the future.

PR #11890, opened 2024-06-24
refactor: seperator env handlers

jayzhang:


packages/vscode-extension/src/extension.ts
```diff
@@ -579,12 +581,12 @@ function registerTreeViewCommandsInHelper(context: vscode.ExtensionContext) {
  * TeamsFx related commands, they will show in command palette after extension is initialized
  */
 function registerTeamsFxCommands(context: vscode.ExtensionContext) {
-  const createNewEnvironment = vscode.commands.registerCommand(
+  const handler = vscode.commands.registerCommand(
```
tecton, line 584: `handler` is a bit ambiguous, why not keep the original variable name?

jayzhang: the name is conflict with the function name 'createNewEnvironment()'

tecton (approved):

PR #11933, opened 2024-07-01
chore: Updates custom copilot for M365 templates to use Graph connectors

waldekmastykarz:


MuyangAmigo:
Close as we will solve this from a doc approach.

PR #11941, opened 2024-07-02
test: add e2e test

huimiu:
E2E:
https://github.com/OfficeDev/teams-toolkit/actions/runs/9850169188/job/27195097412


packages/tests/src/e2e/caseFactory.ts
```diff
@@ -173,7 +182,12 @@ export abstract class CaseFactory {
           expect(result).to.be.true;
           process.env["AZURE_RESOURCE_GROUP_NAME"] = appName + "-rg";
 
-          const { success } = await Executor.provision(projectPath);
+          const { success } = await Executor.provision(
+            projectPath,
+            "dev",
+            true,
+            options?.skipErrorMessage
```
wenytang-ms, line 189: if skip this error message, the pipeline may skip some action

huimiu: Yes, there are some features that are in private preview. The test account cannot run some templates successfully, so we are going to skip this error message for now

wenytang-ms: if so, maybe set skipErrorMessage for specific templates instead of set all rules?

huimiu: Only some templates have passed the `skipErrorMessage` parameter, while others that don't need it haven't. This parameter is optional. The current change provides an option for test cases to choose whether to skip certain actions. Do you think this change is ok?

huimiu: If we don't add this parameter, the `CaseFactory` cannot support scenarios that need to skip some error messages, then I think the `CaseFactory` is not reusable

ayachensiyuan: Can we filter out only the expected errors instead of all errors？Otherwise, the case would be meaningless and no error would be detected during provision.

huimiu: > Can we filter out only the expected errors instead of all errors？Otherwise, the case would be meaningless and no error would be detected during provision.

Sure, we can filter out the expected errors, but I think it seems strange to filter out errors in `CaseFactory` which should not care about specific scenarios. I think each test case should handle this individually. If a test case doesn't want to ignore the error message, it should not set the `skipErrorMessage` parameter. Additionally, we can add some comments in the parameter definition to inform others about the effect of setting this parameter. What do you think?

huimiu: https://github.com/OfficeDev/teams-toolkit/blob/hui/copilot-e2e/packages/tests/src/utils/executor.ts#L56
Also, we don't skip all errors, we only filter out the errors in `skipErrorMessage`

ayachensiyuan: That's fine for me, Thanks.

packages/tests/src/e2e/copilotplugin/CopilotPluginWithNoneAuthForJs.tests.ts
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * @author Yimin Jin <yiminjin@microsoft.com>
+ */
+
+import { ProgrammingLanguage } from "@microsoft/teamsfx-core";
+import { validateFiles } from "./helper";
+import { CopilotPluginCommonTest } from "./copilotPluginCommonTest";
+
+class CopilotPluginWithNoneAuthForJsCase extends CopilotPluginCommonTest {
```
eriolchan, line 12: Similar as above

packages/tests/src/e2e/copilotplugin/CopilotPluginWithApiKeyForJs.tests.ts
```diff
@@ -0,0 +1,32 @@
...
+      "src/keyGen.js",
+    ];
+    await validateFiles(projectPath, files);
+
+    const userFile = path.resolve(projectPath, "env", `.env.dev.user`);
+    await replaceSecretKey(userFile);
+  }
+}
+
+new CopilotPluginWithNoneAuthForJsCase(
```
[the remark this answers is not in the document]
huimiu: We tried using a single file to create test cases for different languages, but we found it hard to find which language's template has a problem since the test cases in the same file were being executed asynchronously and out of order.

So now, we moved some common methods to the `CopilotPluginCommonTest` file and separated the different languages into their own files. This way, we can reduce the duplicate code and makes it easier to identify issues in specific language templates.

packages/tests/src/e2e/caseFactory.ts
huimiu: > If the feature is not ready, is it possible to comment some validation in test case and add the validation back when feature is ready? I have some concern to use this tricky way to work around test case.

Right now, there's an error in an action (extendToM365) during provision. If we skip the provision, we can't proceed with deploying or running further checks, and this test case won't catch any issues because it hasn't done anything yet.

wenytang-ms: is there any ticket to track this issue?

packages/tests/src/e2e/copilotplugin/CopilotPluginWithApiKeyForJs.tests.ts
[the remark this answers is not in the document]
Yimin-Jin: We tried the one test case approach and it worked, the only problem was that the error messages were hard to distinguish, so we ended up splitting into multiple test cases.

packages/tests/src/e2e/caseFactory.ts
huimiu: This feature is in private preview, this issue is expected, we have confirmed with our partners.
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/28501674
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/28542361

BTW, this skip method isn't newly added in this PR, it was already in dev branch.

https://github.com/OfficeDev/teams-toolkit/blob/e08ff04cfd976d3977830c785317e6afbbf1dafa/packages/tests/src/utils/executor.ts#L19-L24

https://github.com/OfficeDev/teams-toolkit/blob/e08ff04cfd976d3977830c785317e6afbbf1dafa/packages/tests/src/utils/executor.ts#L38-L44

huimiu: It seems like everyone has some concerns about this, so let's hold off on adding any test cases related to the Copilot plugin until they've fully rolled it out.

wenytang-ms: If this error is skipped, is it possible to simulate the actual e2e process?

packages/tests/src/commonlib/functionValidator.ts
```diff
@@ -8,6 +8,7 @@ import * as chai from "chai";
 import glob from "glob";
 import path from "path";
 import { EnvConstants, PluginId, StateConfigKey } from "./constants";
+// eslint-disable-next-line import/no-cycle
```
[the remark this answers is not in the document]
huimiu: As some historical reasons, many files in the 'tests/' folder currently have nested import issues. If there are any PR modify any of these files, the pre-commit check will fail. The pre-commit check seems to be added later on.

We need to take a look at all those affected files and discuss with related owners on how we can fix them. Let me follow up with them.

packages/tests/src/e2e/caseFactory.ts
huimiu: Can simulate most, but not all. We'll remove the skip logic once the feature is GA.

packages/tests/src/e2e/caseFactory.ts
```diff
@@ -153,7 +161,8 @@ export abstract class CaseFactory {
           testFolder,
           capability,
           programmingLanguage,
-          custimized
+          custimized,
```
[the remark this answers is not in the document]
huimiu: Fixed, thanks!

packages/tests/src/e2e/caseFactory.ts
huimiu: > if not support provision, the remote for e2e may not stable for the test case, can we use local debug for testing instead of remote? if skip the provision error, the azure resource is really provisioned ?

Sorry, I didn't explain clearly enough. There is only one action, "extendToM365," will fail during provisioning, the arm deploy is work.

hund030: Actually, there is a common function `removeTeamsAppExtendToM365` to work around the issue.
https://github.com/OfficeDev/teams-toolkit/blob/dev/packages/tests/src/e2e/frontend/TestCreateTab.tests.ts#L67

PR #11946, opened 2024-07-03
perf(spec-parser): optimize "None Auth", special media type, AC card gen

SLdragon:
Address below issues:
[Bug 28574537](https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/28574537): API ME scaffold has string "None auth" which is confusing
[Bug 28587522](https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/28587522): TTK fails to load an OpenAPI spec with media type application/json; charset=utf-8
[Bug 28615539](https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/28615539): Type b may generate empty container in adaptive card


yuqizhou77 (approved):

packages/fx-core/resource/package.nls.json
```diff
@@ -255,7 +255,7 @@
   "core.TabSPFxOption.detailNew": "Build UI with SharePoint Framework",
   "core.TabNonSso.label": "Basic Tab",
   "core.TabNonSso.detail": "A simple implementation of a web app that's ready to customize",
-  "core.copilotPlugin.api.noAuth": "None auth",
+  "core.copilotPlugin.api.noAuth": "No authentication",
   "core.copilotPlugin.api.apiKeyAuth": "API key auth(Bearer token auth)",
```
yuqizhou77, line 259: I am wondering whether this need to be API Key authentication. :)

SLdragon: @supkasar , do you have any suggestions about the API Key auth?

supkasar: If you think, users won't understand what authentication we are talking about here, then 'No API key auth' makes sense. But if they would understand it easily, then the current content is enough.

KennethBWSong (approved):

yuqizhou77 (approved):

packages/fx-core/resource/package.nls.json
```diff
@@ -255,9 +255,9 @@
   "core.TabSPFxOption.detailNew": "Build UI with SharePoint Framework",
   "core.TabNonSso.label": "Basic Tab",
   "core.TabNonSso.detail": "A simple implementation of a web app that's ready to customize",
-  "core.copilotPlugin.api.noAuth": "None auth",
-  "core.copilotPlugin.api.apiKeyAuth": "API key auth(Bearer token auth)",
-  "core.copilotPlugin.api.oauth": "OAuth(Auth code flow)",
+  "core.copilotPlugin.api.noAuth": "No authentication",
+  "core.copilotPlugin.api.apiKeyAuth": "API Key authentication(Bearer token authentication)",
+  "core.copilotPlugin.api.oauth": "OAuth(Authorization code flow)",
```
supkasar, line 260: String 260: Is it authorization code flow or authentication code flow?

SLdragon: Hi, @supkasar , the process is known as the authorization code flow in OAuth. For more details, refer to this document: https://learn.microsoft.com/en-us/entra/identity-platform/v2-oauth2-auth-code-flow

supkasar (approved):

xzf0587 (approved):

PR #11975, opened 2024-07-08
refactor: update to API plugin

yuqizhou77:
[Bug 28630397](https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/28630397): Terminology Update: Copilot Plugin -> API Plugin


huimiu (approved):

HuihuiWu-Microsoft (approved):

yuqizhou77:
@MuyangAmigo Could you help check all readme?
Also, I would like to confirm with you whether below two things need to be updated.
- Option id of CLI non-interactive. This feature was released before but being hidden. So it may be a breaking change but may won't impact any customers?
- Description of VSC settings.
- 
image
I believe this settings apply to both bot plugin and API plugin. So keep it as "copilot plugin"

MuyangAmigo:
> @MuyangAmigo Could you help check all readme? Also, I would like to confirm with you whether below two things need to be updated.
> 
> * Option id of CLI non-interactive. This feature was released before but being hidden. So it may be a breaking change but may won't impact any customers? Let me know if you think we should update the option id as well which is not included in this PR.
> * Description of VSC settings.
> 
> image I believe this settings apply to both bot plugin and API plugin. So keep it as "copilot plugin"

Yes we can leave VSC settings unchanged. For CLI changes, since we have not yet exposed to publicly, we can make the change.

yuqizhou77:
> > @MuyangAmigo Could you help check all readme? Also, I would like to confirm with you whether below two things need to be updated.
> > 
> > * Option id of CLI non-interactive. This feature was released before but being hidden. So it may be a breaking change but may won't impact any customers? Let me know if you think we should update the option id as well which is not included in this PR.
> > * Description of VSC settings.
> > 
> > image I believe this settings apply to both bot plugin and API plugin. So keep it as "copilot plugin"
> 
> Yes we can leave VSC settings unchanged. For CLI changes, since we have not yet exposed to publicly, we can make the change.

sure. will have another PR fixing the CLI part.

SLdragon (approved):

jayzhang (approved):

PR #12006, opened 2024-07-15
fix: missing readme for outlook add-in

jayzhang:
fix: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/28693952


tecton (approved):
Please rename the related function names for clarity. It does scaffolding in post() now because the README should not be overwritten by template README.

PR #12035, opened 2024-07-19
refactor: delete AAD after debugging in certain tenant

xiaolang124:
image


supkasar (approved):

HuihuiWu-Microsoft (approved):

packages/vscode-extension/src/debug/teamsfxTaskHandler.ts
```diff
@@ -519,6 +537,7 @@ export function terminateAllRunningTeamsfxTasks(): void {
   }
   allRunningTeamsfxTasks.clear();
   BaseTunnelTaskTerminal.stopAll();
+  void deleteAad();
```
swatDong, line 540: This looks like an async background action. Would it be a corner case if I stop then re-debug in short time?

packages/vscode-extension/src/debug/deleteAadHelper.ts
```diff
@@ -0,0 +1,120 @@
...
+
+const defaultNotificationLocalFile = ".notification.localstore.json";
+export async function deleteAad() {
+  try {
+    if (globalVariables.deleteAadInProgress) {
+      return;
+    }
+    globalVariables.setDeleteAadInProgress(true);
+    const projectPath = globalVariables.workspaceUri!.fsPath;
+    const envFile = path.resolve(projectPath, "env", ".env.local");
```
blackchoey, line 27: The `env` folder name is configurable in teamsapp.yml, please read the folder name from the yaml instead of hard code it.

packages/vscode-extension/src/debug/deleteAadHelper.ts
```diff
@@ -0,0 +1,120 @@
...
+    const userFile = path.resolve(projectPath, "env", ".env.local.user");
+    if (!fs.existsSync(envFile) || !fs.existsSync(userFile)) {
+      return;
+    }
+    const envData = dotenvUtil.deserialize(fs.readFileSync(envFile, "utf-8"));
+    const userEnvData = dotenvUtil.deserialize(fs.readFileSync(userFile, "utf-8"));
+    if (!envData.obj["BOT_ID"] && !envData.obj["AAD_APP_CLIENT_ID"]) {
```
blackchoey, line 34: User can customize the outputted environment variable name in teamsapp.yml, we may also read the actual environment variable name instead of hard code

packages/vscode-extension/src/debug/teamsfxTaskHandler.ts
blackchoey, line 540: Can you share whether this also apply to Python TTK projects? We can discuss more if you need to learn more details for Python project.

blackchoey:
Seems the logic always happen for microsoft users, have we considered exposing some kind of config to enable/disable this behavior? This might help for use existing scenario, which should not trigger the delete logic in my opinion. One example is user manually configured an AAD app with certificate, and then it was deleted by TTK automatically.
Feel free to let me know if that's out of the feature's scope.

packages/fx-core/tests/component/driver/aad/create.test.ts
```diff
@@ -713,4 +713,122 @@ describe("aadAppCreate", async () => {
...
+
+    expect(result.result.isOk()).to.be.true;
+    expect(result.result._unsafeUnwrap().get("MY_CLIENT_ID")).to.equal(expectedClientId);
+    expect(result.result._unsafeUnwrap().get("MY_OBJECT_ID")).to.equal(expectedObjectId);
+    expect(result.result._unsafeUnwrap().get("MY_AUTHORITY_HOST")).to.equal(
+      "https://login.microsoftonline.com"
+    );
+    expect(result.result._unsafeUnwrap().get("MY_CLIENT_SECRET")).to.equal(expectedSecretText);
+    expect(result.result._unsafeUnwrap().size).to.equal(6);
+    expect(result.summaries.length).to.equal(2);
```
blackchoey, line 754: may check the content to avoid the information for microsoft tenant been deleted by accident.

packages/fx-core/tests/component/driver/botAadApp/create.test.ts
```diff
@@ -316,4 +316,88 @@ describe("botAadAppCreate", async () => {
...
+
+    sinon.stub(AadAppClient.prototype, "generateClientSecret").resolves(expectedSecretText);
+
+    const result = await createBotAadAppDriver.handler(
+      args,
+      mockedDriverContext,
+      outputEnvVarNames
+    );
+
+    console.log(JSON.stringify(result));
```
blackchoey, line 342: Can we remove this line?

packages/fx-core/tests/component/driver/botAadApp/create.test.ts
```diff
@@ -316,4 +316,88 @@ describe("botAadAppCreate", async () => {
...
+
+    expect(result.output.get(outputKeys.botId)).to.be.equal(expectedClientId);
```
blackchoey, line 344: Also check the summary strings. If you want to optimize, there may be no need to check the bot id/password as they were covered in other test cases.

packages/fx-core/tests/component/driver/botAadApp/create.test.ts
```diff
@@ -316,4 +316,88 @@ describe("botAadAppCreate", async () => {
...
+
+    sinon.stub(AadAppClient.prototype, "generateClientSecret").resolves(expectedSecretText);
+
+    const result = await createBotAadAppDriver.handler(
+      args,
+      mockedDriverContext,
+      outputEnvVarNames
+    );
+
+    console.log(JSON.stringify(result));
```
blackchoey, line 370: Please also see whether we can remove the console.log here, as well as checking the summary string

packages/fx-core/tests/component/driver/botAadApp/create.test.ts
```diff
@@ -316,4 +316,88 @@ describe("botAadAppCreate", async () => {
...
```
blackchoey, line 398: Please also see whether we can remove the console.log here, as well as checking the summary string

packages/fx-core/tests/component/driver/aad/create.test.ts
```diff
@@ -713,4 +713,122 @@ describe("aadAppCreate", async () => {
...
+        tenantId: "MY_TENANT_ID",
+        authorityHost: "MY_AUTHORITY_HOST",
+        authority: "MY_AUTHORITY",
+        clientSecret: "MY_CLIENT_SECRET",
+      })
+    );
+
+    const result = await createAadAppDriver.execute(args, mockedDriverContext, outputEnvVarNames);
+
+    expect(result.result.isOk()).to.be.false;
```
blackchoey, line 832: I saw the UT for `botAadApp/create` will success and return bot id/password in this case, but the UT for `aadApp/create` will fail under same situation. The code seems the same. Can you share what causes the difference?
In my own opinion, the behavior for `botAadApp/create` seems better, we could just skip the new logic if when cannot determine whether the user is on microsoft tenant or not.

blackchoey (approved):

swatDong (approved):

packages/sdk/src/conversation/middlewares/notificationMiddleware.ts
```diff
@@ -66,7 +66,7 @@ export class NotificationMiddleware implements Middleware {
     const activityType = activity.type;
     if (activityType === "installationUpdate") {
       const action = activity.action?.toLowerCase();
-      if (action === "add") {
+      if (action === "add" || action === "add-upgrade") {
```
tecton, line 69: Is this change a bugfix?

tecton (approved):

PR #12056, opened 2024-07-24
refactor: update aad test case

xiaolang124:
Update test cases.


packages/fx-core/tests/component/driver/botAadApp/create.test.ts
```diff
@@ -367,8 +368,6 @@ describe("botAadAppCreate", async () => {
       outputEnvVarNames
     );
```
blackchoey, line 370: minor: may add below check?
```
expect(result.summaries[0]).not.includes(
      "Teams toolkit will delete the Microsoft Entra application after debugging"
    );
````

blackchoey (approved):

PR #12058, opened 2024-07-25
refactor: only delete ttk created aad

xiaolang124:
Only delete ttk created aad.

### DIFF
--- a/packages/fx-core/src/common/constants.ts
+++ b/packages/fx-core/src/common/constants.ts
@@ -77,5 +77,3 @@ export const GraphReadUserScopes = ["https://graph.microsoft.com/User.ReadBasic.
 export const SPFxScopes = (tenant: string) => [`${tenant}/Sites.FullControl.All`];
 export const AzureScopes = ["https://management.core.windows.net/user_impersonation"];
 export const AppStudioScopes = [`${getAppStudioEndpoint()}/AppDefinitions.ReadWrite`];
-
-export const AadSet: Set<string> = new Set<string>();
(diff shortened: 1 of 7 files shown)


packages/vscode-extension/src/debug/deleteAadHelper.ts
```diff
@@ -84,6 +90,7 @@ export async function deleteAad() {
           localize("teamstoolkit.localDebug.successDeleteNotificationLocalStoreFile")
         );
       }
+      AadSet.clear();
```
swatDong, line 93: Is this redundant? `delete` is already called for created aad.
In addition, would this mis-clear the newly created aad since the whole operation is async?

packages/fx-core/src/common/constants.ts
```diff
@@ -77,3 +77,5 @@ export const GraphReadUserScopes = ["https://graph.microsoft.com/User.ReadBasic.
 export const SPFxScopes = (tenant: string) => [`${tenant}/Sites.FullControl.All`];
 export const AzureScopes = ["https://management.core.windows.net/user_impersonation"];
 export const AppStudioScopes = [`${getAppStudioEndpoint()}/AppDefinitions.ReadWrite`];
+
+export const AadSet: Set<string> = new Set<string>();
```
swatDong, line 81: This seems not a constant. Consider move to other place, e.g., globalVars?

swatDong (approved):

blackchoey (approved):

PR #12123, opened 2024-07-31
fix: upgrade notification bot template based on Teams store policy

JerryYangKai:
Upgrade notification bot template (js ts csharp version)

Based on Teams Store validation guidelines (https://learn.microsoft.com/en-us/microsoftteams/platform/concepts/deploy-and-publish/appsource/prepare/teams-store-validation-guidelines)

### Notification only bots
Notification only bots must send a welcome message that clarifies that the bot is a notification only bot and users won't be able to interact with the bot. [Mandatory Fix]

### DIFF
--- a/templates/csharp/notification-http-timer-trigger-isolated/TeamsBot.cs.tpl
+++ b/templates/csharp/notification-http-timer-trigger-isolated/TeamsBot.cs.tpl
@@ -17,7 +17,7 @@ namespace {{SafeProjectName}}
 
         protected override async Task OnMembersAddedAsync(IList<ChannelAccount> membersAdded, ITurnContext<IConversationUpdateActivity> turnContext, CancellationToken cancellationToken)
         {
-            var welcomeText = "Welcome to the Notification Bot! I am designed to send you updates and alerts using Adaptive Cards triggered by HTTP post requests or timer schedules." +
+            var welcomeText = "Welcome to the Notification Bot! I am designed to send you updates and alerts using Adaptive Cards triggered by HTTP post requests or timer schedules. " +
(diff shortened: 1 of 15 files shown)


JerryYangKai:
@MuyangAmigo  Please help reviewing the content, thanks.

templates/csharp/notification-http-timer-trigger-isolated/TeamsBot.cs.tpl
```diff
@@ -1,15 +1,31 @@
...
     public class TeamsBot : TeamsActivityHandler
     {
-        public override Task OnTurnAsync(ITurnContext turnContext, CancellationToken cancellationToken = default) =>
-            Task.CompletedTask;
+        public override async Task OnTurnAsync(ITurnContext turnContext, CancellationToken cancellationToken = default)
+        {
+            await base.OnTurnAsync(turnContext, cancellationToken);
+        }
+
+        protected override async Task OnMembersAddedAsync(IList<ChannelAccount> membersAdded, ITurnContext<IConversationUpdateActivity> turnContext, CancellationToken cancellationToken)
```
swatDong, line 18: Will this be called on each added member? For example:
- Adding bot to an existing group chat
- Adding User A to the group chat
- Adding User B to the group chat

Will bot send welcome to the group chat twice? Or send to A and B as 1:1 chat?

JerryYangKai: The bot will send welcome message in these three cases. 
If developers want to customize this welcome message scope, developer can change the if condition for different scenarios based on user's need.
What we added in this template is just a general scenario.

swatDong (approved):
Code change looks good to me.
I'm OK if the expected behavior is to send welcome on every member add event.

PR #12160, opened 2024-08-06
fix(custom api): add json path for custom api generator

KennethBWSong:
bug: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/28957038

### DIFF
--- a/packages/fx-core/src/component/generator/apiSpec/helper.ts
+++ b/packages/fx-core/src/component/generator/apiSpec/helper.ts
@@ -823,7 +823,12 @@ async function updateAdaptiveCardForCustomApi(
 
     for (const item of specItems) {
       const name = item.item.operationId!.replace(/[^a-zA-Z0-9]/g, "_");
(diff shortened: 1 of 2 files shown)


packages/fx-core/src/component/generator/apiSpec/helper.ts
```diff
@@ -823,7 +823,12 @@ async function updateAdaptiveCardForCustomApi(
 
     for (const item of specItems) {
       const name = item.item.operationId!.replace(/[^a-zA-Z0-9]/g, "_");
-      const [card] = AdaptiveCardGenerator.generateAdaptiveCard(item.item, true);
+      const [card, jsonPath] = AdaptiveCardGenerator.generateAdaptiveCard(item.item, true);
+      card.body.forEach((value) => {
```
SLdragon, line 827: You should only replace $data in root level, child $data should not be replaced.

PR #12246, opened 2024-08-19
feat: add custom copilot basic template in dotnet

xzf0587:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/29033533


packages/fx-core/src/component/generator/templates/templateReplaceMap.ts
```diff
@@ -47,7 +47,9 @@ export function getTemplateReplaceMap(inputs: Inputs): { [key: string]: string }
     useOpenAI: llmService === "llm-service-openai" ? "true" : "",
     useAzureOpenAI: llmService === "llm-service-azure-openai" ? "true" : "",
     openAIKey: openAIKey ?? "",
+    originalOpenAIKey: inputs[QuestionNames.OpenAIKey] ?? "",
```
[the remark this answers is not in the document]
xzf0587: openAIKey is an TTK encrypted value which will be set in env file.
But in VS project, the key will be set in `appsettings.Testtool.json` which does not support TTK encrypted value. We need to set plain value in `appsettings.Testtool.json` and originalOpenAIKey is the plain value.

hund030 (approved):

hund030 (approved):

blackchoey (commented):
Is `env\.env.testtool` and `env\.env.testtool.user` files needed?

templates/csharp/custom-copilot-basic/README.md.tpl
```diff
@@ -0,0 +1,84 @@
+# Overview of the AI Chat Bot template 12345
```
blackchoey, line 1: Remove the tailing `12345`? By the way, seems this readme file is same with the other readme under `.{{NewProjectTypeName}}/README.md.tpl`. Are both of them required?

jayzhang (approved):

templates/csharp/custom-copilot-basic/README.md.tpl
xzf0587: `{{NewProjectTypeName}}/README.md.tpl` is required only. Fixed.

xzf0587:
> Is `env\.env.testtool` and `env\.env.testtool.user` files needed?

In dotnet projects, ttk use appsettings.TestTool.json instead of env.testtool files.

templates/csharp/custom-copilot-basic/env/.env.local.tpl
```diff
@@ -0,0 +1,10 @@
+# This file includes environment variables that can be committed to git. It's gitignored by default because it represents your local development environment.
```
frankqianms, line 1: it is no need to set `.env.local` as a template file

frankqianms (approved):

templates/csharp/custom-copilot-basic/env/.env.local.tpl
xzf0587: Since the gitignore includes .env.local file, using tpl to walkaround the ignore check in TTK repo.

PR #12267, opened 2024-08-23
fix: fix python env load

frankqianms:
workitem: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/29178873

Api keys in `.env.*.user` are encrypted now. Python needs to decrypt them in scripts if keys are still loaded from `.env.*.user`. 
It is not a good idea to write decript code in opensource scripts, so we change guidances in readme and load paths.


xzf0587 (approved):

templates/python/custom-copilot-assistant-assistants-api/src/utils/creator.py.tpl
```diff
@@ -6,7 +6,7 @@ from openai.types.shared_params import FunctionDefinition
 
 from dotenv import load_dotenv
 
-load_dotenv(f'{os.getcwd()}/env/.env.local.user', override=True)
+load_dotenv(f'{os.getcwd()}/.env', override=True)
```
adashen, line 9: instead of loading env. How about just list the param needed in the script with place holder. User could decide whether to use env or hard code?

adashen (approved):
Please also update the testplan

PR #12268, opened 2024-08-23
fix: replace react-script with vite

huimiu:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/28940900

Migrate `react-script` to `vite` in dashboard tab template.


templates/ts/dashboard-tab/package.json.tpl
```diff
@@ -14,22 +15,24 @@
...
+        "vite": "^5.4.0"
     },
     "scripts": {
         "dev:teamsfx": "env-cmd --silent -f .localConfigs npm run start",
-        "start": "react-scripts start",
-        "build": "react-scripts build",
-        "eject": "react-scripts eject",
+        "start": "vite",
+        "build": "tsc && vite build",
+        "server": "vite preview",
```
hund030, line 35: npm run server? Do you mean `serve`? I think npm scripts are verb in convention.

templates/ts/dashboard-tab/tsconfig.json
```diff
@@ -1,20 +1,7 @@
...
-      "noEmit": true,
-      "jsx": "react-jsx"
-    },
-    "include": ["src"]
-  }
\ No newline at end of file
+  "files": [],
+  "references": [
+    { "path": "./tsconfig.app.json" },
+    { "path": "./tsconfig.node.json" }
```
hund030, line 5: Just curious about why do we need two tsconfig?

hund030 (approved):

templates/ts/dashboard-tab/tsconfig.json
huimiu: Two separate TS configs are needed because the app runs in the browser while Vite and its config run in Node, requiring different settings for each environment, with a master config to unify them.

templates/ts/dashboard-tab/package.json.tpl
huimiu: Thanks for pointing out this issue. After thinking it over, we might not need the `serve` command since we provide a preview launch profile. I'll update this in the next PR.

PR #12273, opened 2024-08-26
feat: enforce service management reference for certain tenant

blackchoey:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/28970426

### DIFF
--- a/packages/fx-core/src/component/driver/aad/create.ts
+++ b/packages/fx-core/src/component/driver/aad/create.ts
@@ -32,6 +32,7 @@ import {
   telemetryKeys,
 } from "./utility/constants";
 import { AadSet } from "../../../common/globalVars";
+import { MissingServiceManagementReferenceError } from "./error/missingServiceManagamentReferenceError";
 
 const actionName = "aadApp/create"; // DO NOT MODIFY the name
 const helpLink = "https://aka.ms/teamsfx-actions/aadapp-create";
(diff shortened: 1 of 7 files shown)


KennethBWSong (approved):

packages/fx-core/src/component/driver/botAadApp/create.ts
```diff
@@ -98,12 +99,25 @@ export class CreateBotAadAppDriver implements StepDriver {
...
+        (tokenJson.value.unique_name as string).includes("@microsoft.com");
+
       const startTime = performance.now();
       if (!botAadAppState.botId) {
         context.logProvider?.info(getLocalizedString(logMessageKeys.startCreateBotAadApp));
+
+        // This hidden environment variable is for internal use only.
+        const serviceManagementReference = process.env.TTK_DEFAULT_SERVICE_MANAGEMENT_REFERENCE;
+        if (isMsftAccount && !serviceManagementReference) {
+          throw new MissingServiceManagementReferenceError(actionName);
```
wenytang-ms, line 115: so the reference all from ENV? is there support file?

blackchoey: We now recommend using `aadApp/create` over the `botAadApp/create` action. The help doc will ask user to change `botAadApp/create` to `aadApp/create` to provide service management reference. `botAadApp/create` won't be updated to support service management reference. `aadApp/create` is updated a long time ago with the service management reference support.
Since we hope the env can set default service management reference, I only include env related logic here.

packages/fx-core/src/component/driver/aad/create.ts
```diff
@@ -79,14 +80,30 @@ export class CreateAadAppDriver implements StepDriver {
           )
         );
         context.addTelemetryProperties({ [telemetryKeys.newAadApp]: "true" });
+
+        const tokenJson = await context.m365TokenProvider.getJsonObject({ scopes: GraphScopes });
+        const isMsftAccount =
+          tokenJson.isOk() &&
+          tokenJson.value.unique_name &&
+          (tokenJson.value.unique_name as string).includes("@microsoft.com");
```
wenytang-ms, line 88: it's better to use endsWith("@microsoft.com")?

blackchoey: Good catch! @xiaolang124 do you have any concerns of using `endsWith("@microsoft.com")` to determine microsoft account? If not, I will make the change.

wenytang-ms (approved):

sffamily (approved):

PR #12288, opened 2024-08-28
perf(kiota): update create project in fx-core for TTK integration

KennethBWSong:


packages/spec-parser/test/specParser.test.ts
```diff
@@ -1136,6 +1152,86 @@ describe("SpecParser", () => {
...
       }
     });
+
+    it(" should generate adaptivecard for existing plugin manifest", async () => {
+      const pluginManifest = {
+        schema_version: "1",
+        name_for_human: "test",
+        description_for_human: "test",
+      };
+      const pluginManifestWithAdaptiveCard = {
```
SLdragon, line 1162: where is the adaptive card?

SLdragon (dismissed):

xzf0587 (approved):

packages/fx-core/src/component/generator/apiSpec/helper.ts
```diff
@@ -389,7 +389,11 @@ export async function generateFromApiSpec(
   projectType: ProjectType,
   outputFilePath: SpecParserOutputFilePath
 ): Promise<Result<SpecParserGenerateResult, FxError>> {
-  const operations = inputs[QuestionNames.ApiOperation] as string[];
+  const operations =
+    featureFlagManager.getBooleanValue(FeatureFlags.KiotaIntegration) &&
+    inputs[QuestionNames.ApiPluginManifestPath]
+      ? (await specParser.list()).APIs.filter((value) => value.isValid).map((value) => value.api)
+      : (inputs[QuestionNames.ApiOperation] as string[]);
```
yuqizhou77, line 396: we do validation in line 397. Feels that it would make more sense if moving list() behind validate()

packages/fx-core/src/question/constants.ts
```diff
@@ -44,6 +44,7 @@ export enum QuestionNames {
   FromExistingApi = "from-existing-api", // group name for creating an App from existing api
   ApiSpecLocation = "openapi-spec-location",
   ApiOperation = "api-operation",
+  ApiPluginManifestPath = "api-plugin-manifest-path",
```
yuqizhou77, line 47: might be better to rename or add a comment indicating that this question is for internal integration and won't for UI.

yuqizhou77 (approved):
left two comment.

packages/fx-core/src/component/generator/apiSpec/helper.ts
KennethBWSong: I remember they are different checks. @SLdragon Do you know what's the difference with isValid and specParser.validate() ?

packages/fx-core/src/question/constants.ts
KennethBWSong: Updated.

xzf0587 (approved):

SLdragon (approved):

yuqizhou77 (approved):

PR #12302, opened 2024-08-29
fix: fix sso-tab

Yimin-Jin:
BUG: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/29256520

The reason I added the ajv dependency is that without it, I encounter the following error during startup:
微信截图_20240829151329

### DIFF
--- a/templates/ts/sso-tab-with-obo-flow/package.json.tpl
+++ b/templates/ts/sso-tab-with-obo-flow/package.json.tpl
@@ -10,8 +10,6 @@
     "@microsoft/teams-js": "^2.22.0",
     "@microsoft/teamsfx": "^2.2.0",
     "@microsoft/teamsfx-react": "^3.0.0",
-    "ajv": "^8.17.1",
-    "ajv-keywords": "^5.1.0",
     "axios": "^0.21.1",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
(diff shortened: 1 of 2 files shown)


templates/ts/sso-tab-with-obo-flow/package.json.tpl
```diff
@@ -10,6 +10,8 @@
     "@microsoft/teams-js": "^2.22.0",
     "@microsoft/teamsfx": "^2.2.0",
     "@microsoft/teamsfx-react": "^3.0.0",
+    "ajv": "^8.17.1",
```
huimiu, line 13: Why we need those two dependencies?

Yimin-Jin: During my local debug, not including this dependency results errors. The detailed reason is writed in the description.

Yimin-Jin: What puzzles me is that no one else has reported this error. Is it possible that it only occurs on my local debug?

hund030: I think it's a bad practice to add any dependency that is not required in our source code.

Yimin-Jin: Okay, I'll remove these dependency.

hund030 (approved):

PR #12318, opened 2024-09-02
perf(kiota): update TTK command and refactor spec parser

KennethBWSong:


packages/vscode-extension/src/handlers/createPluginWithManifestHandler.ts
```diff
@@ -35,7 +35,7 @@ export async function createPluginWithManifest(args?: any[]): Promise<Result<any
     TelemetryEvent.CreatePluginWithManifestStart,
     getTriggerFromProperty(args)
   );
-  if (!args || args.length < 2) {
+  if (!args || args.length < 2 || args.length > 3) {
```
SLdragon, line 38: how about args.length !== 2

KennethBWSong: It could be args.length !== 2 || args.length !==3, but I think the above one is easier to update if we need to add parameters

SLdragon: Oh, I know~

SLdragon (approved):

tecton (approved):

supkasar (approved):

PR #12328, opened 2024-09-04
perf(kiota): add telemetry for vsc settings

KennethBWSong:


tecton (commented):
Only one concern is that it will send a new automatic telemetry event which is not triggered by user. It may affect the BI report.

tecton (approved):

PR #12348, opened 2024-09-06
refactor: remove teams chat participant

Alive-Fish:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/29306950

### DIFF
--- a/packages/vscode-extension/src/chat/prompts.ts
+++ b/packages/vscode-extension/src/chat/prompts.ts
@@ -11,7 +11,7 @@ export const defaultSystemPrompt = () => {
   );
 
   return new vscode.LanguageModelChatMessage(
-    vscode.LanguageModelChatMessageRole.System,
+    vscode.LanguageModelChatMessageRole.User,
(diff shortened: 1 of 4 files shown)


packages/vscode-extension/src/chat/prompts.ts
```diff
@@ -11,7 +11,7 @@ export const defaultSystemPrompt = () => {
   );
 
   return new vscode.LanguageModelChatMessage(
-    vscode.LanguageModelChatMessageRole.System,
+    vscode.LanguageModelChatMessageRole.User,
```
tecton, line 14: I think we could also define a new alias type so that we are able to switch back to system role easily in the future.

Alive-Fish: It is unnecessary because the proposed api may not be stable in the future.

lijie-lee (approved):

HuihuiWu-Microsoft (approved):

PR #12359, opened 2024-09-10
chore: add logic for support single meta os taskpane create option

ZYUN-MSFT:
PM Spec [Here](https://microsoft.sharepoint.com/:w:/t/OfficePlatform/ESc9-gy6-ANOsKdGNfOm4hkBgh6fnjuQ9sezP7bpEHWB2w?e=JhfKqS)

### DIFF
--- a/packages/fx-core/src/question/create.ts
+++ b/packages/fx-core/src/question/create.ts
@@ -124,7 +124,9 @@ export function projectTypeQuestion(): SingleSelectQuestion {
           return [projectType];
         }
       } else if (getRuntime(inputs) === RuntimeOptions.NodeJS().id) {
-        if (featureFlagManager.getBooleanValue(FeatureFlags.OfficeAddin)) {
+        if (featureFlagManager.getBooleanValue(FeatureFlags.OfficeMetaOS)) {
+          staticOptions.push(ProjectTypeOptions.officeMetaOS(inputs.platform));
+        } else if (featureFlagManager.getBooleanValue(FeatureFlags.OfficeAddin)) {
           staticOptions.push(ProjectTypeOptions.officeAddin(inputs.platform));
         } else {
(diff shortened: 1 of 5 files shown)


ZYUN-MSFT:
image

image

hermanwenhe:
> > image
> > image
> 
> So Task Pane can be used across WXPO now correct? Can you share the next screen if I select Task Pane?

Next screen would be folder selection:

image

ZYUN-MSFT:
image

image

This option will be avaliable for wxpo. And no programming lang need to be selected.

packages/fx-core/src/question/create.ts
```diff
@@ -124,7 +124,9 @@ export function projectTypeQuestion(): SingleSelectQuestion {
           return [projectType];
         }
       } else if (getRuntime(inputs) === RuntimeOptions.NodeJS().id) {
-        if (featureFlagManager.getBooleanValue(FeatureFlags.OfficeAddin)) {
+        if (featureFlagManager.getBooleanValue(FeatureFlags.OfficeMetaOS)) {
+          staticOptions.push(ProjectTypeOptions.officeMetaOS(inputs.platform));
+        } else if (featureFlagManager.getBooleanValue(FeatureFlags.OfficeAddin)) {
```
tecton, line 129: It looks like the FeatureFlags.OfficeMetaOS is replacing FeatureFlags.OfficeAddin, is it necessary to keep the office addin feature flag?

ZYUN-MSFT: Not replace. The two feature gate control for two different create question options.

hermanwenhe:
@tecton We will use a separate PR to clean these feature gates after WXP MOS public preview.

MuyangAmigo:
Okay sounds good @ZYUN-MSFT @hermanwenhe Thanks!

jayzhang (approved):

PR #12376, opened 2024-09-12
perf: refresh add plugin treeview item

yuqizhou77:
- refresh isDeclarativeCopilot to update development tree view 5 seconds after user saving manifest.json

### DIFF
--- a/packages/vscode-extension/src/manifestListener.ts
+++ b/packages/vscode-extension/src/manifestListener.ts
@@ -1,0 +1,75 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+"use strict";
+
+import * as vscode from "vscode";
+import {
+  isDeclarativeCopilotApp,
+  updateIsDeclarativeCopilotApp,
+  workspaceUri,
+} from "./globalVariables";
+import path from "path";
+import {
+  AppPackageFolderName,
+  ManifestTemplateFileName,
+  TeamsAppManifest,
+} from "@microsoft/teamsfx-api";
+import TreeViewManagerInstance from "./treeview/treeViewManager";
+import { ExtTelemetry } from "./telemetry/extTelemetry";
+import { TelemetryEvent, TelemetryProperty } from "./telemetry/extTelemetryEvents";
+import { isValidProjectV3 } from "@microsoft/teamsfx-core";
+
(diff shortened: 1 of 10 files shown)


packages/vscode-extension/src/manifestListener.ts
```diff
@@ -0,0 +1,75 @@
...
+  AppPackageFolderName,
+  ManifestTemplateFileName,
+  TeamsAppManifest,
+} from "@microsoft/teamsfx-api";
+import TreeViewManagerInstance from "./treeview/treeViewManager";
+import { ExtTelemetry } from "./telemetry/extTelemetry";
+import { TelemetryEvent, TelemetryProperty } from "./telemetry/extTelemetryEvents";
+import { isValidProjectV3 } from "@microsoft/teamsfx-core";
+
+function setAbortableTimeout(ms: number, signal: any) {
```
tecton, line 23: You could take a look at [debounce](https://www.geeksforgeeks.org/lodash-_-debounce-method/) function.

packages/vscode-extension/src/manifestListener.ts
```diff
@@ -0,0 +1,75 @@
...
+      // Clear the timeout and reject the promise if aborted
+      clearTimeout(timeoutId);
+      reject("resolved after clear");
+    });
+  });
+}
+
+export function manifestListener(): vscode.Disposable {
+  let abortController: undefined | AbortController;
+  const disposable = vscode.workspace.onDidSaveTextDocument(
```
tecton, line 41: AFAIK, this triggers when the manifest file is edited in VS Code. If you want to monitor file changes outside VS Code, you could use `createFileSystemWatcher` to monitor change events.

tecton (approved):

packages/vscode-extension/src/manifestListener.ts
yuqizhou77: Yes, we only need to track when user edit the file in vscode.

packages/vscode-extension/src/manifestListener.ts
yuqizhou77: Thanks for letting me know the alternative :)

HuihuiWu-Microsoft (approved):

anchenyi (approved):

PR #12383, opened 2024-09-12
docs: Add changelog for SEP stable release

MuyangAmigo:


sffamily (approved):
But shall we hold on this given the all the Copilot names changed to Agent? so we just did one shot to avoid another major change after this?

PR #12443, opened 2024-09-24
fix: trim manifest short name

jayzhang:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/29597424

### DIFF
--- a/packages/fx-core/src/component/driver/teamsApp/utils/ManifestUtils.ts
+++ b/packages/fx-core/src/component/driver/teamsApp/utils/ManifestUtils.ts
@@ -393,6 +393,32 @@ export class ManifestUtils {
     const manifest = JSON.parse(manifestString) as TeamsAppManifest;
     return ok(manifest);
   }
+
+  /**
+   * trim the short name in manifest to make sure it is no more than 25 length
+   */
+  async trimManifestShortName(
+    projectPath: string,
+    maxLength = 25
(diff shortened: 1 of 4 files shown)


packages/fx-core/src/component/driver/teamsApp/utils/ManifestUtils.ts
```diff
@@ -393,6 +391,32 @@ export class ManifestUtils {
...
+    let hasSuffix = false;
+    let trimmedName = shortName;
+    if (shortName.includes("${{APP_NAME_SUFFIX}}")) {
+      hasSuffix = true;
+      trimmedName = shortName.replace("${{APP_NAME_SUFFIX}}", "");
+    }
+    if (trimmedName.length <= maxLength) return ok(undefined);
+    let newShortName = trimmedName.replace(/\s/g, "").slice(0, maxLength);
+    if (hasSuffix) {
+      newShortName += "${{APP_NAME_SUFFIX}}";
```
anchenyi, line 414: `newShortName` can exceed 25 length since you add a suffix placeholder?

jayzhang: Max length of suffix is 'local', so max length for the name is 25.

anchenyi (approved):

LongOddCode (approved):

PR #12513, opened 2024-10-14
fix: csharp templates package version

Alive-Fish:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/29860268

### DIFF
--- a/templates/csharp/command-and-response/{{ProjectName}}.csproj.tpl
+++ b/templates/csharp/command-and-response/{{ProjectName}}.csproj.tpl
@@ -27,6 +27,7 @@
       <!-- Exclude TeamsFx wwwroot static files which are for frontend only. -->
       <ExcludeAssets>contentFiles</ExcludeAssets>
     </PackageReference>
+    <PackageReference Include="System.Text.Json" Version="6.0.10" />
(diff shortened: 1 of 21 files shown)


xzf0587 (approved):

templates/csharp/command-and-response/{{ProjectName}}.csproj.tpl
```diff
@@ -27,6 +27,7 @@
       <!-- Exclude TeamsFx wwwroot static files which are for frontend only. -->
       <ExcludeAssets>contentFiles</ExcludeAssets>
     </PackageReference>
+    <PackageReference Include="System.Text.Json" Version="6.0.10" />
```
swatDong, line 30: Why some are 8.x and some are 6.x?

Alive-Fish: I have seen some packages depend on the System.Text.Json@8.x.x and some depend on the System.Text.Json@6.x.x. That's a short solution and for long term your template owner should update your main dependencies and fix it fundamentally.

swatDong (approved):

PR #12522, opened 2024-10-15
fix: split PrerequisitesValidationError into Error types of finer granularity

jayzhang:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/29830908

PrerequisitesValidationError => PortsConflictError, SideloadingDisabledError, CopilotDisabledError, NodejsNotFoundError

### DIFF
--- a/packages/fx-core/src/component/deps-checker/depsChecker.ts
+++ b/packages/fx-core/src/component/deps-checker/depsChecker.ts
@@ -1,6 +1,6 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
-import { FxError, UserError } from "@microsoft/teamsfx-api";
+import { UserError } from "@microsoft/teamsfx-api";
 
 export interface DepsChecker {
   getInstallationInfo(installOptions?: InstallOptions): Promise<DependencyStatus>;


supkasar (approved):

packages/vscode-extension/src/debug/depsChecker/taskChecker.ts
```diff
@@ -82,9 +82,9 @@ function addCheckResultsForTelemetry(
   properties: { [key: string]: string },
   errorProps: string[]
 ): void {
-  const [resultRaw, resultSafe] = convertCheckResultsForTelemetry(checkResults);
-  properties[TelemetryProperty.DebugCheckResultsSafe] = resultSafe;
-  properties[TelemetryProperty.DebugCheckResults] = resultRaw;
+  // const [resultRaw, resultSafe] = convertCheckResultsForTelemetry(checkResults);
```
swatDong, line 85: What's this "comment out" for?

Alive-Fish (approved):

packages/vscode-extension/src/debug/depsChecker/taskChecker.ts
jayzhang: We will use different error code and properties to identify the check result instead of a json string. For example, there is a new error named: PortsConflictError with "ports" and "occupied-ports" properties to identify the port numbers to check and the ones that occupied when local debug. So these two properties is no longer needed.

swatDong (approved):

xzf0587 (approved):

LongOddCode (approved):

PR #12555, opened 2024-10-21
refactor: update teamsapp.js to ts in non-sso-tab ts template

Yimin-Jin:
github issue: https://github.com/OfficeDev/teams-toolkit/issues/12432


templates/ts/non-sso-tab/src/static/scripts/teamsapp.ts
```diff
@@ -1,3 +1,5 @@
+import * as microsoftTeams from "@microsoft/teams-js";
```
hund030, line 1: Also update the source code as it's TypeScript?
For example, adding type for hubName.

Yimin-Jin: Sure.

PR #12583, opened 2024-10-24
fix: update @azure/storage-blob to avoid AbortSignal error in Azure S…

Siglud:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/30001202


yukun-dong (approved):

IAM5K (approved):
Changes looks good. This may solve the issue

JerryYangKai (approved):

PR #12608, opened 2024-10-29
feat: enable stream for CEA(js/ts)

QinghuiMeng-M:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/29180531/

### DIFF
--- a/templates/js/custom-copilot-assistant-new/src/app/app.js.tpl
+++ b/templates/js/custom-copilot-assistant-new/src/app/app.js.tpl
@@ -22,6 +22,9 @@ const model = new OpenAIModel({
 
   useSystemMessages: true,
   logRequests: true,
+  {{#CEAEnabled}} 
+  stream: true,
+  {{/CEAEnabled}}
 });
 const prompts = new PromptManager({
   promptsFolder: path.join(__dirname, "../prompts"),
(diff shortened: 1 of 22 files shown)


templates/js/custom-copilot-assistant-new/src/app/app.js.tpl
```diff
@@ -22,6 +22,9 @@ const model = new OpenAIModel({
 
   useSystemMessages: true,
   logRequests: true,
+  {{#CEAEnabled}} 
+  stream: true,
```
frankqianms, line 26: current teams-ai version is 1.1.0 in package.json. Does this version support stream?

kuojianlu: Agree we can upgrade the teams-ai version in package.json BTW

QinghuiMeng-M: in package-lock.json, the dependencies is "@microsoft/teams-ai": "^1.1.0", it allows changes that do not modify the left-most non-zero digit in the [major, minor, patch] tuple
so in fact, npm will down teams-ai 1.5.3

kuojianlu: Yes I know, but let's upgrade it even if we use ^

kuojianlu (approved):

PR #12629, opened 2024-10-31
feat: multi-tenant support for azure account

HuihuiWu-Microsoft:
1> Implement azure account switch tenant logic
2> Only list subscriptions under the current tenant when provisioning(strings updated)
3> Update error message when provision fail

ADO:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/30046147
Figma:
https://www.figma.com/design/1ugjzYWA4oGBzsXV5g1tXL/VSC-%2B-VS-bug-lists?node-id=0-1&node-type=canvas&t=8QO1rPkBu3bhV7CJ-0
Demo:

https://github.com/user-attachments/assets/6820ee76-ff0e-47f3-9e41-28121d68d894

image


packages/vscode-extension/src/treeview/account/azureNode.ts
```diff
@@ -34,7 +34,7 @@ export class AzureAccountNode extends DynamicNode {
       }
     }
     this.contextValue = "signedinAzure";
-    this.eventEmitter.fire(undefined);
+    this.eventEmitter.fire(this);
```
HuihuiWu-Microsoft, line 37: Only update azure node instead of the whole tree. This can fasten the UI update significantly: **before this change, it could take 10s+ to refresh the whole tree; after this change, the label will get updated in 1s.**

Alive-Fish (approved):

jayzhang (approved):

supkasar (approved):

tecton (approved):

PR #12642, opened 2024-11-01
feat: e2e support for multi-tenant

HuihuiWu-Microsoft:
1> append `tenantId` parameter in debug & preview launch url
2> fallback to displayName when collaborator upn is null
3> fallback to unique_name when upn is null to avoid undefined display name for m365 account in provision confirm dialog
4> read m365 tenant id from env file instead of state.env.json file

ADO:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/29860159

### DIFF
--- a/packages/vscode-extension/src/utils/envTreeUtils.ts
+++ b/packages/vscode-extension/src/utils/envTreeUtils.ts
@@ -5,6 +5,9 @@ import { SubscriptionInfo } from "@microsoft/teamsfx-api";
 import { getProvisionResultJson } from "./fileSystemUtils";
 import { workspaceUri } from "../globalVariables";
 import { getV3TeamsAppId } from "./appDefinitionUtils";
+import path from "path";
+import fs from "fs-extra";
+import { dotenvUtil } from "@microsoft/teamsfx-core/build/component/utils/envUtil";
 
 export async function getSubscriptionInfoFromEnv(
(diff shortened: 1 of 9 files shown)


jayzhang (approved):

packages/vscode-extension/src/utils/envTreeUtils.ts
```diff
@@ -34,20 +37,13 @@ export async function getSubscriptionInfoFromEnv(
 }
 
 export async function getM365TenantFromEnv(env: string): Promise<string | undefined> {
-  let provisionResult: Record<string, any> | undefined;
-
-  try {
-    provisionResult = await getProvisionResultJson(env);
```
tecton, line 40: This function seems not to be used any more. Could we delete it?

HuihuiWu-Microsoft: The function is still used in other code blocks. I'll contact feature owner to update the logics.

tecton: After offline discussion, please help to check the upgrade of `state.env.json` file later. `getProvisionResultJson` is still used in other codes.

tecton (approved):

kuojianlu (approved):

PR #12664, opened 2024-11-05
refactor: add feedback loop for Python RAG&Agent CEAs

frankqianms:
workitem: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/30136938

### DIFF
--- a/templates/python/custom-copilot-assistant-assistants-api/src/bot.py.tpl
+++ b/templates/python/custom-copilot-assistant-assistants-api/src/bot.py.tpl
@@ -35,7 +35,7 @@ bot_app = Application[TurnState](
         bot_app_id=config.APP_ID,
         storage=storage,
         adapter=TeamsAdapter(config),
-        ai=AIOptions(planner=planner),
+        ai=AIOptions(planner=planner, enable_feedback_loop=True),
     )
 )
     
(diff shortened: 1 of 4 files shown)


templates/python/custom-copilot-assistant-assistants-api/src/bot.py.tpl
```diff
@@ -76,4 +76,9 @@ async def on_error(context: TurnContext, error: Exception):
...
 
     # Send a message to the user
-    await context.send_activity("The bot encountered an error or bug.")
\ No newline at end of file
+    await context.send_activity("The bot encountered an error or bug.")
+
+@bot_app.feedback_loop()
+async def feedback_loop(_context: TurnContext, _state: TurnState, feedback_loop_data: FeedbackLoopData):
+    # Add custom feedback process logic here.
+    print(f"Your feedback is:\n{json.dumps(asdict(feedback_loop_data), indent=4)}")
```
blackchoey, line 84: I remember there was a bug for basic copilot template, which is because we didn't import the `json` module. Seems this change also do not import that module, do we need to import it in this PR?

frankqianms: Yes, we do! Very important remind! Thank you very much~

blackchoey (approved):

PR #12672, opened 2024-11-06
fix: agent rename

tecton:


MSFT-yiz (approved):

packages/fx-core/resource/package.nls.json
```diff
@@ -303,8 +303,8 @@
   "core.createProjectQuestion.projectType.officeAddin.framework.placeholder": "Select a framework",
   "core.createProjectQuestion.projectType.tab.detail": "Embed your own web content in Teams, Outlook, and the Microsoft 365 app",
   "core.createProjectQuestion.projectType.tab.title": "App Features Using a Tab",
-  "core.createProjectQuestion.projectType.copilotExtension.label": "Copilot Agent",
-  "core.createProjectQuestion.projectType.copilotExtension.title": "App Features using Copilot Agent",
+  "core.createProjectQuestion.projectType.copilotExtension.label": "Agent",
+  "core.createProjectQuestion.projectType.copilotExtension.title": "App Features using Agent",
```
therealjohn, line 307: Where does this appear in the UI? It might need to be plural and lowercase, like "App Features using agents".

therealjohn (changes requested):

packages/fx-core/resource/package.nls.json
tecton: @therealjohn It's the title after user chooses "Agent" capability. I think it follows the convention that each word is capitalized. ("using" should also be capitalized)
image

tecton: Changed to plural form: "App Features Using Agents"

therealjohn (approved):

PR #12676, opened 2024-11-07
test: running some test cases using node 20

hellyzh:


packages/tests/scripts/pvt.json
```diff
@@ -98,10 +88,20 @@
       "remotedebug-chatdata-customapi-ts-openai-win-only",
       "remotedebug-chatdata-customapi-py-azureopenai-win-only",
       "remotedebug-chatdata-customapi-py-openai-win-only"
-      
-
     ],
-    "node-20": ["localdebug-obo-tab"]
+    "node-18": [
+      "localdebug-obo-tab",
```
yiqing-zhao, line 93: Should OBO tab be in the node-20 section instead?

hellyzh: yes, obo has both node 18 and 20 testing
https://github.com/OfficeDev/teams-toolkit/blob/hellyzh/node20/packages/tests/scripts/pvt.json#L171
image

yiqing-zhao (approved):

PR #12688, opened 2024-11-07
fix: kill process tree sync

jayzhang:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/30162620

The root cause of this bug is that TTK check ports immediately after killing the process that occupies the ports and the killing is not finished. So the kill process behavior should be executed in a sync mode to avoid this issue.

### DIFF
--- a/packages/vscode-extension/src/debug/teamsfxTaskHandler.ts
+++ b/packages/vscode-extension/src/debug/teamsfxTaskHandler.ts
@@ -46,7 +46,7 @@ import {
 } from "./common/localDebugSession";
 import { allRunningDebugSessions } from "./officeTaskHandler";
 import { deleteAad } from "./deleteAadHelper";
-import kill from "tree-kill";
+import { processUtil } from "../utils/processUtil";
 
 class NpmInstallTaskInfo {
   private startTime: number;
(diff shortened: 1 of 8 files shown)


tecton (approved):

packages/vscode-extension/src/debug/teamsfxTaskHandler.ts
```diff
@@ -529,11 +529,11 @@ async function onDidStartDebugSessionHandler(event: vscode.DebugSession): Promis
   }
 }
 
-export function terminateAllRunningTeamsfxTasks(): void {
+export async function terminateAllRunningTeamsfxTasks(): Promise<void> {
   for (const task of allRunningTeamsfxTasks) {
     try {
       if (task[1] > 0) {
-        kill(task[1], "SIGTERM");
```
swatDong, line 536: Any need to add default TIMEOUT in case `killProcess` hangs?

jayzhang: Yes, it makes sense.

a1exwang (dismissed):

swatDong (approved):

tecton (dismissed):

swatDong (approved):

tecton (approved):

PR #12690, opened 2024-11-08
fix: need to upgrade version of team-ai sdk

frankqianms:
workitem: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/30173015

Along with some import fixes

### DIFF
--- a/templates/python/custom-copilot-rag-azure-ai-search/README.md.tpl
+++ b/templates/python/custom-copilot-rag-azure-ai-search/README.md.tpl
@@ -56,7 +56,7 @@ This app template also demonstrates usage of techniques like:
     setup finished
     ```
 1. After creating index, please go to your index portal page to set the `descriptionVector` field as retrievable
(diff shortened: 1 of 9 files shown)


blackchoey (approved):

templates/python/custom-copilot-rag-azure-ai-search/README.md.tpl
```diff
@@ -55,6 +55,9 @@ This app template also demonstrates usage of techniques like:
     Upload new documents succeeded. If they do not exist, wait for several seconds...
     setup finished
     ```
+1. After creating index, please go to your index portal page to set the `descriptionVector` field as retrievable
```
frankqianms, line 58: In our setup.py, descriptionVector field is not default set to "retrievable" anymore. Maybe it is because the latest upgrade of azure.search.document to 11.5.2 on Oct.31. But only downgrading the package doesn't work.

The current workaround is to tell users in Readme to manually set the field as retrievable.

adashen (approved):

PR #12716, opened 2024-11-13
refactor: update templates;

xzf0587:


templates/csharp/custom-copilot-assistant-assistants-api/Program.cs.tpl
```diff
@@ -62,6 +65,11 @@ builder.Services.AddTransient<IBot>(sp =>
     });
 
     app.AI.ImportActions(new ActionHandlers());
+    app.OnFeedbackLoop((turnContext, turnState, feedbackLoopData, _) =>
+    {
+        Console.WriteLine($"Your feedback is {turnContext.Activity.Value.ToString()}");
```
blackchoey, line 70: Other language template needs to serialize the feedback loop data before print the value. Do we need to do similar thing in C#?

frankqianms (approved):

templates/csharp/custom-copilot-assistant-assistants-api/Program.cs.tpl
xzf0587: Thanks. Will use another PR to update it.

PR #12778, opened 2024-11-25
perf: update render template logic to support array element

SLdragon:


yukun-dong (approved):

packages/fx-core/src/component/generator/utils.ts
```diff
@@ -174,7 +174,16 @@ function updateTokens(
     } else if (token[0] === "#" || token[0] === "^") {
       token[2] += accShift;
       token[3] += accShift;
-      accShift += updateTokens(token[4] as any, view, tags, accShift);
+
+      let rootView = view;
+      if (Array.isArray(view[token[1]])) {
+        rootView = view[token[1]][0];
```
hund030, line 180: I think we should merge the child view into root view rather than replace it.
Consider using following:
``
rootView = {...view, ...view[token[1]][0]}
```

File: packages/fx-core/src/component/generator/utils.ts [hunk 2]
```diff
@@ -174,7 +174,16 @@ function updateTokens(
...
+      } else if (typeof view[token[1]] === "object") {
+        rootView = view[token[1]];
```
hund030, line 182: Same here.

hund030 (approved):

PR #12786, opened 2024-11-26
fix(oauth): add app id uri in oauth register action

KennethBWSong:
bug: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/30310193?src=WorkItemMention&src-action=artifact_link

### DIFF
--- a/packages/fx-core/resource/yaml-schema/yaml.schema.json
+++ b/packages/fx-core/resource/yaml-schema/yaml.schema.json
@@ -14,7 +14,7 @@
     "version": {
       "type": "string",
       "description": "The version of the yaml file schema",
-      "const": "v1.6"
+      "const": "v1.8"
     },
     "additionalMetadata": {
       "type": "object",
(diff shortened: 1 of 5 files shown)


packages/fx-core/resource/yaml-schema/yaml.schema.json
```diff
@@ -1782,6 +1782,10 @@
             "configurationId": {
               "description": "Required. The configuration id of created OAuth registration.",
               "$ref": "#/definitions/envVarName"
+            },
+            "applicationIdUri": {
+              "description": "Required. The Application ID URI of created OAuth registration.",
+              "$ref": "#/definitions/envVarName"
```
SLdragon, line 1788: This is not required, and only for SSO, if it is the normal OAuth, there is no identify uri

packages/fx-core/src/component/driver/oauth/create.ts
```diff
@@ -116,6 +116,10 @@ export class CreateOauthDriver implements StepDriver {
           outputEnvVarNames.get(OutputKeys.configurationId)!,
           oauthRegistrationRes.configurationRegistrationId.oAuthConfigId
         );
+        outputs.set(
```
SLdragon, line 119: Should check whether it is oauth sso before output it

SLdragon (approved):

SLdragon (approved):

jayzhang (approved):

PR #12844, opened 2024-12-04
ci: local commit

yuqizhou77:


.github/workflows/cd.yml
```diff
@@ -177,6 +177,12 @@ jobs:
       - name: enable prerelease only features
         if: ${{ github.event_name == 'schedule' || (github.event_name == 'workflow_dispatch' && (github.event.inputs.preid == 'alpha' || github.event.inputs.preid == 'beta')) }}
         run: bash .github/scripts/prerelease-feature-control.sh
+
```
wenytang-ms: please set these two steps into one step. it will be more clear when delete such feature

yuqizhou77: Thanks, updated

wenytang-ms (approved):

PR #12851, opened 2024-12-05
feat: troubleshoot UI

yuqizhou77:
- All things listed below is behind feature, which means they are not visible if stable release.
- Add command that user can troubleshoot with GitHub Copilot Teams Agent if selecting text in Teams Toolkit output.
image

- Will notify user that they can troubleshoot with teamsapp when error notification.
image

- TODO: update string and localize once being finalized

### DIFF
--- a/packages/vscode-extension/src/error/common.ts
+++ b/packages/vscode-extension/src/error/common.ts
@@ -81,11 +81,7 @@ export async function showError(e: UserError | SystemError) {
     VsCodeLogInstance.error(`code:${errorCode}, message: ${e.message}\n Help link: ${e.helpLink}`);
     // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
     VsCodeLogInstance.debug(`Call stack: ${e.stack || e.innerError?.stack || ""}`);
-    const buttons = recommendTestTool
-      ? shouldRecommendTeamsAgent
-        ? []
-        : [runTestTool, help]
-      : [help];
+    const buttons = recommendTestTool ? [runTestTool, help] : [help];
     if (shouldRecommendTeamsAgent) {
       buttons.push(troubleshootErrorWithTeamsAgentButton);
     }
--- a/packages/vscode-extension/src/error/common.ts
+++ b/packages/vscode-extension/src/error/common.ts
@@ -2,7 +2,12 @@
 // Licensed under the MIT license.
 
 import { UserError, SystemError, FxError, Result, err, ok } from "@microsoft/teamsfx-api";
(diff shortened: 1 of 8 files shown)


packages/vscode-extension/src/error/common.ts
```diff
@@ -58,7 +81,14 @@ export async function showError(e: UserError | SystemError) {
...
     // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
     VsCodeLogInstance.debug(`Call stack: ${e.stack || e.innerError?.stack || ""}`);
-    const buttons = recommendTestTool ? [runTestTool, help] : [help];
+    const buttons = recommendTestTool
+      ? shouldRecommendTeamsAgent
+        ? []
+        : [runTestTool, help]
+      : [help];
+    if (shouldRecommendTeamsAgent) {
+      buttons.push(troubleshootErrorWithTeamsAgentButton);
```
tecton, line 86: Can this line90 be merged into line 86? And just to confirm that this means replacing test tool recommendation with troubleshooting button.

packages/vscode-extension/src/error/common.ts
```diff
@@ -95,6 +125,9 @@ export async function showError(e: UserError | SystemError) {
     const buttons = recommendTestTool
       ? [runTestTool, issue, similarIssues]
       : [issue, similarIssues];
+    if (shouldRecommendTeamsAgent) {
```
tecton, line 124: What's the UI effect of showing 3/4 buttons?

packages/vscode-extension/src/error/common.ts
yuqizhou77: oh sorry. I just need line 89 and 90. Will remove line 84-88

packages/vscode-extension/src/error/common.ts
yuqizhou77: If at most four buttons. user can hover on the button to view full text.
image

tecton: OK thanks! Also, I'd suggest discussing with PM/UX to decide the button number and the primary button to achieve better effect because the main purpose is to increase entry point to troubleshooting.

tecton (approved):

packages/vscode-extension/src/error/common.ts
yuqizhou77: just left a comment to UX about button number :) Thanks for your suggestion. 
The troubleshoot button has been discussed that should not be the primary one.

PR #12901, opened 2024-12-11
docs: update changelog for 5.12.0

MuyangAmigo:


sffamily (approved):
looks good, so the content is largely the same as the pre-release and blog.

PR #13004, opened 2025-01-03
refactor: update app generator template to support cec

KennethBWSong:


packages/fx-core/src/component/generator/apiSpec/helper.ts
```diff
@@ -1544,12 +1568,22 @@ async function updatePromptSuggestions(specItems: SpecObject[], manifestPath: st
...
         scopes: ["personal"],
-        commands: descriptions.map((des) => {
-          return {
-            title: des.slice(0, 32),
-            description: des.slice(0, 128),
-          };
-        }),
+        commands: [
+          {
+            title: "Hello, how can you help me?",
```
blackchoey, line 1573: minor: I just realized that the template generates its own command. Let's sync with @MuyangAmigo to determine whether we still want to add these fixed commands.
It's not blocking, so feel free to go ahead and merge first.

blackchoey (approved):

yuqizhou77 (approved):

PR #13063, opened 2025-01-16
refactor: Proposed refactoring of API Plugin from Scratch with OAuth

BobGerman:
## Proposed changes to TTK Scaffolding for Declarative Agent with OAuth

The currently generated code requires modifying the middleware to meet application needs; this is not ideal because:

    * it causes the developer to do extra work to understand and modify the middleware
    * it will cause the middleware to vary between projects
    * it will make it harder to replace the middleware with a supported token validation library in the future

These changes are proposed to better separate the API code (Azure functions) from the "middleware".
**It was my intent not to change the logic or default settings at all** - these changes should only rearrange things to make it less necessary for agent developers to modify the middleware.

IF accepted, I will submit a similar PR for the JavaScript version.

## Summary of changes

1. Instead of hard-coding, pass arguments to the authMiddleware function for things that are likely to vary in different apps. The default values for optional arguments will be the same as the current hard-coded values, which are indeed the most commonly needed.

    a. scope (mandatory) - the authorization scope, currently hard coded to "repairs_read". The proposed argument supports either a single scope (string) or an array of scopes ([string]) since some apps will support many scopes for different permissions.

    b. allowedTenants (optional) - an array of tenant IDs that are authorized to use the app

    c. cloudType (optional) - the cloud type which is used to retrieve the correct JWKS URI to handle common, government, and national clouds

    d. issuer (optional) - the issuer, which will vary for multi-tenant apps and different clouds

2. Return claims from authMiddleware so the API can access the user ID, name, upn, tenant ID, etc. that are commonly used to organize data for each user and/or tenant

3. Other changes:

    a. Made the req argument on authMiddleWare mandatory; it was optional but the code would fail if it wasn't specified

    b. Expose the CloudType enum and EntraJwtPayload interface from authMiddlweare.ts so developers don't take dependencies on the libry functions, which may change

    c. Added commonly needed claims such as name and oid to the EntraJwtPayload interface in tokenValidator.ts

Thanks!

### DIFF
--- a/templates/ts/api-plugin-from-scratch-oauth/src/functions/middleware/authMiddleware.ts
+++ b/templates/ts/api-plugin-from-scratch-oauth/src/functions/middleware/authMiddleware.ts
@@ -1,15 +1,25 @@
 import { HttpRequest } from "@azure/functions";
-import { TokenValidator } from "./tokenValidator";
+import { TokenValidator, EntraJwtPayload } from "./tokenValidator";
 import config from "./config";
 import { getEntraJwksUri, CloudType } from "./utils";
+
+// Export symbols app devs will need to use
+export { CloudType } from "./utils";
+export { EntraJwtPayload } from "./tokenValidator";
 
 /**
  * Middleware function to handle authorization using JWT.
  *
  * @param {HttpRequest} req - The HTTP request.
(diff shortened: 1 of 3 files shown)


therealjohn:
CC @MuyangAmigo

PR #13069, opened 2025-01-20
fix: Fix Local Debug SSO Tab with .NET 9

Yimin-Jin:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/30847984

### DIFF
--- a/templates/csharp/sso-tab-ssr/Program.cs.tpl
+++ b/templates/csharp/sso-tab-ssr/Program.cs.tpl
@@ -36,13 +36,13 @@ app.UseAntiforgery();
 app.UseAuthentication();
 app.UseAuthorization();
 
-{{^Net9Framework}}
+{{#IsNet8Framework}}
 app.MapRazorComponents<App>()
     .AddInteractiveServerRenderMode();
-{{/Net9Framework}}
-{{#Net9Framework}}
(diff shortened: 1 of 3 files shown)


huimiu (dismissed):

templates/csharp/sso-tab-ssr/Program.cs.tpl
```diff
@@ -36,7 +36,13 @@ app.UseAntiforgery();
 app.UseAuthentication();
 app.UseAuthorization();
 
+{{^Net9Framework}}
 app.MapRazorComponents<App>()
     .AddInteractiveServerRenderMode();
+{{/Net9Framework}}
+{{#Net9Framework}}
+app.MapRazorComponents<App>()
+    .AddInteractiveServerRenderMode(o => o.ContentSecurityFrameAncestorsPolicy = "'self' *");
```
hund030, line 45: Does this also work in .net 8? If yes, let's make it consistent for both .net 9 and .net 8.

Yimin-Jin: This is a new feature in .net9, this function was not available in .net8.

hund030 (approved):

huimiu (approved):

PR #13080, opened 2025-01-21
perf: update swagger-parser version, remove workaround (#13025)

KennethBWSong:
* perf: update swagger-parser version, remove workaround

* perf: update lock file

* perf: update test cases

* fix: test issue

---------

### DIFF
--- a/packages/spec-parser/src/specParser.ts
+++ b/packages/spec-parser/src/specParser.ts
@@ -90,8 +90,11 @@ export class SpecParser {
         if (!this.parser.$refs.circular) {
           await this.parser.validate(this.spec!);
         } else {
+          // The following code still hangs for Graph API, support will be added when SwaggerParser is updated.
+          /*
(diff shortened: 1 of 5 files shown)


SLdragon (approved):

packages/spec-parser/src/specParser.ts
```diff
@@ -283,6 +286,11 @@ export class SpecParser {
     }
   }
 
+  private async deReferenceSpec(spec: any): Promise<OpenAPIV3.Document> {
```
adashen, line 289: when is this method used?

packages/spec-parser/src/specParser.ts
```diff
@@ -90,8 +90,11 @@ export class SpecParser {
         if (!this.parser.$refs.circular) {
           await this.parser.validate(this.spec!);
         } else {
+          // The following code still hangs for Graph API, support will be added when SwaggerParser is updated.
```
adashen, line 93: Seems no need to change here for the hotfix

PR #13097, opened 2025-01-22
perf: update swagger parser version

KennethBWSong:


packages/spec-parser/package.json
```diff
@@ -49,9 +49,9 @@
     "access": "public"
   },
   "devDependencies": {
+    "@apidevtools/json-schema-ref-parser": "^9.0.6",
```
adashen, line 52: this dev dependencies is needed?

KennethBWSong: This line is removed in previous pr, restore it for keep minimum changes from last stable version and avoid unexpected error.

SLdragon (approved):

PR #13109, opened 2025-01-26
fix: update AssistantsPlanner parameter due to teams-ai sdk updates

frankqianms:
workitem: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/31063511

teams-ai SDK update PR: https://github.com/microsoft/teams-ai/pull/2153

### DIFF
--- a/templates/python/custom-copilot-assistant-assistants-api/src/utils/creator.py.tpl
+++ b/templates/python/custom-copilot-assistant-assistants-api/src/utils/creator.py.tpl
@@ -81,11 +81,19 @@ async def main():
     )
 
     {{#useOpenAI}}
-    assistant = await AssistantsPlanner.create_assistant(api_key=args.api_key, api_version="", organization="", endpoint="", request=options)
+    assistant = await AssistantsPlanner.create_assistant(
+        api_key=args.api_key,
+        azure_ad_token_provider=None,
+        api_version="",
+        organization="",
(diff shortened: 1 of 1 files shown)


ayachensiyuan (approved):

blackchoey (approved):
Have we reported this breaking change issue to Teams-AI and learnt their thought? I'm not sure whether there will be another release that fixes this breaking change and requires us to change the code again. Please create a follow up task as needed.

blackchoey (changes requested):
By the way, please change the SDK version to latest in requirements.txt to avoid possible errors with old SDK.

frankqianms:
> By the way, please change the SDK version to latest in requirements.txt to avoid possible errors with old SDK.

Our declaration is >=1.4.0, <2.0.0, which will hook the latest SDK version.

blackchoey:
> > By the way, please change the SDK version to latest in requirements.txt to avoid possible errors with old SDK.
> 
> Our declaration is >=1.4.0, <2.0.0, which will hook the latest SDK version.

My understanding is 1.4.0 won't work with the latest code as it does not support managed identity, thus cause build error. So we should update the version in requirements.txt.

blackchoey (approved):
Approving as you will create another PR to update SDK version in requirements.txt per offline discussion.

PR #13132, opened 2025-02-06
feat: Add knowledge to DA

Yimin-Jin:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/30479887


huimiu (dismissed):

huimiu (approved):

jayzhang (approved):

packages/vscode-extension/package.json
```diff
@@ -910,6 +910,12 @@
         "enablement": "fx-extension.isTeamsFx && isWorkspaceTrusted && !fx-extension.commandLocked",
         "category": "Teams"
       },
+      {
```
HuihuiWu-Microsoft, line 913: Should we disable the command for non-dc projects?

HuihuiWu-Microsoft (approved):

packages/vscode-extension/package.json
Yimin-Jin: The judgment logic is added to the treeViewManager.ts

Yimin-Jin: This bug also exists in Add Action, I'll fix it in the next PR.

sffamily (approved):

PR #13164, opened 2025-02-11
perf(da): hide standalone add auth command, fix bugs

SLdragon:
[Bug 31324393](https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/31324393): [DA add auth standalone command] apiSpecPath in action not correct

### DIFF
--- a/packages/fx-core/src/component/generator/apiSpec/helper.ts
+++ b/packages/fx-core/src/component/generator/apiSpec/helper.ts
@@ -649,13 +649,15 @@ export async function injectAuthAction(
     return res;
   } else if (
     (!!authScheme && Utils.isOAuthWithAuthCodeFlow(authScheme)) ||
-    authType === "OAuth2PluginVault"
+    authType === "OAuthPluginVault" ||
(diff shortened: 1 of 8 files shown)


packages/fx-core/src/component/generator/apiSpec/helper.ts
```diff
@@ -649,21 +649,24 @@ export async function injectAuthAction(
...
     );
 
     if (await fs.pathExists(localYamlPath)) {
       await ActionInjector.injectCreateOAuthAction(
         localYamlPath,
         authName,
         relativeSpecPath,
-        forceToAddNew
+        forceToAddNew,
+        authType === "MicrosoftEntra"
```
xzf0587: Since "MicrosoftEntra" is used in many places, it is better to use a constant to replace it

anchenyi (dismissed):

jayzhang (approved):

xzf0587 (approved):

tecton (approved):

PR #13189, opened 2025-02-14
perf(spec-parser): separate adaptive card when generate ai plugin

SLdragon:


packages/spec-parser/src/specParser.ts
```diff
@@ -376,6 +376,28 @@ export class SpecParser {
           existingPluginManifestInfo
         );
 
+      const functions = apiPlugin.functions;
+
+      if (functions) {
+        const adaptiveCardFolder = path.join(path.dirname(pluginFilePath), "adaptiveCards");
```
adashen, line 382: the adaptiveCards folder is hard coded?

SLdragon: Yes, for our template, it is by default inside appPackage/adaptiveCards, user can change it by update the reference inside ai-plugin.json file

adashen (approved):

PR #13214, opened 2025-02-17
feat: add MSI to Python CEA bot templates

frankqianms:
workitem: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/31378062


templates/python/custom-copilot-assistant-assistants-api/src/config.py.tpl
```diff
@@ -15,6 +15,8 @@ class Config:
     PORT = 3978
     APP_ID = os.environ.get("BOT_ID", "")
     APP_PASSWORD = os.environ.get("BOT_PASSWORD", "")
+    APP_TYPE = os.environ.get("MicrosoftAppType", "")
```
blackchoey: The environment variable naming convention only allows upper case, can we use the ENVIRONMENT_VARIABLE_NAME style as the environment variable name?

blackchoey: Please also update other similar environment variable names

frankqianms:
@blackchoey all env variables' name were changed to upper case and all Python templates had been tested manually. Please help review again.

blackchoey (approved):

PR #13215, opened 2025-02-17
fix: missing env args for script execution

jayzhang:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/31378030


Siglud (approved):
LGTM

PR #13227, opened 2025-02-18
fix: generator not found error for VS

jayzhang:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/31442322

### DIFF
--- a/packages/fx-core/src/component/coordinator/index.ts
+++ b/packages/fx-core/src/component/coordinator/index.ts
@@ -11,6 +11,7 @@ import {
   InputsWithProjectPath,
   Platform,
   Result,
+  SystemError,
   err,
   ok,
 } from "@microsoft/teamsfx-api";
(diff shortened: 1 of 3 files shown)


packages/fx-core/src/component/coordinator/index.ts
```diff
@@ -187,7 +188,12 @@ class Coordinator {
       // refactored generator
       const generator = Generators.find((g) => g.activate(context, inputs));
       if (!generator) {
-        return err(new MissingRequiredInputError(QuestionNames.Capabilities, "coordinator"));
+        return err(
+          new InternalError(
+            new Error(`Unable to activate any generator by inputs: ${JSON.stringify(inputs)}`),
```
xzf0587: the error message may show this is an inner error.
so user can figure out it and raise bugs?

jayzhang: Updated to a SystemError with message: No activated generator by inputs: xxx. That is helpful for us to trouble shooting.

xzf0587 (approved):

PR #13235, opened 2025-02-19
fix: fix TestCreateTab e2e test

Yimin-Jin:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/31393707
test case: https://github.com/OfficeDev/teams-toolkit/actions/runs/13404820464


packages/tests/src/e2e/frontend/TestCreateTab.tests.ts
```diff
@@ -50,7 +51,9 @@ describe("Create single tab", function () {
         await CliHelper.createProjectWithCapability(
           appName,
           testFolder,
-          Capability.M365SsoLaunchPage
+          Capability.M365SsoLaunchPage,
+          process.env,
```
hund030: use `{...process.env, TEAMSFX_CLI_DOTNET: "false"}` instead of using `process.env` directly.

Yimin-Jin: you are right, I shouldn't affect global variables.

hund030 (approved):

PR #13276, opened 2025-02-24
perf: add telemetry and validation

KennethBWSong:


packages/fx-core/src/question/other.ts
```diff
@@ -942,6 +957,21 @@ export function oauthTokenUrlQuestion(): TextInputQuestion {
...
+          return getLocalizedString("core.addAuthAction.validation.url");
+        }
+
+        try {
+          new URL(input);
+        } catch (error) {
+          return getLocalizedString("core.addAuthAction.validation.url");
+        }
+
+        return undefined;
```
SLdragon: Seems this logic is same as above, try to extract them to one function

KennethBWSong: Updated, thanks!

jayzhang (approved):

supkasar (approved):

PR #13309, opened 2025-02-27
fix: create from tdp

anchenyi:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/31562103

### DIFF
--- a/packages/fx-core/src/component/developerPortalScaffoldUtils.ts
+++ b/packages/fx-core/src/component/developerPortalScaffoldUtils.ts
@@ -293,10 +293,7 @@ export async function updateEnv(
 ): Promise<Result<undefined, FxError>> {
   const localEnvFilePathRes = await pathUtils.getEnvFilePath(projectPath, "local");
   if (localEnvFilePathRes.isErr()) return err(localEnvFilePathRes.error);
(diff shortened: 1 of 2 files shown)


packages/fx-core/src/component/developerPortalScaffoldUtils.ts
```diff
@@ -285,10 +286,25 @@ async function updateManifest(
...
-    TEAMS_APP_ID: appId,
-  });
+// If the template does not have .env.local, write to .env.dev instead.
+export async function updateEnv(
+  appId: string,
+  projectPath: string
+): Promise<Result<undefined, FxError>> {
+  const localEnvFilePathRes = await pathUtils.getEnvFilePath(projectPath, "local");
+  if (localEnvFilePathRes.isErr()) return err(localEnvFilePathRes.error);
+  const localEnvFileExists = await fs.pathExists(
```
yuqizhou77: should this be !!localEnvFilePathRes.value  && await fs.pathExists(localEnvFilePathRes.value)?

anchenyi: Yes, that's better. Thanks for pointing out.

yuqizhou77 (approved):

PR #13310, opened 2025-02-27
refactor(template+core): rename to DA and ME

tecton:
* Rename copilot extension, api plugin, declarative copilot to declarative agent.
* Rename copilot plugin to message extension.
* Add a new template name `DeclarativeAgentWithExistingAction` to replace `DeclarativeAgentBasic` using existing API.


HuihuiWu-Microsoft (dismissed):

packages/fx-core/src/question/constants.ts
```diff
@@ -100,7 +100,7 @@ export enum QuestionNames {
   DestinationApiSpecFilePath = "destination-api-spec-location",
 
   SyncManifest = "sync-manifest",
-  ApiPluginType = "api-plugin-type",
+  DeclarativeAgentType = "api-plugin-type",
```
yuqizhou77: I personally feel this might be better to stick with "api-plugin-type". The reason is that we divide DA into two types, with action (api plugin) and without action. Then for the with action type, we have three more, new-api, api-spec, existing api spec.

But I do not have strong opinion on this if you want to say DA has four types in total. @anchenyi

yuqizhou77: similar thoughts on "DeclarativeAgentStartOptions" and QuestionNames.DeclarativeAgentManifestPath

SLdragon: @yuqizhou77 , I am a bit confused about the name above. Could you please clarify what the ApiPluginType project type refers to?

yuqizhou77: > @yuqizhou77 , I am a bit confused about the name above. Could you please clarify what the ApiPluginType project type refers to?

The options are new-api, api-spec or existing-plugin.

SLdragon: ApiPluginType means existting plugin?

yuqizhou77: > ApiPluginType means existting plugin?

it means whether user choose to have action starting with scratch, existing api spec or existing plugin. It refers to the question after user selecting DA + with action.

anchenyi: I agree with Yuqi, we can call this `ApiPluginType` or something like `ActionType`. Actions (named as api plugins before) are internal components of an DA app, we can add/create actions in a DA app.

jayzhang (dismissed):

yuqizhou77 (dismissed):

jayzhang (dismissed):

hund030 (dismissed):

huimiu:
Updating the copilot extension template name will affect all related BI reports

HuihuiWu-Microsoft (dismissed):

huimiu (approved):

yuqizhou77 (approved):

HuihuiWu-Microsoft (approved):

hund030 (approved):

jayzhang (approved):

PR #13331, opened 2025-03-04
fix: remind user remove aadApp/update action

nliu-ms:
image


packages/fx-core/src/component/driver/aad/create.ts
```diff
@@ -217,12 +221,18 @@ export class CreateAadAppDriver implements StepDriver {
           if (
             error.response!.status === 403 &&
             message.includes(constants.insufficientPermissionErrorMessage) &&
-            !isTestToolEnabledProject(context.projectPath)
+            !isTestToolEnabledProject(context.projectPath) &&
+            process.env.TEAMSFX_ENV == environmentNameManager.getLocalEnvName()
```
blackchoey, line 225: minor: why only enable the new feature for local environment now?

packages/fx-core/src/component/driver/aad/create.ts
```diff
@@ -321,6 +332,28 @@ export class CreateAadAppDriver implements StepDriver {
...
+      }
+      const teamsAppUpdate = yamlProjectModel.value.provision?.driverDefs.find(
+        (d: any) => d.uses === "aadApp/update"
+      );
+      if (teamsAppUpdate) {
+        const message = getLocalizedString(
+          logMessageKeys.needManualRemoveAadUpdate,
+          teamsappYamlPath
+        );
+        await context.ui!.showMessage("warn", message, true);
```
blackchoey, line 352: should we try to update the manifest first before show the warning? user may already has permission to update the app

blackchoey: By the way, have we reached the UX team to review the experience? We may reach them for feedback if haven't. Asking users to remove a step directly seems a little weird. We probably need to tell them why they need to remove that step and what they need to do before remove.

packages/fx-core/src/component/driver/aad/create.ts
nliu-ms: Because this feature is targeting for F5 success rate, so I'd like to limit the impact.

packages/fx-core/src/component/driver/aad/create.ts
nliu-ms: > should we try to update the manifest first before show the warning? user may already has permission to update the app

We go into this function, because create API respond 403 error with insufficient permission. Which means user don't have permission to update as well.

summzhan (approved):

PR #13334, opened 2025-03-04
fix: spfx project scaffold fails

HuihuiWu-Microsoft:
ADO: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/31618421

### DIFF
--- a/packages/fx-core/src/component/generator/templates/templateNames.ts
+++ b/packages/fx-core/src/component/generator/templates/templateNames.ts
@@ -27,6 +27,7 @@ export enum TemplateNames {
   DashboardTab = "dashboard-tab",
   TabSSR = "non-sso-tab-ssr", // handled by SsrTabGenerator
   SsoTabSSR = "sso-tab-ssr", // handled by SsrTabGenerator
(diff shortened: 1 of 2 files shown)


packages/fx-core/src/component/generator/templates/templateNames.ts
```diff
@@ -27,6 +27,7 @@ export enum TemplateNames {
   DashboardTab = "dashboard-tab",
   TabSSR = "non-sso-tab-ssr", // handled by SsrTabGenerator
   SsoTabSSR = "sso-tab-ssr", // handled by SsrTabGenerator
+  TabSPFx = "spfx-tab", // handled by SPFxGeneratorNew
```
tecton, line 30: Is this template name also be added in question tree node?

HuihuiWu-Microsoft: Template seems hardcoded in SPFx generator for now.
image

tecton: I see, the SPFx generator activation is not triggered by template name yet.

tecton (approved):

PR #13372, opened 2025-03-07
test: test declarative agent with none auth

hellyzh:


packages/tests/src/utils/playwrightOperation.ts
```diff
@@ -263,6 +263,83 @@ export async function initPage(
...
+  }
+): Promise<Page> {
+  let page = await context.newPage();
+  page.setDefaultTimeout(Timeout.playwrightDefaultTimeout);
+
+  // open teams app page
+  // https://github.com/puppeteer/puppeteer/issues/3338
+  await Promise.all([
+    page.goto(
+      `https://m365.cloud.microsoft/chat/entity1-d870f6cd-4aa5-4d42-9626-ab690c041429?auth=2`
```
wenytang-ms, line 282: is this site is fixed? will it change?

wenytang-ms (approved):

packages/tests/src/utils/playwrightOperation.ts
hellyzh: yes, fixed

PR #13375, opened 2025-03-09
perf(tsp): update zipAppPackage action to support tsp

KennethBWSong:


packages/fx-core/src/component/driver/teamsApp/createAppPackage.ts
```diff
@@ -101,7 +110,7 @@ export class CreateAppPackageDriver implements StepDriver {
     }
     await fs.mkdir(jsonFileDir, { recursive: true });
 
-    const appDirectory = path.dirname(manifestPath);
+    const appDirectory = path.dirname(hasTTKGeneratedFolder ? generatedFolder : manifestPath);
```
anchenyi, line 113: `path.dirname(generatedFolder)` is `path.join(context.projectPath, "appPackage")`, which is the same as path.dirname(manifestPath) when the generated folder is not existed. Is this expected?

anchenyi: Can we just use appDirectory = path.dirname(manifestPath) here? Because manifestPath has already contained the .generated path.

packages/fx-core/src/component/driver/teamsApp/createAppPackage.ts
```diff
@@ -286,7 +295,10 @@ export class CreateAppPackageDriver implements StepDriver {
       : manifest.copilotAgents?.declarativeAgents;
     // Copilot GPT
     if (declarativeCopilots?.length && declarativeCopilots[0].file) {
-      const declarativeAgentManifestFile = path.resolve(appDirectory, declarativeCopilots[0].file);
+      const declarativeAgentManifestFile = path.resolve(
+        hasTTKGeneratedFolder ? generatedFolder : appDirectory,
```
anchenyi, line 299: Is passing the appDirectory enough?

packages/fx-core/src/component/driver/teamsApp/createAppPackage.ts
KennethBWSong: Updated. Yes we can use appDirectory = path.dirname(manifestPath) here, so removed this change.

anchenyi (approved):

PR #13452, opened 2025-03-17
docs: March release changelog

summzhan:


CarolKigoonya:
We should add a screenshot with the step to show that the debug experience is launched using the F5 feature 
Screenshot (235)1

CarolKigoonya:
Also add screenshots that show the debug console so that people know where to find the debug info

summzhan:
> We should add a screenshot with the step to show that the debug experience is launched using the F5 feature Screenshot (235)1

Hi @CarolKigoonya  our developers have this knowledge, I think don't need to point out especially. And we need to keep the changelog concise.  I will add to the blog and share to you.

summzhan:
> Also add screenshots that show the debug console so that people know where to find the debug info

made edits per your suggestions. you can see the changes in the later commits.

sffamily (approved):

PR #13510, opened 2025-03-21
fix: add defaultAuthorizationPolicy to azure bicep for OAuth

Yimin-Jin:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/31840257

### DIFF
--- a/templates/js/declarative-agent-with-action-from-scratch-oauth/infra/azure.bicep.tpl
+++ b/templates/js/declarative-agent-with-action-from-scratch-oauth/infra/azure.bicep.tpl
@@ -85,17 +85,12 @@ resource authSettings 'Microsoft.Web/sites/config@2021-02-01' = {
           clientId: aadAppClientId
         }
         validation: {
+          defaultAuthorizationPolicy: {
+            allowedApplications: [
+              aadAppClientId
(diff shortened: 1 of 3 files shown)


templates/js/declarative-agent-with-action-from-scratch-oauth/infra/azure.bicep.tpl
```diff
@@ -93,6 +93,11 @@ resource authSettings 'Microsoft.Web/sites/config@2021-02-01' = {
             ]
           }
 {{/MicrosoftEntra}}
+          defaultAuthorizationPolicy: {
```
huimiu: There will be two defaultAuthorizationPolicy items in Microsoft Entra template, we should wrap those code into a placeholder, because we only need those in the OAuth template.

huimiu:
Also, we may need to update the CSharp template as well.

templates/js/declarative-agent-with-action-from-scratch-oauth/infra/azure.bicep.tpl
Yimin-Jin: sure

huimiu (approved):

PR #13513, opened 2025-03-21
feat: add sensitivity label code lens

anchenyi:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/31854871


supkasar (approved):

packages/fx-core/src/question/other.ts
```diff
@@ -1555,3 +1557,97 @@ export function syncManifestQuestionNode(): IQTreeNode {
...
+        return "./appPackage/declarativeAgent.json";
+      } else {
+        if (!inputs.projectPath) {
+          return undefined;
+        }
+        const manifestPath = path.join(inputs.projectPath, AppPackageFolderName, "manifest.json");
+        if (!fs.pathExistsSync(manifestPath)) {
+          return undefined;
+        }
+        const manifestRes = fs.readJsonSync(manifestPath) as TeamsAppManifest;
```
tecton, line 1597: Minor: do we have manifest parser to reuse? It would be better for maintenance.

tecton (approved):

packages/fx-core/src/question/other.ts
anchenyi: Thanks for pointing out. I will use the functions in the manifest utils in next PR.

jayzhang (approved):

PR #13636, opened 2025-04-08
feat: upgrade fx-core and cli versions

jayzhang:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/32202751

### DIFF
--- a/packages/cli/tests/unit/utils.tests.ts
+++ b/packages/cli/tests/unit/utils.tests.ts
@@ -126,7 +126,7 @@ describe("Utils Tests", function () {
         })
       );
       const templates = await getTemplates();
-      expect(templates.length).equals(1);
+      expect(templates.length).equals(2);
     });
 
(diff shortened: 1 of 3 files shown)


packages/cli/tests/unit/utils.tests.ts
```diff
@@ -126,7 +126,7 @@ describe("Utils Tests", function () {
         })
       );
       const templates = await getTemplates();
-      expect(templates.length).equals(1);
+      expect(templates.length).equals(2);
```
tecton, line 129: Why does this UT fail? It seems that CLI version has already been mocked to 2.0.0.

jayzhang: The start version is not 1.0.0-alpha

wenytang-ms (approved):

Alive-Fish (approved):

PR #13655, opened 2025-04-09
feat: replace botbuilders with agent sdk

nliu-ms:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/32176792
1) Some types changed to optional or unknown
- context.activity.value  any -> unknown
2) Some interface removed. E.g. Response, TokenResponse
3) Some method signature changed:
- sendActivity()
- continueConversationAsync() -> continueConversation()
- CloudAdapter constructor
4) Some type names changed
- BotState -> AgentState
- StatePropertyAccessor -> AgentStatePropertyAccessor
5) No default activity type
image
6) isResponse=true, will create connectorClient in adapter.
image

Issues reports to agent sdk:
https://github.com/microsoft/Agents-for-js/issues/232
https://github.com/microsoft/Agents-for-js/issues/231
https://github.com/microsoft/Agents-for-js/issues/239


rido-min:
> Some types changed to optional or unknown
> context.activity.value any -> unknown

yes, this is intentional to play nicely with TS `any` rules

> Some interface removed. E.g. Response, TokenResponse

As we dont support streaming we are using express.Response
We can evaluate to bring back TokenResponse, can you elaborate in which cases you need it? (seems like an impl. detail of SSO)

> Some method signature changed:
> sendActivity()

The only change is that now we require `Activity` instead of `Partial<Activity>` to improve activity validation. Is this a blocker?

> continueConversationAsync() -> continueConversation()

fair, we can change it back.

> CloudAdapter constructor
> Some type names changed

yes this is a breaking change as we are revisiting the config story

> BotState -> AgentState

yes, you will see there are not more `bot` words on the public API surface

> StatePropertyAccessor -> AgentStatePropertyAccessor

we can change it back if needed

231 and 232 are already merged, and I'm taking care of 239 this week.

nliu-ms:
Thanks @rido-min , only opened issues need to be fixed, others are not brokers.

tecton:
Note: namespace "BotBuilderCloudAdapter" need to be renamed as well.

packages/sdk/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@microsoft/teamsfx",
-  "version": "3.0.2",
+  "version": "4.0.0",
```
tecton: I think this should be 4.0.0-alpha.0, the CD pipeline will bump it to 4.0.0 together with templates. Please double check with @wenytang-ms

tecton (approved):

a1exwang (approved):

PR #13776, opened 2025-04-24
perf(da): better error handling, fix path parsing issue

SLdragon:


KennethBWSong (approved):

supkasar (approved):

packages/fx-core/package.json
```diff
@@ -104,7 +104,7 @@
     "@feathersjs/hooks": "^0.6.5",
     "@microsoft/dev-tunnels-contracts": "1.1.9",
     "@microsoft/dev-tunnels-management": "1.1.9",
-    "@microsoft/kiota": "1.25.1",
+    "@microsoft/kiota": "1.25.1-preview.202504170001",
```
tecton, line 107: Will this package have stable version? A version range might be helpful.

tecton (approved):

packages/fx-core/package.json
SLdragon: This feature is under featureflag, and kiota npm package is not fully ready now and may break for range version, so we use preview here.

PR #13840, opened 2025-05-05
fix(template): update template for TypeSpec projects

sebastienlevert:
Refactoring the initial template that developer use for TypeSpec for Microsoft 365 Copilot.


Onokaev (approved):

samwelkanda (approved):

sebastienlevert:
Dependent on https://github.com/microsoft/typespec-copilot-skills/pull/278

packages/fx-core/package.json
```diff
@@ -105,7 +105,7 @@
     "@feathersjs/hooks": "^0.6.5",
     "@microsoft/dev-tunnels-contracts": "1.1.9",
     "@microsoft/dev-tunnels-management": "1.1.9",
-    "@microsoft/kiota": "1.25.1-preview.202504240001",
+    "@microsoft/kiota": "1.25.1-preview.202505010001",
```
KennethBWSong: @SLdragon @sebastienlevert  I believe we need to test the features first before updating this version.

packages/fx-core/src/question/inputs/AddPluginInputs.ts
```diff
@@ -17,10 +17,16 @@ export interface AddPluginInputs extends Inputs {
   "plugin-manifest-path"?: string;
   /** @description Import OpenAPI Description Document */
   "plugin-opeanapi-spec-path"?: string;
+  /** @description OpenAPI Spec Document */
+  "openapi-spec-type"?: "enter-url-or-open-local-file" | "search-api";
```
KennethBWSong, line 21: What is these changes for? I believe this file and AddPluginOptions.ts is auto-generated when building. Would you please run `npm run setup` under root folder if the file changes are by mistake?

templates/common/declarative-agent-typespec/package.json.tpl
```diff
@@ -2,11 +2,9 @@
   "name": "{{SafeProjectNameLowerCase}}",
   "version": "1.0.0",
   "devDependencies": {
-    "@microsoft/typespec-copilot-skills": "latest",
+    "@microsoft/typespec-m365-copilot": "latest",
```
KennethBWSong: Seems this package is not published yet. Would you please help to clarify what's the target date of publishing this package?

KennethBWSong: Also, we would prefer using specific stable version of `@microsoft/typespec-m365-copilot` and other packages here. Do you have any suggestions?

templates/common/declarative-agent-typespec/actions.tsp.tpl
```diff
@@ -0,0 +1,37 @@
+import "@typespec/http";
```
KennethBWSong: Please rename this file to `actions.tsp` as there is no env variable in this file.

packages/fx-core/package.json
sebastienlevert: Oh. Weird that this was added. It should not have added it. But yes, we absolutely need to update to the latest stable release.

packages/fx-core/src/question/inputs/AddPluginInputs.ts
sebastienlevert: They were added by the automatic scripts I think. Let me see if I can remove them?

templates/common/declarative-agent-typespec/package.json.tpl
sebastienlevert: We will have a RC version considered "stable" for now. I'll update when available.

packages/fx-core/src/question/inputs/AddPluginInputs.ts
sebastienlevert: I appears the latest `npm run setup` is bringing these file changes.

packages/fx-core/package.json
sebastienlevert: They seem related to the latest updates to the flow from @SLdragon.

packages/fx-core/src/question/inputs/AddPluginInputs.ts
sebastienlevert: They seem related to the latest updates to the flow from @SLdragon.

PR #13850, opened 2025-05-06
perf(kiota): update error handling for calling Kiota generatePlugin

KennethBWSong:


packages/fx-core/src/common/kiotaClient.ts
```diff
@@ -113,6 +113,14 @@ export async function kiotageneratePlugin(
       throw new Error("Get empty result from kiota");
     }
 
+    if (!result.isSuccess) {
+      const errorMessage = result.logs
+        .filter((log) => log.level >= 4)
```
SLdragon: you can use this constant already exist: ERROR_LOG_LEVEL

KennethBWSong: Updated.

SLdragon (approved):

PR #13867, opened 2025-05-07
fix(template): enable cea by default with preview tag

tecton:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/32769663

Changes on CEA templates:
1. Use app manifest v1.21.
2. Remove feature flag and enable CEA by default.
3. Add Preview tag to debug option title.
4. Ensure manifest property `bots[0].scopes` count is no more than 3 to pass action "teamsApp/validateManifest".

### DIFF
--- a/templates/js/basic-custom-engine-agent/README.md.tpl
+++ b/templates/js/basic-custom-engine-agent/README.md.tpl
@@ -87,8 +87,6 @@ The following are Microsoft 365 Agents Toolkit specific project files. You can [
 - [Microsoft 365 Agents Toolkit Documentations](https://docs.microsoft.com/microsoftteams/platform/toolkit/teams-toolkit-fundamentals)
 - [Microsoft 365 Agents Toolkit CLI](https://aka.ms/teamsfx-toolkit-cli)
 - [Microsoft 365 Agents Toolkit Samples](https://github.com/OfficeDev/TeamsFx-Samples)
-{{#CEAEnabled}}
 
 ## Known issue
 - The agent is currently not working in any Teams group chats or Teams channels when the stream response is enabled.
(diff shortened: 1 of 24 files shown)


templates/js/basic-custom-engine-agent/README.md.tpl
```diff
@@ -87,8 +87,6 @@ The following are Microsoft 365 Agents Toolkit specific project files. You can [
 - [Microsoft 365 Agents Toolkit Documentations](https://docs.microsoft.com/microsoftteams/platform/toolkit/teams-toolkit-fundamentals)
 - [Microsoft 365 Agents Toolkit CLI](https://aka.ms/teamsfx-toolkit-cli)
 - [Microsoft 365 Agents Toolkit Samples](https://github.com/OfficeDev/TeamsFx-Samples)
-{{#CEAEnabled}}
 
 ## Known issue
 - The agent is currently not working in any Teams group chats or Teams channels when the stream response is enabled.
```
blackchoey, line 92: minor: I'm trying to see whether we can remove this known issue notice. Let's follow up with PM about whether steam is required in the template:
1. If yes, let's see whether the template must be able to run in group chat / channel, if not we can remove these 2 scopes as well as the known issue
2. If not, let's disable stream

@frankqianms

blackchoey (approved):

frankqianms (approved):

adashen (approved):

templates/js/basic-custom-engine-agent/README.md.tpl
frankqianms: In Copilot, agent can response without stream.

PR #13900, opened 2025-05-12
fix: copilot debug compatibility issue

jayzhang:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/32835213

For some accounts, the debugging info is subscribed from the main iframe, and for some acounts, the info is subscribed from the sub iframe. The PR covers both two cases, it will subscribe the main frame first and trying to find the target sub iframe at the same time. If the matched sub iframe is found, it will use the sub iframe to get the debugging info.


nliu-ms:
verified with my MSIT and onmicrosoft tenant, it works.

nliu-ms (approved):

PR #14069, opened 2025-05-28
fix: remove m365 sso in copilot

xzf0587:
Fix: https://dev.azure.com/msazure/Microsoft%20Teams%20Extensibility/_workitems/edit/33041252


frankqianms (commented):
Maybe we need to modify related logic in yaml files, too.

blackchoey (approved):

PR #14162, opened 2025-06-13
refactor: move notification to templates

nliu-ms:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/33033162

Move notification related codes into templates, we will remove dependency of @microsoft/teamsfx. Verified in personal app and Team channel.

### DIFF
--- a/templates/vsc/ts/notification-express/src/notification/middleware.ts
+++ b/templates/vsc/ts/notification-express/src/notification/middleware.ts
@@ -3,7 +3,7 @@
 
 import { Activity, ConversationReference } from "@microsoft/agents-activity";
 import { TurnContext, Middleware } from "@microsoft/agents-hosting";
-import { LocalConversationReferenceStore } from "./storage";
+import { IStorage } from "./interface";
 
 /**
  * @internal
@@ -23,9 +23,9 @@ export function getKey(reference: Partial<ConversationReference>): string {
 }
 
 export class NotificationMiddleware implements Middleware {
-  private readonly conversationReferenceStore: LocalConversationReferenceStore;
+  private readonly conversationReferenceStore: IStorage;
 
-  constructor(storage: LocalConversationReferenceStore) {
+  constructor(storage: IStorage) {
     this.conversationReferenceStore = storage;
   }
 
@@ -35,7 +35,9 @@ export class NotificationMiddleware implements Middleware {
       case ActivityType.CurrentBotInstalled:
       case ActivityType.TeamRestored: {
         const reference = context.activity.getConversationReference();
-        await this.conversationReferenceStore.add(getKey(reference), reference, true);
+        await this.conversationReferenceStore.write({
+          [getKey(reference)]: reference,
+        });
         break;
       }
       case ActivityType.CurrentBotMessaged: {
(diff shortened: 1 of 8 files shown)


tecton (dismissed):

templates/vsc/ts/notification-express/src/notification/middleware.ts
```diff
@@ -0,0 +1,97 @@
...
+
+export function getKey(reference: Partial<ConversationReference>): string {
+  // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+  return `_${reference.conversation?.tenantId}_${reference.conversation?.id}`;
+}
+
+export class NotificationMiddleware implements Middleware {
+  private readonly conversationReferenceStore: LocalConversationReferenceStore;
+
+  constructor(storage: LocalConversationReferenceStore) {
```
blackchoey: minor: for production environment, users may need to use other storage other than a local store? In this case, we may need to define an interface for the store and allow users implement their own production ready store.
Not blocking but I feel it deserves to follow up. We also need to mention this in README after we figured out a solution.

templates/vsc/ts/notification-express/src/notification/notification.ts
```diff
@@ -0,0 +1,831 @@
+// Copyright (c) Microsoft Corporation.
```
blackchoey, line 1: Did you change anything to the original implementation in SDK? If not, I will just skip this file :)

templates/vsc/ts/notification-express/src/notification/storage.ts
```diff
@@ -0,0 +1,137 @@
...
+  /**
+   * The Continuation Token to pass to get the next page of results.
+   *
+   * @remarks
+   * Undefined or empty token means the page reaches the end.
+   */
+  continuationToken?: string;
+}
+
+export class LocalConversationReferenceStore {
```
blackchoey: minor: I remember BotFramework has some kind of `IStorage` interface, which we can use to implement the local store. Did agent SDK provides similar interface for us to use?

templates/vsc/ts/notification-express/src/notification/notification.ts
nliu-ms: The NotificationBot constructor slightly changed.
And Removed NotificationTargetType enum.

templates/vsc/ts/notification-express/src/notification/middleware.ts
tecton: I agree that we should highlight in README because I met 2 cases that customers didn't realize the local file when deployed and the installation data was not persistent. We provide an example of Azure Table Storage in large scale notification sample.

nliu-ms: Add an interface to extend Agents sdk Storage interface (as it doesn't have list method).

Also update readme to highlight the local storage is not persistent, suggest user to implement their own.

templates/vsc/ts/notification-express/src/notification/storage.ts
nliu-ms: Add an interface to extend Agents sdk Storage interface (as it doesn't have list method).

blackchoey (approved):

tecton (approved):

swatDong (approved):

PR #14187, opened 2025-06-20
build: update codeowner

nliu-ms:


adashen (changes requested):
Please also remove the previous owners.

adashen (approved):

PR #14235, opened 2025-07-04
test: add condition as expected if get input title failed

hellyzh:


packages/tests/src/utils/playwrightOperation.ts
```diff
@@ -1532,15 +1529,16 @@ export async function validateEchoBot(
...
-    await RetryHandler.retry(async () => {
-      await frame?.waitForSelector(
-        `p:has-text("${
-          options?.expected || ValidationContent.BotWelcomeInstruction
-        }")`
-      );
-      console.log(options?.expected || ValidationContent.BotWelcomeInstruction);
-      console.log("verified bot that it has sent welcome!!!");
-    }, 2);
+    // No need to validate the welcome message since adding app does not work
```
wenytang-ms: why not remove these lines?

wenytang-ms (approved):

packages/tests/src/utils/playwrightOperation.ts
hellyzh: removed

PR #14261, opened 2025-07-11
feat(core+manifest): support app manifest v1.22 and color32x32 icon

tecton:
1. Add v1.22 schema.
2. Support new property color32x32 when validating and creating the app package.
3. Refactor part of actions to use typed app manifest.

feature: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/33679269


packages/fx-core/src/component/driver/teamsApp/createAppPackage.ts
```diff
@@ -111,43 +116,60 @@ export class CreateAppPackageDriver implements StepDriver {
...
-      const error = new FileNotFoundError(
-        actionName,
-        outlineFile,
-        "https://aka.ms/teamsfx-actions/teamsapp-zipAppPackage"
-      );
-      return err(error);
+    // check and include all relative file paths in manifest
+    const relativeFiles = [manifest.icons.color, manifest.icons.outline];
+    const manifestVersion = semver.coerce(manifest.manifestVersion); // ensure manifestVersion is a valid semver
+    if (manifestVersion && semver.gte(manifestVersion, "1.21.0")) {
```
HuihuiWu-Microsoft, line 122: Do we use the oldest version schema for each of manifest property? These different versions of manifest schema looks hard to maintain.

tecton: It uses the first version that supports the property because in previous manifest version it doesn't exist.

HuihuiWu-Microsoft (approved):

llhmm01 (approved):

PR #14267, opened 2025-07-14
fix: add fallback question for asking missing environment variables

qinzhouxu:
To reduce missing env var error, add fallback question for it.


packages/fx-core/resource/package.nls.json
```diff
@@ -857,6 +857,7 @@
   "driver.file.createOrUpdateEnvironmentFile.OpenAIDeploymentName.validation": "Azure OpenAI deployment name cannot be empty.",
   "driver.file.createOrUpdateEnvironmentFile.OpenAIAssistantID.validation": "OpenAI assistant ID cannot be empty.",
   "driver.file.createOrUpdateEnvironmentFile.OpenAIEmbeddingDeploymentName.validation": "Azure OpenAI embedding deployment name cannot be empty.",
+  "driver.file.createOrUpdateEnvironmentFile.genericEnvVar.validation": "Environment variable value cannot be empty.",
```
tecton, line 860: Will this tell user which variable is empty?

qinzhouxu: Since this error message is displayed under a dialog whose title is env var name, and user can aware the current input variable from title, I think this is fine.

sffamily (approved):

tecton (approved):

kuojianlu (approved):

hund030 (approved):

PR #14342, opened 2025-07-28
feat: support sharing agent with a scope of users

tecton:
Task: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/33835755

1. Update share command to share declarative agent with tenant or selected users.
2. Move `teamsApp/shareToOthers` action to deploy stage in DA template.

New UI:

1. Choose share operation when clicking "Share" button.
image
2. Choose select with tenant or selected users. (ownership is temporarily kept here and will be moved)
image
3. User should input email addresses if selected users is chosen.
image

PR #14384, opened 2025-08-04
feat: move adding agent owner to collaborator

tecton:
1. Move "Add agent owner" from "Share" to "Collaborator".
2. Add the feature to let user list agent owner list.
3. Support group email when sharing agent.

Task: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/33835755
Follow: #14342

Changed UI:
* No "share ownership with users" option when user clicks "Share".
image
* Add a new option "Agent" when user chooses to "Add App Owners" or "List App Owners". (controlled by feature flag)
image

CLI:
add parameter "agent" when SHARE feature flag is on.
`atk collaborator grant --agent true`
`atk collaborator status --agent true`


supkasar (dismissed):

packages/fx-core/resource/package.nls.json
```diff
@@ -64,15 +64,20 @@
   "core.collaboration.AccountUsedToCheck": "Account used to check: ",
   "core.collaboration.StartingListAllTeamsAppOwners": "\nStarting to list all app owners for environment: ",
   "core.collaboration.StartingListAllAadAppOwners": "\nStarting to list all Microsoft Entra app owners for environment: ",
+  "core.collaboration.StartingListAllAgentOwners": "\nStarting to list all agent owners for environment: ",
   "core.collaboration.M365TeamsAppId": "App (ID: ",
   "core.collaboration.SsoAadAppId": "SSO Microsoft Entra App (ID: ",
+  "core.collaboration.AgentTitleId": "Agent (ID: ",
```
HuihuiWu-Microsoft, line 70: This is not good practice to follow, use "core.collaboration.AgentTitleId": "Agent (ID: %s)" instead so localization team could handle this better.

tecton: It's a workaround to support CLI colorized output. Changing to `Agent (ID: %s)` will not work for CLI.

HuihuiWu-Microsoft (dismissed):

supkasar (approved):

Alive-Fish (approved):

llhmm01 (approved):

packages/fx-core/src/question/collaborator.ts
```diff
@@ -0,0 +1,182 @@
...
+        id: CollaborationConstants.TeamsAppQuestionId,
+        label: getLocalizedString("core.teamsAppQuestion.label"),
+        description: getLocalizedString("core.teamsAppQuestion.description"),
+      },
+    ],
+    validation: { minItems: 1 },
+    validationHelp: "Please select at least one app type.",
+  };
+  if (featureFlagManager.getBooleanValue(FeatureFlags.ShareEnabled)) {
+    (options.staticOptions as OptionItem[]).push({
```
HuihuiWu-Microsoft, line 175: Does this option always show even if current app is only a Teams app? What if user  clicks the option but doesn't have an agent?

tecton: Yes, it always shows the option like other 2 options. If it's not an agent, there will be an error message showing `Unable to find the "teamsApp/shareToOthers" action config in the m365agents.yml`.
image

HuihuiWu-Microsoft: Approve first and loop @summzhan to decide whether we need to improve the UX like only show this option for agent app.

HuihuiWu-Microsoft (approved):

wh-alice (approved):

PR #14443, opened 2025-08-25
fix(template): unify declarative agent environments and launch profiles

tecton:
Work item: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/34560656/

Changes:
1. basic DA templates support F5 for local environment that calls provision stage to create a personal scope app.
2. advanced DA templates use personal scope for local and dev environment because shared scope is not supported yet, so that sharing is not working.


templates/vsc/common/declarative-agent-typespec/env/.env.dev
```diff
@@ -3,7 +3,7 @@
 # Built-in environment variables
 TEAMSFX_ENV=dev
 APP_NAME_SUFFIX=dev
-AGENT_SCOPE=personal
+AGENT_SCOPE=shared
```
HuihuiWu-Microsoft: Isn't this template an advanced DA template?

tecton: Thanks for the reminding, fixed.

HuihuiWu-Microsoft (approved):

PR #14485, opened 2025-09-08
fix: add help link when user meet error for uploading da package

HuihuiWu-Microsoft:
Provide help link when user failed to upload package. Reuse the existing aka link and I'll update the wiki content later
ADO: fix #2.1 in https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/34648353/?view=edit
Screenshot:
image

### DIFF
--- a/packages/fx-core/src/component/m365/constants.ts
+++ b/packages/fx-core/src/component/m365/constants.ts
@@ -11,3 +11,4 @@ export const outlookCopilotAppId = "d870f6cd-4aa5-4d42-9626-ab690c041429";
 export const outlookBaseUrl = "https://outlook.office.com";
 export const officeBaseUrl = "https://www.office.com";
 export const advancedDASettingUrl = "https://aka.ms/atk-actions/teamsapp-extendToM365";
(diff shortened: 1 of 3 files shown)


packages/fx-core/src/component/m365/constants.ts
```diff
@@ -11,3 +11,4 @@ export const outlookCopilotAppId = "d870f6cd-4aa5-4d42-9626-ab690c041429";
 export const outlookBaseUrl = "https://outlook.office.com";
 export const officeBaseUrl = "https://www.office.com";
 export const advancedDASettingUrl = "https://aka.ms/atk-actions/teamsapp-extendToM365";
+export const M365HelpLink = "https://aka.ms/teamsfx-actions/teamsapp-extendToM365";
```
tecton, line 14: Shall we avoid using `teamsfx` in URL?

HuihuiWu-Microsoft: This is the existing aka link. I don't think we can remove it for now since it'll break former version.

summzhan (approved):

tecton (approved):

PR #14526, opened 2025-09-12
docs: 6.2.0 changelog

summzhan:


packages/vscode-extension/CHANGELOG.md
```diff
@@ -2,6 +2,62 @@
...
+
+### New Features
+
+- **Declarative agents: expand scope support**
+
+In previous versions, declarative agents were uploaded to Copilot with the default `Personal` scope. With this release, developers can now upload their declarative agents using a `Shared` scope, enabling broader distribution on the Copilot platform.
+
+⚠️ Note: If you plan to share an agent with others later, it must be uploaded with the Shared scope from the start.
+
+This behavior is controlled through the `AGENT_SCOPE` parameter defined in the environment configuration files. The parameter value is passed to the `extendToM365` action during the `Provision` lifecycle stage.
```
HuihuiWu-Microsoft, line 20: Should we mention the limitation for advanced DA + shared scope so users will not report bug to us when failed to provision?

summzhan: I think we don't need to mention the limitation in changelog since we have provided clear error message to let users know when they play with the ATK. And this is not a flaw of ATK itself but a limitation of platform service.

wh-alice (approved):

PR #14558, opened 2025-09-18
fix(template): hide da share feature and multi-env support

tecton:
revert #14553


templates/vsc/common/declarative-agent-typespec/env/.env.dev
```diff
@@ -3,7 +3,7 @@
 # Built-in environment variables
 TEAMSFX_ENV=dev
 APP_NAME_SUFFIX=dev
-AGENT_SCOPE=shared
+AGENT_SCOPE=personal
```
HuihuiWu-Microsoft: This env variable will not be used, right?

tecton: Thanks, fixed.

HuihuiWu-Microsoft (approved):

PR #14569, opened 2025-09-19
feat: Teams AI V2 CSharp (Batch 1)

yuqizhou77:
changes included:
- map to new template id if contains breaking change
- templates include: tab, bot, basic agent, agent with customize data, agent with AI search

TODO:
- message exntesion
- agent with custom API
This will be in another PR later


Alive-Fish (approved):

HuihuiWu-Microsoft (approved):

packages/fx-core/src/component/generator/templates/metadata/special.ts
```diff
@@ -44,7 +44,7 @@ export const specialTemplates: Template[] = [
     description: "",
   },
   {
-    id: "custom-copilot-rag-custom-api-csharp",
+    id: "custom-copilot-rag-custom-api-v2-csharp",
```
tecton, line 47: I think it's better to use folder name like "teams-agent-with-data-custom-api-v2" because custom copilot is hard to connect with UI string.

yuqizhou77: let me update this in batch 2 PR :)

PR #14675, opened 2025-10-27
fix: remove build notification bot tutorial

HuihuiWu-Microsoft:
ADO:https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/35680720
getstarted

### DIFF
--- a/packages/vscode-extension/src/handlers/autoOpenProjectHandler.ts
+++ b/packages/vscode-extension/src/handlers/autoOpenProjectHandler.ts
@@ -12,7 +12,6 @@ import {
   ShowScaffoldingWarningSummary,
 } from "../utils/autoOpenHelper";
 import { updateProjectStatus } from "../utils/projectStatusUtils";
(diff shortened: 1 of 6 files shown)


tecton (commented):
Seems like we also need to remove strings and walkthrough in package.json?

HuihuiWu-Microsoft:
> Seems like we also need to remove strings and walkthrough in package.json?

I keep it for now in case this will be added back later which needs extra localization efforts. If we can confirm that this will no longer needed anymore, I can remove all the strings. @summzhan Any ideas?

HuihuiWu-Microsoft:
> > Seems like we also need to remove strings and walkthrough in package.json?
> 
> I keep it for now in case this will be added back later which needs extra localization efforts. If we can confirm that this will no longer needed anymore, I can remove all the strings. @summzhan Any ideas?

@tecton As offline confirmed with Yu, she prefers to keep it for a while before we can confirm this will never be added back.

tecton (approved):

PR #14725, opened 2025-11-04
feat: Local MCP support

rsr3rs:
Allow user to add local MCP.
- Add menu option of local MCP. Upon selection, will generate a `mcp.json` with local MCP server.
- Update plugin manifest to support local MCP.
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/35869844


packages/fx-core/src/component/utils/odrProvider.ts
```diff
@@ -0,0 +1,83 @@
...
+   */
+  static parseODRListOutput(jsonOutput: any): ODRServer[] {
+    const servers: ODRServer[] = [];
+
+    if (!jsonOutput.servers || !Array.isArray(jsonOutput.servers)) {
+      return servers;
+    }
+
+    for (const server of jsonOutput.servers) {
+      const manifest =
```
rodamore, line 41: Only include servers with a package_family_name

packages/fx-core/src/core/FxCore.ts
```diff
@@ -3111,21 +3111,35 @@ export class FxCore {
...
-      auth:
-        mcpAuth === "OAuthPluginVault" && !!registrationId
-          ? {
-              type: "OAuthPluginVault",
-              reference_id: `$\{\{${registrationId}\}\}`,
-            }
-          : undefined,
-    });
+
+    if (aiPluginContent.namespace == "MCP") {
```
rodamore: We shouldn't set namespace; we decided to use mcp:<server ID> for the local_endpoint instead.

packages/fx-core/src/component/utils/odrProvider.ts
```diff
@@ -0,0 +1,91 @@
...
+  identifier: string;
+  tools: ODRTool[];
+}
+
+export interface ODRTool {
+  name: string;
+  description: string;
+  inputSchema: any;
+  outputSchema?: any;
+}
```
rodamore, line 25:
```suggestion
  _meta?: any
}
```

rodamore: And capture below

packages/fx-core/src/component/driver/teamsApp/utils/PluginManifestUtils.ts
```diff
@@ -253,6 +260,245 @@ export class PluginManifestUtils {
...
+    );
+
+    if (localPluginRuntimes.length === 0) {
+      return errors;
+    }
+
+    // local_endpoint format
+    localPluginRuntimes.forEach((rt) => {
+      const actualIdx = manifest.runtimes?.indexOf(rt) ?? -1;
+      if (!rt.spec.local_endpoint.startsWith(LocalMcpPrefix)) {
```
damorer0118: We should only enforce this if there’s some indication the local plugin is local MCP.

krislopez99: the check is for localplugin runtimes that would only be present in the case of local mcp. So if those runtimes are present it validates it, otherwise it returns no errors for that scenario

packages/fx-core/src/component/driver/teamsApp/publishAppPackage.ts
```diff
@@ -73,6 +79,22 @@ export class PublishAppPackageDriver implements StepDriver {
 
     const zipEntries = new AdmZip(archivedFile).getEntries();
 
+    const pluginFile = zipEntries.find((x) => x.entryName === "ai-plugin.json");
```
rodamore: This file name should be parsed from the DA manifest. ai-plugin.json is the default, but I can rename it and can add another plugin with a separate file.

packages/fx-core/src/component/driver/teamsApp/publishAppPackage.ts
```diff
@@ -199,4 +221,53 @@ export class PublishAppPackageDriver implements StepDriver {
...
+          (x) => x.identifier === runtime.spec.local_endpoint.substring(LocalMcpPrefix.length)
+        );
+        if (serverInfo) {
+          const valid = await this.verifyPackageFamilyCertIsValid(serverInfo.packageFamily);
+
+          if (!valid) {
+            allValidCerts = false;
+          }
+        } else {
+          allValidCerts = false;
```
rodamore: If there's no serverInfo, doesn't that mean it is not a local MCP plugin, and we should just ignore? This looks like it would fail for non-MCP local plugins

packages/fx-core/src/component/driver/teamsApp/utils/PluginManifestUtils.ts
rodamore: Can you explain how that is guaranteed? If I create another local plugin that is not local MCP, why wouldn't this code run?

templates/vsc/common/declarative-agent-with-action-from-mcp/.vscode/mcp.json.tpl
```diff
@@ -1,11 +1,21 @@
 {
   "servers": {
-    // This is the MCP server configuration file for VS Code to use.
-    // Click on Start button to fetch tools from your server
-    // Then, click the “ATK: Update Action with MCP” button from CodeLens to select the wanted tools from the prompt list.
+{{#IsLocalMCP}}
+    "{{MCPLocalServerName}}": {
+      "type": "stdio",
+      "command": "odr.exe",
```
rodamore: Can we parse command and all args from `odr list` instead of hardcoding here? If they make a change to this pattern, it will be breaking.

packages/fx-core/src/component/driver/teamsApp/utils/PluginManifestUtils.ts
rsr3rs: I see, I missed the fact that there are non-MCP local plugins. Will fix this

packages/fx-core/src/component/utils/odrProvider.ts
rsr3rs: What are we capturing here? The tool doesn't have a `_meta` field. Do you mean `_meta` of server?

rsr3rs:
@microsoft-github-policy-service agree company="Microsoft"

packages/fx-core/src/component/driver/teamsApp/publishAppPackage.ts
```diff
@@ -86,6 +92,39 @@ export class PublishAppPackageDriver implements StepDriver {
...
+        const declarativeAgentManifest = JSON.parse(declarativeAgentContent);
+
+        if (declarativeAgentManifest.actions) {
+          for (const action of declarativeAgentManifest.actions) {
+            const actionFile = zipEntries.find((x) => x.entryName === action.file);
+            if (actionFile) {
+              const isValid = await this.verifyLocalMCPPluginCerts(actionFile);
+              if (!isValid) {
+                const message =
+                  "MSIX package certificate verification failed. One or more local mcp servers do not have a cert or have a self-signed cert.";
```
Alive-Fish: Can you move the string into `package.nls.json` for localization.

Alive-Fish (approved):

rodamore (approved):

packages/fx-core/src/component/driver/teamsApp/publishAppPackage.ts
```diff
@@ -86,6 +92,40 @@ export class PublishAppPackageDriver implements StepDriver {
     const manifestString = manifestFile.getData().toString();
     const manifest = JSON.parse(manifestString) as TeamsAppManifest;
 
+    const declarativeAgents =
```
HuihuiWu-Microsoft, line 95: Would prefer to move verification logic to `teamsApp/validateManifest` action since this action is only for publishing.

rsr3rs: > Would prefer to move verification logic to `teamsApp/validateManifest` action since this action is only for publishing.

This validation only happens for publish, and I will have to introduce additional complexity if I move it to teamsApp/validateManifest. I could also refactor it within publishAppPackage.ts if that is better

HuihuiWu-Microsoft: If this validation is bind with publish stage, I'm OK to leave it in publish action. Just curious, before publishing, is it OK for developer to test and try the app without validation?

HuihuiWu-Microsoft (approved):

summzhan (approved):
approve the UI part of changes

packages/fx-core/src/component/driver/teamsApp/publishAppPackage.ts
rsr3rs: Side loading with self signed cert is ok. Adding @rodamore for confirmation

rodamore: That’s correct

PR #14879, opened 2025-12-01
refactor: clean test tool feature flag

tecton:


packages/fx-core/src/component/generator/generator.ts
```diff
@@ -90,9 +90,6 @@ export class Generator {
       SolutionName: solutionNameFromVS ?? appName,
       ApiKey: apiKeyActionData,
       OAuth: oauthActionData,
-      enableTestToolByDefault: featureFlagManager.getBooleanValue(FeatureFlags.TestTool)
-        ? "true"
-        : "",
       enableMETestToolByDefault: featureFlagManager.getBooleanValue(FeatureFlags.METestTool)
```
HuihuiWu-Microsoft, line 93: Is `METestTool` still in use? If not, what about cleaning in the same PR?

HuihuiWu-Microsoft (approved):

packages/fx-core/src/component/generator/generator.ts
tecton: Will clean it in following PR since this PR has many changes.

PR #14917, opened 2025-12-05
fix: reading server name and identifier

rsr3rs:
Fixing bugs in local MCP:
- Read server name from manifest,
- For non-odr servers, use placeholder in local_endpoint.
- Convert server names to lower case to match vs code format https://github.com/microsoft/vscode/blob/a7a6e5c17bb5a9cbea962be9fdbd80ac09c8c488/src/vs/workbench/contrib/mcp/common/mcpService.ts#L231

### DIFF
--- a/packages/fx-core/src/component/utils/odrProvider.ts
+++ b/packages/fx-core/src/component/utils/odrProvider.ts
@@ -26,6 +26,19 @@ export interface ODRTool {
 
 export class ODRProvider {
   /**
+   * Check if a server configuration is ODR-based.
+   * @param serverConfig The server configuration object
+   * @returns True if the server is ODR-based, false otherwise
+   */
+  static isODRServer(serverConfig: any): boolean {
+    if (serverConfig?.type !== "stdio" || !serverConfig.command) {
+      return false;
+    }
+    const configCommand = serverConfig.command;
+    return configCommand.toLowerCase() === "odr" || configCommand.toLowerCase().endsWith("odr.exe");
+  }
+
+  /**
    * Parse the output of 'odr list' command
    * @param jsonOutput The JSON output from 'odr list' command
    * @returns Array of parsed ODR servers with their tools
@@ -97,5 +110,23 @@ export class ODRProvider {
       console.error("Error executing odr list:", error);
       return [];
     }
+  }
+
+  /**
+   * Get tools for a specific ODR server by matching command and arguments.
+   * @param command The command of the ODR server (e.g., "odr")
+   * @param args The arguments array for the ODR server
+   * @returns Array of tools from the matching ODR server, or empty array if no match found
+   */
+  static async getToolsForODRServer(command: string, args: string[] = []): Promise<ODRTool[]> {
+    const odrServers = await ODRProvider.listServers();
+
+    const matchingServer = odrServers.find((odrServer) => {
+      return (
+        odrServer.command === command && JSON.stringify(odrServer.args) === JSON.stringify(args)
+      );
+    });
+
+    return matchingServer?.tools || [];
   }
 }
--- a/packages/fx-core/src/component/utils/odrProvider.ts
+++ b/packages/fx-core/src/component/utils/odrProvider.ts
@@ -34,8 +34,8 @@ export class ODRProvider {
     if (serverConfig?.type !== "stdio" || !serverConfig.command) {
       return false;
     }
-    const configCommand = serverConfig.command;
-    return configCommand.toLowerCase() === "odr" || configCommand.toLowerCase().endsWith("odr.exe");
+    const configCommand = serverConfig.command.toLowerCase();
+    return configCommand === "odr" || configCommand.endsWith("odr.exe");
   }
 
   /**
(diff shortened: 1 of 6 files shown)


packages/vscode-extension/src/handlers/updateActionWithMCP.ts
```diff
@@ -30,11 +30,54 @@ function sanitizeMCPName(name: string): string {
   return name
     .replace(/[^a-zA-Z0-9-]+/g, "_")
     .replace(/^_+|_+$/g, "")
-    .substring(0, 13);
+    .substring(0, 13)
+    .toLowerCase();
```
rsr3rs: @Alive-Fish Added this toLowerCase() conversion based on https://github.com/microsoft/vscode/blob/a7a6e5c17bb5a9cbea962be9fdbd80ac09c8c488/src/vs/workbench/contrib/mcp/common/mcpService.ts#L231.

packages/vscode-extension/src/handlers/updateActionWithMCP.ts
```diff
@@ -140,11 +195,11 @@ export async function updateActionWithMCP(args?: any[]): Promise<Result<any, FxE
         return err(result.error);
       }
       const originalMcpName = result.value.result as string;
-      mcpName = originalMcpName.replace(/[^a-zA-Z0-9]/g, "").substring(0, 10);
+      mcpName = sanitizeMCPName(originalMcpName);
```
rsr3rs: @Alive-Fish Changed this to standard sanitization. Was there a particular reason to use a different sanitization?

Alive-Fish (dismissed):

packages/vscode-extension/src/handlers/updateActionWithMCP.ts
```diff
@@ -30,11 +30,50 @@ function sanitizeMCPName(name: string): string {
...
+ */
+async function extractLocalServerIdentifier(
+  serverConfig: any,
+  originalServerName: string
+): Promise<string> {
+  if (serverConfig?.type !== "stdio") {
+    return originalServerName;
+  }
+
+  const configCommand = serverConfig.command || serverConfig.args?.[0] || "";
```
rodamore: In what scenario is command not provided but args are?

packages/vscode-extension/src/handlers/updateActionWithMCP.ts
```diff
@@ -46,15 +85,17 @@ export async function updateActionWithMCP(args?: any[]): Promise<Result<any, FxE
...
-  let localServerIdentifier =
-    args && args.length > 0 ? extractLocalServerIdentifier(args[0].serverConfig) : undefined;
 
   // Sanitize mcpName if it's provided as an argument
   if (mcpName) {
     mcpName = sanitizeMCPName(mcpName);
   }
 
+  let localServerIdentifier =
+    args && args.length > 0 && isLocalMCP
```
rodamore: isLocalMCP already checks args
```suggestion
    isLocalMCP
```

packages/vscode-extension/src/handlers/updateActionWithMCP.ts
```diff
@@ -112,18 +153,26 @@ export async function updateActionWithMCP(args?: any[]): Promise<Result<any, FxE
...
-              : (serverConfig.url as string);
+          let detail: string;
+          if (serverConfig.type === "stdio") {
+            if (serverConfig.command) {
+              const args = serverConfig.args ? (serverConfig.args as string[]).join(" ") : "";
+              detail = args
+                ? `${serverConfig.command as string} ${args}`
+                : (serverConfig.command as string);
+            } else {
+              detail = "stdio";
```
rodamore: This seems like an error state, can we just skip it?

packages/vscode-extension/src/handlers/updateActionWithMCP.ts
```diff
@@ -162,7 +211,7 @@ export async function updateActionWithMCP(args?: any[]): Promise<Result<any, FxE
 
   inputs[QuestionNames.MCPForDAServerUrl] = server;
   inputs[QuestionNames.MCPForDAServerName] = mcpName;
-  if (isLocalMCP && localServerIdentifier) {
+  if (isLocalMCP) {
     inputs[QuestionNames.MCPLocalServerIdentifier] = localServerIdentifier;
   }
```
rodamore, line 221: Should we use `tool.name.toLowerCase().includes` below to make sure we're normalizing both ways?

rodamore: Or can we use a regex match to capture the tool names with index? Is there anything else we can compare if that returns multiple matches?

rodamore (changes requested):

packages/fx-core/src/component/utils/odrProvider.ts
```diff
@@ -25,6 +25,19 @@ export interface ODRTool {
...
+  /**
+   * Check if a server configuration is ODR-based.
+   * @param serverConfig The server configuration object
+   * @returns True if the server is ODR-based, false otherwise
+   */
+  static isODRServer(serverConfig: any): boolean {
+    if (serverConfig?.type !== "stdio" || !serverConfig.command) {
+      return false;
+    }
+    const configCommand = serverConfig.command;
```
rodamore:
```suggestion
    const configCommand = serverConfig.command.toLowerCase();
```
So you only have to lower case once

rodamore (approved):

Alive-Fish (approved):

summzhan (approved):

HuihuiWu-Microsoft (approved):

PR #14930, opened 2025-12-08
fix: Skip scaffolding for MetaOS upgrade and handle missing env file in Extend to DA

samyakjainmicrosoft:
## Summary
Fixes the "Extend to Declarative Agent" feature that was failing with `TemplateNotFoundError` when upgrading existing Office add-in projects to Declarative Agents.

## Root Cause
1. **Primary Issue**: The `getTemplateInfos()` method was triggering template scaffolding for both new and upgrade scenarios. Upgrade scenarios should not scaffold templates since the project structure already exists and only needs modification.
2. **Secondary Issue**: The `unifyProjectID()` function assumed the `env/.env.dev` file exists, but the file copy operation excludes the `env` folder (via `NOT_COPY_FOLDERS` filter), causing an ENOENT error.

## Validation
Verified that extend to DA is working fine and DA project is also working fine.

Attaching video for the reference:
https://github.com/user-attachments/assets/1ab04fec-4125-45ba-a233-20de1870e553

### DIFF
--- a/packages/fx-core/tests/common/featureFlags.test.ts
+++ b/packages/fx-core/tests/common/featureFlags.test.ts
@@ -42,6 +42,6 @@ describe("FeatureFlagManager", () => {
   it("listEnabled", async () => {
     mockedEnvRestore = mockedEnv({ TEAMSFX_CLI_DOTNET: "true", SME_OAUTH: "true" });
     const list = featureFlagManager.listEnabled();
-    chai.assert.deepEqual(list, ["TEAMSFX_CLI_DOTNET", "SME_OAUTH"]);
+    chai.assert.deepEqual(list, ["TEAMSFX_CLI_DOTNET", "SME_OAUTH", "TEAMSFX_DA_METAOS"]);
   });
 });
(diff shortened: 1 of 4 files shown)


packages/fx-core/tests/common/featureFlags.test.ts
```diff
@@ -42,6 +42,6 @@ describe("FeatureFlagManager", () => {
   it("listEnabled", async () => {
     mockedEnvRestore = mockedEnv({ TEAMSFX_CLI_DOTNET: "true", SME_OAUTH: "true" });
     const list = featureFlagManager.listEnabled();
-    chai.assert.deepEqual(list, ["TEAMSFX_CLI_DOTNET", "SME_OAUTH"]);
+    chai.assert.deepEqual(list, ["TEAMSFX_CLI_DOTNET", "SME_OAUTH", "TEAMSFX_DA_METAOS"]);
```
HuihuiWu-Microsoft: The feature flag 'TEAMSFX_DA_METAOS' is false by default. Why do we need to include it as enabled toggle here?

samyakjainmicrosoft: Oh okay, actually this unit test case was failing for me because it was true for me. got your point will remove this

HuihuiWu-Microsoft (dismissed):

tecton (approved):

PR #14993, opened 2025-12-16
build: do vulnerability check

qinqingxu:


.github/workflows/cd.yml
```diff
@@ -112,6 +114,175 @@ jobs:
 
       - uses: pnpm/action-setup@v4
 
+      - name: Check package.json files for security vulnerabilities(VSC)
+        id: security-check-vsc
+        continue-on-error: true  # Allow pipeline to continue, notification will be sent on failure
+        run: |
+          echo "Checking all package.json and package.json.tpl files under templates/vsc for security vulnerabilities..."
```
tecton, line 121: It would be nice to use a script.

tecton (approved):

PR #15018, opened 2025-12-18
Change variable declaration to `alias` in generate-env.js

### LINKED ISSUE #15017: TSP Template script creates `const` for envvars - suggest `alias` to support string interpolation
**Describe the bug**
When declared as a constant, the envvar can't be used in string concatenation (string interpolation). For example, say you have an authenticated endpoint. When envvars are constants, you can't use it when creating an `alias`. 

The problem is the script in the template, **./scripts/generate-env.js**, generates `const` for each envvar, but it should be an `alias`.

**To Reproduce**
1. In a new DA - TSP project, try to concat an envvar with a string... observe error

    Image

3. Change the variable to a constant, but when used in some TSP objects, like `OAuth2Auth`... observe error

    Image

**Expected behavior**
If the envvar generated was an `alias`, all works:

Image

 - OS: macOS
 - Version  v26.1

andrewconnell:
When declared as a constant, the envvar can't be used in string concatenation (string interpolation). For example, say you have an authenticated endpoint. When envvars are constants, you can't use it when creating an `alias`.

If this PR is merged, this fixes the issue and allows envvars to be used in string interpolation scenarios. See #15017 for the issue that this fixes with examples

fixes #15017

### DIFF
--- a/templates/vsc/common/declarative-agent-typespec/scripts/generate-env.js
+++ b/templates/vsc/common/declarative-agent-typespec/scripts/generate-env.js
@@ -70,7 +70,7 @@
       // Escape backslashes and double quotes for safe TS string literal
       const escaped = value.replace(/\\/g, "\\\\").replace(/\"/g, '\\"');
 
-      tspLines.push(`  const ${ident} = "${escaped}";`);
+      tspLines.push(`  alias ${ident} = "${escaped}";`);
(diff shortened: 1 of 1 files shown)


andrewconnell:
fyi @sebastienlevert

sebastienlevert (approved):

PR #15030, opened 2025-12-22
fix: resolve Invalid id in Terminal add in still installed error

archiesaxenaa:
For the Bug : https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/36076480/

**Fix: Proper add‑in cleanup on debug stop**

`office-addin-debugging start `registers add-ins via `atk install`, but `office-addin-debugging stop` did not call `atk uninstall`, leaving add-ins registered after the debug session ends.

**Change:**
Added a custom uninstall script that extracts the Manifest ID and calls atk uninstall before running office-addin-debugging stop. Updated package.json to route stop through this script.

**Result:**
Clean add-in unregistration when office app is closed and debug is stopped.

image (1)


samyakjainmicrosoft (commented):
you should make changes on office-addin-debugging side, this is not the recommended way

archiesaxenaa:
> you should make changes on office-addin-debugging side, this is not the recommended way

Updated PR: https://github.com/OfficeDev/Office-Addin-Scripts/pull/968

PR #15120, opened 2026-01-12
fix: da templates prerequisites checking

qinzhouxu:
https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/36387195


tecton (approved):

HuihuiWu-Microsoft:
Do we care about VS?

qinqingxu (approved):

PR #15160, opened 2026-01-15
fix: script action parse output from Task API

### LINKED ISSUE #15135: set-teamsfx-env does not expand variables
**Describe the bug**
I'm using a script to gather the current user principal to send it to the bicep template like this:

```yaml
  - uses: script
    name: set environment variables
    with:
      run: $AZURE_PRINCIPAL_ID=(az ad signed-in-user show --query id -o tsv);
        echo "::set-teamsfx-env
        AZURE_PRINCIPAL_ID=${AZURE_PRINCIPAL_ID}"
```

When I run it I can see in the output the correct ID:

```shell
*  Executing task: $AZURE_PRINCIPAL_ID=(az ad signed-in-user show --query id -o tsv); echo $AZURE_PRINCIPAL_ID; echo "::set-teamsfx-env AZURE_PRINCIPAL_ID=${AZURE_PRINCIPAL_ID}" 

::set-teamsfx-env
AZURE_PRINCIPAL_ID=ee******-****-****-****-********46e9
 *  Terminal will be reused by tasks, press any key to close it. 
```

But then, in the .env.dev file, the unexpanded value is stored:

```env
AZURE_PRINCIPAL_ID=${AZURE_PRINCIPAL_ID}
```

**To Reproduce**
1. Write a script
2. Store a value inside a script variable (ex: `$VAR="value"`)
3. Use echo "::set-teamsfx-env ..." to write the variable to the environment
4. See the unexpanded variable in the .env.dev file

**Expected behavior**
Write expanded value into the .env file

 - OS: Windows
 - Version 6.4.2

**CLI Information (please complete the following information):**
 - OS: Windows

**Additional context**

Script is running on powershell

tecton:
Fix #15135

### DIFF
--- a/packages/cli/src/userInteraction.ts
+++ b/packages/cli/src/userInteraction.ts
@@ -35,6 +35,7 @@ import {
   UserCancelError,
   assembleError,
 } from "@microsoft/teamsfx-core";
+import { spawn } from "child_process";
 import fs from "fs-extra";
 import open from "open";
 import path from "path";
@@ -46,7 +47,6 @@ import { cliSource } from "./constants";
 import { CheckboxChoice, SelectChoice, checkbox, select } from "./prompts";
 import { errors } from "./resource";
 import { getColorizedString } from "./utils";
-import { spawn } from "child_process";
 
 export const inquirerPrompts = {
   confirm,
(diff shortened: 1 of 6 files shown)


packages/cli/src/userInteraction.ts
```diff
@@ -641,15 +641,22 @@ class CLIUserInteraction implements UserInteraction {
       const command = isWindows ? "cmd.exe" : "/bin/bash";
       const commandArgs = isWindows ? ["/c", args.cmd] : ["-c", args.cmd];
       logger.info(`Executing task: ${args.cmd}`);
+      let output = "";
       const childProcess = spawn(command, commandArgs, {
-        stdio: "inherit",
+        stdio: ["inherit", "pipe", "pipe"],
```
HuihuiWu-Microsoft, line 646: what's the difference with the added two?

tecton: They are used for stdout and stderr that we can capture the actual script output string.

HuihuiWu-Microsoft (approved):

Alive-Fish (approved):

Issue #14142, opened 2025-06-11
Getting error during Deploy lifecycle for Teams Bot "Cannot read properties of undefined (reading 'enabledHostNames')"


imrankhan-mohdjais:
**Describe the bug**
I have provisioned all the resources. When clicking deploy, an error message comes out from the Output terminal. 

[...]

PR #15348, opened 2026-02-24
fix(fx-core): harden Azure zip deploy endpoint error handling

EmilioAK:
## Summary
This change hardens `AzureZipDeployImpl.getZipDeployEndpoint` so deploy failures return actionable errors instead of crashing on malformed/failed ARM responses.

## Problem
When the App Service details API returned a non-success or unexpected payload, we accessed `responseData.properties.enabledHostNames` directly and could crash with:

`TypeError: Cannot read properties of undefined (reading 'enabledHostNames')`

That hid the real root cause (for example `403 AuthorizationFailed`).

## Changes
- Added explicit handling for fetch/network failures.
- Read response body safely and fail with context if body read fails.
- Check `response.ok` before assuming success payload shape.
- Parse JSON safely and report parse failures with response preview.
- Handle ARM error payloads (`{ error: { code, message } }`) even if status is 2xx.
- Validate `properties.enabledHostNames` is a string array.
- Validate SCM host exists before constructing zip deploy endpoint.
- Throw typed `GetZipDeployEndpointError` for endpoint-resolution failures.
- Added new localized strings for `GetZipDeployEndpointError` in:
  - `package.nls.json`
  - `package.nls.zh-cn.json`
- Added/updated tests in `AzureDeployImpl.test.ts` for:
  - happy path
  - non-success status
  - success status with Azure `error` payload
  - malformed `enabledHostNames` payload

## Validation
- `NODE_OPTIONS="--no-experimental-strip-types" pnpm --filter @microsoft/teamsfx-core exec mocha "tests/component/driver/deploy/azure/AzureDeployImpl.test.ts"`
  - Result: `17 passing`
- Happy path isolation:
  - `NODE_OPTIONS="--no-experimental-strip-types" pnpm --filter @microsoft/teamsfx-core exec mocha "tests/component/driver/deploy/azure/AzureDeployImpl.test.ts" --grep "Get zip deploy endpoint$"`
  - Result: `1 passing`
- Manual repro using local CLI build now reports actionable `GetZipDeployEndpointError` (including `403 AuthorizationFailed`) instead of `TypeError`.

## Additional Notes
- The zh-CN localization strings in this PR were drafted with AI, and I can't confirm if they are accurate.
- I started working on this because of #14142. This doesn't fix that issue, but the fact that there was no error handling here is still relevant, I think.


EmilioAK:
@microsoft-github-policy-service agree

PR #15351, opened 2026-02-25
fix: codeql package downgrade issue

HuihuiWu-Microsoft:


HuihuiWu-Microsoft:
Use https://github.com/OfficeDev/microsoft-365-agents-toolkit/pull/15350#pullrequestreview-3851848887 instead

PR #15564, opened 2026-03-20
feat(fx-core): decouple VS/VSC template release with dynamic version lookup

qfai:
## Summary

This PR decouples the VS and VSC/CLI template release pipelines (Phase 1 of Option C plan).

### Changes

**Dynamic VS template version lookup** (utils.ts, templates-config.json)
- `getTemplateVSUrl` is now async and queries `vsTagListURL` to find the latest `templates-vs@` tag matching `vsVersionPattern` (`~18.4`)
- Adds `getTemplateVSLatestVersion()` export, parallel to the existing `getTemplateLatestVersion()` for VSC
- Adds  `vsVersionPattern` fields to `templates-config.json`

**VS metadata hot-update** (FxCore.ts)
- Adds `fetchOnlineTemplateMetadataForVS()` which downloads `metadata.zip` from the latest `templates-vs@` release
- Caches to `~/.fx/vs-metadata/` (separate from VSC/CLI's `~/.fx/metadata/`)
- Already called from `serverConnection.ts` `listen()` on VS extension startup

**VS metadata read path isolation** (metadata/index.ts)
- `getTemplateMetadataConfig` now reads from `~/.fx/vs-metadata/` for `Platform.VS` and `~/.fx/metadata/` for VSC/CLI

**CLI metadata hot-update** (engine.ts)
- `CLIEngine.start()` now fire-and-forgets `fetchOnlineTemplateMetadata()` on startup, matching VSC extension behavior

**CD pipeline** (cd.yml)
- VS RC and stable template releases now include `metadata.zip` alongside `csharp.zip`
- Adds Generate/Update VS Template Tag List steps after stable VS release (parallel to existing VSC tag list mechanism)

## workitem
[link](https://dev.azure.com/msazure/Microsoft%20Teams%20Extensibility/_workitems/edit/37100162)

### DIFF
--- a/packages/fx-core/src/common/templates-config.json
+++ b/packages/fx-core/src/common/templates-config.json
@@ -4,6 +4,7 @@
     "tagPrefix": "templates@",
     "vstagPrefix": "templates-vs@",
     "vsversion": "18.4.2",
+    "vsVersionPattern": "~18.4",
     "tagListURL": "https://github.com/OfficeDev/microsoft-365-agents-toolkit/releases/download/template-tag-list/template-tags.txt",
     "templateDownloadBaseURL": "https://github.com/OfficeDev/microsoft-365-agents-toolkit/releases/download",
     "templateReleaseURL": "https://github.com/OfficeDev/microsoft-365-agents-toolkit/releases/expanded_assets",
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -354,6 +354,7 @@ jobs:
           clear_attachments: true
           files: |
             ${{ github.workspace }}/templates/build/fallback/csharp.zip
+            ${{ github.workspace }}/templates/build/metadata.zip
       
       - name: Get VS Templates version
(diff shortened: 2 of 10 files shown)


packages/fx-core/src/common/templates-config.json
```diff
@@ -4,6 +4,7 @@
     "tagPrefix": "templates@",
     "vstagPrefix": "templates-vs@",
     "vsversion": "18.4.2",
+    "vsVersionPattern": "~18.4",
```
wh-alice, line 7: Is this dup with vsversion?

qfai: no, this is means how this version will look for the template version, this version will look for patterns begin with 18.4, this make it possible that binary 18.4.2 can look for 18.4.3's template version. Before the change, 18.4.2's binary will always look for 18.4.2's template version, if we want to update the template, we need to update template 18.4.2's content, now we can publish 18.4.3.

HuihuiWu-Microsoft (approved):

PR #15572, opened 2026-03-23
feat: support gcc h and dod clouds and skip teams app related actions

HuihuiWu-Microsoft:
Feature: https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/36995382/?view=edit

Screenshot:
[Note: take 'GCC M' as example for testing as preauth is not done by service yet and blocks e2e test]
When user provision:
image

When user publish:
image

When user preview local/remote in copilot:
Updated bizchat url will be used
image


HuihuiWu-Microsoft:
@summzhan Please help review the strings and UX.

summzhan:
I didn't see the corresponding strings in the file changes hence comment here instead of comment inline.

"Skipping teamsApp/create as target env GCC M is not supported" is misleading and not accurate, giving the wrong impression this step is skipped without doing anything, but in fact in this action ATK is generating ID locally right? Suggest following:

"teamsApp/create generates Teams app ID locally instead of calling Developer Portal API since the API is not yet available in GCC H/DoD"

The other two suggest to change to:
"Skipping XXX as the required Developer Portal API is not yet available in GCC H and DoD."

packages/fx-core/resource/package.nls.json
```diff
@@ -853,6 +853,9 @@
   "driver.teamsApp.progressBar.publishTeamsAppStep1": "Checking if the app is already submitted to tenant App Catalog",
   "driver.teamsApp.progressBar.publishTeamsAppStep2.1": "Update published app",
   "driver.teamsApp.progressBar.publishTeamsAppStep2.2": "Publishing app...",
+  "driver.teamsApp.warning.unsupportedCloud": "Skipping %s as the required Developer Portal API is not yet available in GCC H and DoD.",
+  "driver.teamsApp.warning.publishUnsupportedCloud": "Skipping %s as the required Developer Portal API is not yet available in GCC H and DoD. Please upload to Teams Admin Center manually.",
```
HuihuiWu-Microsoft, line 857: @summzhan Please help review the string for publish to ask user to manually publish agent.

summzhan (approved):
Reviewed Strings in logs look good to me.

qfai (approved):

Alive-Fish (approved):

wh-alice (approved):

PR #15626, opened 2026-03-31
fix: classify mos APIs to send correct event

HuihuiWu-Microsoft:


packages/fx-core/src/common/wrappedAxiosClient.ts
```diff
@@ -155,6 +155,16 @@ export class WrappedAxiosClient {
...
   }
 
+  private static isMOSApi(url: string): boolean {
+    const regex = /titles\.(prod|msit)\.mos\.microsoft\.com/;
+    const matches = regex.exec(url);
+    return matches != null && matches.length > 0;
+  }
+
+  static getMOSApiRelativePath(url: string): string {
+    return url.split(/titles\.(prod|msit)\.mos\.microsoft\.com/)[2] ?? "";
```
qinzhouxu: Seems the pattern we are getting is different than before, will this cause issue

packages/fx-core/src/component/m365/serviceConstant.ts
```diff
@@ -77,4 +77,24 @@ export const MOS3ApiDefinitions: any = {
     method: "GET",
     path: /^\/marketplace\/v1\/users\/titles\/[^\/]+\/preview/,
   },
+  PostBuilderTitleAllowed: {
+    key: "builder_share_title_with_users",
+    method: "POST",
+    path: /^\/builder\/v1\/users\/titles\/[^\/]+\/allowed\?idType=(TitleId|AppId)/,
```
qinzhouxu, line 83: How about mark query parameter as optional as in #15624, since in function `onRequest`, the request.url does not contain query parameter, it's stored apart in `request.params` by axois

Image

HuihuiWu-Microsoft: I think it does as we got records like `<mos_unclassified__builder_v1_users_titles_T_f05589ae-88ec-6198-ae36-8b929d2edb72_allowed?idType=TitleId-url>`. I've verified locally and it's working well.

tecton (approved):

PR #15828, opened 2026-04-30
fix: declarative agent with copilot connector fails when opened from TDP portal

qfai:
## Problem

When opening a declarative agent from TDP portal and selecting **declarative - add a copilot connector**, the scaffold fails with:

```
code:gen.TemplateNotFoundError, message: Unable to find template: declarative-agent-with-graph-connector
```

## Root Cause

`TdpGenerator` is the first generator in the chain and activates whenever `inputs.teamsAppFromTdp !== undefined` (i.e. any TDP flow). This takes priority over `CombinedProjectGenerator`.

The template name `declarative-agent-with-graph-connector` does **not exist as a standalone folder** in any template zip -- it is a virtual combined template that `CombinedProjectGenerator` handles by splitting into:
- `graph-connector` (TypeScript)
- `declarative-agent-basic` (Common)

When `TdpGenerator` intercepts this, it gets 0 files from the zip and throws `TemplateNotFoundError`.

The "declarative - no action" path works because its template (e.g. `declarative-agent-basic`) exists directly in the zip.

## Fix

1. `TdpGenerator.activate()`: Exclude `DeclarativeAgentWithGraphConnector` so `CombinedProjectGenerator` handles it.
2. `CombinedProjectGenerator.post()`: Call `developerPortalScaffoldUtils.updateFilesForTdp()` when `inputs.teamsAppFromTdp` is set, so TDP manifest patching still runs.


qfai:
Closing in favor of new PR from clean branch based on dev.

PR #15859, opened 2026-05-08
fix: update hover message for playground env

neil-yechenwei:
For playground env, now it indicates to execute provision to see the placeholder which doesn't make sense since playground doesn't need provision. So I updated hover message in this PR.

The related ADO bug is https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/33482927/?view=edit

### DIFF
--- a/packages/vscode-extension/src/hoverProvider.ts
+++ b/packages/vscode-extension/src/hoverProvider.ts
@@ -69,7 +69,9 @@ export class ManifestTemplateHoverProvider implements vscode.HoverProvider {
       }
       args = [{ type: "env", env: envName }];
     } else {
-      for (const envName of envNames) {
+      for (const envName of envNames.filter(
+        (name) => name !== environmentNameManager.getPlaygroundEnvName()
+      )) {
         const envInfo = envInfos[envName];
         const value = envInfo[key];
         if (value) {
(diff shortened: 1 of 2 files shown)


packages/vscode-extension/src/hoverProvider.ts
```diff
@@ -77,6 +77,9 @@ export class ManifestTemplateHoverProvider implements vscode.HoverProvider {
         } else {
           if (envName === environmentNameManager.getLocalEnvName()) {
             message += `**${envName}** Trigger debug to see placeholder value \n\n`;
+          } else if (envName === environmentNameManager.getPlaygroundEnvName()) {
```
HuihuiWu-Microsoft: @neil-yechenwei But under playground env, actually remote resources will not be provisioned. So trigger debug in playground will not generate these envs either. 
My proposal is to filter playground env. Need @summzhan to confirm the user experience.

summzhan: In the playground env there is no BotID right?

So playground could be none. or just remove playground option in BotID's hover message.

packages/vscode-extension/src/hoverProvider.ts
```diff
@@ -77,6 +77,9 @@ export class ManifestTemplateHoverProvider implements vscode.HoverProvider {
...
+            const commandUri = vscode.Uri.parse("command:fx-extension.localdebug");
+            message += `**${envName}**: [Trigger Debug in Microsoft 365 Agents Playground to see placeholder value](${commandUri.toString()}) \n\n`;
           } else {
             const commandUri = vscode.Uri.parse("command:fx-extension.provision");
             message += `**${envName}**: [Trigger Teams: Provision in the cloud command to see placeholder value](${commandUri.toString()}) \n\n`;
```
summzhan: happens to noticed that in line 85: "Trigger Teams :provision in the cloud command ..."

the command should have been renamed back to 1 year ago. Sepurised that in our code there is still left over old naming remained unupdated.

The new name of command is: "Microsoft 365 Agents: Provision"

Image

neil-yechenwei: Updated the hover message

packages/vscode-extension/src/hoverProvider.ts
neil-yechenwei: Filtered the playground.

neil-yechenwei:
@HuihuiWu-Microsoft , @summzhan , I updated PR according to your suggestions. Please take another look. Thanks.

HuihuiWu-Microsoft (approved):

PR #15953, opened 2026-05-18
feat: add `atk import/export openplugin` for interop with OpenPlugins

rsullivan00:
Tracked by: https://dev.azure.com/msazure/Microsoft%20Teams%20Extensibility/_workitems/edit/37883091

Adds two new CLI commands for bidirectional interop between Microsoft 365 Agents Toolkit projects and the [Open Plugin Spec](https://openplugin.dev) (including the Claude Code and Cursor plugin variants). Together they form a lossless round-trip.

## Commands

- `atk import openplugin --path <plugin-dir> [--output <project-dir>] [--privacy-url ...] [--terms-url ...] [--website-url ...] [--default-auth-type Auto|None|OAuthPluginVault|ApiKeyPluginVault] [--manifest-kind open-plugin|claude-plugin|cursor-plugin]`
  - Scaffolds a full ATK project from an existing `.plugin/plugin.json` (or `.claude-plugin/`, `.cursor-plugin/`) directory: `appPackage/manifest.json`, `agentSkills`, `agentConnectors` (built from `.mcp.json` remote MCP servers), icons, and the static baseline (`m365agents.yml`, `README`, `.vscode`, `env`) from the new `open-plugin-import` template.
- `atk export openplugin --path <project-dir> [--output <plugin-dir>] [--manifest-kind open-plugin|claude-plugin|cursor-plugin]`
  - Writes an ATK project back out as an Open Plugin directory: `plugin.json`, `.mcp.json`, `skills/`, `commands/`, `color.png`, `outline.png`.

Both are gated by the `TEAMSFX_OPENPLUGIN_IMPORT_EXPORT` feature flag (default `true`).

## Lossless round-trip

`atk export openplugin` writes an `x-microsoft-365-agents-toolkit` extension block at the root of `plugin.json` containing every field the Open Plugin Spec cannot natively represent: `manifestVersion`, `id`, `packageName`, `accentColor`, `developer.{name,websiteUrl,privacyUrl,termsOfUseUrl}`, `name.{short,full}`, `description.{short,full}`, and per-connector overrides (`displayName`, `description`, `authorization`). The importer's parser reads this block on a subsequent import so `--privacy-url` / `--terms-url` / `--website-url` become optional in the round-trip case.

E2E verified on `knowledge-work-plugins/bio-research`: `import` -> `export` -> re-`import` (no flags the second time) produced a **byte-identical** `appPackage/manifest.json`.

## Test summary

- fx-core `openPlugin` tests: **98 passing** (including new `exporter.test.ts` and `roundtrip.test.ts`)
- `FxCore.openPlugin.test.ts`: 10 passing
- CLI `commands.tests.ts`: 138 passing
- Lint: zero errors on touched files


packages/fx-core/src/component/generator/openPlugin/assets.ts
```diff
@@ -0,0 +1,126 @@
+// Copyright (c) Microsoft Corporation.
```
HuihuiWu-Microsoft: Current generator design is to have default template with placeholders and these placeholders can be replaced runtime. The benefit is that template release can be dynamic and separated from ATK release. What about follow the same pattern to have a default template for openPlugin convert? Those variables can be parsed and replaced in post() interface.

rsullivan00: Done in 638a1e4 — moved the baseline files (m365agents.yml, README, .gitignore, .vscode, env) into a real template at `templates/vsc/common/openplugin-convert/` with mustache `{{appName}}` placeholders. The converter now calls `Generator.generateTemplate(ctx, projectPath, TemplateNames.OpenPluginConvert, "common")` to scaffold the baseline, and the variable-length outputs (manifest, skill folders, icons) are written as a post-scaffold step. `assets.ts` is removed. This way the template can be updated through the normal template release pipeline without an fx-core change.

packages/fx-core/src/common/featureFlags.ts
```diff
@@ -26,6 +26,9 @@ export class FeatureFlagName {
   static readonly SensitivityLabelEnabled = "TEAMSFX_SENSITIVITY_LABEL";
   static readonly DAMetaOS = "TEAMSFX_DA_METAOS";
   static readonly CFShortcutMetaOS = "TEAMSFX_CF_SHORTCUT_METAOS";
+  static readonly MCPForDA = "TEAMSFX_MCP_FOR_DA";
```
HuihuiWu-Microsoft: This change to add back 'TEAMSFX_MCP_FOR_DA' feature flag seems not related with current PR

templates/vsc/common/open-plugin-import/m365agents.yml.tpl
```diff
@@ -0,0 +1,39 @@
+# yaml-language-server: $schema=https://aka.ms/m365-agents-toolkits/v1.11/yaml.schema.json
+# Generated by 'atk import openplugin'.
+version: v1.11
+
+environmentFolderPath: ./env
+
+# Triggered when 'atk provision' is executed
+provision:
```
HuihuiWu-Microsoft, line 8: In provision stage, `copilotAgent/publish` action is needed for the generated app so the app can be uploaded to MOS3.
Also for publish stage, it should also be  `copilotAgent/publish` action with `tenant` scope instead of the legacy `teamsApp/publishAppPackage` action.

packages/fx-core/src/common/featureFlags.ts
rsullivan00: Removed in cbc56d99cb — the flag was already deleted upstream in #15992 and our branch reintroduced it accidentally during the rebase. Good catch.

templates/vsc/common/open-plugin-import/m365agents.yml.tpl
rsullivan00: Done in cbc56d99cb.

- `publish` now uses `copilotAgent/publish` with `scope: tenant` and writes `M365_PUBLISHED_TITLE_ID` / `M365_PUBLISHED_APP_ID` (drops `teamsApp/publishAppPackage`).
- `provision` adds a trailing `copilotAgent/publish` step with `scope: ${{AGENT_SCOPE}}` that writes `M365_TITLE_ID`, `M365_APP_ID`, `SHARE_LINK` so the app uploads to MOS3.
- `env/.env.dev` gained `AGENT_SCOPE=shared` plus the new env vars.

Verified end-to-end by re-importing `knowledge-work-plugins/productivity` and `atk install` — Title/App IDs returned successfully.

HuihuiWu-Microsoft (approved):

PR #15838, opened 2026-05-01
feat: Added evals/prompts.json file to DA templates and updated the READMEs…

sakov-ms:
This pull request adds agent evaluation support to all Microsoft 365 Agents Toolkit project templates. It introduces documentation and sample configuration files to help developers test and measure the quality of their agents using the Microsoft 365 Copilot Agent Evaluations CLI. Each template now includes a sample `evals/prompts.json` file and updated README instructions.

AZDO Ticket - https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/36981324

**Agent Evaluation Support:**

* Added a new "Evaluating Agents" section to the README files for all declarative agent project templates, guiding users to install and use the `@microsoft/m365-copilot-eval` CLI for structured agent evaluations. [[1]](diffhunk://#diff-30507386453aeec2818b029fb02b0017269a76814ced7f2e6858d7322b3fb846R53-R56) [[2]](diffhunk://#diff-b556ce42d9d05cfcc24067b0db090e11acfbcd6d1c8337662a22746b1a176adbR34-R37) [[3]](diffhunk://#diff-82761c3c075b979b8800d033224b0a967d923606a25113591bd539bbb70bdeedR52-R55) [[4]](diffhunk://#diff-8986324c6c0df473e15d9fb3261572d20d160663307ae0baf2f0f8a970087591R72-R75) [[5]](diffhunk://#diff-61e04d7e65099ee327fd5f9c73c945183807fa5db395cba9588babd3e99ed49dR75-R78) [[6]](diffhunk://#diff-27f0f373ca83264c45d46719010a0ea15f7f800581bbc9b936934fa3913e1f31R82-R85)
* Introduced a sample `evals/prompts.json` file in each template, providing an example dataset for agent evaluation with default evaluators and a test prompt/response.… with details on Evaluating Agents.

PR #15962, opened 2026-05-19
feat: Added evals/prompts.json file to DA templates and updated the READMEs

sakov-ms:
Here is the Azure work item - https://msazure.visualstudio.com/Microsoft%20Teams%20Extensibility/_workitems/edit/36981324

Added evals/prompts.json file to DA templates and updated the READMEs.


templates/vsc/common/declarative-agent-basic/env/.env.dev.tpl
tecton, line 12: KEY is usually stored in `.env.dev.user` so that it get encrypted and ignored in git.

qfai:
close due to duplicate with #15838

PR #16086, opened 2026-06-04
feat(fx-core): route v3 scaffold through v4 distribution channel

Alive-Fish:
## What

Wires v3's scaffold main path through the **merged v4 distribution channel** (#16082), behind the default-off `TEAMSFX_V4_ENABLED` feature flag.

When the flag is **off** (default), `DefaultTemplateGenerator.scaffolding` runs the unchanged v3 `Generator.generate(..., TemplateActionSeq)` path — zero behavior change.

When the flag is **on**, scaffolding resolves and renders templates via the v4 channel: `resolveTemplateSource` → `loadResolvedPackage` → `openTemplatePackage` → render.

## How

- **`v4TemplateBridge.ts`** (new): `scaffoldFromV4Channel` + `renderTemplateEntries`. The renderer re-adds the `${name}/` prefix that `openTemplatePackage` strips, then applies the v3 `filterFn` / `fileNameReplaceFn` / `fileDataReplaceFn` verbatim, so output is **byte-identical** to the legacy `unzip` path.
- **`defaultGenerator.ts`**: flag-gated branch in `scaffolding`; v4 failures throw `FxError` (same semantics as the legacy local-template action).
- **Telemetry**: same `GenerateTemplate` event; `template-name` unchanged (query continuity). New `template-channel` (= `v3`/`v4`, written before the branch so failures are attributable) + `template-package-source` / `-version` / `-digest` (written right after resolution so digest/render failures still carry origin).
- **Flag**: single global `TEAMSFX_V4_ENABLED` (no per-template / per-scenario sub-flags by design).

## Scope

Only the v3→v4 integration. The v4 distribution channel itself already merged in #16082.

## Tests

- `v4TemplateBridge.test.ts` (5): prefix rebuild, `.tpl` mustache render + suffix strip, binary verbatim, `filterFn` exclusion, trailing-slash locator boundary.
- Regression: full `generator.test.ts` (99) green with flag off — v3 path unchanged.

### Follow-up (not in this PR)

End-to-end test with flag on + a real floor zip (byte-for-byte vs v3 output across representative scenarios), and the release-time `v4.bundled` → `false` flip.


packages/fx-core/src/component/generator/v4TemplateBridge.ts
```diff
@@ -0,0 +1,126 @@
...
+ * directory (zip-slip). Throws `TemplateOutputPathError` on escape.
+ */
+function resolveTemplateOutputPath(destination: string, entryName: string): string {
+  const base = path.resolve(destination);
+  const outputPath = path.resolve(base, entryName);
+  const relative = path.relative(base, outputPath);
+  if (relative === "" || relative.startsWith("..") || path.isAbsolute(relative)) {
+    throw new TemplateOutputPathError(entryName);
+  }
+  return outputPath;
```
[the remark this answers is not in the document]
Alive-Fish: Good catch — fixed in 7ec66e0838. The guard now rejects only a true parent-directory escape: relative === \..\ or relative.startsWith(\..\ + path.sep) (plus absolute paths). An in-root filename that merely starts with \..\ (e.g. \..foo\) is now allowed, with a regression test covering it.

packages/fx-core/src/component/generator/defaultGenerator.ts
```diff
@@ -153,7 +154,22 @@ export class DefaultTemplateGenerator implements IGenerator {
 
     await actionContext?.progressBar?.next(ProgressMessages.generateTemplate);
     context.logProvider.debug(`Downloading app template "${templateName}" to ${destinationPath}`);
-    await Generator.generate(generatorContext, TemplateActionSeq);
+    const useV4Channel = featureFlagManager.getBooleanValue(FeatureFlags.V4Enabled);
+    merge(actionContext?.telemetryProps, {
+      [TelemetryProperty.TemplateChannel]: useV4Channel ? "v4" : "v3",
+    });
+    if (useV4Channel) {
```
HuihuiWu-Microsoft, line 161: Does this affect VS templates?

Alive-Fish: Yes, will consider VS together in the future. Currently the feature flag is disabled by default.

HuihuiWu-Microsoft (approved):

summzhan (approved):

PR #16089, opened 2026-06-04
feat(fx-core): pull metadata from v4 release when TEAMSFX_V4_ENABLED

Alive-Fish:
## What

Transitional follow-up to #16086. When the v4 channel flag (`TEAMSFX_V4_ENABLED`) is on, `fetchOnlineTemplateMetadata` downloads `metadata.zip` from the `templates-v4@<ver>` release instead of `templates@<ver>`. The v4 release shares the exact same version string as the v3 release, so only the tag prefix differs; version resolution, cache, unzip target (`~/.fx`) and the metadata/ui readers are untouched.

The v4 publish script (`publish-v4-channel.sh`) now mirrors `metadata.zip` onto the v4 release, and `cd.yml` passes its path.

## Why

Keeps v4-channel metadata version-consistent with v4 template content while the v3 channel stays the default. Explicitly transitional — remove once `selector.json` drives metadata distribution.

## Tests

Added `should download metadata from the v4 release tag when V4 channel is enabled`; all 23 `fetchOnlineTemplateMetadata*` tests pass.

### DIFF
--- a/packages/fx-core/src/core/FxCore.ts
+++ b/packages/fx-core/src/core/FxCore.ts
@@ -66,6 +66,7 @@ import {
   setErrorContext,
   setTools,
 } from "../common/globalVars";
+import { featureFlagManager, FeatureFlags } from "../common/featureFlags";
 import { clearLocaleCache, getLocalizedString } from "../common/localizeUtils";
 import { ListCollaboratorResult, PermissionsResult } from "../common/permissionInterface";
 import { getProjectMetadata, isValidProjectV3 } from "../common/projectSettingsHelper";
@@ -2675,8 +2676,16 @@ export class FxCore extends FxCoreOpenPluginPart {
         return ok(undefined); // Already up-to-date
       }
 
-      // Construct metadata.zip download URL based on tag prefix and version
-      const tag = `${templateConfig.tagPrefix}${latestVersion}`;
+      // Construct metadata.zip download URL based on tag prefix and version.
+      // Transitional: when the v4 channel is enabled, pull metadata from the
+      // v4 release tag (`templates-v4@<ver>`). The v4 release shares the exact
+      // same version string as the v3 `templates@<ver>` release (both minted
+      // from the same templates version), so only the tag prefix differs.
+      // Remove this branch once selector.json drives metadata distribution.
+      const tagPrefix = featureFlagManager.getBooleanValue(FeatureFlags.V4Enabled)
+        ? templateConfig.v4tagPrefix
+        : templateConfig.tagPrefix;
+      const tag = `${tagPrefix}${latestVersion}`;
       const metadataZipUrl = `${templateConfig.templateDownloadBaseURL}/${tag}/metadata.zip`;
 
       const zip = await fetchZipFromUrl(metadataZipUrl);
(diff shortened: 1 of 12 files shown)


packages/fx-core/src/core/FxCore.ts
```diff
@@ -2675,8 +2676,16 @@ export class FxCore extends FxCoreOpenPluginPart {
...
 
-      // Construct metadata.zip download URL based on tag prefix and version
-      const tag = `${templateConfig.tagPrefix}${latestVersion}`;
+      // Construct metadata.zip download URL based on tag prefix and version.
+      // Transitional: when the v4 channel is enabled, pull metadata from the
+      // v4 release tag (`templates-v4@<ver>`). The v4 release shares the exact
+      // same version string as the v3 `templates@<ver>` release (both minted
+      // from the same templates version), so only the tag prefix differs.
+      // Remove this branch once selector.json drives metadata distribution.
+      const tagPrefix = featureFlagManager.getBooleanValue(FeatureFlags.V4Enabled)
```
qfai: if (!(await needDownload())) {
        return ok(undefined); // Already up-to-date
      } I think if this is true it won't trigger your code. it will make test hard.

qfai: better to skip that if the featureflag is on.

packages/fx-core/src/core/FxCore.ts
```diff
@@ -2675,8 +2676,16 @@ export class FxCore extends FxCoreOpenPluginPart {
...
+        ? templateConfig.v4tagPrefix
+        : templateConfig.tagPrefix;
+      const tag = `${tagPrefix}${latestVersion}`;
       const metadataZipUrl = `${templateConfig.templateDownloadBaseURL}/${tag}/metadata.zip`;
```
qfai, line 2716: what about bundled case? it is seems always download

qfai: the logic do have if download failure I use bundled but seems not respect your config

packages/fx-core/src/core/FxCore.ts
Alive-Fish: Good catch. The root cause is that the v3 and v4 metadata releases share the same version string, so the shared `template-version.txt` cache made `needDownload()` return `false` right after flipping the flag — leaving the user on stale v3 metadata. Fixed in f0f7af1eb8 by giving the v4 channel its own `template-version-v4.txt` cache file: switching channels now triggers a fresh download, while each channel still caches independently (no re-download every activation).

packages/fx-core/src/core/FxCore.ts
Alive-Fish: Same fix addresses this: with the dedicated `template-version-v4.txt`, a stable build still caches after the first download, so it is not 'always download'. Re-download only happens on a genuine version change or the first time the v4 channel is used (or for the mutable `0.0.0-rc` pre-release tag, same as v3). Bundled/local templates still short-circuit earlier via `useLocalTemplate()`.

Alive-Fish: You're right that the bundled fallback is v3's — there is no v4 bundled metadata in this package yet. Two notes: (1) this function itself returns an error on download failure rather than silently swapping channels; the bundled fallback only kicks in later in the readers (`getTemplateMetadataConfig` / `loadUiNode`) when `~/.fx/metadata` is absent. (2) Shipping a v4 bundled metadata copy is intentionally out of scope for this transitional PR — it belongs with the `selector.json` work that makes v4 a real bundled channel. Until then, a failed v4 metadata download degrading to v3 bundled metadata is the accepted fallback (the flag still drives the download source, which is the part this PR owns). I'll call this out in the PR description so it's tracked.

qfai (approved):

HuihuiWu-Microsoft (approved):

tecton (approved):

PR #16250, opened 2026-07-01
fix(vscuse): attempt 1 - In step_d0c53170, changed the click Y from 7…

qfai:
…0 to 125 so it selects "OAuth (with static registration)" (2nd item) instead of "OAuth (with dynamic registration)" (1st item); dynamic registration caused provision to fail ("missing registration_endpoint") so the client-ID prompt never appeared and step_73df5e59 timed out.


qinzhouxu (approved):

xiaolang124 (approved):

xiaolang124:
Seems the fix does not work, will have another pr to fix this case.

PR #16269, opened 2026-07-02
fix(fx-core): recover from corrupted template metadata cache

### LINKED ISSUE #16245: bug report: Metadata can become corrupted and requires manual clearing
**Describe the bug**
If template metadata cache JSON is corrupted, CLI commands that read template metadata can fail with a JSON parse exception instead of recovering.

This impacts flows that depend on template metadata loading (for example, template listing and command initialization paths that read template metadata).

**To Reproduce**
1. Corrupt the template metadata cache file:
   - File: `~/.fx/metadata/allTemplates.json`
   - Example invalid content: `{"badJson":`
   - I've seen this organically several times, probably due to cancelled execution.
2. Ensure metadata version marker files exist:
   - `~/.fx/template-version-v4.txt`
   - `~/.fx/template-version.txt`
3. Run a CLI command that reads templates metadata, for example:

```bash
atk list templates --format json
```
- CLI exits with an error.
- Error contains a JSON parse failure (for example, `SyntaxError: Unexpected end of JSON input`).
- Corrupted cache file and version markers remain, so the bad state persists.

**Expected behavior**
- CLI should not crash due to corrupted cache JSON.
- On corrupted cache detection, CLI should recover automatically by:
  - removing invalid cached metadata JSON,
  - clearing relevant metadata version marker files so metadata can be repopulated,
  - falling back to bundled metadata for the current operation.
- Command should complete successfully when bundled metadata is available.

**CLI Information (please complete the following information):**
 - OS: Windows 11
 - Version : Latest commit at time of filing

**Additional context**
PR incoming with fix.

neowenmsft:
## Summary
- recover from corrupted ~/.fx/metadata/*.json reads by falling back to bundled metadata
- recover from corrupted ~/.fx/ui/*.json reads by falling back to bundled UI nodes
- clear version marker files when corruption is detected so metadata can be repopulated
- add regression tests for corrupted cache recovery behavior

## Validation
- verified corrupted cache no longer crashes template listing in the fix branch
- verified baseline behavior still reproduces JSON parse failure

Fixes #16245

### DIFF
--- a/packages/fx-core/src/component/generator/templates/metadata/index.ts
+++ b/packages/fx-core/src/component/generator/templates/metadata/index.ts
@@ -9,6 +9,32 @@ import * as folder from "../../../../folder";
 import * as templateHelper from "../../templateHelper";
 import { Template } from "./interface";
 
+function getMetadataVersionFiles(platform?: Platform): string[] {
+  const configRoot = path.join(os.homedir(), `.${String(ConfigFolderName)}`);
+  if (platform === Platform.VS) {
+    return [path.join(configRoot, "vs-metadata", "template-vs-version.txt")];
+  }
+  return [
+    path.join(configRoot, "template-version.txt"),
+    path.join(configRoot, "template-version-v4.txt"),
+  ];
+}
+
+function invalidateCorruptedMetadataCache(cacheFilePath: string, platform?: Platform): void {
+  try {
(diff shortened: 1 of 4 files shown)


neowenmsft:
@microsoft-github-policy-service agree company="Microsoft"

PR #16291, opened 2026-07-06
fix(fx-core): gate bundled vs downloaded templates/metadata on useLocalTemplate flag

qfai:
## Why

Today the extension decides **at runtime** whether to use the **bundled (in-tree)** template zips + `metadata.zip` contents or to **download** them from GitHub Releases. That decision is driven only by the fx-core `package.json` version string. As a result, when a PR edits template/metadata/wizard JSON (e.g. `allTemplates.json`, `defaultGeneratorTemplates.json`, `ui/wizardNode.json`, `ui/tdpNode.json`, `ui/resource/package.nls*.json`), the PR's own CD-built vsix silently **downloads** the published metadata instead of using the in-tree copy — so the change is **not verifiable** in that build.

This PR gates the local-vs-download decision on the explicit **`useLocalTemplate`** flag already present in `templates-config.json`. The flag is committed `true` on `dev` and only flipped to `false` at build time for a **stable production release**, so every PR/CD build is **self-contained** and reflects its own template/metadata/wizard/NLS changes.

## What changed

**1. `packages/fx-core/src/component/generator/templateHelper.ts` — `useLocalTemplate()`**
Falls through to `templateConfig.useLocalTemplate` when no env/version override applies. Order:
- `TEMPLATE_VERSION=local` → `true` (unchanged)
- version contains `alpha` → `true` (unchanged)
- an explicit non-`local` `TEMPLATE_VERSION` → `false` (still downloads that exact version)
- otherwise → `templateConfig.useLocalTemplate`

This single change propagates to all five consumers that key off it: the metadata reader (`templates/metadata/index.ts`), the wizard tree reader (`scaffold/vsc/rootNode.ts`), both `fetchOnlineTemplateMetadata` / `…ForVS` download guards (`FxCore.ts`), and the NLS loader (`localizeUtils.ts`).

**2. `packages/fx-core/src/component/generator/utils.ts` — `getTemplateVSCUrl()`**
Honors the same flag so the **template ZIP** path also uses the bundled zips unless it is a real release. (`getTemplateVSUrl()` already read this flag.)

**3. `.github/scripts/fxcore-sync-up-version.js` — `updateUseLocalFlag()`**
Sets `useLocalTemplate = false` **only when the release is stable (non-prerelease) AND `goproduct=true`** (the `PRODUCTION` env). Every other build — alpha, preview/beta, rc, and any non-goproduct build — stays `true` (self-contained). The flip stays build-time only (`git update-index --assume-unchanged`), so `dev` remains `true` in the repo and esbuild bakes the release value into the artifact.

## Behavior matrix (`updateUseLocalFlag`)

| Build | stable? | goproduct | `useLocalTemplate` | templates/metadata source |
|-------|---------|-----------|--------------------|---------------------------|
| dev / PR CD (alpha) | no | false | **true** | bundled (in-tree) |
| preview / beta | no | any | **true** | bundled |
| rc | no | any | **true** | bundled |
| stable, non-goproduct | yes | false | **true** | bundled |
| **stable + goproduct** | yes | **true** | **false** | **downloaded (published)** |

## Reviewer call-outs (intentional behavior changes vs. the old rule)

The previous rule set `false` when `!prerelease || includes("rc") || VS_RELEASE==="true"`. The new rule drops the `rc` and `VS_RELEASE` conditions:

- **rc** builds now bundle instead of downloading `templates@0.0.0-rc`. These are content-equivalent (the rc CD rebuilds the same tree it would publish), and the rc vsix becomes self-contained.
- **VS release** (`vsrelease=true`) now depends on `goproduct`: a VS **stable + goproduct=true** release still downloads published `templates-vs@` (via `getTemplateVSUrl`), but a VS release run **without** `goproduct` would bundle. This assumes VS product releases are dispatched with `goproduct=true`. If that invariant isn't guaranteed, we should either re-add `VS_RELEASE` to the false-condition or enforce `goproduct=true` for `vsrelease` in `cd.yml`.

## Out of scope

- The **v4 template channel** (`FeatureFlags.V4Enabled`, default off) has its own `bundled = !goproduct` sync (`sync-v4-template-config.js`) and its scaffold bridge; not touched here.
- Pinning **metadata** to an explicit `TEMPLATE_VERSION` (the download path still selects the metadata version by channel) — pre-existing, unchanged.
- **Samples** resolution is unchanged.

## Tests

- Rewrote the `templateHelper` unit tests to deterministically cover the config-flag logic (local override, explicit version override, flag true → bundled, flag false → download).
- Added a `getTemplateVSCUrl` test proving the flag short-circuits to bundled even when the upstream version is higher.
- `packages/fx-core` suites run locally: `utils.test.ts` (22), `FxCore.templateMetadata.test.ts` (22), `generator.test.ts` (101), plus the generator/metadata/scaffold suites — all green (one pre-existing flaky "limit concurrency" timing test passes on isolated re-run). ESLint: no new errors.

### DIFF
--- a/.github/scripts/fxcore-sync-up-version.js
+++ b/.github/scripts/fxcore-sync-up-version.js
@@ -53,7 +53,15 @@ function syncTemplateVersion(templateVersion, templateConfigs) {
 }
 
 function updateUseLocalFlag(templateVersion, templateConfigs) {
-  if (!semver.prerelease(templateVersion) || templateVersion.includes("rc") || process.env.VS_RELEASE === "true") {
+  // Bundle the in-tree templates & metadata (useLocalTemplate = true) for every
+  // build EXCEPT a stable production release, i.e. a stable (non-prerelease)
+  // template version AND goproduct=true (PRODUCTION env). A stable production
+  // release downloads the published templates instead; every other build (alpha,
+  // preview/beta, rc, and any non-goproduct build) stays self-contained so a PR's
+  // CD-built vsix reflects its own template/metadata/wizard changes.
+  const isStableRelease = !semver.prerelease(templateVersion);
+  const isProduction = process.env.PRODUCTION === "true";
+  if (isStableRelease && isProduction) {
     templateConfigs.useLocalTemplate = false;
   } else {
     templateConfigs.useLocalTemplate = true;
--- a/.github/scripts/fxcore-sync-up-version.js
+++ b/.github/scripts/fxcore-sync-up-version.js
@@ -54,14 +54,19 @@ function syncTemplateVersion(templateVersion, templateConfigs) {
 
 function updateUseLocalFlag(templateVersion, templateConfigs) {
   // Bundle the in-tree templates & metadata (useLocalTemplate = true) for every
-  // build EXCEPT a stable production release, i.e. a stable (non-prerelease)
-  // template version AND goproduct=true (PRODUCTION env). A stable production
-  // release downloads the published templates instead; every other build (alpha,
-  // preview/beta, rc, and any non-goproduct build) stays self-contained so a PR's
-  // CD-built vsix reflects its own template/metadata/wizard changes.
+  // build EXCEPT a stable production release or a VS release. A stable production
+  // release is a stable (non-prerelease) template version AND goproduct=true
+  // (PRODUCTION env). A VS release (VS_RELEASE env) always downloads its published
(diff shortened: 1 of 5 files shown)


qfai:
@copilot resolve the merge conflicts in this pull request

HuihuiWu-Microsoft (approved):

PR #16306, opened 2026-07-07
fix(fx-core): generate DT add-MCP action inline instead of v4 modify front door

Alive-Fish:
## Summary

Fixes the Dynamic Tool Discovery (DT) "add MCP action" flow (`atk add action --api-plugin-type mcp`). Previously, with DT on, the flow routed through the v4 modify front door + `add-mcp-server` template, which produced an **incomplete `oauth/register`** action in `m365agents.yml` and coupled DT to `TEAMSFX_V4_ENABLED`. DT now generates the plugin manifest and auth wiring **directly in `core.addPlugin`**, independent of v4.

## What changed

- **Inline DT add-MCP** (`FxCore.declarativeAgent.ts`): DT-on `add action --api-plugin-type mcp` writes the URL-only `ai-plugin.json` (no static tools list, no `enable_dynamic_discovery`) and wires `oauth/register` / `dcr/register` via the shared injector — no longer via `modifyProjectFrontDoor`. Works with `TEAMSFX_V4_ENABLED` off.
- **URL-derived naming**: the `oauth/register` action `name` and `registrationId` are derived from the MCP server URL (shared verbatim with the create flow), so re-adding the same server is **idempotent** (one registration per server) and different servers get distinct registrations. Fixes the previous `action_<n>`-based naming that could skip creating a needed registration.
- **Auth-metadata probe**: when the caller doesn't pass `--mcp-da-auth-metadata-url`, the MCP server is probed to discover it (mirrors the create flow) so `oauth` / `oauth-dynamic` resolve real authorization/token/well-known URLs instead of empty ones.
- **Feature flag**: `TEAMSFX_MCP_FOR_DA_DT` is enabled by default; `FeatureFlagManager.listEnabled` now includes default-on flags.
- **CLI**: regenerated `AddPluginInputs` / `AddPluginOptions` for the DT auth-type + credential options.
- **e2e**: added MCP no-auth / with-auth / edge-case coverage.

> The v4 `add-mcp-server` modify template is intentionally **untouched** — the shipped DT flow no longer depends on it.

## Testing

- `packages/fx-core` unit: `tests/core/FxCore.declarativeAgent.test.ts` (inline flow, `oauth`/`entra-sso`/`oauth-dynamic` naming, metadata probe) + `tests/common/featureFlags.test.ts` — green.
- `tsc -p ./` clean.
- e2e env matrix validated locally with `atk new`: dynamic (V4 on) vs static (V4 off).

### DIFF
--- a/packages/fx-core/src/common/featureFlags.ts
+++ b/packages/fx-core/src/common/featureFlags.ts
@@ -125,7 +125,7 @@ export class FeatureFlags {
   };
   static readonly MCPForDADT = {
     name: FeatureFlagName.MCPForDADT,
-    defaultValue: "false",
+    defaultValue: "true",
   };
   static readonly MCPForDADCR = {
     name: FeatureFlagName.MCPForDADCR,
@@ -165,7 +165,7 @@ export class FeatureFlagManager {
   }
   listEnabled(): string[] {
     return this.list()
-      .filter((f) => isFeatureFlagEnabled(f.name))
+      .filter((f) => this.getBooleanValue(f))
       .map((f) => f.name);
   }
 }
--- a/packages/fx-core/src/core/FxCore.declarativeAgent.ts
+++ b/packages/fx-core/src/core/FxCore.declarativeAgent.ts
@@ -678,28 +678,42 @@ export class FxCoreDeclarativeAgentPart {
         };
 
         if (authType && authType !== "none") {
-          try {
-            // The add question tree does not collect the auth-server metadata /
-            // well-known URL, so — mirroring the create flow — probe the MCP
-            // server to discover it when the caller didn't pass one. The result
-            // feeds resolveMCPAuthEndpoints so oauth/register + dcr/register get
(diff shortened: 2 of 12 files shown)


packages/fx-core/src/common/featureFlags.ts
```diff
@@ -165,7 +165,7 @@ export class FeatureFlagManager {
   }
   listEnabled(): string[] {
     return this.list()
-      .filter((f) => isFeatureFlagEnabled(f.name))
+      .filter((f) => this.getBooleanValue(f))
```
HuihuiWu-Microsoft, line 168: Any reason for this change?

Alive-Fish: The previous way used the default value of the feature flag as false. It is wrong.
{312C3AE7-D26C-4097-AA32-9E90E18F2D89}

HuihuiWu-Microsoft (approved):

HuihuiWu-Microsoft (approved):

PR #16316, opened 2026-07-08
fix(fx-core): persist MCP add-action OAuth credentials to env files

Alive-Fish:
## Problem
When using **Add Action -> MCP server** (Dynamic Tool Discovery on, the default), the client id / secret / scopes entered during the flow were **not written to the env files**, so the injected `oauth/register` env-var references resolved to nothing at provision time.

## Root cause
`addPlugin`''s `useMCPDTModifyFlow` branch (`packages/fx-core/src/core/FxCore.declarativeAgent.ts`) collected the credentials via `MCPForDAAuthCredentialNodes()` but:
- called `injectMCPAuthActionToYml` **without** `persistCredentialEnvRefs` / `serverName`, and
- never called `persistMCPAuthCredentialEnvVars`.

The create flow (`component/generator/declarativeAgent/helper.ts`) already does this correctly; the add-action flow had drifted.

## Fix
Mirror the create flow in the add-action DT branch:
- derive `serverName = deriveMCPServerNameFromUrl(url).toUpperCase()`,
- pass `persistCredentialEnvRefs: true` + `serverName` to `injectMCPAuthActionToYml`,
- call `persistMCPAuthCredentialEnvVars(...)`.

Per auth type:
- **oauth** -> writes client id + secret (SECRET_ key) + scopes
- **entra-sso** -> writes only the client id
- **oauth-dynamic / none** -> no-op

## Tests
Updated the DT-on `oauth` and `entra-sso` `addPlugin` tests to assert the env writes and the injector credential env-ref args.
`npx vitest run tests/core/FxCore.declarativeAgent.test.ts` -> 66 passed, 6 skipped.

### DIFF
--- a/packages/fx-core/src/core/FxCore.declarativeAgent.ts
+++ b/packages/fx-core/src/core/FxCore.declarativeAgent.ts
@@ -53,6 +53,7 @@ import {
   ResolvedMCPAuthEndpoints,
   deriveMCPManifestOAuth,
   injectMCPAuthActionToYml,
+  persistMCPAuthCredentialEnvVars,
   resolveMCPAuthEndpoints,
 } from "../component/utils/mcpAuthScaffolder";
 import { pathUtils } from "../component/utils/pathUtils";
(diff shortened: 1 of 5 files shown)


packages/fx-core/src/core/FxCore.declarativeAgent.ts
```diff
@@ -734,6 +742,27 @@ export class FxCoreDeclarativeAgentPart {
               ),
             });
           }
+          // Persist the client id / secret / scopes the user entered into env
```
HuihuiWu-Microsoft, line 745: @copilot where is client secret persisted?

HuihuiWu-Microsoft (approved):

packages/fx-core/src/core/FxCore.declarativeAgent.ts
Alive-Fish: It will be persisted on env file (.env and .env.user)

PR #16340, opened 2026-07-14
test(vscuse): disable explorer hover tooltip in docker setting

qfai:
## Summary
- add workbench.hover.enabled: false in vscuse docker VS Code setting
- preserve existing workbench.hover.delay

## Purpose
- hide Explorer path hover tooltip in vscuse environment


qfai:
Superseded by same branch from OfficeDev origin.

PR #16437, opened 2026-07-23
test(vscuse): separate driver model + reduce AOAI load (parallelism 50, single attempt)

qfai:
## What / Why

Manage the Azure OpenAI load + model used by the nightly vscuse UI-test chain, so the vscuse
automation driver can run on a stronger (but rate-limited) model without saturating its TPM.

Three independent knobs, all on the test side (no vscuse package change, works with the
currently shipped wheel):

### 1. Separate the vscuse driver's model from the product under test
The vscuse setup used a **single** deployment name (`AZURE_OPENAI_MODEL`) for **two** consumers:
the vscuse driver's own agents (UI action + assertion / precondition verification, via
`config.yaml`) **and** the product under test (typed into the scaffolded app's `.env` by the
plan steps). This decouples them:

- `packages/tests/vscuse/vscode-test-cases/config.yaml`: `azure_openai.model` now reads a
  dedicated **`AZURE_OPENAI_VSCUSE_MODEL`**.
- `.github/workflows/ui-test-vscuse-common.yml`:
  `AZURE_OPENAI_VSCUSE_MODEL: ${{ vars.TEST_TENANT_AZURE_OPENAI_VSCUSE_MODEL || vars.TEST_TENANT_AZURE_OPENAI_MODEL }}`.

The product under test keeps its own deployment via the unchanged `${{env:AZURE_OPENAI_MODEL}}`
plan steps. No-op until the `TEST_TENANT_AZURE_OPENAI_VSCUSE_MODEL` repo variable is set.

### 2. Halve UI-test parallelism (100 → 50)
`ui-test-vscuse-common.yml` matrix `max-parallel: 100 → 50`. Fewer plans in flight at once =
lower peak token/request rate against the shared Azure OpenAI deployment.

### 3. Default to a single attempt (max_retries 7 → 1)
`max_retries` default `7 → 1` across `common` + the `template`/`features`/`samples` callers.
A failing stage now runs a single attempt instead of re-running up to 7 times.

The chain-advancement and report gates use `run_attempt == max_retries`, which still holds at 1
(`run_attempt` starts at 1), so **template → features → samples still advances** and the **daily
report still fires** — there is just no automatic retry by default. Both `max-parallel` (edit) and
`max_retries` (workflow_dispatch input) remain overridable.

## Behavior / Safety
- All three are **no-ops or conservative by default**: the model split falls back to
  `AZURE_OPENAI_MODEL` when the new repo variable is unset; parallelism/retry just lower load.
- Works with the **currently shipped** vscuse wheel — no wheel rebuild.

## Activation
Set repo variable `TEST_TENANT_AZURE_OPENAI_VSCUSE_MODEL` (e.g. `gpt-5.4`); it also doubles as
the throttle dial (clear/lower it if the stronger deployment hits its rate limit).


qfai:
```
Only test plan: Basic_Custom_Engine_Agent_Azure_OpenAI_py_Remote_Debug
```

Context: this push only swaps `packages/tests/vscuse/local-wheel/*.whl` (rebuilt from vsc-use PR #27 at db95769 — reconciled AI precondition retarget branch). Goal is to validate `step_25ce8a2f`'s Add-button click (originally blind-clicked due to the ocr:true empty-preconditions off-switch bug, now fixed) end-to-end in this exact plan.

PR #16464, opened 2026-07-27
refactor: export list templates for fxcore consumer

yuqizhou77:
- refactor list templates to be a function exported from fxcore
- tmp should be a production dependency


packages/fx-core/src/core/FxCore.ts
```diff
@@ -366,6 +371,21 @@ export class FxCore extends FxCoreOpenPluginPart {
     return res;
   }
 
+  /**
+   * List all scaffolding templates grouped by name. Thin, stateless delegate
+   * over the shared util so instance-based consumers get the same data.
+   */
+  listTemplates(): Result<TemplateGroup[], FxError> {
```
tecton: It seems that the 2 functions are not bound to fxcore instance. They are already exported at package level so it's better to prevent code bloat here.

yuqizhou77: removed

tecton (approved):

PR #16509, opened 2026-08-04
fix(fx-core): make listCollaborator collaborators field opt-in

### LINKED ISSUE #16512: fx-core: listCollaborator() discards resolved collaborator data despite ListCollaboratorResult.collaborators being declared
## Description

`listCollaborator()` in `packages/fx-core/src/core/collaborator.ts` fetches real Teams app owners and Aad owners, but discards them - it always returns `ok({ state: CollaborationState.OK })` even though `collaborators` is already declared on the `ListCollaboratorResult` type contract (`packages/fx-core/src/common/permissionInterface.ts`).

## Expected behavior

Callers that need the resolved collaborator list should be able to get it back from `listCollaborator()` instead of the field always being empty/absent.

## Repro

1. Call `listCollaborator()` with a project that has resolvable Teams app id and/or Aad object id.
2. Observe the returned `ListCollaboratorResult.collaborators` is never populated, even though the underlying Teams/Aad owner lookups succeed.

llhmm01:
## Summary

Closes #16512

`listCollaborator()` in fx-core fetches real Teams/Aad owner data but previously discarded it, always returning `ok({ state: CollaborationState.OK })` even though `ListCollaboratorResult.collaborators` was already declared in the type contract.

## Change

- Added a new opt-in input flag `CollaborationConstants.IncludeCollaborators` (`"includeCollaborators"`).
- When `inputs[CollaborationConstants.IncludeCollaborators] === true`, `listCollaborator()` now populates `collaborators` via a new `mergeCollaborators()` helper that merges Teams app owners and Aad owners by `userObjectId`.
- When the flag is absent/false (default), the return shape is unchanged: exactly `{ state: CollaborationState.OK }`, with no `collaborators` key at all — zero impact on existing CLI/VS Code callers.

## Tests

- Updated existing "happy path" and distinct-owners merge tests to opt in via the new flag, and added a mock for `CollaborationUtil.getTeamsAppIdAndAadObjectId` so the Teams/Aad owner lookups are actually exercised.
- Added a test asserting the default (no flag) behavior returns no `collaborators` key.

## Verification

CI `Unit Test/unit-test` and `E2E test` checks pass on the latest commit.

### DIFF
--- a/packages/fx-core/src/common/permissionInterface.ts
+++ b/packages/fx-core/src/common/permissionInterface.ts
@@ -40,10 +40,16 @@ export interface PermissionsResult {
   permissions?: ResourcePermission[];
 }
 
+export enum CollaboratorType {
+  TeamsApp = "teamsApp",
+  AadApp = "aadApp",
+  Agent = "agent",
+}
+
 export interface Collaborator {
   userPrincipalName: string;
   userObjectId: string;
-  collaboratorType: string;
+  collaboratorType: CollaboratorType;
   collaboratorResourceId: string;
 }
 
--- a/packages/fx-core/src/common/permissionInterface.ts
+++ b/packages/fx-core/src/common/permissionInterface.ts
@@ -46,8 +46,6 @@ export interface Collaborator {
   isAadOwner: boolean;
   teamsAppResourceId: string;
   aadResourceId?: string;
-  isAgentOwner: boolean;
-  agentResourceId?: string;
 }
 
 export interface AadOwner {
--- a/packages/fx-core/src/common/permissionInterface.ts
+++ b/packages/fx-core/src/common/permissionInterface.ts
@@ -24,6 +24,11 @@ export interface CollaborationStateResult {
 export interface ListCollaboratorResult {
   state: CollaborationState;
   message?: string;
+  /**
+   * Structured list of collaborators. Only populated when the caller opts in
+   * via `inputs[CollaborationConstants.IncludeCollaborators] = true`;
+   * omitted otherwise for backward compatibility.
+   */
   collaborators?: Collaborator[];
   error?: any;
 }
--- a/packages/fx-core/src/common/permissionInterface.ts
+++ b/packages/fx-core/src/common/permissionInterface.ts
@@ -43,9 +43,8 @@ export interface PermissionsResult {
 export interface Collaborator {
   userPrincipalName: string;
   userObjectId: string;
-  isAadOwner: boolean;
-  teamsAppResourceId: string;
-  aadResourceId?: string;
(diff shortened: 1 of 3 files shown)


packages/fx-core/src/common/permissionInterface.ts
```diff
@@ -24,6 +24,11 @@ export interface CollaborationStateResult {
 export interface ListCollaboratorResult {
   state: CollaborationState;
   message?: string;
+  /**
+   * Structured list of collaborators. Only populated when the caller opts in
```
tecton, line 28: If the `collaborators` is empty, the original behavior should also be fixed. It might be better to specify the collaborator type here?

packages/fx-core/src/core/collaborator.ts
```diff
@@ -468,11 +474,48 @@ export async function listCollaborator(
     telemetryProps[SolutionTelemetryProperty.CollaboratorCount] = teamsOwnerCount.toString();
     telemetryProps[SolutionTelemetryProperty.AadOwnerCount] = aadOwnerCount.toString();
   }
+  if (inputs[CollaborationConstants.IncludeCollaborators] === true) {
+    return ok({
+      state: CollaborationState.OK,
+      collaborators: mergeCollaborators(teamsAppOwners, aadOwners),
```
tecton: Agent owners are people have write permission to declarative agent, but `agentOwners` are missed here.

packages/fx-core/src/common/permissionInterface.ts
llhmm01: Thanks for the review! For this one, I'd prefer to keep it opt-in rather than changing the default behavior:

- The goal here is to satisfy wiqd's requirement for structured collaborator data, without needing any change on ATK's side.
- No existing caller reads `collaborators` today - both the CLI (`permissionStatus.ts`) and the VS Code extension (`collaboratorHandlers.ts`) only check `result.isErr()` and rely on the log/console output `listCollaborator()` already prints unconditionally. Neither touches `result.value.collaborators`.
- Making it opt-in via `inputs[CollaborationConstants.IncludeCollaborators]` lets wiqd get what it needs while guaranteeing zero behavior change for ATK's existing call sites.

Let me know if you see a concrete case where the default needs fixing - happy to revisit.

packages/fx-core/src/core/collaborator.ts
llhmm01: Good catch - fixed. `agentOwners` is now merged into the result via `mergeCollaborators()`, using the existing `Collaborator` fields (no new fields added, to keep the interface change minimal).

tecton (approved):

PR #16515, opened 2026-08-04
test(vscuse): migrate tab cases to YAML

Alive-Fish:
## What

Migrates the three basic tab plans into the generated semantic case engine.

| Legacy plan | Generated case |
| --- | --- |
| `Basic_Tab_Local_Debug.json` | `tab-ts-local-teams` |
| `Basic_Tab_Remote_Debug.json` | `tab-ts-remote-teams` |
| `Tab_Local_Debug_Env_Local_Creation.json` | `tab-ts-local-teams-env-recreated` |

## Why the engine needed extending

Every generated case so far reached a conversation. A tab reaches a page, and the same launch titles serve both: `Debug in Teams (Chrome)` opens a conversation for a bot and a tab page for a tab, so only the case knows which it scaffolded. Five capabilities were added for that, each with an AC row and a test named after it:

- **VCB-115** - the `non-sso-tab` selector path, which prompts no language because the package ships one.
- **VCB-116 / VCB-117** - a target profile registers at most one activation adapter per destination, and the shared add-and-open transition closes on the converged subject its adapter supplies.
- **VCB-118** - the local tab open accepts the self-signed certificate at `https://localhost:3978` and reloads the app inside Teams before claiming `page-ready`; the remote profile emits neither step.
- **VCB-119** - a `page` check asserts visible page content after `page-ready`.
- **VCB-120** - `removeWorkspaceFile` deletes one project-relative file, so a case can observe the local lifecycle recreating `env/.env.local`.

## Not migrated

The negative input flows stay with their legacy plans, which are therefore kept and marked **Partial** in `legacy-case-mapping.md`:

- the invalid and over-length app name flow, because a scaffold operation authors one answer per prompted question;
- the provision-before-login flow, which the compiler now rejects at compile time (`VCB_PROVISION_PREREQUISITE`) rather than at runtime;
- leftover notification cleanup, which is environment noise.

## Verification

- `pnpm --dir packages/tests run test:vscuse-engine` passes.
- `pnpm --dir packages/tests run generate:vscuse-cases` is idempotent; the 88 pre-existing plans have zero diff, including after the add-and-open component was parameterized.
- Prettier clean.
- Runtime validation against the pinned image is in progress; results will be posted in a comment.


Alive-Fish:
## CY260804 runtime validation

Validated `non-sso-tab--tab-ts-local-teams` after adding the Chrome local-device permission step.

- Image: `ghcr.io/officedev/vscuse-atk-vscode:CY260804`
- Image digest: `sha256:5a10cba810155cba0f7b87d69e48c4358c9fd0f5b257d60595a98036202771ee`
- Plan ID: `plan_d9538c0aaec9`
- Execution ID: `exec_20260805_101541`
- Duration: 1222.61 s
- Result: **93/93 passed (100%)**

The new `allowLocalDeviceAccess` component:
1. Asserted the `teams.cloud.microsoft` local-device access prompt.
2. Used `ocr:true` to resolve the screenshot-grounded Allow coordinate from `(389, 241)` to `(389, 240)`.
3. Clicked Allow and verified the prompt was dismissed.
4. Verified the tab content contains `Your app is running in TeamsModern`.

Artifacts: `.local/test-reports/20260805-101527-non-sso-tab--tab-ts-local-teams-pr16515-allow-run4/`

xiaolang124 (approved):

qfai (approved):

PR #16531, opened 2026-08-06
fix(cli): initialize region for ROPC authentication

qfai:
## Summary
- Initialize the Teams Developer Portal region after every successful ROPC M365 token acquisition.
- Add a regression test for the separate-but-equal `AppStudioScopes()` array case.

## Root cause
`M365ProviderUserPassword` gated AuthSvc region discovery on `tokenRequest.scopes === AppStudioScopes()`. `AppStudioScopes()` creates a new array on each call, so the identity comparison is always false. Consequently, CI ROPC authentication never initialized `teamsDevPortalClient.regionEndpoint`; `listApps` and `deleteApp` then failed with `Failed to get region`.

The normal code-flow provider already requests `AuthSvcScopes()` and calls `setRegionEndpointByToken` after authentication. This change aligns the ROPC provider with that behavior while retaining its headless username/password grant.

## Validation
- `pnpm exec vitest run tests/unit/commonlib/m365LoginUserPassword.tests.ts --config vitest.config.ts --maxWorkers=75%`
- `pnpm run build` in `packages/cli`
- Local ROPC AuthSvc discovery returned `regionGtms.teamsDevPortal` as `https://dev.teams.microsoft.com`.

### DIFF
--- a/packages/cli/src/commonlib/m365LoginUserPassword.ts
+++ b/packages/cli/src/commonlib/m365LoginUserPassword.ts
@@ -73,8 +73,7 @@ export class M365ProviderUserPassword extends BasicLogin implements M365TokenPro
     if (M365ProviderUserPassword.accessToken) {
       const m365Token = M365ProviderUserPassword.accessToken;
 
-      // Set region for App Studio API
-      if (tokenRequest.scopes === AppStudioScopes()) {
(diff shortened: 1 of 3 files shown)


qfai:
Superseded by #16533, which uses the upstream OfficeDev branch.

PR #16590, opened 2026-08-11
fix(vscuse): verify image feature flag hook

Alive-Fish:
## Summary

- derive ATK vscuse image source-ref/source-SHA tags and OCI revision from the checked-out Docker context rather than the workflow dispatch ref
- verify the published image by digest and reject it when the revision, feature-flag entrypoint, or settings synchronization script is missing
- add VCB-104 coverage for image provenance and startup-hook verification

## Root cause

A build dispatched on `dev` could check out another `source_ref` while `docker/metadata-action` still labeled the image from the workflow dispatch context. This allowed an image to claim a newer revision while containing stale Docker infrastructure without `vscuse-atk-entrypoint.sh`. Feature-flagged Copilot cases then passed `TEAMSFX_CEA_ENABLED=true`, but VS Code settings were never updated before scaffold, so the Copilot debug profile was absent.

## Validation

- `pnpm --dir packages/tests run generate:vscuse-cases` (twice, no generated plan changes)
- `pnpm --dir packages/tests run test:vscuse-engine` (178/178 passed)
- `node --test packages/tests/vscuse/docker/vscuse-atk/*.test.cjs` (3/3 passed)
- targeted feature-flag isolation/template-scope tests (6/6 passed)
- Prettier, editor diagnostics, and `git diff --check`
- startup injection smoke confirmed `PLAN_FLAG_INJECTED=true` and `VSCODE_SETTINGS_SYNCED=true`

The GitHub Actions build remains the final published-image verification; the workflow now fails before success if the pushed digest lacks the required hook.


Alive-Fish:
/atk-vscuse-test run general-teams-py-azure-openai-local-copilot

Alive-Fish:
/atk-vscuse-test run the test case "general-teams-py-azure-openai-local-copilot"

xiaolang124 (approved):

PR #16652, opened 2026-08-21
fix(fx-core): update local YAML for MCP auth

yuqizhou77:
## Description

When adding an authenticated MCP action to a declarative agent, update `m365agents.local.yml` when it exists in addition to the primary `m365agents.yml`. This matches the existing add-auth behavior while avoiding duplicate writes when both paths resolve to the same file.

## Changes

- Reuse the MCP auth injection arguments for the primary and local workflow files.
- Skip local injection when the local YAML does not exist or resolves to the primary path.
- Extend the existing MCP DCR test to verify both YAML files are updated.

## Validation

- `pnpm run build` in `packages/fx-core`
- `pnpm run build` in `packages/cli`
- `pnpm exec vitest run tests/core/FxCore.declarativeAgent.test.ts --config vitest.config.ts --maxWorkers=75%`
  - 72 passed, 6 skipped
- Real CLI E2E:
  - Scaffolded a declarative agent with `node cli.js new`.
  - Added a static OAuth MCP action with `node cli.js add action`.
  - Verified matching `oauth/register` actions in `m365agents.yml` and `m365agents.local.yml`.
  - Verified the declarative-agent action, `RemoteMCPServer` runtime, OAuth reference, and dev/local environment values.

The E2E project is preserved locally under `e2e-output/mcp-local-yaml-e2e-20260821-1608` for manual inspection.

### DIFF
--- a/packages/fx-core/src/core/FxCore.declarativeAgent.ts
+++ b/packages/fx-core/src/core/FxCore.declarativeAgent.ts
@@ -817,8 +817,7 @@ export class FxCoreDeclarativeAgentPart {
           try {
             const ymlPath = pathUtils.getYmlFilePath(inputs.projectPath);
             if (ymlPath) {
-              const injectResult = await injectMCPAuthActionToYml({
-                ymlPath,
+              const injectArgs = {
                 authType,
                 authName: namespace,
                 registrationId,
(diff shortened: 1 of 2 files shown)


yuqizhou77:
E2E failed on publish action, which has nothing to do with this PR change on adding plugin

tecton (approved):

PR #16667, opened 2026-08-27
feat(fx-core): support bearer token auth for MCP actions

yuqizhou77:
## Summary

- expose `bearer-token` authentication for MCP add actions in interactive and non-interactive CLI flows
- generate `ApiKeyPluginVault` manifest auth and `apiKey/register` provisioning actions for legacy and v4 scaffolding
- defer bearer secret collection to provisioning
- add scenario, compatibility, CLI, question, injector, and auth-action coverage

## Validation

- affected fx-core matrix: 345 passed, 6 skipped
- affected CLI tests: 10 passed
- v4 scenario and parity tests: 15 passed
- rebuilt `fx-core`, `cli`, and `vscode-extension`
- scaffolded and inspected v4-off and v4-on projects with the target MCP URL

## Notes

No `apiKey/update`, OAuth/DCR actions, `apiSpecPath`, or bearer secret values are generated. Package-wide unit runs retain unrelated dependency-checker failures and one stale CLI add-command count assertion.


tecton (commented):
## Review Findings

1. **High: Re-adding bearer-token authentication is not idempotent in the v4 flow.**

   The `registrationId()` helper in `mcpAuthAction.ts:59-67` only reads `writeToEnvironmentFile.configurationId`, while the newly added API-key action writes `writeToEnvironmentFile.registrationId` in `mcpAuthAction.ts:175-187`. As a result, the `mcpAuthAction.ts:211-214` never recognizes an existing bearer-token registration.

   Calling the injector twice with identical arguments produces two actions with the same registration ID:

   ```json
   {
     "apiKeyActionCount": 2,
     "registrationIds": [
       "MCP_DA_AUTH_ID_EXAMPLECOM",
       "MCP_DA_AUTH_ID_EXAMPLECOM"
     ]
   }
   ```

   This violates `SCN-ADD-MCP-05`. The helper should read either `registrationId` or `configurationId`, depending on the action type, and the tests should cover rerunning the bearer-token flow. The current idempotency test only covers OAuth.

2. **High: The v4 add flow updates only the main YAML file, leaving local authentication configuration incomplete.**

   The v4 pipeline injects the authentication action only into m365agents.yml. It does not update an existing `m365agents.local.yml`. Because the local YAML overrides the main YAML during local development, this causes the authentication registration action to be missing from local provisioning. The legacy flow explicitly updates both files in `FxCore.declarativeAgent.ts:830-834`.

   For an existing DA project containing `m365agents.local.yml`, adding bearer-token authentication through v4 results in:

   - `m365agents.yml` containing `apiKey/register`
   - `m365agents.local.yml` not containing the action
   - Local provisioning not producing `MCP_DA_AUTH_ID_*`
   - The plugin’s `ApiKeyPluginVault.reference_id` remaining unresolved

   This also contradicts the v3/v4 equivalence required by `SCN-ADD-MCP-17`. The current scenario test neither seeds nor checks a local YAML file, and the test labeled `SCN-ADD-MCP-17` only executes the v4 path instead of comparing it with v4 disabled: `addMcpServer.test.ts:228-249`.

3. **Medium: The CLI unintentionally exposes bearer-token authentication to `atk new`, although the v4 create descriptor rejects it.**

   `common.ts:19-22` is shared by both the add and create commands. After this change, `atk new --mcp-da-auth-type bearer-token` passes CLI option parsing. However, the v4 MCP create descriptor does not include `bearer-token` in its allowed enum: `descriptor.json:16`.

   With `TEAMSFX_V4_ENABLED` enabled, the CLI therefore accepts the command but the template input validation later rejects it. `SCN-ADD-MCP-14` only requires this value for `atk add action`.

   The gate should either accept a caller-specific set of choices or bearer-token should be added only to the add command’s options. The existing CLI test uses a synthetic option array and does not exercise the actual create command.

yuqizhou77:
All comments are addressed,

tecton (approved):

